Issue #17, opened 2022-02-04
TestJsonDiff not working correctly


slayerjain:
**Describe the bug**
I found a case in which `TestJsonDiff` doesn't work correctly. I've also added a test case for it in the test file. 

**To Reproduce**
Run this test case - https://github.com/keploy/keploy/blob/f930bf342c26510d746dd8a55591d24ac19b3f8c/pkg/match_test.go#L20

**Expected behavior**
It should return false for the inputs but returns true.

PR #18, opened 2022-02-05
fix: made TestJsonDiff function work correctly solving #17

Sarthak160:
Made some changes in jsonMatch function in match.go for making comparison between actual and expected Maps to fix bugs and added some comments for better understanding .


slayerjain:
LGTM

PR #69, opened 2022-03-01
fix(graphiql): fix the url for the graphql server

slayerjain:


nehagup (commented):
Can we please change the "Johari backend" param to "keploy backend"

nehagup (approved):
LGTM

PR #76, opened 2022-03-04
fix (testCaseDb) : fixes uri filter for reading testcases from mongo collection and sets the modified anchors instead of unset in mongo update.

re-Tick:


slayerjain:
@re-Tick the commit title is too long. please try to keep the titles short and you can utilize the message body to add more details

slayerjain:
@re-Tick also, we should add an option to disable dedup algo (and keep it disabled by default, until we create a separate mode for it)

slayerjain (changes requested):
added review in previous comments

re-Tick:
Added a boolean to enable deduplication logic for keploy server.

slayerjain (approved):

PR #93, opened 2022-03-26
ci: add commit linting and automatic release

slayerjain:
Although we are using conventional commits, there are situations when the commit messages can have unwanted spaces. These spaces can be problematic for automatic releaser like `semantic releaser` - which is also added in this commit.


nehagup (approved):
Approving as requested by @slayerjain

Issue #107, opened 2022-04-13
Fix code scanning alert - Log entries created from user input


nehagup:
Tracking issue for:
- [ ] https://github.com/keploy/keploy/security/code-scanning/1

###  **If the above issue returns 404 please find the details below.** 

https://github.com/keploy/keploy/blob/bbca9dbf8e5addf7e4aa87cb053bc2614b735925/pkg/service/regression/regression.go#L88-L88

This log write receives unsanitized user input from 
https://github.com/keploy/keploy/blob/bbca9dbf8e5addf7e4aa87cb053bc2614b735925/http/regression/regression.go#L129-L129

If unsanitized user input is written to a log entry, a malicious user may be able to forge new log entries.

Forgery can occur if a user provides some input with characters that are interpreted when the log output is displayed. If the log is displayed as a plain text file, then new line characters can be used by a malicious user. If the log is displayed as HTML, then arbitrary HTML may be included to spoof log entries.

**Recommendation**

User input should be suitably encoded before it is logged.

If the log entries are plain text then line breaks should be removed from user input, using strings.Replace or similar. Care should also be taken that user input is clearly marked in log entries, and that a malicious user cannot cause confusion in other ways.

For log entries that will be displayed in HTML, user input should be HTML encoded using html.EscapeString or similar before being logged, to prevent forgery and other forms of HTML injection.

**Example**

In the following example, a user name, provided by the user, is logged using a logging framework without any sanitization.

```
package main

import (
	"log"
	"net/http"
)

// BAD: A user-provided value is written directly to a log.
func handler(req *http.Request) {
	username := req.URL.Query()["username"][0]
	log.Printf("user %s logged in.\n", username)
}
```

In the next example, strings.Replace is used to ensure no line endings are present in the user input.

```
package main

import (
	"log"
	"net/http"
	"strings"
)

// GOOD: The user-provided value is escaped before being written to the log.
func handlerGood(req *http.Request) {
	username := req.URL.Query()["username"][0]
	escapedUsername := strings.Replace(username, "\n", "", -1)
	escapedUsername = strings.Replace(escapedUsername, "\r", "", -1)
	log.Printf("user %s logged in.\n", escapedUsername)
}
```

References

OWASP: [Log Injection](https://www.owasp.org/index.php/Log_Injection).
Common Weakness Enumeration: [CWE-117](https://cwe.mitre.org/data/definitions/117.html).

nehagup:
[...]

PR #109, opened 2022-04-15
fix(logging): Escape logged user input for security

ditsuke:
## Related Issue
Closes: #107
Closes: #106

This commit adds safe escaping for logged user input in the `service/regression` package, replacing newlines with spaces and escaping HTML.

## Type of change
Logging, security.

None, but I could work on it if we need any.

N/A

## Checklist:

## Screenshots (if any)
None.

### DIFF
--- a/pkg/service/regression/regression.go
+++ b/pkg/service/regression/regression.go
@@ -83,6 +83,9 @@ func (r *Regression) GetApps(ctx context.Context, cid string) ([]string, error) 
 	return r.tdb.GetApps(ctx, cid)
 }
 
+// sanitiseInput sanitises user input strings before logging them for safety, removing newlines
+// and escaping HTML tags. This is to prevent log injection, including forgery of log records.
+// Reference: https://www.owasp.org/index.php/Log_Injection
 func sanitiseInput(s string) string {
 	re := regexp.MustCompile(`(\n|\n\r|\r\n|\r)`)
 	return html.EscapeString(string(re.ReplaceAll([]byte(s), []byte(""))))
--- a/pkg/service/regression/regression.go
+++ b/pkg/service/regression/regression.go
@@ -5,8 +5,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"html"
 	"net/http"
 	"reflect"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/service/regression/regression.go
```diff
@@ -81,11 +83,16 @@ func (r *Regression) GetApps(ctx context.Context, cid string) ([]string, error)
 	return r.tdb.GetApps(ctx, cid)
 }
 
-func (r *Regression) Get(ctx context.Context, cid, appID, id string) (models.TestCase, error) {
+func sanitiseInput(s string) string {
```
slayerjain, line 86: @ditsuke Thanks for the PR. How about we add some documentation on top of this method to explain why we added this. You can also reference the issue links or the security vulnerability. It'll inform future contributors especially when refactoring the code base :)

slayerjain (changes requested):

pkg/service/regression/regression.go
```diff
@@ -81,11 +83,16 @@ func (r *Regression) GetApps(ctx context.Context, cid string) ([]string, error)
...
+	re := regexp.MustCompile(`(\n|\n\r|\r\n|\r)`)
+	return html.EscapeString(string(re.ReplaceAll([]byte(s), []byte(" "))))
+}
 
+func (r *Regression) Get(ctx context.Context, cid, appID, id string) (models.TestCase, error) {
 	tcs, err := r.tdb.Get(ctx, cid, id)
 	if err != nil {
-		r.log.Error("failed to get testcases from the DB", zap.String("cid", cid), zap.String("appID", appID), zap.Error(err))
+		sanitizedAppID := sanitiseInput(appID)
+		r.log.Error("failed to get testcases from the DB", zap.String("cid", cid), zap.String("appID", sanitizedAppID), zap.Error(err))
```
nehagup, line 95: Check failure on line 95 in pkg/service/regression/regression.go

**Log entries created from user input**
This log write receives unsanitized user input from
https://github.com/keploy/keploy/blob/7206b0797938cc1119158a348da6bba8b0856708/http/regression/regression.go#L129-L129

pkg/service/regression/regression.go
```diff
@@ -103,7 +110,8 @@ func (r *Regression) GetAll(ctx context.Context, cid, appID string, offset *int,
 	tcs, err := r.tdb.GetAll(ctx, cid, appID, false, off, lim)
 
 	if err != nil {
-		r.log.Error("failed to get testcases from the DB", zap.String("cid", cid), zap.String("appID", appID), zap.Error(err))
+		sanitizedAppID := sanitiseInput(appID)
+		r.log.Error("failed to get testcases from the DB", zap.String("cid", cid), zap.String("appID", sanitizedAppID), zap.Error(err))
```
nehagup, line 114: Check failure on line 114 in pkg/service/regression/regression.go

**Log entries created from user input**

This log write receives unsanitized user input from 
https://github.com/keploy/keploy/blob/7206b0797938cc1119158a348da6bba8b0856708/http/regression/regression.go#L129-L129

nehagup:
Please also add yourselves here https://github.com/keploy/keploy#-our-valuable-contributors-

pkg/service/regression/regression.go
ditsuke: @slayerjain thanks, I absolutely agree! Will push a godoc to go with the function

pkg/service/regression/regression.go
ditsuke: @nehagup I might be totally be missing something here, but could this be a false positive? The `app` query is passed as the `appID` param to `regression.Get()` which we're now sanitizing before passing to the logger. I believe this should have fixed this warning.

slayerjain: @ditsuke can you please correct the `sanitiseInput` methods replace to replace empty string (eg: `""`) rather than a space( eg: `" "`). Maybe that fixes it?

ditsuke: @slayerjain giving it a try, fingers crossed.

ditsuke: Thanks, that worked fine! I believe that was a false positive indeed. 😄

pkg/service/regression/regression.go
ditsuke: Resolved

pkg/service/regression/regression.go
ditsuke: I've added the godoc. PTAL.

slayerjain:
@ditsuke please resolve the conflict. We're almost there :)

ditsuke:
@slayerjain should be good to go now

slayerjain (approved):

ditsuke:
Thanks guys!

PR #136, opened 2022-06-05
Modify GetAll API of testCaseDB to return all entries instead of considering index and limit

anirudhjain75:
## Related Issue
https://github.com/keploy/ui/issues/24 /testlist testcases part

Closes: NA

Removed limit and offset consideration of getAll for testcasedb. Therefore allowing pagination to be handled on client side.

Added ToDo comment for future purpose.

## Type of change

None

My code generates warning as I added a ToDo comment for future reference.

## Checklist:

## Screenshots (if any)
NA

### DIFF
--- a/pkg/platform/mgo/tdb.go
+++ b/pkg/platform/mgo/tdb.go
@@ -258,6 +258,8 @@ func (t *testCaseDB) getAll(ctx context.Context, filter bson.M, findOptions *opt
 
 func (t *testCaseDB) GetAll(ctx context.Context, cid, app string, anchors bool, offset int, limit int) ([]models.TestCase, error) {
 
+	// ToDo: Implement Pagination Once its required. Currently Fetching all entries from Database
+
 	filter := bson.M{"cid": cid, "app_id": app}
 	findOptions := options.Find()
 	if !anchors {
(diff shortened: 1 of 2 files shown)


nehagup:
@anirudhjain75 How about we keep 1K limit on the frontend and not change the backend implentation?

anirudhjain75:
Hmm. Then what about the times when there are more than 1k testcases ?

anirudhjain75:
@nehagup 

Here are the drawbacks with implementation

Handling pagination on Server -> Requires amount of total item exist for particular fitler
*Handling pagination on Client -> Requires all the entries in client so the count and pages are automatically configured.

* - Implemented by this PR

Incase the req input size is set to 1000, The pagination can be made to work correctly but the count displayed in pagination part of DataGrid would be wrong.

A better solution would be updating the server to send PageSize data with App data with getTc API which currently only sends a list of testcases.

Possible implication of this would be that an extra query is run on database to get the number of testCases of an App.

nehagup:
Thanks for the PR @anirudhjain75. While this was a good approach, we found an approach where the schema changes were not required and using the beauty of GraphQL within the same complexity we were able to fetch the total number of records. Closing this one as this is resolved now. :)

PR https://github.com/keploy/ui/pull/68

Issue #444, opened 2023-03-06
[feature]: add configuration to pipe logs to a file


re-Tick:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

Currently Keploy server logs are logged on CLI which Keploy binary is running. Due to which upon starting Keploy server from SDK asynchronously, the server logs are logged along with sample application logs. 

 We can add a logPath field in our configurations to pipe the logs into the text file instead of CLI. 

### Why should this be worked on?

The server logs are mixed up with the user application logs due to which it becomes to debug.

achintya-7:
Is this issue related to the Keploy GO SDK

achintya-7:
Would like to work on this issue. Can I just directly start with it?

slayerjain:
@achintya-7 yes feel free :)

Sarthak160:
We don't need this feature in the newer keploy version . Since there is no SDK involved and keploy will be running the user application .

PR #457, opened 2023-03-07
Implemented method to pipe logs to a custom log file

achintya-7:
## Related Issue
Resolves the feature to add a way to pipe logs to a file

Closes: #444 

If a user passes a value for `config.LogPath`, a new file will be generated of that name and a new zap object is also created to write logs in that particular file

## Type of change

`NA`

## Checklist:

## Screenshots (if any)


achintya-7:

server/server.go
```diff
@@ -66,6 +67,7 @@ type config struct {
 	Port             string `envconfig:"PORT" default:"6789"`
 	ReportPath       string `envconfig:"REPORT_PATH" default:""`
 	PathPrefix       string `envconfig:"KEPLOY_PATH_PREFIX" default:"/"`
+	LogPath          string `envconfig:"LOG_PATH"`
```
re-Tick, line 70: Is this path for logs is absolute? Can it check for relative path and resolve to absolute path.

achintya-7: Indeed, it will create a file in the same dir as that of go.mod file. I have been using this at my work place as well and it has been working fine.

server/server.go
```diff
@@ -84,6 +86,28 @@ func Server(ver string) *chi.Mux {
...
+		logPath := fmt.Sprintf("%s.log", conf.LogPath)
+		logFile, err := os.Create(logPath)
+		if err != nil {
+			logger.Error("failed to create log file", zap.Error(err))
+		}
+		defer logFile.Close()
+
+		fileSyncer := zapcore.AddSync(logFile)
+
+		core := zapcore.NewCore(
```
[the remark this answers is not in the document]
achintya-7: Yes
If the `conf.LogPath` is empty, it will just log it on the console, but
If the `conf.LogPath` is not empty, it will write the logs in the file name given.

I am adding a demo script which follow the same procedure as written here for testing 
```
package main

import (
	"fmt"
	"os"
	"time"

	"go.uber.org/zap"
	"go.uber.org/zap/zapcore"
)

func main() {
	logConf := zap.NewDevelopmentConfig()
	logConf.Level = zap.NewAtomicLevelAt(zap.InfoLevel)
	logger, err := logConf.Build()
	if err != nil {
		panic(err)
	}
	defer logger.Sync()

	logPath := "" // if given a path, will write to file 
	// logPath = "test"

	if logPath != "" {
		logPath := fmt.Sprintf("%s.log", logPath)
		logFile, err := os.Create(logPath)
		if err != nil {
			logger.Error("failed to create log file", zap.Error(err))
		}
		defer logFile.Close()

		fileSyncer := zapcore.AddSync(logFile)

		core := zapcore.NewCore(
			zapcore.NewConsoleEncoder(logConf.EncoderConfig),
			fileSyncer,
			logConf.Level,
		)

		logger = zap.New(core)
	}

	logger.Info("hello world")
	time.Sleep(1 * time.Second)
	logger.Error("Goodbye world")
}
```

Hope this helps

achintya-7:
There is one doubt that I have. Every time we run the program, do we want to rewrite the new logs over the existing file? Or write new logs preserving the previous ones. 
For the current PR, it will rewrite the logs and we will lose the previous logs in it.
I can implement the later functionality as well if that's what is needed here.

re-Tick:
> There is one doubt that I have. Every time we run the program, do we want to rewrite the new logs over the existing file? Or write new logs preserving the previous ones. For the current PR, it will rewrite the logs and we will lose the previous logs in it. I can implement the later functionality as well if that's what is needed here.

We need the later one because the test-runs result summary are logged. Therefore we need older logs of 1 running instance.

server/server.go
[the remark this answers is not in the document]
re-Tick: Got it. just tried this locally. It works perfectly. LGTM

server/server.go
re-Tick: LGTM. Please add comments for the format of the logPath.

achintya-7: Sure 👍

achintya-7:
>  
Ok, So I have to only preserve all the logs of the last running instance rather than all the logs of all instances.

re-Tick:
I have tried the changes locally. But we should have logs of all instances.

re-Tick:
I have tried the changes locally. But we should have logs of all instances.

achintya-7:
> We need the later one because the test-runs result summary is logged. Therefore we need older logs of 1 running instance. 

Ok, Got it.
I will also add a separator for better clarification of running instances.

re-Tick:
Hi @achintya-7, Keploy also uses a pp library to log the difference in response. We should also redirect that logs to a file.

achintya-7:
Yeah, I have come across them in regression.go quite many times for `test` mode. I believe we can go with a global logger object or use it as a pkg. We can then use it at multiple places.

achintya-7:
Yeah, I have come across them in regression.go quite many times for `test` mode. I think we can use the regression struct which has a log field for using the same logger object. Being a receiver function, this should be easily implemented.

re-Tick:
I think that zap don't provide the functionality to pretty print the logs.

achintya-7:
Well, I am facing some issues when writing logs to a file that is formed by pp.
image
Have tried both the ways i.e. zap-logger and native os.Write but this same problem persists. It may be due to the colour coding of pp.

re-Tick:
@achintya-7, yes this is because of the logging system of pp library. This is acceptable because we can view the formatted logs using cat cmd.

achintya-7:
@re-Tick  Ok, I will add the required commit soon.

Sarthak160:
This PR is stale since we migrated to V2 which has newer code structure .

Issue #469, opened 2023-03-13
[bug]: Keploy server doesn't start if we use PORT env variable for our server


Frey0-0:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When we use the `PORT` variable instead of `KEPLOY_APP_PORT` for the sample app then the Keploy server will try to start on the same port as that of the sample app as `PORT` is a generic name so we change to some other variable [here](https://github.com/keploy/keploy/blob/ee0e9c33ebeedaf213713e03205343146e21449d/server/server.go#L66). This related conversation can is in this [PR](https://github.com/keploy/typescript-sdk/pull/62).

### Steps to reproduce

1. Use node-fetch sample of samples/typescript.
2. Add `export PORT=8080` in the [.env](https://github.com/keploy/samples-typescript/blob/main/node-fetch/.env) file of node-fetch.
3. Now, Change the env varibale name from `KEPLOY_APP_PORT` to `PORT` in [this](https://github.com/keploy/samples-typescript/blob/7a26737be3aa0f1f65a898fc0710095f1e34a3e1/node-fetch/server.js#L31) line of app.listen
4. Try recording some testcases.
    Error will be thrown` failed to post testcase to keploy server`.

### Environment

Release

### Version

Self-hosted

PR #470, opened 2023-03-13
update environment variable name for port of keploy server

Frey0-0:
Closes: #469 
Reference: [#62](https://github.com/keploy/typescript-sdk/pull/62)

Just change the name of the environment variable for the port of the Keploy server

## Type of change

## Checklist:

## Screenshots (if any)

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -63,7 +63,7 @@ type config struct {
 	EnableDebugger   bool   `envconfig:"ENABLE_DEBUG" default:"false"`
 	EnableTestExport bool   `envconfig:"ENABLE_TEST_EXPORT" default:"true"`
 	KeployApp        string `envconfig:"APP_NAME" default:"Keploy-Test-App"`
-	Port             string `envconfig:"PORT" default:"6789"`
+	Port             string `envconfig:"KPORT" default:"6789"`
 	ReportPath       string `envconfig:"REPORT_PATH" default:""`
 	PathPrefix       string `envconfig:"KEPLOY_PATH_PREFIX" default:"/"`
 }


nehagup (changes requested):
I think this will be a breaking change for existing users. Also, change is also required where this env variable is being read.

Sarthak160:
I guess , This we already have an alternative for this problem . @Frey0-0 Since we are using [envconfig](https://pkg.go.dev/github.com/kelseyhightower/envconfig@v1.4.0) library for managing env variables , You can simply use `KEPLOY_PORT` env variable to set your custom keploy port as it has more priority than `PORT` env variable. 
Similarly you can prefix `KEPLOY_`  with any other keploy env variable available in the config.

Frey0-0:
Hey @Sarthak160 @nehagup, Actually if a person doesn't use the variable `KEPLOY_PORT` and they are using the variable `PORT` in their project's because it is so common and source it then the keploy server will also try to run on the same port as that of the app. I was actually trying to solve another issue where we start the keploy server automatically so @re-Tick encountered this issue while reviewing so if we change the variable name from env even if a person uses PORT it will still be default 6789 unless they mention `KPORT.`
I might be wrong( most probably am) but the variable name read everywhere is still `Port` of the `config` struct, just variable name it is mapped to in the env will be `KPORT`. It is not a breaking change :sweat_smile: I used the dev docker version mentioned in the repo. Please tell me if I am wrong and what else can be a better solution to this :smile:

Sarthak160:
Hey @Frey0-0 , Can you please try by using `KEPLOY_PORT` env variable for keploy's port and `PORT` env variable for your own application . I think this should work and should not overlap the keploy's own port .

Sarthak160:
Hey @Frey0-0 , Can you please try by using `KEPLOY_PORT` env variable for keploy's port and `PORT` env variable for your own application . I think this should work and should not overlap the keploy's own port .

Frey0-0:
@Sarthak160 I know that :sweat_smile: but it is for a user who is just using keploy and has no idea of the internals. I didn't face this problem actually @re-Tick was trying a change I made for an issue in typescript-sdk and faced this issue :sweat_smile: so I raised it.

re-Tick:
@Sarthak160 let me try the changes. Maybe we don't require a "KPORT".

re-Tick:
@Frey0-0 I have tried using KEPLOY_PORT env variable while running keploy from TS-SDK in your changes. And it works. Please close this PR.

Issue #334, opened 2023-03-02
[feature]: beautify response diff viewer on CLI   


nehagup:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

Currently after running the test-cases the test-result shows the differences of the responses in the failing test cases which looks messy for a long response. This task is to improve that view of showing the differences in the responses on the CLI logs to make more readable. 

It's an open-ended task since differences can be shown in many different representations like git diff checker, diff checker, line by line etc.. 

Here's the current way: 

   
<img width="1132" alt="Screenshot 2023-03-02 at 4 34 36 PM" src="https://user-images.githubusercontent.com/15074229/222411748-f121217d-12fc-433c-ba1c-d80b9fa0ae90.png">

### Why should this be worked on?

To enhance the readability of failing test cases.

Sukriti-m:
Hi @nehagup 
I want to work on this issue.
I'd be grateful if you could assign this issue to me.

SanskritiHarmukh:
Hi @Sukriti-m , since this is a gsoc issue, it will not be assigned to anyone rather you can start working on it directly and raise a PR for the same. After creating a PR, add the PR ID to the task table under [the CLI Refactoring project](https://github.com/keploy/gsoc/tree/main/2023#1-keploy-cli-refactoring) corresponding to the issue.

Sukriti-m:
> Hi @Sukriti-m , since this is a gsoc issue, it will not be assigned to anyone rather you can start working on it directly and raise a PR for the same. After creating a PR, add the PR ID to the task table under [the CLI Refactoring project](https://github.com/keploy/gsoc/tree/main/2023#1-keploy-cli-refactoring) corresponding to the issue.

Thanks for telling. I will start working on this issue then. 😊

ADV1K:
How about putting the expected outcome and actual outcome inside a box?

https://github.com/Delta456/box-cli-maker

SanskritiHarmukh:
@ADV1K totally on you how you wanna style and bring out the creativity.

nehagup:
@achintya-7 we're open for both. However, the diff viewer in output logs is the main task.

achintya-7:
Ok got it

Sarthak160:
#492 has been accepted and merged in the main code base .

PR #481, opened 2023-03-19
Add human readable prettified DIFFs for failing test cases

aerowisca:
Closes: #334

Keploy simply displays LHS and RHS for the changed components, which are not easy to look at and interpret. This commit adds support to view the diffs in a more human readable and colorful format.

## Type of change

NA

## Checklist:

## Behaviour before the PR
before

## Behaviour after the PR
after-1

after-2


nehagup:
@aerowisca while this will look good incase when differences are small, how about the cases when there's a whole json object diff or 2,3 elements of an array are diff?

aerowisca:
@nehagup  This should  be easy to visualize even in case of large json diff (recursively) or when some elements of the array are missing/replaced. Here's a sample.

large-json-diff

nehagup:
> @nehagup This should be easy to visualize even in case of large json diff (recursively) or when some elements of the array are missing/replaced. Here's a sample.
> 
> large-json-diff
@aerowisca Assume there are more than 10 cases and the differences are displayed like you showed, it'll be a huge log pipe. How about we showcase the expected and actual test-response side-by-side like popular code diff checkers does(like we also see on GitHub)? IMO that'll reduce the length of the diff-logs to almost half. WDYT?

aerowisca:
@nehagup  I did some reasearch and I was able to accomplish the goal. Could you please take a look?

Here's the final version: 
p1
p2

nehagup (commented):
@aerowisca mostly looks good! Some formatting can enhance the readability of it though. 

2 cases I see might look odd in this formatting.
1. There's an extra object in either of the jsons. Maybe we can make that more identifiable by clearly mentioning which is expected and actual response in 2 halves of the output screen.

2. Have you considered the case where the noisy fields marked by keploy are not highlighted as differences?

aerowisca:
Yes, @nehagup  The CR already handles removal of noisy fields for difference (It piggy-backs on the current implementation of removal of noisy fields, instead of adding its own code). 

As you can see [here](https://github.com/keploy/keploy/blob/ad53d4fa95ef1be3b97514aee642b08b0d4c4436/pkg/service/regression/regression.go#L192) and [here](https://github.com/keploy/keploy/blob/ad53d4fa95ef1be3b97514aee642b08b0d4c4436/pkg/service/regression/regression.go#L196), `cleanExp` and `cleanAct` are supposed to contain data after removing the noisy fields.  You can see that in my PR, I only apply the algo to [denoised fields](https://github.com/aerowisca/keploy/blob/8320af83239b968ddaf2945cfe6fa599fb143752/pkg/service/regression/regression.go#L267)

I don't see any issue with the extra obejct in the JSON (assuming you are talking about `message: "field"`). When there are no differences, we are still displaying the fields without any highlights, so that the structure of the JSON remains intact. Doesn't make sense to print the same value on both halfs.

By the way, what is the acceptance criteria for the required labels for GSoC? If this PR meets those criterias, could you add the required labels so that I can mention it in the proposal doc? Thanks.

nehagup (changes requested):
@aerowisca 

Can you please add some sample screenshots with some sample apps that showcase the noisy fields being ignored by your diff comparator? I suspected that from the screenshot you posted (`timestamp` field is generally noisy). I re-run the java-sample SDK with your changes and found no luck with ignoring noisy fields. Here are some screenshots. 
Screenshot 2023-04-03 at 11 33 06 PM

 
Screenshot 2023-04-03 at 11 35 42 PM

> Yes, @nehagup The CR already handles removal of noisy fields for difference (It piggy-
> backs on the current implementation of removal of noisy fields, instead of adding its own 
> code).

It'll be great if you can mention the pointers/reasons for your opinion formation. I'm sharing the thought process of where I come from. IMO you can either show differences in SPLIT or UNIFIED popular diff components (attaching screenshots from a GitHub, tool that sets a great example of dev experience, only to showcase the difference). 

Screenshot 2023-04-03 at 11 45 38 PM

Screenshot 2023-04-03 at 11 46 20 PM

IMO most of the code-diff tools are designed this way. Also, since our code(response) differences are huge, it's better to save on the number of lines to keep the output logs shorter with a more crisp and concise experience.   

> I don't see any issue with the extra obejct in the JSON (assuming you are talking about 
> message: "field"). When there are no differences, we are still displaying the fields without 
> any highlights, so that the structure of the JSON remains intact. Doesn't make sense to 
> print the same value on both halfs.



I also found that your changes were breaking for a case and the diff showed empty array.  Please find the attached screenshot for that as well.
Screenshot 2023-04-03 at 11 32 47 PM

This is an open-ended task, you can draft and refine the PR as much as you want to. The `accepted` label is the criterion to make it mergeable and to be considered `complete` for Keploy. You can mention this PR as part of the `WIP` or `Draft` section since the proposals are closing soon, and can still work on it later..   

> By the way, what is the acceptance criteria for the required labels for GSoC? If this PR 
> meets those criterias, could you add the required labels so that I can mention it in the 
> proposal doc? Thanks.

aerowisca:
> The accepted label is the criterion to make it mergeable and to be considered complete for Keploy

Thanks for the clarification. I was under the impression that it was for selection criteria.

About the noisy fields, as I stated earlier, I didn't handle any specific scenario of removing them, since the code to remove them was already there (along with documentation). Let me investigate on whether there's a bug in the current implementation of removing noisy field removal during difference display (or possibly bug in my CR).

About the breaking change, thanks for the heads up. Let me take a more detailed look and run some more experiments (once I'm done with my uni exams).

Sarthak160:
#492 Already resolved by this PR

PR #482, opened 2023-03-19
Add feature to pipe or redirect the logs via user flags

aerowisca:
Closes: #444

The current implementation prints the entire logs in the console. This is not effective when integrated with other applications. To give the user complete flexibility, this commit adds support for a flag using which the user can ask the logs to either be printed on the console, or to a file of their choice (via another flag), or pipe the logs to both console and file.

## Type of change

## Checklist:


server/server.go
```diff
@@ -66,24 +81,47 @@ type config struct {
...
 
 func Server(ver string) *chi.Mux {
 	rand.Seed(time.Now().UTC().UnixNano())
 
+	// Parse the config before creating the logger to figure out whether we need
+	// to pipe the logs or not.
+	var conf config
+	err := envconfig.Process("keploy", &conf)
+	if err != nil {
+		panic(fmt.Errorf("failed to read/process configuration with error %v", err))
```
re-Tick, line 102: We should not panic in the program. Please throw an error and return.

re-Tick:
Keploy also uses a prettify logging library (called pp) which is used to log the test results in regression service. Those logs should also be transfer or pipe in the file.

aerowisca:
@re-Tick  Addressed the review comments and added support for the pretty printer logs to be piped/redirected as per user's choice.

Please take a look.

server/server.go
```diff
@@ -66,24 +67,48 @@ type config struct {
...
 	logConf := zap.NewDevelopmentConfig()
 	logConf.Level = zap.NewAtomicLevelAt(zap.InfoLevel)
+
+	// If the user has requested to the pipe the logs in a file,
+	// modify the internal config for zap which supports multiple syncers.
+	switch conf.LogLocation {
+	case models.Console:
+		logConf.OutputPaths = []string{"stderr"}
+	case models.File:
+		logConf.OutputPaths = []string{conf.LogFileLocation}
```
re-Tick, line 101: Please create a file if already not exist for pipe and redirect the logs.

aerowisca: Hi, @re-Tick  can you please elaborate? My code already handles this scenario, i.e, it creates the file if that file does not exist. (You just need to pass in the file names to conf, and the internals of `zap` would ensure file creation.

re-Tick: It throws following error when there is no tmp directory in keploy repo.
```
panic: couldn't open sink "tmp/keploy.log": open tmp/keploy.log: no such file or directory

goroutine 1 [running]:
go.keploy.io/server/server.Server({0x1049c2144, 0x9})
        /Users/xxx/Desktop/go-practice/tmp/aerowisca/keploy/server/server.go:108 +0x1c98
main.main()
        /Users/xxx/Desktop/go-practice/tmp/aerowisca/keploy/cmd/server/main.go:14 +0x30
```

 Steps to reproduce:
1. Use this env variables and make sure that there is no tmp folder in the keploy repo
```bash
export KEPLOY_LOG_LOCATION=file
export LOG_FILE_LOCATION="tmp/keploy.log"
```
2. source the env variables and run the keploy server
```bash
source .env && go run cmd/server/main.go
```

aerowisca: Yes, that understanding is correct, and is in line with what I've seen in other projects (i.e, external packages should only create missing files, instead of creating entire directories, since we do not know about the specific permissions of the nested directories that we are going to create).

It's the user's responsibility to specify a correct directory (regardless of file being present there). (In documentation, that's a typo, On Unix systems, I was planning to default it to `/tmp` directory).

I can of course add the feature for creating the nested directories as well, but just curious on what other's thoughts are. Do you think I should add it?

aerowisca: @re-Tick  What about the other features (assuming the directory exists)? Did you get a chance to experiment with those by creating `tmp` folder? Let me know if any changes/modifications are needed there.

re-Tick: Yes, I have tried after creating the tmp folder, and it is working.

re-Tick: I think we should create the missing directory or nested directory because this will decrease a step of creating the nested directory and be more quick.

aerowisca: Makes sense. I've added the feature to create nested directories in my new commit. Can you please do one more round of experiments at your end?

Yes, the current implementation should handle absolute as well as relative paths.

server/server.go
```diff
@@ -66,24 +67,62 @@ type config struct {
...
 
 func Server(ver string) *chi.Mux {
 	rand.Seed(time.Now().UTC().UnixNano())
 
+	// Parse the config before creating the logger to figure out whether we need
+	// to pipe the logs or not.
+	var conf config
+	err := envconfig.Process("keploy", &conf)
+	if err != nil {
+		fmt.Printf("failed to read/process configuration with error %v", err)
```
re-Tick, line 88: Please use log.Fatal to log error and return

server/server.go
```diff
@@ -66,24 +67,62 @@ type config struct {
...
+	// Convert the separators in the filepath to a platform agnostic version.
+	// For example, if user passed in "/C/Downloads", it'd be converted to
+	// "\C\Downloads" in Windows.
+	conf.LogFileLocation = filepath.FromSlash(conf.LogFileLocation)
+
+	// If we need to pipe or redirect the logs, create the relevant directories.
+	if conf.LogLocation != models.Console {
+		err := CreateNestedDirectoriesIfNotExists(conf.LogFileLocation)
+		if err != nil {
+			fmt.Println("Could not create nested directories for filepath due to error %v", err)
```
re-Tick, line 101: Please use log.Fatal to log error

re-Tick:
There is a unit test in regression service which calls the New function of regression.go. As you have added some parameters to New function, the unit test build is failing.

server/server.go
aerowisca: The reason for using `fmt.Print` instead of `logger.Print` was that the logger wasn't created by that point.

Sure, we can temporarily use inbuilt `log` package while our logger is initialized. Updated.

aerowisca:
@re-Tick  Sorry, my bad. Forgot to update the test case files. Added them now.

aerowisca:
@re-Tick  Can you add the required labels for GSoC if eveything looks good to you?

re-Tick:
LGTM.

SanskritiHarmukh:
Hi @aerowisca can you add the PR link in [gsoc/2023 repo](https://github.com/keploy/gsoc/tree/main/2023#1-keploy-cli-refactoring) corresponding to the [issue](https://github.com/keploy/keploy/issues/444) making a column on your username under CLI Refactoring project?

Sarthak160:
This PR is stale since we migrated to V2 which has newer code structure .

PR #487, opened 2023-03-20
feat: Add keploy version, os in telemetry events

charankamarapu:

Add keploy version, os in telemetry events :
1. Added OS and KeployVersion attributes in Telemetry - [https://github.com/keploy/telemetry/pull/13](PR)
2. Added OS and KeployVersion telemetry.go 
3. Added version variable in telemetry object from main
4. Added Default version from executing git command , so that when Keploy is run by cloning the repository we can know the release where user has cloned Keploy.

## Type of change

## Checklist:

## Screenshots (if any)

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -1,15 +1,29 @@
 package main
 
 import (
+	"os/exec"
+	"strings"
+
 	"go.keploy.io/server/server"
 )
 
 // version is the version of the server and will be injected during build by ldflags
 // see https://goreleaser.com/customization/build/
 
-var version = server.DefaultVersion
+var version = getKeployVersion()
 
 func main() {
 	// main method to start Keploy server
 	server.Server(version)
+}
+
+func getKeployVersion() string {
+	cmd := exec.Command("git", "describe", "--tags")
+	out, err := cmd.Output()
+	if err != nil {
+		return "0.1.0-dev"
+	}
+	version := strings.TrimSpace(string(out))
+	version = version[0:6] + "-dev"
+	return version
 }
--- a/pkg/models/event.go
+++ b/pkg/models/event.go
@@ -6,4 +6,6 @@ type TeleEvent struct {
 	Meta           map[string]interface{} `json:"meta"`
 	CreatedAt      int64                  `json:"createdAt"`
 	TeleCheck      bool                   `json:"tele_check"`
+	OS             string                 `json:"os"`
+	KeployVersion string                  `json:"keploy_version"`
 }
--- a/pkg/platform/telemetry/telemetry.go
+++ b/pkg/platform/telemetry/telemetry.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"net/http"
+	"runtime"
 	"time"
 
 	// "go.mongodb.org/mongo-driver/bson"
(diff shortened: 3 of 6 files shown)


cmd/server/main.go
```diff
@@ -1,15 +1,29 @@
...
-var version = server.DefaultVersion
+var version = getKeployVersion()
 
 func main() {
 	// main method to start Keploy server
 	server.Server(version)
 }
+
+func getKeployVersion() string {
+	cmd := exec.Command("git", "describe", "--tags")
```
aerowisca, line 21: I don't think this is correct, for the following reasons:
1. It assumes that the user has `git` installed on their system (Adds unnecessary dependency on the statically linked binary).
2. Even if `git` were to be installed, it assumes that the user would always run the binary from the `keploy` repo? What if they don't even have the repo locally. (Running this command from outside a repo throws an error, conflicting with the idea of a binary being added to path so that it can be run from "anywhere").
3. Even if they do have the repo, this would produce misleading results. It only shows the version of the repo that you are currently in, not the version on which the binary was built.

The correct version would only be known by the person who built and distributed the binary. Hence, to retrieve this version, we need to embed it in the binary (via `ldflags`) during build phase and then retrieve it using runtime/debug pacakge.

A sample might look like

```bash
CGO_ENABLED=0 GOOS=linux go build -ldflags="version=$(git describe --tags)" -o keploy cmd/server/main.go
```

And to retrieve the version in code, we can use
```go
info, _ := debug.ReadBuildInfo()

// Iterate over info.Setting and retrieve the "value" of key="ldflags". Then split it via the "=" separator,
// and look for the value corresponding to "version". This is the version that we are looking for.
```

This would work out of the box for non-docker binaries. Of course, if you want to inject the version during build phase in docker, you might need to temporarily add the git repo inside the docker container (and run `git describe` and pass an environment variable to docker).

charankamarapu: 1. This change is particularly for people who clone the repo and runs Keploy. The user whoever clones I guess they have git in their system .
2. For Binary , version is already loaded by Keploy . When someone downloads binary,  version variable is loaded using ldflags. But when user clones the repo and runs Keploy we don't have a way to deal with the version or know the version. So that's why git command is added . 
3. . For the 3rd point actually that command outputs the tag for the keploy repository i.e nothing but build version tag that was added to it. We have added "-dev" if you see, that is to tell that the user has cloned the repo when this xxx-dev was released but there can be changes which are not yet released .   

Does this solves you question or is there anything you feel needs to be changed.

charankamarapu: Good Catch ! corrected it and tested.

re-Tick: Hey @charankamarapu, can we use [this](https://github.com/go-git/go-git) library here in place of git cmd.

cmd/server/main.go
```diff
@@ -1,15 +1,32 @@
...
+	out, err := cmd.Output()
+	if err != nil {
+		return "0.1.0-dev"
+	}
+	version := strings.TrimSpace(string(out))
+	version = version[0:6] + "-dev"
```
aerowisca, line 30: This would cause segmentation fault if something goes wrong, instead of seg fault, we can add graceful failure handling (like only accessing `version[0:6]` when the length is atleast "x". In other cases, we default to some value

aerowisca (commented):
LGTM. (Feel free to explore the alternative approach suggested by @re-Tick )

cmd/server/main.go
charankamarapu: Hey @re-Tick thanks for the suggestion. Used go-git library instead git command.

cmd/server/main.go
charankamarapu: This issue will get resolved as go-get is used now.

re-Tick (approved):
LGTM

PR #492, opened 2023-03-22
Beautiful diff output

cassiozareck:
## Related Issue
  - https://github.com/keploy/keploy/issues/334

Ive updated the output of diffs in the command line after running tests cases. It can match the difference between the the body response, headers and status by coloring where they start to not match. It uses box cli maker to align texts and use borders. 
Also I added the DiffBox utility function in the utils.go file and an algorithm to check where the strings starts to become unmatched and add colors to it (I used ascii codes for it).

## Type of change

NA

## Checklist:

## Screenshots (if any)

| original | updated  |


SanskritiHarmukh:
@nehagup can you please review this PR?

nehagup (commented):
while this will look good incase when differences are small, how about the cases when there's a whole json object diff or 2,3 elements of an array are diff?

 Assume there are more than 10 cases and the differences are displayed like you showed, it'll be a huge log pipe. How about we showcase the expected and actual test-response side-by-side like popular code diff checkers does(like we also see on GitHub)? IMO that'll reduce the length of the diff-logs to almost half. WDYT?

cassiozareck:
Hmm makes sense. Of course we can do it. Ill change the package as this package doesnt handle ascii colors very well and its almost impossible to make it run diffs side-by-side. Go-pretty seems a good choice for it as its table-based (what means we can make diffs side-by-side) and as I saw it can handle ascii colors very well and its flexible what mean we can keep improving UI in future pull requests without need to change packages.
Im thinking in trunk it someway if the chars length of the response > 500 as nobody would read an entire expect/response diff too large. Do you think this idea is good?

nehagup:
> Hi @nehagup! Hmm that makes sense. Of course we can do it. Ill change the package as this package doesnt handle ascii colors very well and its almost impossible to make it run diffs side-by-side. Go-pretty seems a good choice for it as its table-based (what means we can make diffs side-by-side) and as I saw it can handle ascii colors very well and its flexible what mean we can keep improving UI in future pull requests without need to change packages.

One way to showcase this could look like this - 
 
Screenshot 2023-03-28 at 12 17 18 PM



> Im thinking in trunk the diffs someway if the string length of the response > 500 as nobody would read an entire expect/response diff too large. Do you think this idea is good?

Yea, that makes sense. We can maybe filter out the test cases with body more than X characters/lines and just provide a _`See More`_  link for those. WDYT?

nehagup:
I generated this from ChatGPT to produce the above output. You can maybe make changes to it to ignore noisy fields marked by keploy in the diff. 

```
package main

import (
	"encoding/json"
	"fmt"
	"strings"
)

// Define a function to compare two JSON objects recursively and find the differences
func compareJSON(path string, obj1 interface{}, obj2 interface{}, diff *[]string) {
	switch obj1.(type) {
	case map[string]interface{}:
		for key, val1 := range obj1.(map[string]interface{}) {
			val2, exists := obj2.(map[string]interface{})[key]
			if !exists {
				diffStr := fmt.Sprintf("- %s.%s: %v", path, key, val1)
				*diff = append(*diff, diffStr)
			} else {
				compareJSON(fmt.Sprintf("%s.%s", path, key), val1, val2, diff)
			}
		}
		for key, val2 := range obj2.(map[string]interface{}) {
			_, exists := obj1.(map[string]interface{})[key]
			if !exists {
				diffStr := fmt.Sprintf("+ %s.%s: %v", path, key, val2)
				*diff = append(*diff, diffStr)
			}
		}
	case []interface{}:
		for i, val1 := range obj1.([]interface{}) {
			val2 := obj2.([]interface{})[i]
			compareJSON(fmt.Sprintf("%s[%d]", path, i), val1, val2, diff)
		}
	default:
		if obj1 != obj2 {
			diffStr := fmt.Sprintf("- %s: %v", path, obj1)
			*diff = append(*diff, diffStr)
			diffStr = fmt.Sprintf("+ %s: %v", path, obj2)
			*diff = append(*diff, diffStr)
		}
	}
}

// Define a function to format the JSON with differences highlighted
func formatJSON(json string, diff []string, color string) string {
	var obj interface{}
	json.Unmarshal([]byte(json), &obj)
	formatted, _ := json.MarshalIndent(obj, "", "  ")
	lines := make([]string, 0)
	for _, line := range strings.Split(string(formatted), "\n") {
		if len(line) == 0 {
			continue
		}
		for _, d := range diff {
			if strings.Contains(line, strings.TrimPrefix(d, "- ")) {
				line = strings.Replace(line, strings.TrimPrefix(d, "- "), d, 1)
				line = colorize(line, color)
			}
			if strings.Contains(line, strings.TrimPrefix(d, "+ ")) {
				line = strings.Replace(line, strings.TrimPrefix(d, "+ "), d, 1)
				line = colorize(line, color)
			}
		}
		lines = append(lines, line)
	}
	return strings.Join(lines, "\n")
}

// Define a function to colorize text
func colorize(text string, color string) string {
	colorCode := ""
	switch color {
	case "green":
		colorCode = "\033[32m"
	case "red":
		colorCode = "\033[31m"
	default:
		colorCode = ""
	}
	reset := "\033[0m"
	return colorCode + text + reset
}

// Define the main function
func main() {
	// Define the two JSON strings to compare
	json1 := `{
		"name": "John",
		"age": 30,
		"city": "New York",
		"pets": ["dog", "cat"]
	}`

	json2 := `{
	"name": "Jane",
	"age": 35,
	"city": "San Francisco",
	"pets": ["dog", "bird"],
	"spouse": "John"
}`

// Parse the JSON strings into maps
var obj1, obj2 interface{}
json.Unmarshal([]byte(json1), &obj1)
json.Unmarshal([]byte(json2), &obj2)

// Compare the JSON objects and get the differences
diff := make([]string, 0)
compareJSON("", obj1, obj2, &diff)

// Format the JSON strings with differences highlighted
formatted1 := formatJSON(json1, diff, "red")
formatted2 := formatJSON(json2, diff, "green")

// Print the formatted JSON strings side by side
lines1 := strings.Split(formatted1, "\n")
lines2 := strings.Split(formatted2, "\n")
for i := 0; i < len(lines1) || i < len(lines2); i++ {
	if i < len(lines1) {
		fmt.Printf("%-50s", lines1[i])
	} else {
		fmt.Printf("%-50s", "")
	}
	if i < len(lines2) {
		fmt.Printf("%s\n", lines2[i])
	} else {
		fmt.Printf("\n")
	}
}
}
```

achintya-7:
This looks good. With some changes and using the `gojsondiff` package, I have got to something like this. 
image
We can parse in the string generated by the current error finder or can even implement a new one.

achintya-7:
@nehagup  How about this....
image
It does require some work on formatting but it got the feel of it.

nehagup:
@achintya-7 looks interesting, another optimisation could be instead of highlighting the complete line we can maybe highlight just the word/character change? and we put the formatting such that the similar fields are side by side. 

> @nehagup How about this....
> image
> It does require some work on formatting but it got the feel of it.

achintya-7:
Ok, I'll try to implement it in.

cassiozareck:
Hey @nehagup I've update the diffs now they are very beautiful and support json and generics strings. Have a look:

## Json-based:
Screenshot from 2023-04-09 23-18-41


## Generic string:
Screenshot from 2023-04-09 23-21-08

nehagup (commented):
@cassiozareck this looks very clean!  I have 2 questions: 

1. Why is it highlighting the whole JSON component and not just the field which is different? It took me some time to figure out which part of the coloured body is exactly different. 
2. Have you tried it with a sample in which there are keploy highlighted noisy fields (for example timestamps, random ids..). The diff viewer should ignore the noisy field differences and should not colour those.

cassiozareck:
Hmm sure I'll do it. Yea makes more sense to just highlight the exact diff. The second question seems a bit more complicated but I'll study how to implement it

cassiozareck:
Hi @nehagup I've updated the PR again and now the output can relate more the fields like body, status and header into a single test-case doing the appropriate formatting. Like the image below:
Screenshot from 2023-04-15 14-57-13

Also now it can handle noise.

Its important to mention that for large jsons input and lot of test-cases the output may get messed or mixed. Its not output-related bug. In fact I've tested it with even larger jsons  and tons of tests in a standalone package created by mine and it was not happening.

What I suppose is that there is a problem of race conditions where tests are being running asynchronously and one ends up messing the stdout of another. Anyway a workaround I found for now was to put a small delay before the Render() methods runs so it can output properly.

Sarthak160:
Hey @cassiozareck , logging the diff table looks awesome . Can you also remove the previous diff log from the code base and use mutex to handle race condition running tests while logging in the CLI .

cassiozareck:
@Sarthak160 good idea! I'm on It.

cassiozareck:
Hello devs. I wanted to give you an update on this PR. Unfortunately, I discovered that the package I used relies on generics, which are just supported on go version 1.8. I've searched for alternative packages, but they all seem to require generic support as well. (go-pretty: Screenshot from 2023-04-22 19-50-53). The problem is in the uniseg package that uses generics and a lot of packages rely on it.
Im really sorry for that, it was mine mistake because I ended up running this in go 1.20 instead of go 1.17. Given the current situation, we have a few options: 
1- Wait till Keploy gets an upgrade to use 1.18 go version. 
2- Try to implement ourselves the side-by-side diffs renderer (this would require a lot of work and I cannot guarantee it will have the same quality as using the tablewriter package that is very customizable). 
3- We also have a third option that is to try vendorize the table-writer package so I can try to modify functions that call the runewidth package (that by itself calls the uniseg package). 
Those lasts 2 options is a bit risk but if you want I can give a try however it would require more time to implement them

aerowisca:
I don't think option (1) is feasible. We were already on Go 1.18, and we purposefully reverted to Go 1.16 for backward compatibility.

cassiozareck:
@aerowisca Yeah I remembered right after I go to bed it was tworking before even with the build action. Yea the thing was the downgrade that was made a few days. I was not aware of any plans to downgrade the go version  I couldn't have predicted this change. Generics seems important in some packages. But I'll wait a mentor say their thoughts on it

aerowisca:
Tagging @slayerjain  for visibility.

Sarthak160:
@cassiozareck Have you tried by using older version of your diff viewer library which supports Go 1.16 , If it works the same way we are good to use older version as well .

cassiozareck:
Hi @Sarthak160. Good news! it seems there's a package that can help us, it does run on go 1.16! Give me a bit of time I'll implement it

Sarthak160 (approved):
LGTM

Issue #448, opened 2023-03-06
[feature]: Define basic library interfaces for the JS/Go DSL


slayerjain:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

Once we have a format for the data definitions and are able to read and parse it using the keploy server (#447), we need to define some basic interface to define assertions and other interactions with the keploy server. 

### Why should this be worked on?

This is where we would start seeing the advantages of using a language-based DSL. The library would make various customisations easy and transparent.

Arnab7456:
can i start working on it

nehagup:
@Arnab7456 It's a GSoC task. Anyone can contribute and make a PR.

sayantan135:
can i work on it ?

nehagup:
@sayantan135 It's a GSoC task. Anyone can contribute and make a PR.

charankamarapu:
Issue is irrelevant as keploy v1 is no more supported.

Issue #447, opened 2023-03-06
[feature]: add support for reading js or go based testcases 


slayerjain:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

Defining a js or go-based test case format alternative to the current keploy YAML would be good. We can take inspiration from `k6.io` or `rest assured`. We can also explore leveraging an existing interpreter for example from k6s and just create a js based interpreter for tests. 

### Why should this be worked on?

JS or Go-based tests would be more flexible than YAML-based tests. It can make adding custom assertions easier.

Frey0-0:
Hello @slayerjain @SanskritiHarmukh  I find this issue interesting and would like to work on it. I have used k6.io once before and I feel the JS-based test cases would be more readable personally. For the approach,h I would prefer to use JS as I have less experience in Go but considering that whole of keploy is in Go what should it be ideally? And also we will need to write a parser right?
I am exploring the code to understand how a yaml file is processed to maybe use the same methods. I would love to hear your opinion on this :)  if there are any resources that can help.

charankamarapu:
Issue is irrelevant as keploy v1 is no more supported.

PR #494, opened 2023-03-22
added a dsl for keploy and a way to parse it .

Hitenjain14:
## Related Issue
  - #447 #448 

Used `HCL` to create a DSL for `keploy` . HCL already has an interpreter for go , so using that we can extend the language and add the required  functionality to it . In this I have created a simple library to be able to make get and post requests using HCL . If this idea is liked , I will start adding more functionality like assertion for the DSL . I have created a [test.hcl](https://github.com/Hitenjain14/keploy/blob/main/dsl/test.hcl) file and explained the working of it in [readme](https://github.com/Hitenjain14/keploy/blob/main/dsl/README.md)

## Type of change

`NA`


Hitenjain14:

Hitenjain14:
recheck

slayerjain:
@Hitenjain14 Thanks for creating the PR. The DSL actually looks very good, but what I meant was a DSL based on a general purpose programming language like go or js. We already have a YAML based custom DSL defined by Keploy.

The core problem with HCL or YAML based DSL is that it's hard to add customisation, but instead if we generate js or go code, the user would be able to potentially use existing libraries in go or js to add better assertions, logs, etc.

We would need to generate go/js code instead of YAML, and that code should be using interfaces that allows users to change the assertions while still allowing Keploy to execute it. The closest example for this would be the js code used in [k6s](https://k6.io) scripts

Hitenjain14:
@slayerjain I was taking inspiration from [rest-assured](https://rest-assured.io/) and [karate dsl](https://github.com/karatelabs/karate) by using something like this , the tests become more readable and easily customizable , HCL already has alot of types built into it and we can extend that to build something like karate and rest-assured. HCL is also a very easy language to understand so no related programming knowledge will be required to read or add tests to it .

slayerjain:
Doesn't restassured use java instead of defining an independent DSL? Karate does have a DSL. But to add any custom assertions we would need to specifically define the syntax and add support. 

The benefit of using just go or js is the extra power the user gets to customise or extend functionalities without too much involvement from Keploy project.

slayerjain:
@Hitenjain14 yes they don’t have universal support because they don’t actually run node.js (although they do support external libraries via http as per this - https://k6.io/docs/using-k6/modules/). For most usecases having simple js support by using a js interpreter like https://github.com/robertkrimen/otto . While this may not give complete npm support, it’s still better than custom DSL (because user can write code and we can enable support for some popular libraries). 

If universal supports becomes important, we could actually running it from nodejs and then the library could communicate with the keploy go code either via http/rpc or standard shared object binary file (.so) route.

Hitenjain14:
@slayerjain when i tried to parse js using [https://github.com/robertkrimen/otto ](https://github.com/robertkrimen/otto) , it throws an error while parsing this 
```
 let req = {
	body: {
	   name: 'John',
	   },
	  };
	  console.log(req)  
```

Seems like it is not expecting `let` or `const` as it is able to run this 
```
req = {
	body: {
	   name: 'John',
	   },
	  };
	  console.log(req)  
```
It seems like k6 have their own implementation of parser with the methods they have defined . If we have to support libraries we can do it in dsl too , check this out [orion](https://ivan-corrales-solera.medium.com/orion-a-next-generation-automation-testing-tool-4ea53eeb2517)  , this is similar to karate and we can extend and  implement this for keploy . Meanwhile I will keep exploring more options to integrate tests in js .

slayerjain:
@Hitenjain14 thanks for sharing. yes, it looks like `otto ` only support es5 and not es6 - https://github.com/robertkrimen/otto/issues/406

I don't think that should be too much of a problem. The example I shared above was just an example and not the exact DSL, so feel free to design  more compact and readable interfaces.

Orion looks interesting, I'll explore more and share my thoughts.

slayerjain:
I explored Orion and yes it seems to be similar to karate. 

I strongly feel that making a library instead of a DSL would be better long term for developers. You maybe not need to really parse the js file. You just need to run it and call Keploy methods from the library.

Sarthak160:
This PR is outdated as we have migrated our main branch acc to KeployV2 !

Issue #500, opened 2023-03-23
[bug]: Keploy Fails in testing mode 


Sonichigo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When I run command's
```
export KEPLOY_MODE=test 
go test -coverpkg=./... -covermode=atomic  ./...
```
Either the test case fail or showcase connection timed out... This issue is coming with latest pull with Echo-Sql Sample-go application, also even though there are only 2 testcases made, it shows 4 tests.

![image](https://user-images.githubusercontent.com/53110238/227153850-a896f4f9-9d49-4f74-974e-7ad0a03b7627.png)

![image](https://user-images.githubusercontent.com/53110238/227154085-0c85e64c-bff4-44eb-9996-c53804cd6d54.png)

### Steps to reproduce

1. Run the Keploy in record mode and creates the testcases
2. Start the Keploy Server in test mode.
3. Run the command ```go test -coverpkg=./... -covermode=atomic  ./...```
4. Error is visible on both Keploy CLI and Terminal

### Environment

None

### Version

Local

PR #503, opened 2023-03-24
Fix grpc

re-Tick:
## Related Issue
  -  Changes in the GetTCS of `v0.8.2` returns testcases irrespective due to which go-sdk versions before `0.8.2` fetches the same set of tcs twice resulting in the error `unlock of unlocked mutex`. 
  
Closes: #500
Makes the `reqType` query param as optional. This will return testcases of same type when `reqType` is not empty and all tcs if `reqType` param is empty
## Type of change

## Checklist:

## Screenshots (if any)


aerowisca:
Why are switching to a generic string for `tcsType` when we already made the effort to switch to a more stable and predictable [models.Kind](https://github.com/keploy/keploy/blob/bb7dd253b5c90056a1d145835d59094a8dd30d00/pkg/models/mock.go#L22)? IMO, it's better to modify the signature to accept a `models.Kind` and in our handler, we do a sanity test on whether the given request parameter fits into one of the known models.

Also, we don't need to do the filtering after unnecessarily collecting a large number of results and sorting them. You can instead modify the function signature for [toTestCase](https://github.com/keploy/keploy/blob/bb7dd253b5c90056a1d145835d59094a8dd30d00/pkg/platform/fs/mock.go#L151) so that results are [filtered](https://github.com/keploy/keploy/blob/bb7dd253b5c90056a1d145835d59094a8dd30d00/pkg/platform/fs/mock.go#L69) even before they are sorted.

re-Tick:
These changes are made to maintain the backward compatibility with older go-sdk versions. 
 
 That can be good but since, the reqType is used to filter the array of testcases only, it is not needed to sanitise.

Sarthak160 (approved):
LGTM 😇

slayerjain:
LGTM

Issue #726, opened 2023-08-21
[feature]: Add manual tests/ mock Library feature to keploy v2


charankamarapu:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

In Keploy v1 there is a feature which allows users to record studs and use it in unit test , similar feature should be added to keploy v2 .

### Why should this be worked on?

Feature Parity

### Repository

keploy

charankamarapu:
PR - [727](https://github.com/keploy/keploy/pull/727) , [192](https://github.com/keploy/go-sdk/pull/192)  resolves this issue

PR #727, opened 2023-08-21
Add support for recording stubs in unit tests

charankamarapu:
## Related Issue
 Add support for recording stubs using keploy v2 for using in unit tests

Closes: #726 

1. decoupled path from hooks and attached into yaml .
2. Remove config.yaml and added them in mocks.yaml.
3. create two commands :
   sub cmd : mockRecord  flags : path, mockName, pid
   sub cmd : mockTest  flags : path, mockName, pid

In both the commands pid of the running unit test is given and accordingly other flag are given by the user . In mockRecord we take in mockName name, rather than randomly creating one session and we store only mocks over there. tests will not be created as unit tests don't hit handlers they hit the handler functions. Similarly, for mockTest mockName name is taken and only those mocks are loaded.

In this way we can support recording stubs in unit tests

## Type of change

Tested using gin-mongo app details are provided in go-samples [PR](https://github.com/keploy/go-sdk/pull/194) 

## Checklist:

## Screenshots (if any)

### DIFF
--- a/pkg/platform/interface.go
+++ b/pkg/platform/interface.go
@@ -3,11 +3,8 @@ package platform
 import "go.keploy.io/server/pkg/models"
 
 type TestCaseDB interface {
-	WriteTestcase(path string, tc *models.TestCase) error
-	WriteMock(path string, tc *models.Mock) error
-
-	NewSessionIndex(path string) (string, error)
-	ReadSessionIndices(path string) ([]string, error)
+	WriteTestcase(tc *models.TestCase) error
+	WriteMock(tc *models.Mock) error
 
 	ReadTestcase(path string, options interface{}) ([]*models.TestCase, error)
 	ReadMocks(path string) ([]*models.Mock, []*models.Mock, error)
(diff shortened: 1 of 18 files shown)


pkg/platform/interface.go
```diff
@@ -3,11 +3,8 @@ package platform
 import "go.keploy.io/server/pkg/models"
 
 type TestCaseDB interface {
-	WriteTestcase(path string, tc *models.TestCase) error
-	WriteMock(path string, tc *models.Mock) error
-
-	NewSessionIndex(path string) (string, error)
-	ReadSessionIndices(path string) ([]string, error)
```
re-Tick, line 10: I think we should have ReadSessionIndices and NewSessionIndex in the interface because implementation would be different for each type of data store.

charankamarapu: I have removed them because I didn't see them depending on any other elements of the struct other than path which is being passed in the function . I think sessionIndices are more independent from writing mocks and tests although it can be used to be a part of path .

charankamarapu: I have also made a recent change 

Removed config.yaml and added them in mocks.yaml.

please review this change too

re-Tick (approved):
LGTM

charankamarapu:
Thanks @re-Tick

Issue #750, opened 2023-09-02
[bug]:  Keploy fails in the test mode when expect header is involved


PranshuSrivastava:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When I capture a testcase with Keploy that uses the expect header, it breaks in the test mode in the decode outgoing function because the function is not able to decode the body of the request.

### Steps to reproduce

1. Make an application that uploads some data to s3.
2. Start the application using Keploy.
3. Record some testcases for the application.
4. Test the application using Keploy

### Environment

None

### Version

Latest Binary

### Repository

keploy

PR #751, opened 2023-09-02
fix: added support for expect header in the test mode

PranshuSrivastava:
## Related Issue
When we capture a testcase with Keploy that uses the expect header, it breaks in the test mode in the decode outgoing function because the function is not able to decode the body of the request.

Closes: #750 

I added support for the expect header in the decodeOutgoingHttp function wherein , if the request contains the expect header, we write the 100-continue response to the user and then append the body in the request and send it back to the decodeOutgoingHttp function.

## Type of change

## Checklist:


gouravkrosx (approved):
LGTM

Issue #799, opened 2023-09-12
[bug]: Keploy doesn't kill child nodes of application server


shivamsouravjha:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When I run the node app with Keploy, the child node of npm isn't killed after the test phase. The pid of the application is killed not the child processes it started.

### Steps to reproduce

* Run Keploy test with Node app.
* View the port of the running node app.

### Environment

Production

### Version

Cloud

### Repository

keploy

PR #800, opened 2023-09-12
fix: cleanup all the child processes started by app

shivamsouravjha:
## Related Issue
  -Recursively kills the child process started by the application.

Closes: #799

Instead of just killing the PID of the application we kill the child processes started by the application as well.

# How to Kill the child processes?
* Approach 1: PGID-Based Termination
This method involves identifying and terminating all processes sharing the same Process Group ID (PGID) as the user application. While this approach can efficiently kill all processes within the same group simultaneously, it presents a significant risk: it may inadvertently terminate unrelated processes—possibly even the parent processes—since they share the same PGID as the target application. Consequently, this approach is not recommended as it does not guarantee the isolation of the user application's processes and may lead to unintended system disruptions.

* Approach 2: Recursive PID Fetching and Termination
This approach adopts a more refined strategy to terminate processes initiated by the user application. It starts by identifying the child processes directly spawned by the user application using its Process ID (PID), and then recursively identifying and storing the PIDs of all subsequent child processes. Once all relevant PIDs are accumulated, the method proceeds to terminate these processes in a parent-first sequence. This sequence is critical as it prevents the triggering of any potential cleanup mechanisms that might activate if a child process is terminated before its parent. This strategy ensures a safer and more controlled termination of processes directly associated with the user application, without affecting unrelated processes sharing the same PGID.

Conclusion Approach 2 is the most suitable approach.

StopUserApplication is being called twice while stopping loaded hooks in case of force stop and RunTestSet. 

By monitoring the ProcessState of the userAppCmd, we can ensure that the child process of an already killed process isn't called recursively as it would cause.

Findings for userAppCmd:
* Since wait is already being used we can't use wait for the process to finish in order to see the final process state to avoid going on the pre-killed child process.
* Move StopUserApplication inside the force-stop method, as during the test phase user application is already stopped while stopping loaded hooks.
 
## Type of change

## Checklist:

- [X] My code follows the style guidelines of this project.

### DIFF
--- a/pkg/hooks/loader.go
+++ b/pkg/hooks/loader.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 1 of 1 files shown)


gouravkrosx (approved):
LGTM

PR #865, opened 2023-09-22
fix: Checking for parent distro in case of failure

PranshuSrivastava:
## Related Issue
 Keploy often failed to add custom CA certificates to the trusted store of Linux distros. This is because the path for those distros was not found in the map. 

Now if the name of that distro is not found in the map, we check for the parent distro and use the path of that distro. Even if after all this, we fail to add the ca to the path, we output the name of the distro along with the error.

## Type of change

## Checklist:


pkg/proxy/proxy.go
```diff
@@ -125,14 +125,22 @@ func (c *Conn) Read(b []byte) (n int, err error) {
 func getDistroInfo() string {
 	osRelease, err := ioutil.ReadFile("/etc/os-release")
 	if err != nil {
-		fmt.Println(Emoji+"Error reading /etc/os-release:", err)
 		return ""
```
gouravkrosx, line 128: Add a Debug log before returning.

gouravkrosx (approved):
LGTM

Issue #901, opened 2023-09-27
[bug]: unprecedented Header for outgoing HTTP request 


shivamsouravjha:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

underscore is sent in http call header i.e, Keploy_test_id

### Steps to reproduce

Make an outgoing call in http request

### Environment

Production

### Version

Cloud

### Repository

keploy

PR #902, opened 2023-09-27
update:header set for outoging http call

shivamsouravjha:
## Related Issue
  - Updating header for outgoing HTTP calls

Closes: #901 

Using _ (underscore) or . (dot) in HTTP header fields is technically possible, but it's not a common or recommended practice. The naming conventions for HTTP header fields are generally to use hyphens (-) to separate words, like Content-Type or User-Agent.
# Considerations:
## Standards Compliance:
According to the [HTTP/1.1 specification (RFC 2616)](https://tools.ietf.org/html/rfc2616#section-4.2), header field names are case-insensitive and can contain any printable ASCII character except colon (:). However, the convention is to use hyphen-separated words in Camel-Case format.
## Compatibility:
Some web servers, proxies, and libraries may expect header field names to follow the common convention and may not handle underscores or dots well. For example, Nginx by default does not allow underscores in HTTP header field names.
Readability and Maintenance:
Using uncommon characters can lead to confusion and make maintenance more difficult, as developers might not expect to see them in header field names.
## Client Behavior:
 Some HTTP clients may have unexpected behavior when encountering uncommon characters in header field names.
# Conclusion:
It's best to stick to the commonly used naming convention for HTTP header fields, which is hyphen-separated words, like X-Request-ID, to ensure the broadest compatibility and readability.

## Type of change

## Checklist:

### DIFF
--- a/pkg/util.go
+++ b/pkg/util.go
@@ -69,7 +69,7 @@ func SimulateHttp(tc models.TestCase, logger *zap.Logger, apiTimeout uint64) (*m
 		return nil, err
 	}
 	req.Header = ToHttpHeader(tc.HttpReq.Header)
-	req.Header.Set("KEPLOY_TEST_ID", tc.Name)
+	req.Header.Set("KEPLOY-TEST-ID", tc.Name)
 	req.ProtoMajor = tc.HttpReq.ProtoMajor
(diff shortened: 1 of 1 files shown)


charankamarapu (approved):
ship it!

Issue #881, opened 2023-09-26
[feature]: Add curl section in the testcase yamls


charankamarapu:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

Lets say user has hit a requests a testcase is created . Lets say user is trying after quite a few days or lost the curl user was using . To reconstruct the testcase which user has created user wants to hit the request again but he needs to create the curl from scratch looking at the created testcase . This problem can also occur when user has too many curls or testcases. So it will be better if we add a curl section in the testcase where user can use this curl recreate that request in no time.

### Why should this be worked on?

For good user exeperince

### Repository

keploy

AkashKumar7902:
Hey @charankamarapu, what fields will the curl section contain ?

charankamarapu:
Hi @AkashKumar7902 , It will contain the curl command of the test case recorded

AkashKumar7902:
@charankamarapu is this looking good ?
![image](https://github.com/keploy/keploy/assets/91385321/346704cf-2e6a-40c8-8adc-2890318016d2)

AkashKumar7902:
Also, do you want me to remove the headers like Content-Length which are calculated when request is sent ?
Simply copy pasting the above curl is giving "failed to decode req" error because of these headers
![image](https://github.com/keploy/keploy/assets/91385321/b788e216-8445-4abe-ba41-1c82c1743285)

charankamarapu:
Yes you can remove those headers which are calculated at run time because we have saved the request which was sent by the user but we may not be sure that we got everything that user sent majorly headers like the content-length may vary . So you can remove the content length header and Great work!

PR #929, opened 2023-09-30
feat: add curl command in yaml

AkashKumar7902:
## Related Issue
  - Info about Issue or bug

Closes: #881 
I have added a new entry in HttpReq struct named `Curl` in which string representation of curl command will be stored. I have defined a function named `makeCurlCommand` to get the string representation of the curl command

## Type of change

N/A

## Checklist:

## Screenshots (if any)

| **image** | <b>image </b> |

### DIFF
--- a/pkg/util.go
+++ b/pkg/util.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
+	"regexp"
 	"io"
 	"math/rand"
 	"net/http"
@@ -137,11 +138,12 @@ func MakeCurlCommand(method string, url string, header map[string]string, body s
 	curl := fmt.Sprintf("curl --request %s \\\n", method)
 	curl = curl + fmt.Sprintf("  --url %s \\\n", url)
 	for k, v := range header {
-		if k == "Content-Length" {
-			continue
-		}
 		curl = curl + fmt.Sprintf("  --header '%s: %s' \\\n", k, v)
 	}
-	curl = curl + fmt.Sprintf("  --data '%s'", body)
+	if body != "" {
+		pattern := regexp.MustCompile(`\n\s+`)
+		body := pattern.ReplaceAllString(body, "\n")
+		curl = curl + fmt.Sprintf("  --data '%s'", body)
+	}
 	return curl
 }
--- a/pkg/util.go
+++ b/pkg/util.go
@@ -132,3 +132,16 @@ func GenerateRandomID() int {
 	id := rand.Intn(1000000000) // Adjust the range as needed
 	return id
 }
+
+func MakeCurlCommand(method string, url string, header map[string]string, body string) string {
+	curl := fmt.Sprintf("curl --request %s \\\n", method)
+	curl = curl + fmt.Sprintf("  --url %s \\\n", url)
+	for k, v := range header {
+		if k == "Content-Length" {
+			continue
+		}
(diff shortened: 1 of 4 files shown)


pkg/util.go
```diff
@@ -132,3 +132,16 @@ func GenerateRandomID() int {
 	id := rand.Intn(1000000000) // Adjust the range as needed
 	return id
 }
+
+func MakeCurlCommand(method string, url string, header map[string]string, body string) string {
+	curl := fmt.Sprintf("curl --request %s \\\n", method)
+	curl = curl + fmt.Sprintf("  --url %s \\\n", url)
+	for k, v := range header {
+		if k == "Content-Length" {
```
charankamarapu, line 140: Have you debugged why providing content-length in curl is throwing an error when you use the curl..?

pkg/util.go
```diff
@@ -132,3 +132,16 @@ func GenerateRandomID() int {
 	id := rand.Intn(1000000000) // Adjust the range as needed
 	return id
 }
+
+func MakeCurlCommand(method string, url string, header map[string]string, body string) string {
```
charankamarapu, line 136: Add checks for param and type of methods . Like if there is a body then only you need to add `--body` etc . These kind of checks.

pkg/util.go
```diff
@@ -132,3 +132,16 @@ func GenerateRandomID() int {
...
+	curl := fmt.Sprintf("curl --request %s \\\n", method)
```
charankamarapu, line 137: Add error handling.

charankamarapu (commented):
Please resolve these comments

pkg/util.go
AkashKumar7902: @charankamarapu I found the issue. It is happening because of the extra spaces before "url" field of --data.
with extra space             |  without extra space
:-------------------------:|:-------------------------:
image |  image

pkg/util.go
AkashKumar7902: @charankamarapu I have added a body check in makeCurlCommand like this :
image
any invalid method type would result in 400 error and it won't make a .yaml file.

pkg/util.go
charankamarapu: Ok then try to find out why there is an extra space and remove it. Add content-length header also in the the curl.

AkashKumar7902: the extra spaces is because curl command is part of req in .yaml. The only solution is to make curl command a top level field in .yaml. 
when selecting curl command, you can see that there will also be extra spaces :
image

charankamarapu: Oh I see, I guess keeping curl in top level field looks fine for me. Because curl section shouldn't be under spec, as spec majorly tells how keploy decoded its req and resp. Put the curl section before spec or may be at the last.

AkashKumar7902:
@charankamarapu At any level at least 4 spaces will also be present: 
image
Should i remove the content-length as i did earlier ?

charankamarapu:
> @charankamarapu At any level at least 4 spaces will always be present: image Should i remove the content-length as i did earlier ?

I guess there is no other way . Yes do it as eariler.

pkg/util.go
charankamarapu: Ok got it.

charankamarapu:
Hi @AkashKumar7902 are all changes done can I review the PR..?

charankamarapu (approved):
Ship it!

PR #1058, opened 2023-11-02
fix: updates the decode function for OpMsg section

re-Tick:
## Related Issue
  - Test fails for the mongo call which have have a substring containing ` }` in the sectionSingle of OpMsg like:
  ``` "errmsg":"Error in specification { background: true, name: \"merchant_id_text_url_text_hash_hashed\", key: { merchant_id: \"text\", url: \"text\", hash: \"hashed\" } } :: caused by :: Can''t use more than one index plugin for a single index." ```.
  Due to which keploy is not able to extract and encode into the message back to the binary Mongo wire message.

The decode function is not able to distinguish the closing braces of the string json and sectionSingle, so used the prefix and suffix in place of regex to extract the document correctly

## Type of change

Tested the changes with gin-mongo API. And the problamatic example of OpMsg in isolation

## Checklist:

### DIFF
--- a/pkg/proxy/integrations/mongoparser/mongoparser.go
+++ b/pkg/proxy/integrations/mongoparser/mongoparser.go
@@ -689,7 +689,7 @@ func compareOpMsgSection(expectedSection, actualSection string, logger *zap.Logg
 		return score
 	case strings.HasPrefix(expectedSection, "{ SectionSingle msg:"):
 		var expectedMsgsStr string
-		expectedMsgsStr, err := decodeOpMsgSectionSingle(actualSection)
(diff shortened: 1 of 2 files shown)


slayerjain:
@re-Tick would this same problem happen in other dependencies too if there are double quotes in strings?

slayerjain (approved):
LGTM!

re-Tick:
> @re-Tick would this same problem happen in other dependencies too if there are double quotes in strings?

No the regex is used for encoding/decoding the mongo bson document of the OpMsg sections, so this error is limited to mongo only.

Issue #1051, opened 2023-10-31
add automatic port change if the port is busy


EraKin575:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

since test and record mode run at same port, it will be beneficial to add a automatic port changing feature so both modes can be run at same time

### Why should this be worked on?

..

### Repository

keploy

EraKin575:
@charankamarapu can I have a hacktoberfest label on this?😁

charankamarapu:
Hi @EraKin575 we already have an option to change the port using --proxyport flag. So if some one wants to change the port in some mode like record or test they can change it using that flag . So I dont think there is any issue here. If you are satisfied with the answer we can close the issue.

EraKin575:
@charankamarapu I am actually talking about automatically changing port if the port is busy

charankamarapu:
Oh okay, then I guess we can take that up.  But you can change the description of the issue because if someone want to run test and record they can use proxyport flag and do that, so we are not solving that problem. We are solving the problem which mentioned in the above comment.

EraKin575:
Thanks @charankamarapu

nakulbh:
what is the default port that test and records run on ?

charankamarapu:
The default port is 16789

EraKin575:
@charankamarapu Please review on my PR

EraKin575:
thanks @AkashKumar7902

nakulbh:
@charankamarapu @EraKin575, I believe that simply incrementing the port won't resolve the issue. I've made some changes that I think are better suited to address this. Please review and share your thoughts on these modifications

charankamarapu:
I guess there is some misunderstanding here, before raising a PR please ask to assign the issue to you. Right now there are two PRs and according to hackboterfest we decided to go by first come first serve. So @EraKin575 Pr will be reviewed . @nakulbh @EraKin575 please ask to assign before working on the issue. @nakulbh you choose some other issue. If this issue is not resolved even after completely reviewing after multiple revisions then I will consider @nakulbh PR.

charankamarapu:
> thanks @AkashKumar7902

🤔

nakulbh:
@charankamarapu why my PR is closed

charankamarapu:
Like I mentioned above only one PR will be merged out of both which are submitted. PR submitted by @EraKin575 is about to merge. So I have to close yours.

PR #1066, opened 2023-11-03
added automatic portChange feature if current port is busy

EraKin575:
## Related Issue
  - Info about Issue or bug

Closes: #1051 
added a automatic port changing feature for test and record mode if current port is busy
image

## Type of change

## Checklist:

## Screenshots (if any)


nakulbh (commented):
I Things there is no better way than this to change port automatically, Looks good to me

AkashKumar7902:
What if the next port is in use by some other running service? A while loop with a random function to generate port would be better I believe

nakulbh:
> What if the next port is in use by some other running service? A while loop with a random function to generate port would be better I believe

it will loop over until it will find a free port

pkg/proxy/proxy.go
```diff
@@ -348,6 +348,9 @@ func BootProxy(logger *zap.Logger, opt Option, appCmd, appContainer string, pid
 	if opt.Port == 0 {
 		opt.Port = 16789
 	}
+	for !isPortAvailable(opt.Port) {
```
charankamarapu, line 351: Define upper and lower bounds. For loop running continuously is not preferred, define max attempts.

charankamarapu (commented):
Please address the comments

pkg/proxy/proxy.go
```diff
@@ -349,7 +349,19 @@ func BootProxy(logger *zap.Logger, opt Option, appCmd, appContainer string, pid
...
 	}
+	maxAttempts := 1000
+	attemptsDone:=0
 
+
+    if !isPortAvailable(opt.Port) {
+		for i := 1024; i <= 65535 && attemptsDone < maxAttempts; i++ {
+			if isPortAvailable(uint32(i)) {
+				opt.Port = uint32(i)
+				break
```
charankamarapu, line 360: If all ports are busy you should throw an error right ..? missed this in last review my bad.

charankamarapu (commented):
Please address the comment

pkg/proxy/proxy.go
```diff
@@ -349,7 +349,26 @@ func BootProxy(logger *zap.Logger, opt Option, appCmd, appContainer string, pid
...
 	}
+	maxAttempts := 1000
+	attemptsDone:=0
 
+
+    if !isPortAvailable(opt.Port) {
+		for i := 1024; i <= 65535 && attemptsDone < maxAttempts; i++ {
+			if isPortAvailable(uint32(i)) {
+				opt.Port = uint32(i)
+				logger.Info(Emoji+"Port %v is already in use, switching to next available port");
```
charankamarapu, line 360: I don't think we need emoji here it is already appended when we initialise logger and include the port number which you are switching to in the log.

pkg/proxy/proxy.go
```diff
@@ -349,7 +349,26 @@ func BootProxy(logger *zap.Logger, opt Option, appCmd, appContainer string, pid
...
+
+				break
+			}
+			attemptsDone++
+		}
+	}
+	
+	if maxAttempts<=attemptsDone{
+		logger.Error(Emoji+"Failed to find an available port to start proxy server")
```
charankamarapu, line 369: Remove the emoji.

pkg/proxy/proxy.go
```diff
@@ -349,7 +349,26 @@ func BootProxy(logger *zap.Logger, opt Option, appCmd, appContainer string, pid
...
+				opt.Port = uint32(i)
+				logger.Info(Emoji+"Port %v is already in use, switching to next available port");
+
+				break
+			}
+			attemptsDone++
+		}
+	}
+	
+	if maxAttempts<=attemptsDone{
```
charankamarapu, line 368: beautify the statement I guess there must be spaces before and after the <= symbol. There must be some extension in the IDE you can use it and make sure it is indentation is correct.

charankamarapu (commented):
Please address the comments, rest looks fine.

pkg/proxy/proxy.go
```diff
@@ -349,11 +349,30 @@ func BootProxy(logger *zap.Logger, opt Option, appCmd, appContainer string, pid
...
+	
+	if maxAttempts <= attemptsDone{
+		logger.Error("Failed to find an available port to start proxy server")
+		return nil
+	}
 	//IPv4
 	localIp4, err := util.GetLocalIPv4()
 	if err != nil {
-		log.Fatalln(Emoji+"Failed to get the local Ip4 address", err)
+		log.Fatalln("Failed to get the local Ip4 address", err)
```
charankamarapu, line 375: Why is it removed from here? This is not zap.logger, It is just log. Here we should use emoji.

charankamarapu (commented):
Please address the comment.

pkg/proxy/proxy.go
```diff
@@ -349,7 +348,26 @@ func BootProxy(logger *zap.Logger, opt Option, appCmd, appContainer string, pid
 	if opt.Port == 0 {
 		opt.Port = 16789
 	}
+	maxAttempts := 1000
+	attemptsDone:=0
+
 
+    if !isPortAvailable(opt.Port) {
+		logger.Info("Port is already in use, switching to next available port",zap.Uint32("port",opt.Port));
```
charankamarapu, line 356: `Port is already in use, switching to next available port {port : 16789}` this is how the logger looks when you run keploy. It gives a sense that you you found that port is already in use and you switching to port 16789 which is wrong. You can delete this log or correct it . Deleting this log makes more sense for me.

charankamarapu (approved):
ship it!

PR #1078, opened 2023-11-05
refactor: Refactored the code to make it more modular

PranshuSrivastava:
## Related Issue
 The code did not have many modules so it became difficult to use certain parts of it, and the same work needed to be repeated again and again.

Made the code more modular by dividing the code into two parts, the code which is supposed the same everywhere, and the code which is supposed to be different.

## Type of change

## Checklist:


pkg/service/record/record.go
```diff
@@ -19,38 +19,37 @@ import (
 var Emoji = "\U0001F430" + " Keploy:"
 
 type recorder struct {
-	logger *zap.Logger
+	Logger *zap.Logger
```
re-Tick, line 22: Why do we need to export logger, since it is called internally by the exported functions.

pkg/service/test/util.go
```diff
@@ -271,6 +298,11 @@ func filterTcsMocks(tc *models.TestCase, m []*models.Mock, logger *zap.Logger) [
 	}
 
 	for _, mock := range m {
+		if mock.Version == "api.keploy-enterprise.io/v1beta1" {
+			logger.Info("This mock was recorded using the the enterprise version, may not work properly with the open source version", zap.String("mock kind:", string(mock.Kind)))
```
re-Tick, line 302: Please remove the typo here, and can the info be logged once for a test-set.

pkg/service/test/service.go
```diff
@@ -11,5 +11,9 @@ import (
 
 type Tester interface {
 	Test(path string, proxyPort uint32, testReportPath string, appCmd string, testsets []string, appContainer, networkName string, Delay uint64, passThorughPorts []uint, apiTimeout uint64, noiseConfig map[string]interface{}) bool
-	RunTestSet(testSet, path, testReportPath, appCmd, appContainer, appNetwork string, delay uint64, pid uint32, ys platform.TestCaseDB, loadedHook *hooks.Hook, testReportfs yaml.TestReportFS, testRunChan chan string, apiTimeout uint64, ctx context.Context, noiseConfig map[string]interface{}) models.TestRunStatus
+	RunTestSet(testSet, path, testReportPath, appCmd, appContainer, appNetwork string, delay uint64, pid uint32, ys platform.TestCaseDB, loadedHook *hooks.Hook, testReportfs yaml.TestReportFS, testRunChan chan string, apiTimeout uint64, ctx context.Context, noiseConfig map[string]interface{})models.TestRunStatus
```
re-Tick, line 14: Here, since the parameters of this function is long. You can add an option struct for this function. And for the required fields can use a check for null values.

pkg/service/test/util.go
```diff
@@ -10,10 +11,37 @@ import (
...
+	SessionsMap              map[string]string
+	TestReportFS             *yaml.TestReport
+	Ctx                      context.Context
+	AbortStopHooksForcefully bool
+	Ps                       *proxy.ProxySet
+	ExitCmd                  chan bool
+	Ys                       platform.TestCaseDB
+	LoadedHooks              *hooks.Hook
+	AbortStopHooksInterrupt  chan bool
+	Ok                       bool
```
re-Tick, line 42: In the changes ok is used for returning false for errors in starting ap or loading hooks so, instead of boolean we can just return the error along with InitialiseTestReturn for the InitialiseTestReturn.

pkg/service/test/util.go
```diff
@@ -10,10 +11,37 @@ import (
...
+}
+
+type InitialiseTestReturn struct {
+	SessionsMap              map[string]string
+	TestReportFS             *yaml.TestReport
+	Ctx                      context.Context
+	AbortStopHooksForcefully bool
+	Ps                       *proxy.ProxySet
+	ExitCmd                  chan bool
+	Ys                       platform.TestCaseDB
```
re-Tick, line 39: Here since the fields are exported, you can use descriptive name or add comment for the field

pkg/service/mocktest/mocktest.go
```diff
@@ -21,13 +21,13 @@ import (
 var Emoji = "\U0001F430" + " Keploy:"
 
 type mockTester struct {
-	logger *zap.Logger
+	Logger *zap.Logger
```
re-Tick, line 24: If logger is not used externally, it could be unexported.

pkg/service/mockrecord/mockrecord.go
```diff
@@ -21,13 +21,13 @@ import (
 var Emoji = "\U0001F430" + " Keploy:"
 
 type mockRecorder struct {
-	logger *zap.Logger
+	Logger *zap.Logger
```
re-Tick, line 24: If it is not used by the external packages, this can be unexported.

pkg/proxy/proxy.go
```diff
@@ -55,6 +54,13 @@ func getNextID() int64 {
 	return atomic.AddInt64(&idCounter, 1)
 }
 
+type DepInterface interface {
```
re-Tick, line 57: Remove interface from the name of the type

cmd/record.go
```diff
@@ -19,7 +19,7 @@ func NewCmdRecord(logger *zap.Logger) *Record {
 	recorder := record.NewRecorder(logger)
 	return &Record{
 		recorder: recorder,
-		logger:   logger,
+		Logger:   logger,
```
re-Tick, line 22: If this is not used by the external packages, then it can be unexported.

pkg/hooks/connection/factory.go
```diff
@@ -98,6 +98,7 @@ func (factory *Factory) GetOrCreate(connectionID structs.ConnID) *Tracker {
 }
 
 func capture(db platform.TestCaseDB, req *http.Request, resp *http.Response, logger *zap.Logger, ctx context.Context, reqTimeTest time.Time, resTimeTest time.Time) {
+	V1Beta1 := models.GetVersion()
```
re-Tick, line 101: Here, you can use version as the variable name

pkg/models/testcase.go
```diff
@@ -4,11 +4,18 @@ type Kind string
 type BodyType string
 type Version string
 
-const (
-	V1Beta1 Version = Version("api.keploy.io/v1beta1")
-	V1Beta2 Version = Version("api.keploy.io/v1beta2")
+var (
+	V1Beta1 = Version("api.keploy.io/v1beta1")
```
re-Tick, line 8: This should be a constant and to store the current version, an un-exported global scoped variable.

pkg/proxy/integrations/mysqlparser/mysqlparser.go
```diff
@@ -643,7 +643,7 @@ func recordMySQLMessage(h *hooks.Hook, mysqlRequests []models.MySQLRequest, mysq
 			"responseOperation": responseOperation,
 		}
 		mysqlMock := &models.Mock{
-			Version: models.V1Beta2,
+			Version: models.V1Beta1,
```
re-Tick, line 646: Handle for the mysql server to register it as a dependency in the proxy.

pkg/service/record/record.go
PranshuSrivastava: reverted.

pkg/models/testcase.go
PranshuSrivastava: Done

pkg/hooks/connection/factory.go
PranshuSrivastava: Done

cmd/record.go
PranshuSrivastava: Done

pkg/proxy/proxy.go
PranshuSrivastava: Modified

pkg/service/mockrecord/mockrecord.go
PranshuSrivastava: done

pkg/service/mocktest/mocktest.go
PranshuSrivastava: done

pkg/service/test/util.go
PranshuSrivastava: done

pkg/service/test/service.go
PranshuSrivastava: don

pkg/service/test/util.go
PranshuSrivastava: Done

pkg/service/test/util.go
PranshuSrivastava: Change ps, ys and dIDE

re-Tick (approved):
LGTM

PR #1133, opened 2023-11-25
fix: refactor code

officialasishkumar:
## Related Issue
  - N/A

Closes: N/A

Refactor the code in record.go and test.go to eliminate if conditions 

if we just change the order of gettestconfig, we won't be needing the if conditions

## Type of change

NA

## Checklist:

## Screenshots (if any)

| **original screenshot** | <b>updated screenshot </b> |s

### DIFF
--- a/cmd/record.go
+++ b/cmd/record.go
@@ -49,25 +49,16 @@ func (t *Record) GetRecordConfig(path *string, proxyPort *uint32, appCmd *string
 	if err != nil {
 		return fmt.Errorf("failed to get the record config from config file due to error: %s", err)
 	}
-	if len(*path) == 0 {
-		*path = confRecord.Path
-	}
+
 	filters = confRecord.Filters
-	if *proxyPort == 0 {
-		*proxyPort = confRecord.ProxyPort
-	}
-	if *appCmd == "" {
-		*appCmd = confRecord.Command
-	}
-	if *appContainer == "" {
-		*appContainer = confRecord.ContainerName
-	}
-	if *networkName == "" {
-		*networkName = confRecord.NetworkName
-	}
-	if *Delay == 5 {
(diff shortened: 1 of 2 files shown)


AkashKumar7902:
Nice work! 
This also solves the loophole in  delay flag as well. (Earlier If the user passed 5 second in delay then getconfig function would override that 5 second as if(delay == 5) evaluates to true).

cmd/record.go
```diff
@@ -88,49 +79,61 @@ func (r *Record) GetCmd() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			isDockerCmd := len(os.Getenv("IS_DOCKER_CMD")) > 0
 
-			path, err := cmd.Flags().GetString("path")
+			var (
```
charankamarapu, line 82: what is significance of this code ? Declaring variables before assigning them to a value has its own advantage and declaring and assigning at the same time has its own advantage. Can you specify any advantage that your approach creates in this function.

cmd/record.go
```diff
@@ -49,25 +49,16 @@ func (t *Record) GetRecordConfig(path *string, proxyPort *uint32, appCmd *string
 	if err != nil {
 		return fmt.Errorf("failed to get the record config from config file due to error: %s", err)
 	}
-	if len(*path) == 0 {
-		*path = confRecord.Path
-	}
+
```
charankamarapu, line 52: This change is not needed. We have specifically did this because we wanted to give more preference to flag over config file. Like if we have mentioned -c flag in command line and mention appcmd in config file, Keploy should prefer cmd specified in flag over config file.

charankamarapu (commented):
Please address the comments

officialasishkumar:
Sorry, I misunderstood.

Issue #1170, opened 2023-12-04
[bug]:  Update error concat with Keploy emoji


shivamsouravjha:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

In case of setting custom error at many places the custom error is created in ``` fmt.Errorf(Emoji + "failed to get Resolver")``` format. This format is incorrect and thus logs unexpected log.

### Steps to reproduce

Fix the concat across the repo.

### Environment

Production

### Version

Cloud

### Repository

keploy

shivamsouravjha:
solved in https://github.com/keploy/keploy/pull/1248

PR #1183, opened 2023-12-06
fixed error concat with keploy emoji through %s

jaimin001:
## Related Issue

Closes: #1170 

The issue was using `fmt.Errorf` incorrectly by directly concatenating an emoji with a string. The fix involves using a format string with %s to properly insert the emoji. Example: `fmt.Errorf("%s failed to get Resolver", Emoji)`.

## Type of change

`NA`
## Checklist:


shivamsouravjha:
Hey @jaimin001 thanks for raising the PR.

A suggestion for a better contribution would be to start with asking someone to assign you the issue,then create the PR, as to avoid multiple folks working on same issue.

Also ask for the review of the person who created the issue.

jaimin001:
Hey @shivamsouravjha, 
I'm sorry for not asking for the assignment of the issue before working on the PR. Could you kindly review my PR so that I can receive feedback and make necessary changes? Thank you.

jaimin001:
Hey @shivamsouravjha,
Sorry to bother you but, could you please have look at the PR so that I can make necessary changes for the issue you created?

pkg/hooks/launch.go
```diff
@@ -45,10 +45,10 @@ func (h *Hook) LaunchUserApplication(appCmd, appContainer, appNetwork string, De
 
 	if appCmd == "" {
 		if len(appContainer) == 0 {
-			return fmt.Errorf(Emoji + "please provide container name when running application container in isolation")
+			return fmt.Errorf("%s please provide container name when running application container in isolation", Emoji)
```
shivamsouravjha, line 48: incase adding the string like here replacement is not required. Like this log won't generate the error

shivamsouravjha:
Hi @jaimin001,

My apologies for the delayed reply. Please go ahead and take a fresh pull from the main branch, and then update your PR accordingly. Thank you for your patience and understanding. I'm looking forward to seeing the updated changes.

Best regards

pkg/hooks/launch.go
jaimin001: Yes, I understand that here, it won't generate the error, but I think even though `Emoji + "` is syntactically correct, it can lead to issues if the Emoji variable is empty or not a string. It also makes the code less readable.
That's why I made these changes also.

So should I make changes `fmt.Errorf(Emoji + "failed to get Resolver")` only?

shivamsouravjha: That's a great approach @jaimin001 but the emoji in this context is always a string that has a defined value. Thus keeping it to the original syntax is the better thing to do.

jaimin001:
Hey @shivamsouravjha , I have made some small changes as you suggested, please let me know of something else is needed?

pkg/hooks/launch.go
shivamsouravjha: Seems there had been some confusion @jaimin001. What I meant was to replace the ```fmt.Errorof(Emoji, ``` and not the addition of string.

jaimin001:
> Hey @jaimin001 are you working on this?

Yes, I am working

Issue #1188, opened 2023-12-11
[bug]: unable to record custom name test cases


gouravkrosx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Keploy has a feature to rename the test cases by adding a custom header `Keploy-Test-Name:<your test case name>`. Right now when I try to record subsequent request, it is throwing the below error.
```
🐰 Keploy: 2023-12-11T10:03:58Z         ERROR   yaml/yaml.go:67 failed to decode the last sequence number from yaml test        {"for the file": "test my case.yaml", "at path": "/workspaces/zpi/keploy/test-set-0/tests"}
🐰 Keploy: 2023-12-11T10:03:58Z         ERROR   connection/factory.go:140       failed to record the ingress requests   {"error": "failed to decode the last sequence number from yaml test"}
```

### Steps to reproduce

1. Setup keploy locally
2. Record a custom name test case for your application by adding `Keploy-Test-Name:<your test case name>`.
3. Again try to record a new subsequent test case.

### Environment

None

### Version

Local

### Repository

keploy

PR #1165, opened 2023-12-01
Fix #915: add support for running particular test

AkashKumar7902:
## Related Issue
  - Info about Issue or bug

Closes: #915 

#### Describe the changes you've made
added support to run specific testset and testcases.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Code style update (formatting, local variables)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Please let us know if any test cases are added

Please describe the tests(if any). Provide instructions how its affecting the coverage.

#### Describe if there is any unusual behaviour of your code(Write `NA` if there isn't)
A clear and concise description of it.

## Checklist:

- [ ] My code follows the style guidelines of this project.
- [ ] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes.

## Screenshots (if any)

|        Original         |          Updated           |
|:-----------------------:|:--------------------------:|
| **original screenshot** | <b>updated screenshot </b> |

PR #1189, opened 2023-12-11
fix unable to record custom name test cases

AkashKumar7902:
## Related Issue
  - Info about Issue or bug

Closes: #1188 

the `findlastindex` must be called only when no custom test case name is provided. so I have changed the order of function call

## Type of change

## Checklist:

## Screenshots (if any)

| **image** | <b>image</b> |


pkg/platform/yaml/yaml.go
```diff
@@ -182,12 +182,12 @@ func (ys *Yaml) WriteTestcase(tc *models.TestCase, ctx context.Context, filters
 		ys.mutex.Unlock()
 		var tcsName string
 		if ys.TcsName == "" {
-			// finds the recently generated testcase to derive the sequence number for the current testcase
-			lastIndx, err := findLastIndex(ys.TcsPath, ys.Logger)
-			if err != nil {
-				return err
-			}
 			if tc.Name == "" {
+				// finds the recently generated testcase to derive the sequence number for the current testcase
```
charankamarapu, line 186: Please test all the cases once for this feature.

1. record tests normally and run test mode. Don't use this feature and don't change any test case name.
2. record tests normally and record using this feature somewhere in the middle and run in test mode.
3. Along with second case rename test case manually in between recording and run in test mode.

charankamarapu (commented):
Please resolve the comments

pkg/platform/yaml/yaml.go
AkashKumar7902: here are the test-reports for all the specified scenarios:
1. https://pastebin.com/9nXVC3UL
2. https://pastebin.com/VrkciP8D
3. https://pastebin.com/VBKjrFLw

in the 3rd case, if we just rename the test case file and if we try running only that particular file (refer [PR 1165
](https://github.com/keploy/keploy/pull/1165)), then it will result in an error: 
```cmd
🐰 Keploy: 2023-12-12T17:59:27+05:30    ERROR   test/test.go:465        failed to fetch test results    {"error": "🐰 Keploy:%!(EXTRA string=found no test results for test report with id: %v, string=report-24)"}
```
we need to go inside the test case file and change the name parameter as well.

charankamarapu:
Is the PR ready for review..?

pkg/platform/yaml/yaml.go
charankamarapu: why does this happen any analysis..? I mean we just get the file and run the content right. Do we do any validation like checking if parameter in the file is equal to file name..?

charankamarapu (commented):
Please address the comments

pkg/platform/yaml/yaml.go
AkashKumar7902: in line [555](https://github.com/keploy/keploy/blob/f29564d0752836e33e509e84db77857e3646551a/pkg/service/test/test.go#L555) of test.go, we are iterating through ` initialisedValues.Tcs` which contains information from inside the file. so to run a particular testcase, the value of the name field present in the file will be compared to the testcase name you provided. renaming alone won't work as the name of the comparison is done with the name field not the filename. 

the error will arise if there are testcases and you want to run something that does not match with the name of any available testcase. This error is caused because in `FetchTestReport` , the [error condition](https://github.com/keploy/keploy/blob/f27b82fad82dcd16815c602ed73fca126294367f/pkg/service/test/test.go#L464) (passed + failed == 0) is true, as no testcase ran.

charankamarapu: Through this feature like renaming using header we will change the tc.name also right along with file name...? leave about renaming manually we can skip it.

AkashKumar7902: yes

charankamarapu (approved):
Ship it!

Issue #1209, opened 2023-12-15
[bug]: Keploy is exhibiting unexpected behavior when used with applications that have two or more databases


charankamarapu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

1. In record mode, When Generic dependency(Redis) is used in the application the expected behaviour is that the application run without any problem, but the dependency is getting timed out and giving error.

2. In test mode, when Generic dependency(Redis) is used in the application and few mock requests like ping calls and quit calls which are not recorded in record mode are encountered, the expected behaviour should be that parser calls the real database via passthrough. But as approximate matching is implemented the calls are getting matched to some other request and incorrect response is supplied in the parser.

3. In test mode when I run the the application, The application behaves unexpected because there are 2 mocks with same request but different response lets say MockRequest-1 ~ MockResponse-1 and MockRequest-1 ~ MockResponse-2, according to the application when it sends MockRequest-1 for first time it should receive MockResponse-1 and for second  time it should receive MockResponse-2. But the situation is vice versa in actual which is causing the error.

4. In test mode when I run the application in one of the dependency, a mock gets matched but when comes to reset the mock list some other mock gets deleted. There can be few race conditions with resetting the mock list.

### Steps to reproduce

1. Try an application with two dependencies Mongo and Redis with Keploy and record and test few calls. The issue will be replciated.

### Environment

Release

### Version

Cloud

### Repository

keploy

PR #1214, opened 2023-12-16
fix: Keploy unexpected behaviour with multiple dependencies

charankamarapu:
## Related Issue
  - Info about Issue or bug

Closes: #1209

PS : The line numbers below are corresponding to the line numbers in the issue mentioned above.

1. Time out is encountered in generic dependency because of the logic of writing mocks into yaml in generic. If any other dependencies is appending the mocks and as mutex is put, the generic parser had to wait till it gets its turn to add mock. Due to this there is a delay in receiving data from client or server in generic parser which is causing client timeouts in generic. Added go routine in generic parser where keploy appends mocks in mocks.yaml. This is go routine logic is already present in other parsers expect in generic.

2. Increased similarity index to improve approximate matching in generic dependency.

3. Implemented sorting the mocks in the `GetTcsMocks` function according to the req timestamp of the mock.

4. There are two race conditions obeserved for this mismatch or misDelete of the mocks while matching.
i) `GetTcsMocks` is returning the `tcsMocks` slice variable of the struct instead returning a copy it. So as different parsers update the `tcsMocks` it will be reflected in all other parsers suddenly and causes race conditions
ii) Updating the `tcsMocks` via `SetTcsMocks` function after every parser get its match with a mock. Due to this there is a race condition.

Lets say 
- parser a has read `tcsMocks` via `GetTcsMocks` 
- parser b has read `tcsMocks` via `GetTcsMocks`
- parser a has deleted one mock after matching and updates `tcsMocks` via `setTcsMocks`
- parser b has deleted one mock after matching and updates `tcsMocks` via `setTcsMocks`

What happened here is the update made by parser a is lost.

To deal with these two race condition we are implementing a in-memory db go-memdb(which implements mvcc) and instead of resetting the tcsMocks after every match, we will delete that particular mock from the in-memory db. The only case where we can get a conflict is when some parser wants to delete the same mock which some other parser has already deleted. This cannot happen in different type of parsers. This can only happen in the case of connection pooling. In this case of conflict we will restart the matching process for that parser which faced the conflict. As we only have delete behaviour and not update we will not be implementing version based mvcc approach here.

## Type of change

## Checklist:

## Screenshots (if any)


Sarthak160 (approved):
@charankamarapu This looks good to me and the only method where I was updating mocks for change to MD5 Auth type has been also merged with Match function only, And that is used in case of latest postgres versions. Since test bench is also passing   using that version only so we can conclude to be fine .

gouravkrosx (approved):
LGTM

re-Tick (approved):

re-Tick:
LGTM

Issue #1301, opened 2024-01-07
[bug]: authentication fails for java app with DocumentDb.


re-Tick:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When I run the test cmd, the auth fails for a larger java application which is making connection with the DocumentDB. Error logs: 
```sh
java.util.concurrent.ExecutionException: org.springfxramework.data.mongodb.UncategorizedMongoDbException: Exception authenticating MongoCredential{mechanism=SCRAM-SHA-1, userName='<username>', source='user-db', password=<hidden>, mechanismProperties=<hidden>}; nested exception is com.mongodb.MongoSecurityException: Exception authenticating MongoCredential{mechanism=SCRAM-SHA-1, userName='<username>', source='user-db', password=<hidden>, mechanismProperties=<hidden>}

```

### Steps to reproduce

1. Install the keploy CLI.
2. Record the test case for a big java app making external calls on documentDb and other dependency.
3. Run the test cases in test mode.
4. The error occurs.

### Environment

None

### Version

Local

### Repository

keploy

PR #1302, opened 2024-01-07
Fix mongo scram

re-Tick:
## Related Issue
  - Test mode fails for big java app that depends on documentDB and other reps.

Closes: #1301

Updates the extraction of client nonce from the saslStart request. And updates the conversationId in the response of saslStart and request of saslContinue.

For a client nonce in r like this: `n,,n=svoc-owner,r=9'sji{=${C"StM|SZWM1>Vpk`, the current extractClientNonce fails due to the presence of `=` char in the nonce. This is fixed by using `strings.TrimPrefix(part, "r=")` in place of the `strings.Split(part, "=")[1]` of the **extractClientNonce**. 

And the `saslContinue` is failing because we are returning the same conversationId in the response of `saslStart` due to which the key collision occurs for `authMessageMap` since, the key is used **conversationId** to get the authMessage for generating the server signature for response of saslContinue. So, the conversationId is updated for the response of **saslStart** and conversationId is extracted from the **saslContinue** request to fetch the authMessage from the  authMessageMap. 

## Type of change

- [ ] This change requires a documentation 

## Checklist:


pkg/proxy/integrations/mongoparser/mongoparser.go
```diff
@@ -596,7 +601,9 @@ func isHeartBeat(opReq Operation, requestHeader models.MongoHeader, mongoRequest
 	case wiremessage.OpMsg:
 		_, ok := mongoRequest.(*models.MongoOpMessage)
 		if ok {
-			return (opReq.IsIsAdminDB() && (strings.Contains(opReq.String(), "hello") || (isScramAuthRequest(mongoRequest.(*models.MongoOpMessage).Sections, logger)))) || opReq.IsIsMaster()
+			return (opReq.IsIsAdminDB() && strings.Contains(opReq.String(), "hello")) ||
```
charankamarapu, line 604: Please correct the indentation

pkg/proxy/integrations/mongoparser/mongoparser.go
```diff
@@ -55,8 +55,9 @@ func (m *MongoParser) ProcessOutgoing(requestBuffer []byte, clientConn, destConn
 		m.logger.Debug("the outgoing mongo in record mode")
 		encodeOutgoingMongo(requestBuffer, clientConn, destConn, m.hooks, m.logger, ctx)
 	case models.MODE_TEST:
+		logger := m.logger.With(zap.Any("Client IP Address", clientConn.RemoteAddr().String()), zap.Any("Client ConnectionID", util.GetNextID()), zap.Any("Destination ConnectionID", util.GetNextID()))
```
charankamarapu, line 58: Is this change required? Looks like you are adding few fields to the logger by default. So will these default fields appear in info logs too..?

pkg/proxy/integrations/mongoparser/scramAuth.go
```diff
@@ -22,14 +23,19 @@ func isScramAuthRequest(actualRequestSections []string, logger *zap.Logger) bool
 			return false
 		}
 
+		conversationId, _ := extractConversationId(actualMsg)
 		// Check if the message is for starting the SASL (authentication) process
 		if _, exists := actualMsg["saslStart"]; exists {
-			logger.Info("the recieved request is saslStart", zap.Any("OpMsg", actualMsg))
+			logger.Info("the recieved request is saslStart",
```
charankamarapu, line 29: shouldn't this be a debug log..?

pkg/proxy/integrations/mongoparser/scramAuth.go
```diff
@@ -22,14 +23,19 @@ func isScramAuthRequest(actualRequestSections []string, logger *zap.Logger) bool
...
+				zap.Any("OpMsg", actualMsg),
+				zap.Any("conversationId", conversationId))
 			return true
 			// Check if the message is for final request of the SASL (authentication) process
 		} else if _, exists := actualMsg["saslContinue"]; exists {
-			logger.Info("the recieved request is saslContinue", zap.Any("OpMsg", actualMsg))
-
+			logger.Info("the recieved request is saslContinue",
```
charankamarapu, line 35: shouldn't this be a debug log..?

pkg/proxy/integrations/scram/util.go
```diff
@@ -29,7 +29,11 @@ func extractClientNonce(firstMsg string) (string, error) {
 	for _, part := range parts {
 		if strings.HasPrefix(part, "r=") {
 			// Split based on "=" and get the value of "r"
-			value := strings.Split(part, "=")[1]
+			// value := strings.Split(part, "=")[1]
+			value := strings.TrimPrefix(part, "r=")
+			if value == part {
```
charankamarapu, line 34: Is there any chance that this case occurs..?

pkg/proxy/integrations/mongoparser/scramAuth.go
```diff
@@ -266,6 +292,11 @@ func handleSaslStart(i int, actualMsg map[string]interface{}, expectedRequestSec
 	logger.Debug("after replacing the new client nonce in auth response", zap.String("first response", newFirstAuthResponse))
 	// replace the payload with new first response auth
 	responseMsg["payload"].(map[string]interface{})["$binary"].(map[string]interface{})["base64"] = base64.StdEncoding.EncodeToString([]byte(newFirstAuthResponse))
+	responseMsg, err = updateConversationId(responseMsg, int(util.GetNextID()))
```
charankamarapu, line 295: For my clarity, the thing we should consider here mainly is that, the conversationId shouldn't be same. Every time it needs to change. Is this the only change done here ? and doesn't that depend on the client request message ..? Some more info regarding conversationId helps me understand  more regarding this change mainly like why it is used etc.

charankamarapu (commented):
Please address the comments. Most of the comments are questions for my understanding.

pkg/proxy/integrations/mongoparser/mongoparser.go
re-Tick: Yes these fields are required and they will only appear for error and debug logs of decodeOutgoingMongo function only. Other info logs will not be effected by this

pkg/proxy/integrations/scram/util.go
re-Tick: Yes I have mentioned an example for this in the PR comments.

pkg/proxy/integrations/mongoparser/mongoparser.go
charankamarapu: ok got it.

pkg/proxy/integrations/scram/util.go
charankamarapu: Yeah I have seen that but that is example for difference between `strings.TrimPrefix(part, "r=")` and `strings.Split(part, "=")[1]` right ? My comment was regarding even after using TrimPrefix will `value == part` anytime ?

pkg/proxy/integrations/mongoparser/scramAuth.go
re-Tick: Since scram is a challenge based authentication, so the challenge is done in conversation format. There are 2 request-response cycles: saslStart and saslContinue. During the response of saslStart the conversationId is passed by db server, and that is used by the saslContinue request. Using this conversationId the authMessage are mapped to generate the server proof. So this has to be unique so that we can map concurrent authMessage's. Earlier the field is not updated for the saslStart response.

charankamarapu: Got it.

pkg/proxy/integrations/mongoparser/mongoparser.go
re-Tick: Indentation is correct. I have done this because line was getting bigger. And it is a good practice.☺️

charankamarapu: Got it.

pkg/proxy/integrations/scram/util.go
re-Tick: That is a check, which can be true if the decoding of base64 of a wrong payload. Using this the error can be handled.

charankamarapu: I mean there is a `if strings.HasPrefix(part, "r=")` then we used `strings.TrimPrefix(part, "r=")` how can value == part..?

gouravkrosx (approved):
LGTM

charankamarapu:
I think there are few other changes added in the PR from other's commit.

re-Tick:
> I think there are few other changes added in the PR from other's commit.

Yes they were made to merge to main branch. but there is some issue with the main branch.

Issue #1346, opened 2024-01-12
[bug]: keploy fails to Handle HTTP Headers if Not Received in First Chunk


gouravkrosx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

I've observed cases where my HTTP headers also come in chunks. Keploy fails to handle this case.

### Steps to reproduce

It was reproduced using my local keploy test-bench changes. It never occurred in normal cases but can be a potential future bug.

### Environment

None

### Version

Local

### Repository

keploy

PR #1347, opened 2024-01-12
fix: HTTP handling for partial header reception

gouravkrosx:
## Related Issue
  - Keploy HttpParser fails to handle when HTTP headers also come in chunks

Closes: https://github.com/keploy/keploy/issues/1346

- Our proxy operates as a TCP server rather than an HTTP server, making it possible for HTTP headers not to arrive in the first chunk. Our HTTP parser observed headers arriving in multiple chunks due to TCP fragmentation based on chunk length, especially when the sender buffer is smaller. I've implemented a fix to handle these header chunks until complete headers are received.

- I've refactored error handling, identifying cases where certain errors need to be propagated to the caller function.
- Addressed an issue with the HTTP mock response headers, where two content-length values were being received. Previously, the code used the `add` function, but it should use the `set` function of `Headers`. This has been corrected.

## Type of change

- I've tested the changes with java and Golang applications.

`NA`

## Checklist:

## Screenshots (if any)

### DIFF
--- a/pkg/proxy/integrations/httpparser/httpparser.go
+++ b/pkg/proxy/integrations/httpparser/httpparser.go
@@ -129,10 +129,14 @@ func contentLengthRequest(finalReq *[]byte, clientConn, destConn net.Conn, logge
 		logger.Debug("This is a chunk of request[content-length]: " + string(requestChunked))
 		*finalReq = append(*finalReq, requestChunked...)
 		contentLength -= len(requestChunked)
-		_, err = destConn.Write(requestChunked)
-		if err != nil {
-			logger.Error("failed to write request message to the destination server")
-			return err
+
+		// destConn is nil in case of test mode.
+		if destConn != nil {
+			_, err = destConn.Write(requestChunked)
+			if err != nil {
+				logger.Error("failed to write request message to the destination server")
+				return err
+			}
 		}
 	}
 	return nil
@@ -156,10 +160,13 @@ func chunkedRequest(finalReq *[]byte, clientConn, destConn net.Conn, logger *zap
 			}
 
 			*finalReq = append(*finalReq, requestChunked...)
-			_, err = destConn.Write(requestChunked)
-			if err != nil {
-				logger.Error("failed to write request message to the destination server")
-				return err
+			// destConn is nil in case of test mode.
+			if destConn != nil {
+				_, err = destConn.Write(requestChunked)
+				if err != nil {
+					logger.Error("failed to write request message to the destination server")
+					return err
+				}
 			}
 
 			//check if the intial request is completed
@@ -267,10 +274,13 @@ func handleChunkedRequests(finalReq *[]byte, clientConn, destConn net.Conn, logg
 			logger.Error("failed to read the request message from the client")
 			return err
 		} else {
-			_, err = destConn.Write(reqHeader)
-			if err != nil {
-				logger.Error("failed to write request message to the destination server")
-				return err
+			// destConn is nil in case of test mode
+			if destConn != nil {
+				_, err = destConn.Write(reqHeader)
+				if err != nil {
+					logger.Error("failed to write request message to the destination server")
+					return err
+				}
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


charankamarapu:
will review by EOD

pkg/proxy/integrations/httpparser/httpparser.go
```diff
@@ -110,122 +110,183 @@ func ProcessOutgoingHttp(request []byte, clientConn, destConn net.Conn, h *hooks
...
 		}
 		logger.Debug("This is a chunk of request[content-length]: " + string(requestChunked))
 		*finalReq = append(*finalReq, requestChunked...)
 		contentLength -= len(requestChunked)
-		_, err = destConn.Write(requestChunked)
-		if err != nil {
-			logger.Error("failed to write request message to the destination server", zap.Error(err))
-			return
+
+		// destConn is nil in case of test mode.
```
charankamarapu, line 133: How was this working before, Like did we used to get this error `logger.Error("failed to write request message to the destination server", zap.Error(err))` before in test mode..?

pkg/proxy/integrations/httpparser/httpparser.go
```diff
@@ -110,122 +110,183 @@ func ProcessOutgoingHttp(request []byte, clientConn, destConn net.Conn, h *hooks
...
 			if err == io.EOF {
-				logger.Error("connection closed by the destination server", zap.Error(err))
-				break
+				isEOF = true
+				logger.Debug("recieved EOF, connection closed by the destination server")
+				if len(resp) == 0 {
+					break
+				}
 			} else if netErr, ok := err.(net.Error); ok && netErr.Timeout() {
 				logger.Info("Stopped getting data from the connection", zap.Error(err))
```
charankamarapu, line 196: Shouldn't this be a debug log?

pkg/proxy/integrations/httpparser/httpparser.go
```diff
@@ -637,3 +734,113 @@ func encodeOutgoingHttp(request []byte, clientConn, destConn net.Conn, logger *z
...
+		"operation": req.Method,
+	}
+	passthroughHost := false
+	for _, host := range models.PassThroughHosts {
+		if req.Host == host {
+			passthroughHost = true
+		}
+	}
+	if !passthroughHost {
+		err = h.AppendMocks(&models.Mock{
```
charankamarapu, line 804: This should be in a go routine else parser will get struck here sometimes.

charankamarapu (commented):
Please address the comments. Rest of the things looks good to me.

pkg/proxy/integrations/httpparser/httpparser.go
gouravkrosx: We might have not encountered this chunking during test mode before, otherwise, it would have panicked

pkg/proxy/integrations/httpparser/httpparser.go
gouravkrosx: I guess we should tell the user that may be due to slow internet, the response couldn't be completed in time, hence this error.

charankamarapu: Will user understand that this is because of slow internet through that log?

pkg/proxy/integrations/httpparser/httpparser.go
gouravkrosx: Okay will add go routine here.

pkg/proxy/integrations/httpparser/httpparser.go
charankamarapu: got it

pkg/proxy/integrations/httpparser/httpparser.go
gouravkrosx: Since it is a timeout error, it will also be reflected in its application. So I guess the user will be able to know

charankamarapu: Got it.

charankamarapu (approved):
Ship it!

PR #1352, opened 2024-01-14
fix: sorting of config mocks acc to req timestamp

Sarthak160:
## Related Issue
  - #1354
  - #1353

Closes:  #1354 and #1353

1. Generic mocks are not sorted based on the timestamp of the request to significantly improve accuracy of matching. 
2. Testcase Response is not recorded as soon as the server starts sending data. 
3. The Mock datastore now has sorting information to ensure that the mocks are matched according the timestamps. 
4. The mocks now also have an id to ensure correlation with the logs. 
5. We have made Postgres use ConfigMocks by default instead of TCS (Data mocks) since it's hard to pinpoint exact request related mock. This would be further improved in a future version. 
6. Extra Debug and info logs added for better visibility to debug future issues. 

## Type of change

### DIFF
--- a/pkg/hooks/connection/tracker.go
+++ b/pkg/hooks/connection/tracker.go
@@ -48,7 +48,7 @@ type Tracker struct {
 	logger *zap.Logger
 
 	reqTimestampTest []time.Time
-	resTimestampTest time.Time
+	resTimestampTest []time.Time
 	isNewRequest     bool
 }
 
@@ -138,8 +138,6 @@ func (conn *Tracker) IsComplete() (bool, []byte, []byte, time.Time, time.Time) {
 
 			if conn.verifyResponseData(expectedSentBytes, actualSentBytes) {
 				validRes = true
-				// Capturing the response timestamp as response is verified
-				conn.resTimestampTest = time.Now()
 			} else {
 				conn.logger.Debug("Malformed response", zap.Any("ExpectedSentBytes", expectedSentBytes), zap.Any("ActualSentBytes", actualSentBytes))
 				recordTraffic = false
@@ -193,8 +191,6 @@ func (conn *Tracker) IsComplete() (bool, []byte, []byte, time.Time, time.Time) {
 				conn.currentRecvBufQ = conn.currentRecvBufQ[1:]
 
 				responseBuf = conn.SentBuf
-
-				conn.resTimestampTest = time.Now()
 			} else {
 				conn.logger.Debug("no data buffer for request", zap.Any("Length of RecvBufQueue", len(conn.currentRecvBufQ)))
 				recordTraffic = false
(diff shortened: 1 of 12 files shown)


slayerjain:
@Sarthak160 is [pkg/hooks/bpf_bpfel_arm64.o](https://github.com/keploy/keploy/pull/1352#diff-c8781ca316f03c430b95ec56c7d441deb7f4dbc244d9d5f2bf7fb8a89f8ec06d) intentional or accidental ?

Sarthak160:
> @Sarthak160 is [pkg/hooks/bpf_bpfel_arm64.o](https://github.com/keploy/keploy/pull/1352#diff-c8781ca316f03c430b95ec56c7d441deb7f4dbc244d9d5f2bf7fb8a89f8ec06d) intentional or accidental ?

Added some ebpf logs for req timestamp

slayerjain:
won't we need to do this for amd64 then too?

pkg/hooks/connection/tracker.go
```diff
@@ -213,19 +209,32 @@ func (conn *Tracker) IsComplete() (bool, []byte, []byte, time.Time, time.Time) {
...
+	var resTimestampTest time.Time
+	// Checking if record traffic is recorded and request & response timestamp is captured or not.
+	if recordTraffic {
+		if len(conn.reqTimestampTest) > 0 {
+			// Get the timestamp of current request
+			reqTimestampTest = conn.reqTimestampTest[0]
+			// Pop the timestamp of current request
+			conn.reqTimestampTest = conn.reqTimestampTest[1:]
+		} else {
+			conn.logger.Debug("no request timestamp found, skipping recording")
```
slayerjain, line 221: Please add some context about the request that are skipping. This will the user understand which request we skipped. Maybe url, method. 

The message can also be something like, "failed to capture request timestamp for a request. Please record it again if important"

pkg/hooks/connection/tracker.go
```diff
@@ -213,19 +209,32 @@ func (conn *Tracker) IsComplete() (bool, []byte, []byte, time.Time, time.Time) {
...
+		if len(conn.resTimestampTest) > 0 {
+			// Get the timestamp of current request
+			resTimestampTest = conn.resTimestampTest[0]
+			// Pop the timestamp of current request
+			conn.resTimestampTest = conn.resTimestampTest[1:]
+		} else {
+			conn.logger.Debug("no response timestamp found, skipping recording")
+			recordTraffic = false
+		}
+		conn.logger.Debug(fmt.Sprintf("TestRequestTimestamp:%v || TestResponseTimestamp:%v", reqTimestampTest, resTimestampTest))
```
slayerjain, line 234: same as above

slayerjain (changes requested):

pkg/hooks/connection/tracker.go
slayerjain: Also, please change this to info rather than debug. The user should know about it.

Sarthak160: Sure, on it

slayerjain:
LGTM!

Issue #1338, opened 2024-01-10
[feature]: Create Keploy version command


charankamarapu:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

As a command line tool Keploy should have a --version command.

### Why should this be worked on?

For good user experience

### Repository

keploy

Akash-Singh04:
Already have finished an implementation for this!

PR #1376, opened 2024-01-16
[Feature]: Create Keploy  --version Flag to CLI

Akash-Singh04:
## Related Issue
  - Created Keploy --version flag.
  
Closes: #1338 

Implemented the ```--version``` flag functionality to display the current Keploy version when triggered. Removed the display of version information from regular command outputs.

## Type of change

No additional test cases are added as this change doesn't affect existing functionality.

NA

## Checklist:

## Screenshots (if any)
image


charankamarapu:
Will review by EOD

cmd/root.go
```diff
@@ -242,13 +251,24 @@ func (r *Root) execute() {
 		Example: rootExamples,
 	}
 	rootCmd.CompletionOptions.DisableDefaultCmd = true
+
 	rootCmd.SetHelpTemplate(rootCustomHelpTemplate)
 
 	rootCmd.PersistentFlags().BoolVar(&debugMode, "debug", false, "Run in debug mode")
 
-	// Manually parse flags to determine debug mode early
+	rootCmd.PersistentFlags().Bool("version", false, "Fetch the latest version")
```
charankamarapu, line 259: Why is version flag added as PersistentFlag..? PersistentFlag is a flag which can be used in other commands also I guess. But we require --version flag only for root command right?

cmd/root.go
```diff
@@ -242,13 +251,24 @@ func (r *Root) execute() {
...
+
+	// Manually parse flags to determine debug mode and version flag early
 	debugMode = checkForDebugFlag(os.Args[1:])
-	// Now that flags are parsed, set up the l722ogger
+	versionFlag := checkForVersionFlag(os.Args[1:])
```
charankamarapu, line 263: If you have defined a flag in the above code, why do you have to parse again for the version flag?

cmd/root.go
```diff
@@ -242,13 +251,24 @@ func (r *Root) execute() {
...
+	if versionFlag {
+		// Fetch the version and print it
+		currentVersion := utils.KeployVersion
+		fmt.Println("Current version:", currentVersion)
```
charankamarapu, line 267: Print something like this:

Keploy xxx

not this current version : xxx and this print statement should come before that box present in the image.

main.go
```diff
@@ -35,7 +36,23 @@ func main() {
 	}
 	utils.KeployVersion = version
 	fmt.Println(logo, " ")
-	fmt.Printf("version: %v\n\n", version)
+	releaseInfo, err2 := utils.GetLatestGitHubRelease()
```
charankamarapu, line 39: Try to do this at root.go so that the current version print statement comes before this statement.

main.go
```diff
@@ -35,7 +36,23 @@ func main() {
...
+	if err2 != nil {
+		log.Debug("Failed to fetch latest release version", err2)
+	}
+	updatetext := models.GrayString.Sprint("keploy update")
```
charankamarapu, line 43: Why are you still showing this box when err has occurred? We have to skip this right..?

main.go
```diff
@@ -35,7 +36,23 @@ func main() {
 	}
 	utils.KeployVersion = version
 	fmt.Println(logo, " ")
-	fmt.Printf("version: %v\n\n", version)
+	releaseInfo, err2 := utils.GetLatestGitHubRelease()
+	if err2 != nil {
```
charankamarapu, line 40: Never name the error as err2. It is named as err only.

charankamarapu (commented):
Please address the comments

charankamarapu:
Also in the case of -dev as prefix in the current we shouldn't show that update version box because, that person will be contributor or developer of Keploy.

cmd/root.go
Akash-Singh04: My approach for using a PersistentFlag here was to make it accessible not only for the root command but also for any subcommands that might need to access or check the version flag. Since you have made it clear that isn't the case , we can change the PersistantFlags() to Flags() to make it accessible only to the root command

main.go
Akash-Singh04: Refactored the code to reflect the same

cmd/root.go
Akash-Singh04: The manual parsing of flags is performed to determine the value of versionFlag variables before setting up the logger and executing the root command. In my opinion it is the correct way to handle this feature. Let me know if there is an alternate or a better way to do this

main.go
Akash-Singh04: If we do this at root.go, then the update message wont be visible everytime we run keploy, which was one of the original goals of this feature. Let me know if this behaviour is unintended or needs to be revised.

Akash-Singh04:
@charankamarapu Kindly have a look!

main.go
charankamarapu: Every command in keploy goes through root cmd right..?

cmd/root.go
charankamarapu: reference: 

https://chromium.googlesource.com/external/github.com/spf13/cobra/+/4fd30b69ee2b62cf3bbecf0a423f8a1ee47f5f24/user_guide.md#version-flag

charankamarapu:
Replied to your comments. Pleas have a look

cmd/root.go
Akash-Singh04: Refactored

main.go
Akash-Singh04: Refactored

Akash-Singh04:
@charankamarapu let me know if it is all good now!

cmd/root.go
charankamarapu: I don't think you have got the point I am trying make here. Cobra actually has --version inbuilt you have to provide some template to it show the version properly Please go through the reference again.

charankamarapu (commented):
There is one comment which is not yet resolved please resolve it

Akash-Singh04:
@charankamarapu Refactored, Kindly Check. Thank You

cmd/root.go
Akash-Singh04: Sorry for the confusion, It is now done.

cmd/root.go
```diff
@@ -240,15 +243,44 @@ func (r *Root) execute() {
...
 	}
+
 	rootCmd.CompletionOptions.DisableDefaultCmd = true
+
 	rootCmd.SetHelpTemplate(rootCustomHelpTemplate)
 
 	rootCmd.PersistentFlags().BoolVar(&debugMode, "debug", false, "Run in debug mode")
 
-	// Manually parse flags to determine debug mode early
+	rootCmd.PersistentFlags().Bool("version", false, "Fetch the latest version")
```
charankamarapu, line 255: As you have added SetVersionTemplate this line is no more needed.

cmd/root.go
```diff
@@ -240,15 +243,44 @@ func (r *Root) execute() {
...
+
+	// Manually parse flags to determine debug mode and version flag early
 	debugMode = checkForDebugFlag(os.Args[1:])
-	// Now that flags are parsed, set up the l722ogger
+
+	rootCmd.SetVersionTemplate(`
+	{{with .Version}}{{printf "Keploy %s" .}}{{end}}
```
charankamarapu, line 261: Screenshot 2024-01-27 at 12 01 58 AM

I have tested the changes I have observed few things 

1. Once I have ran the command you can see that the indentation is lost in the version sentence and also next prompt from the terminal. Please correct the indentation in the template.
 2. Update info is given before version statement.
 3. When I was trying the command actually it took 1 or 2 secs to give out the command output. This is because of update function which is doing a  https call. We should take a call whether we should show this update thing. Because I can clearly see a lag after typing the command. This may affect the user experience a lot. Let me discuss internally too and I will let you know.
 
 If you want you can remove this update thing from this PR as these changes are present in update PR too. So that I will merge this PR and we can work regarding this update fucntion in update PR.

charankamarapu (commented):
Please address the comments and rest looks good for me

cmd/root.go
Akash-Singh04: I think the correct decision would be to take a step back and reflect upon the goals of this Pull Request. I agree with you that the update message should therefore then be shifted to the update command PR. Refactoring the same

Akash-Singh04:
@charankamarapu Addressed the review comments. PR is ready to be merged, Kindly Check

cmd/root.go
```diff
@@ -240,15 +240,22 @@ func (r *Root) execute() {
...
 	}
+
 	rootCmd.CompletionOptions.DisableDefaultCmd = true
+
 	rootCmd.SetHelpTemplate(rootCustomHelpTemplate)
 
 	rootCmd.PersistentFlags().BoolVar(&debugMode, "debug", false, "Run in debug mode")
 
-	// Manually parse flags to determine debug mode early
+	// Manually parse flags to determine debug mode and version flag early
```
charankamarapu, line 252: please correct the comment

main.go
```diff
@@ -35,7 +35,6 @@ func main() {
 	}
 	utils.Version = version
 	fmt.Println(logo, " ")
-	fmt.Printf("version: %v\n\n", version)
```
charankamarapu, line 38: For time being I think we should Let it be. When ever we get the error logs from user if it contains the version it will be useful for debugging.

utils/utils.go
```diff
@@ -14,6 +17,13 @@ import (
 	sentry "github.com/getsentry/sentry-go"
 )
 
+type GitHubRelease struct {
```
charankamarapu, line 20: Do you need these changes in this PR any more..? I think this not needed right?

charankamarapu (commented):
Please address the comments.

utils/utils.go
Akash-Singh04: Yep removed them from this PR

cmd/root.go
Akash-Singh04: We are not manually parsing for the version flag here, hence the corrected comment

main.go
Akash-Singh04: Alright

Akash-Singh04:
@charankamarapu Kindly review

main.go
charankamarapu: This is not yet reverted. I think I was not clear last time, we need to specify version every time we run any command in keploy for debugging purposes.

Akash-Singh04: Apologies for the oversight, Refactored

charankamarapu (approved):
Ship it!

Issue #1418, opened 2024-01-23
[bug]: Gin-mongo throwing error in Test mode in POST calls 


charankamarapu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

1. When we record 5-6 POST call in gin-mongo and run in Test mode, the POST will fail due to an error in mongo dependency.

### Steps to reproduce

1. Record 5-6 POST calls with gin-mongo
2. Run Test mode
3. Test mode fails

### Environment

None

### Version

Cloud

### Repository

keploy

PR #1433, opened 2024-01-25
refactor:setting time stamp for response only after verified response

shivamsouravjha:
## Related Issue
  - Instead of using an array to save the response time reverted to set the response time of response only after the data is verified.
Closes: #1418

Because of maintaining an array of mock and picking the first element of the array we were unable to get certain mock files as a result of which  connection would end and test cases would fail.

## Type of change

## Checklist:

## Screenshots (if any)

image

### DIFF
--- a/pkg/hooks/connection/tracker.go
+++ b/pkg/hooks/connection/tracker.go
@@ -66,9 +66,8 @@ type Tracker struct {
 	mutex  sync.RWMutex
 	logger *zap.Logger
 
-	reqTimestamps  []time.Time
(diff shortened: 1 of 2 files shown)


charankamarapu (commented):
Looks good to me but Can you explain why is the last chuck timestamp approach is not implemented? Like maintaining an array and instead of first timestamp taking the last one.

shivamsouravjha:
updated the logic to update the last element of the array rather

charankamarapu (approved):
ship it!

Issue #1485, opened 2024-02-03
[feature]: Add New Version Available Message to keploy


Akash-Singh04:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

 Add New Version Available Message everytime keploy runs, the message would look like this:
 
 
![image](https://github.com/keploy/keploy/assets/114267538/e04d3663-3f10-4a06-a902-b7a805229a9f)
 

### Why should this be worked on?

Helps users be updated with the latest keploy updates.

### Repository

keploy

Akash-Singh04:
I was already working on this as a part of #1325 , What needs to be discussed is the delay that will happen(~2-3s) due to an Http request being made everytime keploy is run

gouravkrosx:
@charankamarapu is this completed after the v2-refactor?

charankamarapu:
no

PR #1325, opened 2024-01-09
Feature: Added Keploy Update to CLI

Akash-Singh04:
## Related Issue
  - This pull request enhances the update functionality in the Keploy CLI tool to ensure smoother and more efficient updates of the Keploy binary file. 

Resolves: #1303 

#### Describe the changes you've made
The updates include:

# Addition of Update Command 
-  Integrated the update command in the Keploy CLI to fetch the latest version from the GitHub repository.
- Refactored the update process to verify the current installed version against the latest available version fetched from the repository.
- Integrated GitHub API requests to retrieve the latest release version and get the changelog
- Updated the binary file 
- Display the changelog after the version has been successfully updated

# Addition of notification for new update
- Checks for a new update whenever keploy CLI is used
- If a new update is available, prompts the user in a non invasive way to update the binary file.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Code style update (formatting, local variables)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Please let us know if any test cases are added

NIL

#### Describe if there is any unusual behaviour of your code(Write NA if there isn't) 
NA

A clear and concise description of it.

## Checklist:

- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes.

## Screenshots (if any)

Screenshot 2024-01-11 061617

PR #1486, opened 2024-02-04
[feature]: Add Update Available Message to keploy

Akash-Singh04:
## Related Issue
  - Added the update available message to Keploy CLI

Closes: #1485 

- Added Update Available message to notify users about new updates
- Message only shows up if current version of keploy is not updated 
- Message does not show up on dev versions

## Type of change

NIL

NIL

## Checklist:

## Screenshots (if any)
image


Akash-Singh04:
What needs to be discussed is the delay that will happen(~2-3s) due to an Http request being made everytime keploy is run


There are two possible solutions:

- The update message is shown at the end of execution of the command

- The update message is shown only once a day.


Let us discuss on how this can be implemented.

Akash-Singh04:
@charankamarapu The review and discussion for this pull request is pending. Kindly look into it at your earliest convenience

cmd/root.go
```diff
@@ -279,6 +282,28 @@ func (r *Root) execute() {
...
 	//Set the version template for version command
 	rootCmd.SetVersionTemplate(`{{with .Version}}{{printf "Keploy %s" .}}{{end}}{{"\n"}}`)
+	currentVersion := utils.Version
+	// Show update message only if it's not a dev version
+	if !strings.HasSuffix(currentVersion, "-dev") {
+		// Check for the latest release version
+		releaseInfo, err := utils.GetLatestGitHubRelease()
+		if err != nil {
+			r.logger.Debug("Failed to fetch the latest release version", zap.Error(err))
+			return
```
PranshuSrivastava, line 292: why are we returning if we failed to fetch the latest release, keploy should still work isn't it? This will confuse the user as to why keploy is not running

main.go
```diff
@@ -37,7 +37,6 @@ func main() {
 	utils.Version = version
 	if binaryToDocker := os.Getenv("BINARY_TO_DOCKER"); binaryToDocker != "true" {
 		fmt.Println(logo, " ")
-		fmt.Printf("version: %v\n\n", version)
```
PranshuSrivastava, line 40: why did you remove this?

cmd/root.go
```diff
@@ -279,6 +282,28 @@ func (r *Root) execute() {
...
+			updatetext := models.HighlightGrayString("keploy update")
+			const msg string = `
+               ╭─────────────────────────────────────╮
+               │ New version available:              │
+               │ %v  ---->   %v       │
+               │ Run %v to update         │
+               ╰─────────────────────────────────────╯
+			   `
+			versionmsg := fmt.Sprintf(msg, currentVersion, releaseInfo.TagName, updatetext)
+			fmt.Printf(versionmsg)
```
PranshuSrivastava, line 304: shouldn't we also give an option to the user to update keploy? or atleast tell him how he can do so by mentioning the command here.

PranshuSrivastava (changes requested):

main.go
Akash-Singh04: Refrence: https://github.com/keploy/keploy/issues/1488

cmd/root.go
Akash-Singh04: In my opinion asking the user to update immediately would be a bad user experience. We can however mention the command here to do so.

main.go
PranshuSrivastava: I don't see any comments on that issue, did you discuss this with any organisation member before implementing? We usually print the version along with logo because whenever someone faces a problem with Keploy, we ask for the logs, and there we can check what specific version is causing the issue.

Akash-Singh04: Yes this was prior discussed with @charankamarapu , It was supposed to be implemented as a part of #1376, but was asked to be removed as some discussions needed to be done before the implementation. 

Can we have a discussion over this on slack and decide whats the best course of action regarding the version message?

Akash-Singh04: > What needs to be discussed is the delay that will happen(~2-3s) due to an Http request being made everytime keploy is run
> 
> There are two possible solutions:
> 
> * The update message is shown at the end of execution of the command
> * The update message is shown only once a day.
> 
> Let us discuss on how this can be implemented.


This also needs to be discussed with the community

PranshuSrivastava: I had a chat with him and I think we are now supposed to run keploy --version everytime we start Keploy. Is that correct? Please let me know when youa add that change

PranshuSrivastava (changes requested):

Akash-Singh04:
@PranshuSrivastava It has been added again

main.go
Akash-Singh04: Updated, Thank you for resolving

Akash-Singh04: As it turns out, Keploy automatically adds in a --version flag upon initialisation, which was customised in #1376. Instead of running keploy with --version command everytime, i think its best we just update the message to 
```Keploy version: 2-xyz```


Let me know if this is an optimal approach

PranshuSrivastava: I think it was finaliased that the version flag should be run everytime because showing the version like this is not a good pracice so I think we should stick to that.

PranshuSrivastava (changes requested):

charankamarapu:
Hey @Akash-Singh04 please update the branch with main.

Akash-Singh04:
Hey @charankamarapu , Branch has been updated.
I was wondering if there is a way we can run the update message only when a new terminal session is created? Do you know a way we can get that information in our CLI? I tried searching for some solutions but wasn't able to get anything that fits our need.

PranshuSrivastava:
Hey @Akash-Singh04 , the checks are failing, can you please take a look? Also, the update command that you added has not been working correctly for users. Even after they run keploy update, and then type Keploy again, it shows the same version.

Akash-Singh04:
@PranshuSrivastava I am not sure about your bug report for the update command. I used used it to update from ```Keploy 2.0.0-alpha78``` to  ```Keploy 2.1.0-alpha2```. Here are the screenshots:
image
image

re-Tick:
There is an issue with the closing `|`. Please have a look.

Akash-Singh04:
@re-Tick @PranshuSrivastava Both of the requested changes have been made
- The update command is now refactored to work reliably on all platforms as done in v1 in #1325
- Fixed #1608 Keploy update command updates even on latest version 
- Added a function to calculate whitespaces so that the closing braces of the update message are always inline

PranshuSrivastava:
image
There is still an error in the graphics, it is aligned towards the right too much. I tried running the update command again, and even though it said it had updated the keploy version, when i opened a new terminal tab, it showed the same version. Can you try to just simply download the one touch install and execute if, instead of updating the binary inside the "/usr/ocal/bin" directory?

Akash-Singh04:
Will look into this

nehagup:
Hi @Akash-Singh04 this is great initiative. One good addition to `keploy update` that users suggested is - we prompt "keploy update available.. (y/n)" for the first time, however, if the user enters "no" we should just log the warning (update available) and not prompt every time.

Akash-Singh04:
@PranshuSrivastava Your requested changes have been made

nehagup:
@Sonichigo please review

Issue #1469, opened 2024-01-31
[bug]: only add sudo to docker alias for linux if required


slayerjain:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

By default docker requires using `sudo` in most linux distros. There are ways to avoid adding the sudo flag everytime. Keploy should check if docker is executable without `sudo` and use the alias (used during installation accordingly"  

### Steps to reproduce

1. install keploy on linux select docker using -  curl -O https://raw.githubusercontent.com/keploy/keploy/main/keploy.sh && source keploy.sh
2. keploy will ask for sudo everytime. can be verified with `alias docker`

### Environment

None

### Version

Cloud

### Repository

keploy

Akash-Singh04:
Hey, I would like to work on this!

Akash-Singh04:
@slayerjain Kindly review the PR!

PR #1492, opened 2024-02-04
Updated Docker Installation Alias for linux

Akash-Singh04:
## Related Issue
  - Only add sudo to docker alias for linux if required

Closes: #1469 

- We check if ```docker ps``` runs successfully without sudo.
- If it does, the alias is created without sudo, otherwise, it's created with sudo, otherwise not
## Type of change

NIL
NIL
## Checklist:


Akash-Singh04:
@slayerjain Kindly look into this PR for the issue you opened

slayerjain:
@Akash-Singh04 thanks for the PR. 

@PranshuSrivastava since you were working on this script, can you please and test this?

PranshuSrivastava:
Hey @Akash-Singh04, thanks for the PR. But this is not a good solution for the problem. This is kind of a hack to check if docker runs with sudo or not. So in macos sudo is generally not required, but in linux sudo is required but it can be removed with a specific configuration. So take a look at how the need for sudo is removed in Linux, and check if the user's system has that configuration or not, and if it does then don't use sudo

Akash-Singh04:
In this updated implementation, we check if the current user belongs to the ```docker``` group using the ```groups``` command since users inside the ```docker``` group have permission to run docker without the ```sudo``` command

Reference: https://askubuntu.com/a/477554

keploy.sh
```diff
@@ -71,20 +71,25 @@ installKeploy (){
...
-            if ! docker volume inspect debugfs &>/dev/null; then
-                docker volume create --driver local --opt type=debugfs --opt device=debugfs debugfs
-            fi
-            alias keploy='sudo docker run --pull always --name keploy-v2 -p 16789:16789 --privileged --pid=host -it -v $(pwd):$(pwd) -w $(pwd) -v /sys/fs/cgroup:/sys/fs/cgroup -v debugfs:/sys/kernel/debug:rw -v /sys/fs/bpf:/sys/fs/bpf -v /var/run/docker.sock:/var/run/docker.sock -v '"$HOME"'/.keploy-config:/root/.keploy-config -v '"$HOME"'/.keploy:/root/.keploy --rm ghcr.io/keploy/keploy'
+    if [ "$OS_NAME" = "Darwin" ]; then
+        if ! docker volume inspect debugfs &>/dev/null; then
+            docker volume create --driver local --opt type=debugfs --opt device=debugfs debugfs
+        fi
+        alias keploy='docker run --pull always --name keploy-v2 -p 16789:16789 --privileged --pid=host -it -v $(pwd):$(pwd) -w $(pwd) -v /sys/fs/cgroup:/sys/fs/cgroup -v debugfs:/sys/kernel/debug:rw -v /sys/fs/bpf:/sys/fs/bpf -v /var/run/docker.sock:/var/run/docker.sock -v '"$HOME"'/.keploy-config:/root/.keploy-config -v '"$HOME"'/.keploy:/root/.keploy --rm ghcr.io/keploy/keploy'
+    else
```
PranshuSrivastava, line 83: use else if here, we are doing this specifically on linux

PranshuSrivastava (changes requested):
The rest of the changes look good, there is just one comment that you need to address.

keploy.sh
Akash-Singh04: Kindly check if this was the intended change

PranshuSrivastava (approved):
LGTM

Issue #1393, opened 2024-01-18
[feature]: Make generating testReports files optional


tomtastico:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

Right now, whenever a test run is finished, a `keploy/testReports/report-[number].yaml` file is generated with the test results.

In general these files are not really necessary for an end user, as we can see the test results in the console output. They should only be generated on request (by a CLI / config flag) in case the user needs to use them in some way.

### Why should this be worked on?

The `testReports` folder grows unchecked otherwise. I have to clean it manually now and then.

### Repository

keploy

slayerjain:
@probablyArth Please try

Akash-Singh04:
Hey there! @slayerjain , I have already been workimg on Two PR's related to adding CLI commands to keploy, which is why i feel like i should pick up this issue as i am already well versed with the required approach.

Here is my approach for the same:


We can add in a persistant flag to check if a command is run with --testreport flag inorder to check if the user wants these files to be generated

```
    rootCmd.PersistentFlags().BoolVar(&generateTestReportsFlag, "generate-test-reports", false, "Generate test reports")
```

We can then use this flag to check if the files are to be generated

Let me know if this approach works! Looking forward to getting started immediately!

slayerjain:
anyone is free to open PR. we will merge the first PR that works well. 

@gouravkrosx please review this approach.

gouravkrosx:
@Akash-Singh04 You can use normal flags so that the user can use the config file also to add this information.

Akash-Singh04:
Great, working on this!

Yaxhveer:
Is this issue still open?

Akash-Singh04:
> Is this issue still open?

Hey there, Currently working on the final implementation for this!

Akash-Singh04:
PR is ready for review! Awaiting feedback

gouravkrosx:
@Akash-Singh04 In my opinion the default feature should generate the test reports and can be disabled using CLI/Config flag.

Akash-Singh04:
Hey there, made a new PR with the new flag as per your request. Kindly look into it

PR #1513, opened 2024-02-06
Feature: added generateTestReport flag

Yaxhveer:
## Related Issue
  - Info about Issue or bug

Closes: #1393 

Added a Flag ***--generateTestReports*** to keploy test and config file which is true by default. 

## Type of change

## Checklist:

## Screenshots (if any)


Yaxhveer:
@gouravkrosx test reports will be generated in serve command since they are used to show status.

gouravkrosx:
@Yaxhveer 
You can delete the TestReport folder and files after the completion in case of the serve command.

Yaxhveer:
@gouravkrosx Please review the changes.

cmd/test.go
```diff
@@ -384,17 +391,21 @@ func (t *Test) GetCmd() *cobra.Command {
 			}
 
 			path += "/keploy"
+			t.logger.Info("", zap.Any("keploy test and mock path", path))
```
PranshuSrivastava, line 394: Why has this line been added

cmd/test.go
```diff
@@ -384,17 +391,21 @@ func (t *Test) GetCmd() *cobra.Command {
...
+	
+				testReportPath, err = pkg.GetNextTestReportDir(testReportPath, models.TestRunTemplateName)
+					t.logger.Info("", zap.Any("keploy testReport path", testReportPath))
+					if err != nil {
+						t.logger.Error("failed to get the next test report directory", zap.Error(err))
+						return err
+					}
 			}
-
-			t.logger.Info("", zap.Any("keploy test and mock path", path), zap.Any("keploy testReport path", testReportPath))
```
PranshuSrivastava, line 396: Why have these essential logs been removed?

pkg/service/test/test.go
```diff
@@ -537,9 +543,10 @@ func (t *tester) FetchTestResults(cfg *FetchTestResultsConfig) models.TestRunSta
 	(*resultForTele)[0] += *cfg.Success
 	(*resultForTele)[1] += *cfg.Failure
 
-	err = cfg.TestReportFS.Write(context.Background(), cfg.TestReportPath, cfg.TestReport)
-
-	t.logger.Info("test report for "+cfg.TestSet+": ", zap.Any("name: ", cfg.TestReport.Name), zap.Any("path: ", cfg.Path+"/"+cfg.TestReport.Name))
+	if cfg.GenerateTestReport {
```
PranshuSrivastava, line 546: if we are not generating the testreport we should atleast log once that we are skipping the test report since the flag has been set to false. Do this in all the services with this if condition

pkg/graph/serve.go
```diff
@@ -40,9 +40,11 @@ func NewGraph(logger *zap.Logger) graphInterface {
 const defaultPort = 6789
 
 // Serve is called by the serve command and is used to run a graphql server, to run tests separately via apis.
-func (g *graph) Serve(path string, proxyPort uint32, mongopassword, testReportPath string, Delay uint64, pid, port uint32, lang string, passThroughPorts []uint, apiTimeout uint64, appCmd string, enableTele bool) {
+func (g *graph) Serve(path string, proxyPort uint32, mongopassword, testReportPath string, generateTestReport bool, Delay uint64, pid, port uint32, lang string, passThroughPorts []uint, apiTimeout uint64, appCmd string, enableTele bool) {
 	var ps *proxy.ProxySet
 
+	defer pkg.DeleteTestReports(g.logger, generateTestReport)
```
PranshuSrivastava, line 46: why exactly are we deleting the reports at the end of the function?

PranshuSrivastava (changes requested):

cmd/test.go
Yaxhveer: Actually ```t.logger.Info("", zap.Any("keploy test and mock path", path), zap.Any("keploy testReport path", testReportPath))``` was a combined log and I have divided it into two different log for path and testReport Path.

cmd/test.go
Yaxhveer: Similar reason divided it into two different logs

pkg/service/test/test.go
Yaxhveer: ok

pkg/graph/serve.go
Yaxhveer: serve command require testReports to show status. So I removed the testReports when function ends

PranshuSrivastava: Can't we just get the status without writing to a file, or checking the condition before writing to the file?

Yaxhveer: Actually this method was recommended here https://github.com/keploy/keploy/pull/1513#issuecomment-1931478001

PranshuSrivastava: ok

PranshuSrivastava:
@Yaxhveer the functionality is working correctly, there is just one problem with the logs
image
Why is this same log being printed 3 times? It should be printed only once, or if you are logging three logs, they should atleast contain useful information that was not in the previous log, although I don't think that will be the case.

Yaxhveer:
@PranshuSrivastava fixed the extra logs issue

PranshuSrivastava (approved):
LGTM

Issue #1520, opened 2024-02-07
feat: option to disable ANSI color codes


slayerjain:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Currently when we are logging to places where ANSI color codes are not supported eg: piping to a file - the output is not readable. We use ANSI color codes to show colors in terminals, but they are not supported in log files and CI. 

We should provide an option to disable ANSI color codes. 

### Steps to reproduce

1. Record tests for any application with keploy. 
2. run testmode by pipping to a file. eg:`keploy test -c "some app command" &> log.txt`
3. You will see random charaters appened instead of colors. 

eg: 

```
  TESTRUN SUMMARY. For testrun with id: [31m[1m"[0m[31mtest-set-0[0m[31m[1m"[0m
	Total tests: [34m[1m79[0m
	Total test passed: [34m[1m66[0m
	Total test failed: [34m[1m13[0m
	```

### Environment

None

### Version

Cloud

### Repository

keploy

Akash-Singh04:
I think we can add a CLI flag and a config flag to do this, Working on the implementation right away

Akash-Singh04:
@slayerjain PR has been made, along with relevant screenshots to satisfy the flag is working as intended. Kindly check it and let me know if any changes are required

PR #1483, opened 2024-02-02
[feature]: Implemented Struct instead of passing in variable in CaptureTraffic and GetRecordConfig

Akash-Singh04:
## Related Issue
  - Info about Issue or bug

Closes: #1482 

#### Describe the changes you've made
Implemented Struct ```Record Options``` to pass params into ```Capture Traffic``` and ```GetRecordConfig``` instead of passing in variables, inorder to have better coding practice and less merge conflicts.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Code style update (formatting, local variables)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Please let us know if any test cases are added
NIL
#### Describe if there is any unusual behaviour of your code(Write `NA` if there isn't)
NA
## Checklist:

- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes.

## Screenshots (if any)

image

PR #1526, opened 2024-02-07
Feat: Added --disableANSI Flag to CLI to enable/disable ANSI Color Codes

Akash-Singh04:
## Related Issue
  - Added --enableANSI CLI Flag to enable/disable colour encoding.

Closes: #1520 

-  Added ```--enableANSIColor``` PersistantFlag to CLI to enable/disable ASNI Color Codes
Example of enabled ANSI Colors in Log files:
image
- Added it to testCmd inorder to prevent ASNI Color Codes being rendered in this part of report:
image

## Type of change

NIL
NIL
## Checklist:

### Usage:
image

| **
image
** | <b>
image
 </b> |
 | **
 
image

** | <b>
image

 </b> |
 
 Below is an attached video of the implementation in action:
 

https://github.com/keploy/keploy/assets/114267538/eac44ce0-fbfb-41a1-90a8-b62d908e22a5


Akash-Singh04:
@slayerjain Kindly review the PR and let me know if any changes are required.

PranshuSrivastava (changes requested):

cmd/test.go
```diff
@@ -42,7 +42,7 @@ func ReadTestConfig(configPath string) (*models.Test, error) {
 	return &doc.Test, nil
 }
 
-func (t *Test) getTestConfig(path *string, proxyPort *uint32, appCmd *string, tests *map[string][]string, appContainer, networkName *string, Delay *uint64, buildDelay *time.Duration, passThroughPorts *[]uint, apiTimeout *uint64, globalNoise *models.GlobalNoise, testSetNoise *models.TestsetNoise, coverageReportPath *string, withCoverage *bool, configPath string, ignoreOrdering *bool, passThroughHosts *[]models.Filters) error {
+func (t *Test) getTestConfig(path *string, proxyPort *uint32, appCmd *string, tests *map[string][]string, appContainer, networkName *string, Delay *uint64, buildDelay *time.Duration, passThroughPorts *[]uint, apiTimeout *uint64, globalNoise *models.GlobalNoise, testSetNoise *models.TestsetNoise, coverageReportPath *string, withCoverage *bool, configPath string, ignoreOrdering *bool, passThroughHosts *[]models.Filters, EnableColor *bool) error {
```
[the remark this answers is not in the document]
Akash-Singh04: We are already working on that in #1483

cmd/test.go
```diff
@@ -250,7 +256,7 @@ func (t *Test) GetCmd() *cobra.Command {
 
 			passThroughHosts := []models.Filters{}
 
-			err = t.getTestConfig(&path, &proxyPort, &appCmd, &tests, &appContainer, &networkName, &delay, &buildDelay, &ports, &apiTimeout, &globalNoise, &testsetNoise, &coverageReportPath, &withCoverage, configPath, &ignoreOrdering, &passThroughHosts)
+			err = t.getTestConfig(&path, &proxyPort, &appCmd, &tests, &appContainer, &networkName, &delay, &buildDelay, &ports, &apiTimeout, &globalNoise, &testsetNoise, &coverageReportPath, &withCoverage, configPath, &ignoreOrdering, &passThroughHosts, &EnableColor)
```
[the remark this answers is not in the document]
Akash-Singh04: Already working on that in #1483

PranshuSrivastava:
Also fix the merge conflicts please

cmd/root.go
```diff
@@ -33,17 +33,21 @@ type Root struct {
...
 	zapcore.Encoder
 }
 
-func NewColorConsole(cfg zapcore.EncoderConfig) (enc zapcore.Encoder) {
-	return colorConsoleEncoder{
-		EncoderConfig: &cfg,
-		// Using the default ConsoleEncoder can avoid rewriting interfaces such as ObjectEncoder
-		Encoder: zapcore.NewConsoleEncoder(cfg),
+// Define a global flag to control ANSI color code usage
+var EnableColor bool
```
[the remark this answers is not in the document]
Akash-Singh04: Refactored to enableASNIColor flag

pkg/service/test/test.go
```diff
@@ -541,14 +542,18 @@ func (t *tester) FetchTestResults(cfg *FetchTestResultsConfig) models.TestRunSta
...
+		// Print without ANSI color codes
+		fmt.Printf("\n <=========================================> \n  TESTRUN SUMMARY. For testrun with id: %s\n"+"\tTotal tests: %d\n"+"\tTotal test passed: %d\n"+"\tTotal test failed: %d\n <=========================================> \n\n", cfg.TestReport.TestSet, cfg.TestReport.Total, cfg.TestReport.Success, cfg.TestReport.Failure)
 	} else {
-		pp.SetColorScheme(models.PassingColorScheme)
+		if *cfg.Status == models.TestRunStatusFailed {
+			pp.SetColorScheme(models.FailingColorScheme)
+		} else {
+			pp.SetColorScheme(models.PassingColorScheme)
+		}
+		pp.Printf("\n <=========================================> \n  TESTRUN SUMMARY. For testrun with id: %s\n"+"\tTotal tests: %s\n"+"\tTotal test passed: %s\n"+"\tTotal test failed: %s\n <=========================================> \n\n", cfg.TestReport.TestSet, cfg.TestReport.Total, cfg.TestReport.Success, cfg.TestReport.Failure)
```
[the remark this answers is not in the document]
Akash-Singh04: The way pp works according to my testing is that it adds certain ASNI color codes regardless of which ColorScheme we set it, hence it is imperative that we use fmt.Printf() in the if else condition.

I agree that it is redundant code but it is the workaround i found for this problem. Let me know if there is an optimised approach to this. 

Note: I have already tried using Sprintf to set the string prior and that doesn't work either.

Akash-Singh04:
@PranshuSrivastava Kindly review this PR and have a look at the attached video to better understand the implementation, Thank You!

cmd/root.go
```diff
@@ -32,17 +32,20 @@ type Root struct {
 }
 
 var debugMode bool
+var EnableASNIColor bool
```
PranshuSrivastava, line 35: It's "ANSI" not "ASNI", correct it everywhere

pkg/service/test/test.go
```diff
@@ -707,21 +713,22 @@ func (t *tester) RunTestSet(testSet, path, testReportPath, appCmd, appContainer,
...
+		TestReportPath:  testReportPath,
+		Path:            path,
+		EnableASNIColor: EnableASNIColor,
 	}
 	status = t.FetchTestResults(resultsCfg)
 	return status
 }
 
-func (t *tester) testHttp(tc models.TestCase, actualResponse *models.HttpResp, noiseConfig models.GlobalNoise, ignoreOrdering bool) (bool, *models.Result) {
+func (t *tester) testHttp(tc models.TestCase, actualResponse *models.HttpResp, noiseConfig models.GlobalNoise, ignoreOrdering bool, EnableASCIColor *bool) (bool, *models.Result) {
```
PranshuSrivastava, line 731: why did the name of the flag suddenly change?

pkg/service/test/test.go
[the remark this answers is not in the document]
PranshuSrivastava: It doesn't work that way. There is a way to disable the colors altogether. Check the library again.

PranshuSrivastava (changes requested):

pkg/service/test/test.go
Akash-Singh04: Typo, Fixed

cmd/root.go
Akash-Singh04: Corrected

pkg/service/test/test.go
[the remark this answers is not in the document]
Akash-Singh04: I stand corrected, there is an inbuilt property ```pp.ColoringEnabled = false```. Thank you for the learning opportunity!

pkg/service/test/test.go
```diff
@@ -867,8 +870,10 @@ func (t *tester) testHttp(tc models.TestCase, actualResponse *models.HttpResp, n
 
 	if !pass {
 		logDiffs := NewDiffsPrinter(tc.Name)
-
 		logger := pp.New()
+		if !*EnableANSIColor {
```
PranshuSrivastava, line 874: why are you using a pointer here?

cmd/root.go
```diff
@@ -245,6 +248,15 @@ func checkForDebugFlag(args []string) bool {
 	return false
 }
 
+func checkForEnableANSIColorFlag(args []string) bool {
```
PranshuSrivastava, line 251: This is not a correct way of reading flags, can't you just parse it and then check its value?

pkg/service/test/test.go
```diff
@@ -922,19 +927,14 @@ func (t *tester) testHttp(tc models.TestCase, actualResponse *models.HttpResp, n
 				logDiffs.PushBodyDiff(fmt.Sprint(tc.HttpResp.Body), fmt.Sprint(actualResponse.Body), bodyNoise)
 			}
 		}
-		t.mutex.Lock()
-		logger.Printf(logs)
-		err := logDiffs.Render()
```
PranshuSrivastava, line 927: why exactly have you removed this render function?

pkg/service/test/test.go
```diff
@@ -54,6 +54,7 @@ type TestOptions struct {
 	CoverageReportPath string
 	IgnoreOrdering     bool
 	PassthroughHosts   []models.Filters
+	EnableANSIColor    *bool
```
PranshuSrivastava, line 57: why pointer?

utils/utils.go
```diff
@@ -356,4 +358,15 @@ func UpdateKeployToDocker(cmdName string, isDockerCompose bool, flags interface{
 
 }
 
+// RemoveANSIEscapeCodes removes ANSI escape codes from a string
+func RemoveANSIEscapeCodes(input string) string {
```
PranshuSrivastava, line 362: where is this being referenced?

PranshuSrivastava (changes requested):

cmd/root.go
Akash-Singh04: The reason for this is because similar to the debug flag, we need to manually parse and get the value of the flag early as in root command the colorCodeConsole is initialised first and then the flags are parsed.

I had tested it without manually parsing it and the flag value wasn't updated until after the console was created.

pkg/service/test/test.go
Akash-Singh04: The reason for using a pointer is that we are passing a pointer to the const enableANSIColor which holds the value of the flag ,a pointer to which is being passed down to this function, instead of passing the value.

pkg/service/test/test.go
Akash-Singh04: Answered in prev comment.

pkg/service/test/test.go
Akash-Singh04: Revised

utils/utils.go
Akash-Singh04: Revised

pkg/service/test/test.go
PranshuSrivastava: But if it is a const variables why do you need to pass it by reference? you are not going to change the value inside the function, which is what leads to duplication, you are just going to use it right?

cmd/root.go
PranshuSrivastava: We are currently refactoring the cmd package, then you will be able to parse it properly, without having to read it like that.

PranshuSrivastava:
Also please fix the merge conflicts

pkg/service/test/test.go
Akash-Singh04: Will address this once the cmd layer is refactored and we can move ahead with this pull request

charankamarapu:
Hey @Akash-Singh04 please update your branch with the refactoring changes. Mostly the code where the color is decided is same. Major change would be around cli. you can add the flag in this file https://github.com/keploy/keploy/blob/7febf76bd7b1230604d9b8afb5add32e703ba34d/cli/provider/cmd.go#L153
please add the flag in the config also and I think it is better to use persistent flag for this flag. You can discuss with the reviewer and go with.

Akash-Singh04:
@charankamarapu How can we get the value of the --enableANSIColor flag before the logger is initialised? Since we are initialising the logger at `func start(ctx context.Context) {
	logger, err := log.New()`
But the flags are being added here at ```AddFlags``` as you mentioned. How can we pass the value of the flag to ```main.go``` so that it can be passed to ```New()``` function in ```logger.go```?

Akash-Singh04:
@charankamarapu @PranshuSrivastava The pr is ready after refactoring! Do let me know if there are any further deliverables

Sarthak160 (approved):

PranshuSrivastava:
image
Looks the color is not being changed in the correct place, I am still getting colored logs in the beginning

PranshuSrivastava (changes requested):

cli/provider/cmd.go
```diff
@@ -222,6 +223,12 @@ func (c *CmdConfigurator) AddFlags(cmd *cobra.Command) error {
 			utils.LogError(c.logger, err, errMsg)
 			return errors.New(errMsg)
 		}
+		err := viper.BindPFlag("debug", cmd.PersistentFlags().Lookup("debug"))
```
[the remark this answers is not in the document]
Akash-Singh04: This was present in an early version of the refactoring and got added in. Removed

Akash-Singh04:
> image Looks the color is not being changed in the correct place, I am still getting colored logs in the beginning

The reason being that the logger is initially set-up here:
`func NewColor(cfg zapcore.EncoderConfig, enableColor bool) (enc zapcore.Encoder) {
	if enableColor {
		return color{
			EncoderConfig: &cfg,
			Encoder:       zapcore.NewConsoleEncoder(cfg),
		}
	}
	// fmt.Println("Color is disabled")
	return zapcore.NewConsoleEncoder(zap.NewDevelopmentEncoderConfig())
}`

Then when the command runs, and the flags are read, we change the color of the logger:
`if !c.cfg.EnableANSIColor {
		c.logger.Info("Color encoding is disabled")
		logger, err := log.ChangeColorEncoding()
		*c.logger = *logger
		if err != nil {
			errMsg := "failed to change color encoding"
			utils.LogError(c.logger, err, errMsg)
			return errors.New(errMsg)
		}
	}`

This is due to the current flow of commands in the v2 codebase. For the most part it correctly disables the ANSI color codes except in the beginning while the flags are being setup.

CC: @PranshuSrivastava

PranshuSrivastava:
Ok, the first log can be left colored as we haven't fully read the flags by that time, but the second log is being printed by you right? So why can't you just print it after changing the colors instead of before it?

Akash-Singh04:
@PranshuSrivastava Your requested changes have been made

cli/provider/cmd.go
```diff
@@ -219,6 +219,7 @@ func (c *CmdConfigurator) AddFlags(cmd *cobra.Command) error {
 	case "keploy":
 		cmd.PersistentFlags().Bool("debug", c.cfg.Debug, "Run in debug mode")
 		cmd.PersistentFlags().Bool("disableTele", c.cfg.DisableTele, "Run in telemetry mode")
+		cmd.PersistentFlags().Bool("enableANSIColor", c.cfg.EnableANSIColor, "Enable ANSI color in logs")
```
PranshuSrivastava, line 222: The changes are working correctly, just change the name of this flag to "enableANSI". Currently the name feels too long to type for a flag.

PranshuSrivastava (changes requested):

Akash-Singh04:
@PranshuSrivastava Your requested changes have been made

cli/provider/cmd.go
Akash-Singh04: refactored

PranshuSrivastava:
@re-Tick Yes thats good. I just want the name to be short

Akash-Singh04:
@re-Tick @PranshuSrivastava Your requested changes have been made

PranshuSrivastava (approved):
LGTM

Issue #1557, opened 2024-02-11
[bug]: Sort test suite names for complete summary


shivamsouravjha:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

* In `test.go` testSuiteNames are sorted as string using `sort.Strings(testSuiteNames)`.
* 
<img width="674" alt="image" src="https://github.com/keploy/keploy/assets/60891544/6474a7f4-625a-4b1c-84dd-09da663a7602">

### Steps to reproduce

* Instead of sorting as string it should be sorted as number.

### Environment

Production

### Version

Cloud

### Repository

keploy

officialasishkumar:
A possible fix is to store the numbers after "test-set-" in a slice of integers and then sort it.

PR #1559, opened 2024-02-12
[Bug]: Sort test suite names for complete summary

Yaxhveer:
## Related Issue
  - Info about Issue or bug

Closes: #1557 

Sorted test suit based on the number.

## Type of change

## Checklist:

## Screenshots (if any)


pkg/service/test/test.go
```diff
@@ -274,7 +275,13 @@ func (t *tester) Test(path string, testReportPath string, generateTestReport boo
 		testSuiteNames = append(testSuiteNames, testSuiteName)
 	}
 
-	sort.Strings(testSuiteNames)
+	sort.SliceStable(testSuiteNames, func(i, j int) bool {
+
+        num1, _ := strconv.Atoi(strings.Split(testSuiteNames[i], "-")[2])
+        num2, _ := strconv.Atoi(strings.Split(testSuiteNames[j], "-")[2])
+
```
shivamsouravjha, line 282: is it wise to ignore the errors here?

pkg/service/test/test.go
```diff
@@ -274,7 +275,13 @@ func (t *tester) Test(path string, testReportPath string, generateTestReport boo
 		testSuiteNames = append(testSuiteNames, testSuiteName)
 	}
 
-	sort.Strings(testSuiteNames)
+	sort.SliceStable(testSuiteNames, func(i, j int) bool {
+
+        num1, _ := strconv.Atoi(strings.Split(testSuiteNames[i], "-")[2])
```
shivamsouravjha, line 280: To make it not throw error in case someone updates the` testSuiteNames` better to put a length check as well as address the errors from `strconv.Atoi` (returning false in the case would be a good idea)

Yaxhveer: what should be the max length?

shivamsouravjha: so when you split the testsuitename it goes into array of 3 (as of now expected format is `test-set-X` thus its fair to check if length is 3 atleast as to order the last element X of the name.

pkg/service/test/test.go
```diff
@@ -274,7 +275,26 @@ func (t *tester) Test(path string, testReportPath string, generateTestReport boo
...
+
+		test2 := strings.Split(testSuiteNames[j], "-")
+		lenEl = len(test2)-1
+
+		num2, err := strconv.Atoi(test2[lenEl])
+		if err != nil {
+			return false
+		}
+
+		return num1 < num2
```
shivamsouravjha, line 296: a better way of doing thing could be 
```
parts1 := strings.Split(testSuiteNames[i], "-")
    parts2 := strings.Split(testSuiteNames[j], "-")

    if len(parts1) < 3 || len(parts2) < 3 {
        // Handle the case where there are not enough parts after the split
        return false // or true, or some other logic that makes sense for your use case
    }

    num1, err1 := strconv.Atoi(parts1[2])
    num2, err2 := strconv.Atoi(parts2[2])

    if err1 != nil || err2 != nil {
        // Handle the error case where the conversion from string to int fails
        return false // or true, or some other logic that makes sense for your use case
    }

    return num1 < num2
    
    ```
    
    Offcourse with names that are more suited for our context.

Comment (Yaxhveer) [reply pkg/service/test/test.go:296]: case where length is less than 3 should be directly compare the string instead of returning false?

File: pkg/service/test/test.go [hunk 4]
```diff
@@ -274,7 +275,25 @@ func (t *tester) Test(path string, testReportPath string, generateTestReport boo
 		testSuiteNames = append(testSuiteNames, testSuiteName)
 	}
 
-	sort.Strings(testSuiteNames)
+	sort.SliceStable(testSuiteNames, func(i, j int) bool {
+
+		test1 := strings.Split(testSuiteNames[i], "-")
+		test2 := strings.Split(testSuiteNames[j], "-")
```
shivamsouravjha, line 281: rather than test1 ~ testSuitePartsI and testSuitePartsJ can be a better. similarly instead of num1 something more specific testSuiteIDNumberI and testSuiteIDNumberJ can be used a better explainer.

Yaxhveer: ok

shivamsouravjha (approved):
LGTM!

Issue #1565, opened 2024-02-14
[bug]: Docker fails when ran in linux installation from installation script


charankamarapu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

1. Docker fails when ran in linux installation from installation script

### Steps to reproduce

1. Run docker in linux mahcine

### Environment

None

### Version

Cloud

### Repository

keploy

PR #1566, opened 2024-02-14
fix: add network in linux installation script

charankamarapu:
## Related Issue
  - Info about Issue or bug

Closes: #1565

## Type of change

## Checklist:

## Screenshots (if any)

### DIFF
--- a/keploy.sh
+++ b/keploy.sh
@@ -20,6 +20,8 @@ installKeploy (){
         sudo mkdir -p /usr/local/bin && sudo mv /tmp/keploy /usr/local/bin/keploybin
 
         set_alias 'sudo -E env PATH="$PATH" keploybin'
+
+        add_network
     }
 
     check_sudo(){
@@ -36,6 +38,8 @@ installKeploy (){
         sudo mkdir -p /usr/local/bin && sudo mv /tmp/keploy /usr/local/bin/keploybin
 
         set_alias 'sudo -E env PATH="$PATH" keploybin'
+
+        add_network
     }
 
     append_to_rc() {
@@ -96,9 +100,15 @@ installKeploy (){
         fi
     }
 
-
-
-    install_docker() {
+    add_network() {
+        if ! which docker &> /dev/null; then
+            echo -n "Docker not found on device, please install docker and reinstall keploy if you are willing to use applications with docker"
+            return
+        fi
+        if ! docker info &> /dev/null; then
+            echo "Please start Docker and reinstall keploy if you are willing to use applications with docker"
+            return
+        fi
         check_sudo
         sudoCheck=$?
         network_alias=""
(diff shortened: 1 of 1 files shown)


keploy.sh
```diff
@@ -109,7 +119,9 @@ installKeploy (){
         if ! $network_alias docker network ls | grep -q 'keploy-network'; then
             $network_alias docker network create keploy-network
         fi
+    }
 
+    install_docker() {
```
PranshuSrivastava, line 124: why haven't you called add_network in this function? In macos, we still install keploy via docker, so it will be needed in that case. In the future when we add a binary for macos, we can remove the install_docker completely

PranshuSrivastava (changes requested):
Please address the comments

keploy.sh
charankamarapu: Added add_network in darwin

keploy.sh
```diff
@@ -96,9 +100,15 @@ installKeploy (){
...
 
-
-
-    install_docker() {
+    add_network() {
+        if ! which docker &> /dev/null; then
+            echo -n "Docker not found on device, please install docker and reinstall keploy if you are willing to use applications with docker"
+            return
+        fi
+        if ! docker info &> /dev/null; then
```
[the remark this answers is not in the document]
charankamarapu: Added two functions `check docker status `for  Darwin as well as linux. Created separate functions because their logs are different.

keploy.sh
```diff
@@ -96,21 +110,59 @@ installKeploy (){
...
+        return 1
+    }
 
-
-    install_docker() {
+     check_docker_status_for_Darwin() {
         check_sudo
         sudoCheck=$?
         network_alias=""
         if [ "$sudoCheck" -eq 0 ] && [ $OS_NAME = "Linux" ]; then
```
PranshuSrivastava, line 136: os name should be darwin here.

keploy.sh
```diff
@@ -96,21 +110,59 @@ installKeploy (){
...
+    check_docker_status_for_linux() {
+        check_sudo
+        sudoCheck=$?
+        network_alias=""
+        if [ "$sudoCheck" -eq 0 ] && [ $OS_NAME = "Linux" ]; then
+            # Add sudo to docker
+            network_alias="sudo"
+        fi
+        if ! $network_alias which docker &> /dev/null; then
+            echo -n "Docker not found on device, please install docker and reinstall keploy if you are willing to use applications with docker"
```
PranshuSrivastava, line 122: The logs for mac and linux shouldn't be so different in my opinion. and I think "willing" doesn't fit well in a log message. I think just change both to "Docker not found on device. Please install docker and reinstall keploy". Same thing for starting docker.

keploy.sh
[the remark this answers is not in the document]
PranshuSrivastava: I think we can standardize that, instead of creating a seperate function just for that one thing.

PranshuSrivastava (changes requested):

keploy.sh
charankamarapu: added

keploy.sh
charankamarapu: we shouldn't force user to use docker on linux if he has no application with docker

charankamarapu:
Regarding  - "standardize thing, instead of creating a seperate function just for that one thing." I think this is the standard way, branching should happen in the parent function only. You can not use same conditions in the child functions also which you have used in parent. Then there is no purpose of branching in parent. and more over they are not identical the log messages are different.

PranshuSrivastava (approved):
LGTM

Issue #1534, opened 2024-02-08
[bug]: remove explicit chmod commands from keploy codebase


slayerjain:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

We currently have multiple `chmod` commands within the keploy codebase. This creates dependency on the library and so we must handle this programatically. 

Since keploy runs as root (using the sudo command), the files created by keploy are not editable by other users by default (even after setting broader file permission during create). The `chmod` command ensures that generated yamls are user editable. 

### Steps to reproduce

eg: https://github.com/keploy/keploy/blob/main/pkg/proxy/util/util.go

### Environment

None

### Version

Cloud

### Repository

keploy

Akash-Singh04:
Working on this!

shivamsouravjha:
Issue seems to be resolved after refactor.

PR #1567, opened 2024-02-15
[bug]: remove explicit chmod commands from keploy codebase

Akash-Singh04:
## Related Issue
  - Bug fix

Closes: #1534 

- Implemented ```syscall.umask``` to set the file permissions to 0 before creating the file and then restoring it to normal.
- This effectively eliminates the need for ```chmod``` command which is used to set the file permissions to editable. 

## Type of change

NIL
NIL
## Checklist:


Akash-Singh04:
@PranshuSrivastava Let me know if this implementation is apt!

pkg/proxy/proxy.go
```diff
@@ -158,14 +158,17 @@ func isJavaInstalled() bool {
 
 // to extract ca certificate to temp
 func ExtractCertToTemp() (string, error) {
+	// Set the umask to allow access to the file for all users
+	err := utils.SetUmask(0)
```
PranshuSrivastava, line 162: why are you changing the permission of the cert file, nobody tries to access that. It is not needed

pkg/proxy/util/util.go
```diff
@@ -70,6 +69,12 @@ func CreateYamlFile(path string, fileName string, Logger *zap.Logger) (bool, err
 		return false, err
 	}
 	if _, err := os.Stat(yamlPath); err != nil {
+		// Temporarily set umask to allow broader permissions
+		err = utils.SetUmask(0)
```
PranshuSrivastava, line 73: why are you setting it as 0 here? It just sets it to the original permissions of the program right?

pkg/service/generateConfig/generateConfig.go
```diff
@@ -47,17 +46,17 @@ func (g *generatorConfig) GenerateConfig(filePath string, options GenerateConfig
 
 	finalOutput := append(results, []byte(utils.ConfigGuide)...)
 
+	err = utils.SetUmask(0)
```
PranshuSrivastava, line 49: why are you first setting to 0 and then 0022. What is the problem in directly setting to 0022?

PranshuSrivastava (changes requested):
Also, you have missed the cmd/test.go file.

pkg/proxy/util/util.go
Akash-Singh04: That would be incorrect. Setting Umask to 0 temporarily  before creating the file, allows the file to have the  maximum permissions (read, write, and execute) for the owner, group, and others, thus effectively removing the need for explicitly using the chmod command to handle the file permissions

pkg/service/generateConfig/generateConfig.go
Akash-Singh04: After creating the file, the umask is restored to 0022 to apply the default permissions for subsequent file creations.

pkg/proxy/proxy.go
Akash-Singh04: Previously there was an implementation of chmod command here,which is why i had changed the permissions here. Reverting as you deem it unnecessary.

cmd/test.go
```diff
@@ -448,17 +448,6 @@ func (t *Test) GetCmd() *cobra.Command {
 					IgnoreOrdering:     ignoreOrdering,
 					PassthroughHosts:   passThroughHosts,
 				}, enableTele)
-
-				fileExist := utils.CheckFileExists(path)
```
PranshuSrivastava, line 452: You have still not added the changes to this file

PranshuSrivastava (changes requested):

cmd/test.go
Akash-Singh04: Incase it wasn't clear, earlier we were checking if the file exists here and if it did we would use chmod to change the permissions of the file. Now, it has been refactored so that the permissions are set appropriately right when the testReport file is being generated

gouravkrosx:
@Akash-Singh04 Please update your PR according to the refactored changes.

Akash-Singh04:
@gouravkrosx @PranshuSrivastava The v2 refactoring of keploy removed most of the chmod commands from the codebase. In this PR i have further removed any Chmod commands present. Do review this

utils/utils.go
```diff
@@ -559,6 +559,15 @@ func InterruptProcessTree(logger *zap.Logger, ppid int, sig syscall.Signal) erro
 	return nil
 }
 
+func SetUmask(mask int) error {
+	prevUmask := syscall.Umask(mask)
```
PranshuSrivastava, line 563: We are still using a linux command to replace for chmod. It would be better if we can use some api or sdk from a linux command to do this. Like now we are using docker sdk instead of docker cli commands for executing docker related operations. Research on it a bit and let me know

PranshuSrivastava (changes requested):

pkg/core/proxy/ca.go
```diff
@@ -91,6 +91,15 @@ func getCaPaths() ([]string, error) {
 
 // to extract ca certificate to temp
 func extractCertToTemp() (string, error) {
+
+	if err := utils.SetUmask(0); err != nil {
+		return "failed to set Umask", err
+	}
+
+	if err := utils.SetUmask(0022); err != nil {
```
re-Tick, line 99: You have to make sure that the umask is set back to its before value which can be (027, etc). So, need to first extract the before umask of system.

re-Tick: And here setUmask should be after CreateTemp method.

Akash-Singh04: Your requested changes have been made

shivamsouravjha:
closing this PR has a lot of unnecessary changes, feel free to open a new PR

Issue #1538, opened 2024-02-09
[feature]: Create a normalisation command in keploy CLI


charankamarapu:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

Normalisation here means that when ever users application is changed a bit and it is tested against the old test cases the responses may vary and mocks may also vary in few cases . If there is only change in the test case response and not in the mocks then we should make a feature where user can replace the response in the testcase with new response we got in test mode. 

### Why should this be worked on?

Generally an application will have slight change every month or week while developing . To record test cases every time we change the application code is difficult. This feature wouldn't require re-record and saves developer time.

### Repository

keploy

Akash-Singh04:
Hey @charankamarapu , I have worked on CLI commands and flags before in Keploy and i believe this issue falls under my skillset. Could you kindly elaborate upon how this command is supposed to work?
- Is it supposed to be an add-on to the test command? That is, after the test has failed, the user has the option to change the responses?
-  Or is it supposed to be a standalone command where the user will manually choose the test case that failed and input a new response header to it?

Thank you for the insights!

charankamarapu:
1. Yes this command is something user is going to use after test command. If test fails and user thinks the new response is correct we need to provide option or a command that user can change the response in the yaml.
2. Once user runs normalisation command you will not ask user to enter input, as test is already done, you will be getting the new response from the test-report and updating the yaml.

gouravkrosx:
closed by https://github.com/keploy/keploy/pull/1859

PR #1589, opened 2024-02-17
[Feature]: Create a normalisation command in keploy CLI

Akash-Singh04:
## Related Issue
  - New CLI Command Feature

Closes: #1538 

Added ```keploy normalise``` command to the Keploy CLI to allow users to normalise their test cases. The Function of the command is as follows:
- The command checks for ```test-sets``` , ```test-cases``` and ```test-run``` flags from the CLI
-  The command reads YAML files containing test data and  unmarshals it..
- The command iterates over each failed test case in the test data and retrieves the corresponding test data from the struct. 
- The command then updates the expected response in the test case file with the actual response obtained
- Finally, the program writes the updated YAML data back to the original testcase files. 

## Type of change

NIL
NA
## Checklist:

## Screenshots (if any)
image
image


Akash-Singh04:
@charankamarapu Kindly review this PR and let me know if this implementation works!

charankamarapu:
From the screenshot I can say there is a bit of feature missing. User should give the testcases and testSet name which user want to normalise.

charankamarapu (commented):
Please address the comment

Akash-Singh04:
> From the screenshot I can say there is a bit of feature missing. User should give the testcases and testSet name which user want to normalise.

Alright, if that is how we would want to implement this feature it should not be a problem at all. Will update asap and revert!

Akash-Singh04:
@charankamarapu Let me know if this implementation is ready to go or there is a need to further include more features in this!

Akash-Singh04:
@charankamarapu If the CI pipeline will require time to fix, this PR is ready to be merged

charankamarapu:
cool will do a review and let you know.

cli/normalise.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// Normalise retrieves the command to normalise Keploy
+func Normalise(ctx context.Context, logger *zap.Logger, conf *config.Config, serviceFactory ServiceFactory, cmdConfigurator CmdConfigurator) *cobra.Command {
+	var normaliseCmd = &cobra.Command{
+		Use:     "normalise",
+		Short:   "Normalise Keploy",
+		Example: "keploy normalise --path /path/to/localdir --test-set testset --test-cases testcases",
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			path, err := cmd.Flags().GetString("path")
```
charankamarapu, line 38: Getting the value from flag is no more needed after refactored changes because viper is dealing with it - https://github.com/keploy/keploy/blob/2554c2f393fa3849b10372bb3449250aee8c901c/cli/provider/cmd.go#L225

cli/normalise.go
```diff
@@ -0,0 +1,90 @@
...
+			if err != nil {
+				utils.LogError(logger, err, "Error in getting path")
+				return err
+			}
+			//if user provides relative path
```
charankamarapu, line 43: These validations should also be done in here https://github.com/keploy/keploy/blob/2554c2f393fa3849b10372bb3449250aee8c901c/cli/provider/cmd.go#L223

cli/normalise.go
```diff
@@ -0,0 +1,90 @@
...
+			if err != nil {
+				return err
+			}
+			var tools toolsSvc.Service
+			var ok bool
+			if tools, ok = svc.(toolsSvc.Service); !ok {
+				utils.LogError(logger, nil, "service doesn't satisfy normalise service interface")
+				return nil
+			}
+			if err := tools.Normalise(ctx, path, testSet, testCases); err != nil {
```
charankamarapu, line 78: Need to take a call on whether normalise will be in tools service or as separate service or in test..? Because normalise is going to update test which is in testDB. Majorly tools deals with smaller tasks like update, version, config generate etc.

cli/provider/cmd.go
```diff
@@ -155,6 +155,10 @@ func (c *CmdConfigurator) AddFlags(cmd *cobra.Command, cfg *config.Config) error
 	switch cmd.Name() {
 	case "update":
 		return nil
+	case "normalise":
```
charankamarapu, line 158: I think test report name (test-set-0-report) and test case name is enough to get to this normalisation because you can get whole information from that. And please add these required things in config also.

pkg/service/tools/service.go
```diff
@@ -1,12 +1,90 @@
...
 
 type Service interface {
 	Update(ctx context.Context) error
 	CreateConfig(ctx context.Context, filePath string, config string) error
+	Normalise(ctx context.Context, path string, testSet string, testCases string) error
 }
 
 type teleDB interface {
 }
+type TestCaseFile struct {
```
charankamarapu, line 18: I think this is already present in models

pkg/service/tools/tools.go
```diff
@@ -274,3 +277,109 @@ func (t *Tools) CreateConfig(_ context.Context, filePath string, configData stri
 	t.logger.Info("Config file generated successfully")
 	return nil
 }
+
+// Normalise initiates the normalise process for normalising the test cases.
```
charankamarapu, line 281: This will significantly change after resolving above comments so will review this file once above comments are addressed .

charankamarapu (commented):
Please address the comments

pkg/service/tools/service.go
Akash-Singh04: In models we have a strut for TestCase, not one for the entire testcase yaml file. Having this struct makes it easier for unmarshalling the testcase file to a struct and back to the yaml file.

cli/normalise.go
Akash-Singh04: Refactored

cli/normalise.go
Akash-Singh04: In my opinion, normalise command should indeed be a part of the tools service. While it does indeed update test which is in testDB, it serves more like an utility tool in my opinion , which is to be used whenever required. What is your opinion on this?

cli/provider/cmd.go
Akash-Singh04: Yes, those two flags would be compulsory, i.e. the test-report-name and the test cases(prural) to be normalised. On top of them, i thought adding the --path flag would allow the user to provide the path to the test report folder incase it wasn't the default "./".

Akash-Singh04:
@charankamarapu I have addressed the review comments. Looking forward to your review on them

cli/provider/cmd.go
Akash-Singh04: Also could you elaborate on what you mean by adding these to config?

charankamarapu: Actually we should give option to the user to provide the test-run also so the flags would test-run, test-set, test-case. regarding path flag it should be normal path flag which we get in record and replay commands. It shouldn't be specific to reports. We have a config file which will be used for provided info like path etc without flags. There is a package for it named config.

cli/provider/cmd.go
```diff
@@ -377,6 +381,35 @@ func (c CmdConfigurator) ValidateFlags(ctx context.Context, cmd *cobra.Command)
 				}
 			}
 		}
+	case "Normalise":
+		path := viper.GetString("path")
```
charankamarapu, line 385: You need not do this if add things in config please go through test and record command

pkg/service/tools/tools.go
```diff
@@ -283,3 +287,113 @@ func (t *Tools) CreateConfig(_ context.Context, filePath string, configData stri
...
+func (t *Tools) Normalise(_ context.Context) error {
```
charankamarapu, line 292: You should use yaml functions to edit the test-cases. You should how it is done in record and replay via interfaces of DBs . you can add this function in record or replay service.

charankamarapu (commented):
please address the comments

cli/normalise.go
```diff
@@ -0,0 +1,62 @@
...
+	"context"
+
+	"github.com/spf13/cobra"
+	"go.keploy.io/server/v2/config"
+	toolsSvc "go.keploy.io/server/v2/pkg/service/tools"
+	"go.keploy.io/server/v2/utils"
+	"go.uber.org/zap"
+)
+
+// func NewCmdNormalise(logger *zap.Logger) *Normalise {
```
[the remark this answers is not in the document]
Akash-Singh04: Refactored

cli/provider/cmd.go
Akash-Singh04: Refactored

cli/provider/cmd.go
Akash-Singh04: Refactored

pkg/service/tools/tools.go
Akash-Singh04: Added it to replay service

pkg/service/replay/replay.go
```diff
@@ -711,3 +713,84 @@ func (r *Replayer) ProvideMocks(ctx context.Context) error {
 	<-ctx.Done()
 	return nil
 }
+
+// Normalise initiates the normalise process for normalising the test cases.
+func (r *Replayer) Normalise(_ context.Context, cfg *config.Config) error {
+	path := cfg.Path
+	testSets := viper.GetString("test-sets")
```
charankamarapu, line 720: These are not needed use config struct instead of getting them like this. Please refer to how test or record command are accessing these kind of flags

pkg/service/replay/replay.go
```diff
@@ -711,3 +713,84 @@ func (r *Replayer) ProvideMocks(ctx context.Context) error {
...
+	if err != nil {
+		utils.LogError(r.logger, err, "Failed to read directory")
+		return err
+	}
+	// Iterate over each YAML file in the test run folder
+	for _, file := range files {
+		if strings.HasSuffix(file.Name(), ".yaml") {
+			filePath := filepath.Join(testRunPath, file.Name())
+			// Read the YAML file
+			yamlData, err := os.ReadFile(filePath)
```
charankamarapu, line 737: why are you not using the test or report dbs..?

charankamarapu (commented):
Please address the comments

pkg/service/replay/replay.go
Akash-Singh04: Will look into this

shivamsouravjha:
@Akash-Singh04 if the comments are resolved feel free to resolve them @charankamarapu would review the updated PR

Akash-Singh04:
@shivamsouravjha Need some time to properly look into the refactored codebase to resolve the comments,

shivamsouravjha:
Sure thing @Akash-Singh04 in the meanwhile you can move it as a draft,

PR #1719, opened 2024-03-21
fix: added bool variable in proxy struct to check if user app is closed

AkashKumar7902:
…n terminated or not

## Related Issue
  - Info about Issue or bug

## Type of change

## Checklist:

## Screenshots (if any)

### DIFF
--- a/pkg/core/core.go
+++ b/pkg/core/core.go
@@ -222,6 +222,8 @@ func (c *Core) Run(ctx context.Context, id uint64, opts models.RunOptions) model
 		if appErr.Err != nil {
 			utils.LogError(c.logger, appErr, "error while running the app")
 			appErrCh <- appErr
+		} else if appErr.AppErrorType == models.ErrCtxCanceled {
+			c.Proxy.SetUserAppTerminateInitiated(true)
 		}
 		return nil
 	})
--- a/pkg/core/proxy/proxy.go
+++ b/pkg/core/proxy/proxy.go
@@ -64,6 +64,7 @@ func New(logger *zap.Logger, info core.DestInfo, opts config.Config) *Proxy {
 		IP6:          "::1",          //default: "::1" <-> ([4]uint32{0000, 0000, 0000, 0001})
 		ipMutex:      &sync.Mutex{},
 		connMutex:    &sync.Mutex{},
+		appMutex:     &sync.Mutex{},
 		DestInfo:     info,
 		sessions:     core.NewSessions(),
 		MockManagers: sync.Map{},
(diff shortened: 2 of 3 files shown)


pkg/core/core.go
```diff
@@ -222,6 +222,8 @@ func (c *Core) Run(ctx context.Context, id uint64, opts models.RunOptions) model
 		if appErr.Err != nil {
 			utils.LogError(c.logger, appErr, "error while running the app")
 			appErrCh <- appErr
+		} else if appErr.AppErrorType == models.ErrCtxCanceled {
```
charankamarapu, line 225: I think checking here is a bit late. Because for some apps it takes some time to close . We can actually check on the context which this function is using and also when app itself has an error then also we can do that. So there will be two checks 
1. On the context of this function once context is cancelled we should signal proxy and 
2. When application stops and if it is not context err we need to send signal to proxy. Why we ar ignoring context cancel here is that if it is context cancel we will already know it in the first case.

pkg/core/core.go
```diff
@@ -222,6 +222,8 @@ func (c *Core) Run(ctx context.Context, id uint64, opts models.RunOptions) model
...
+			c.Proxy.SetUserAppTerminateInitiated(true)
```
charankamarapu, line 226: include app id in it. Use https://github.com/keploy/keploy/blob/e35909158d3728b25d58304be55ba1ff7c7e2972/pkg/core/service.go#L113 for storing where the app is closed or not. May be create a channel param for it.

pkg/core/proxy/proxy.go
```diff
@@ -333,7 +337,7 @@ func (p *Proxy) handleConnection(ctx context.Context, srcConn net.Conn) error {
 
 		//mock the outgoing message
 		err := p.Integrations["mysql"].MockOutgoing(parserCtx, srcConn, &integrations.ConditionalDstCfg{Addr: dstAddr}, m.(*MockManager), rule.OutgoingOptions)
-		if err != nil {
+		if err != nil && !p.IsUserAppTerminateInitiated() {
```
charankamarapu, line 340: Instead of here . Cancel the context [here](https://github.com/keploy/keploy/blob/e35909158d3728b25d58304be55ba1ff7c7e2972/pkg/core/proxy/proxy.go#L235) so that errors will automatically taken care.

charankamarapu (commented):
Please address the comments

pkg/core/proxy/proxy.go
AkashKumar7902: should I cancel "clientConnCtx" ?

charankamarapu:
closing this as this just needs a EOF check in err and is being done in https://github.com/keploy/keploy/pull/1727

PR #1723, opened 2024-03-26
Fix: Concurrent error in removeUnusedMocks

charankamarapu:
## Related Issue
  - Info about Issue or bug
issue :
image
changes:
1. instead of Map use sync.Map 
2. use update mocks instead of delete mocks

## Type of change

## Checklist:

## Screenshots (if any)

### DIFF
--- a/pkg/core/proxy/mockmanager.go
+++ b/pkg/core/proxy/mockmanager.go
@@ -2,33 +2,24 @@ package proxy
 
 import (
 	"fmt"
-	"sort"
-	"strconv"
-	"strings"
+	"sync"
 
 	"go.keploy.io/server/v2/pkg/models"
-)
-
-const (
-	filteredMock   = "filtered"
-	unFilteredMock = "unfiltered"
-	totalMock      = "total"
+	"go.uber.org/zap"
 )
 
 type MockManager struct {
-	filtered   *TreeDb
-	unfiltered *TreeDb
-	// usedMocks contains the name of the mocks as key which were used by the parsers during the test execution.
-	//
-	// value is an array that will contain the type of mock
-	usedMocks map[string][]string
+	filtered      *TreeDb
+	unfiltered    *TreeDb
+	logger        *zap.Logger
+	utilizedMocks sync.Map
 }
 
-func NewMockManager(filtered, unfiltered *TreeDb) *MockManager {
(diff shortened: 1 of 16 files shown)


pkg/service/replay/replay.go
```diff
@@ -380,13 +381,19 @@ func (r *replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
 			utils.LogError(r.logger, err, "failed to simulate request")
 			break
 		}
+		consumedFilteredMocks, err := r.instrumentation.GetConsumedFilteredMocks(runTestSetCtx, appID)
```
re-Tick, line 384: We need to print the unfiltered mocks as well because the dependencies like postgreSQL, generic have only config mocks `unfiltered` mocks. So need to merge both the consumed filtered and unfiltered mocks in sorted fashion.

charankamarapu: got it but I think we don't need sorting because when we update mocks we are going to read the mock file and  from that array of mocks that are read we will remove few. I guess we don't need sorting.

charankamarapu: Are suggesting sorting for printing the log..?

charankamarapu: done the changes you can review now

pkg/core/proxy/mockmanager.go
```diff
@@ -2,33 +2,33 @@ package proxy
...
+		filtered:      filtered,
+		unfiltered:    unfiltered,
+		logger:        logger,
+		consumedMocks: sync.Map{},
 	}
 }
 
 func (m *MockManager) SetFilteredMocks(mocks []*models.Mock) {
 	m.filtered.deleteAll()
-	m.utilizedFilteredMocks = sync.Map{}
```
re-Tick: I think it should clear the consumed mocks to print the mocks for a test-case run. Otherwise, mocks of test-1 will also be logged as consumed mocks for test-2

charankamarapu: clearing it in getconusmedMocks only, If we clear it in setfiltered mocks we will miss the mocks used up just after starting app and before running 1st test case

re-Tick: Got it. That is correct.

re-Tick (approved):

PR #1808, opened 2024-04-10
fix: folder permissions

charankamarapu:
## Related Issue
  - Info about Issue or bug

1. Keploy folder not getting deleted
2. This is happening because permissions were added to the top directories which are created but not all the directories which are created.
3. Added recursive adding of permissions

## Type of change

## Checklist:

## Screenshots (if any)


pkg/platform/yaml/yaml.go
```diff
@@ -188,3 +183,17 @@ func ReadSessionIndices(_ context.Context, path string, Logger *zap.Logger) ([]s
 	}
 	return indices, nil
 }
+
+func chmodRecursive(path string, mode os.FileMode) error {
```
slayerjain, line 187: @charankamarapu instead of changing the permissions on the file, can't we set the right permissions while creating the file itself?

charankamarapu: that is already done - https://github.com/keploy/keploy/blob/aaa712e2e8751f9b42cad4937f05d12ed26972fe/pkg/platform/yaml/yaml.go#L149

charankamarapu: I have created the files and these are the permissions

Screenshot 2024-04-11 at 12 13 04 PM

charankamarapu: can you please check the permissions for your folders..?

slayerjain: that is weird.

charankamarapu: please delete the complete folder and record from scratch.

slayerjain: I think the problem was in my laptop, I might have missed deleting the keploy folder properly. It worked as described after that.

slayerjain:
In summary, we should look into: 

1. Using the original user instead of root
2. set the permissions to 664
3. Try setting these permissions during create time itself, instead of setting it later.

charankamarapu:
1. regarding setting pemissions while creating file :
using os library functions like mkdirAll, OpenFile we cannot set the exact permissions while creating it we can set the permissions before unmask. that is we can give he permission like os.MkdirAll(path, 0777) but the final permissions will be  bitwise AND NOT with unmask bits which are decided by parent folder and the permissions get changed. So it is must that we should be using chmod function to change the permissions at the end.

2. Using the original user instead of root :
this will also be done by os.chown fucntion in the same way which we are using for chmod.

There are few other libraries like syscall but there are not many functions like mkdirAll etc which are useful in keploy.

We can go with the same approach which is already there and change the permissions to 664 and in the chmodRecursive function we can change the owner.

what are your thoughts on this @slayerjain ?

slayerjain (approved):
LGTM! 

Please just add some comments to explain how it works, and the implications of using via lima above  `syscall.Umask(0)`

Issue #1851, opened 2024-04-26
[bug]: race condition in keploy record


shivamsouravjha:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

* testCount is concurrently written and read in case of keploy record.

### Steps to reproduce

* Build the binary with `go build -race -tags=viper_bind_struct -o keploybin . && sudo mv keploybin /usr/local/bin/keploybin`
* Try running express-mongoose in record.
* While closing race condition occurs

### Environment

None

### Version

Cloud

### Repository

keploy

dxtym:
Hi! May I give this a try?

shivamsouravjha:
sure @dxtym  you can work on the issue!

PR #1867, opened 2024-05-07
fix: race condition in keploy record

dxtym:
## Related Issue
  - Race condition with `testCount` while exiting keploy record 

Closes: #1851

Put `sync.Mutex` locks over the sharing value so that concurrent activities don't cause race condition. Add `sync.waitGroup` to control the flow of actions between the main and sub gorountines.

## Type of change

No additional tests. Only the build procedure at #1851 

NA

## Checklist:

## Screenshots (if any)

| image | image |


dxtym:
Singed-off-by: Dilmurod Abdusamadov dilmurod.abdusamadov2004@gmail.com

dxtym:

pkg/service/record/record.go
```diff
@@ -66,13 +67,18 @@ func (r *Recorder) Start(ctx context.Context) error {
...
+	var waitGroup = &sync.WaitGroup{}
 	var mockCountMap = make(map[string]int)
 
 	// defering the stop function to stop keploy in case of any error in record or in case of context cancellation
 	defer func() {
 		select {
 		case <-ctx.Done():
+			waitGroup.Wait()
+			mutex.Lock()
 			r.telemetry.RecordedTestSuite(newTestSetID, testCount, mockCountMap)
```
slayerjain, line 80: I might be missing something but not really sure if this defer could be called concurrently 🤔

slayerjain: aah, so its the testCount probably..

slayerjain: I think in the defer function if we wait for the other goroutines to complete/cancel - this race condition can be better resolved. Maybe we can move the `runAppErrGrp.Wait()` in the defer to the top. 

@charankamarapu @gouravkrosx what do you think?

slayerjain (changes requested):
Hi @dxtym , Thanks for the PR! I have left some comments and I think @gouravkrosx / @charankamarapu should give more context for a better solution.

pkg/service/record/record.go
gouravkrosx: @dxtym , we can move the `r.telemetry.RecordedTestSuite(newTestSetID, testCount, mockCountMap)` to the bottom of the defer, this way we are waiting for all the go routines to complete their execution.

dxtym: Thank you for the comments @slayerjain @gouravkrosx! I moved the `r.telemetry.RecordedTestSuite(newTestSetID, testCount, mockCountMap)` to the bottom and removed mutex and waitGroups. Did I get your feedback right?

slayerjain: @dxtym yes, I think you've made the changes according to the feedback. 

It seems like `ctx` cancellation is not working correctly based on the this pipeline - https://github.com/keploy/keploy/actions/runs/9034964803/job/24829105617?pr=1867


It seems to be stuck on the last line: 

Screenshot 2024-05-10 at 9 55 22 PM

dxtym: image
I checked one of the previous PRs that passed all the tests. It had the same line. Could it be something else?

dxtym: What if I undo the changes to `ctx` cancelling part by not moving down? I did the previous solution with `mutex` while removing  the `waitGroup`, and no race condition occurred.  Perhaps, we could go with that )

dxtym: I tried removing `waitGroup` from my previous solution, and the binary worked the same way. Perhaps we could try that )

slayerjain:
@gouravkrosx please help close this

dxtym:
> @gouravkrosx please help close this

I would greatly appreciate support 🙏

gouravkrosx:
@dxtym, Sorry for the delay, will try to review this PR before EOD.

gouravkrosx:
@dxtym , In this [commit](https://github.com/keploy/keploy/pull/1867/commits/708485c049e675231d9eaa21dc01f79e26a2ad06), you not only moved the `r.telemetry.RecordedTestSuite(newTestSetID, testCount, mockCountMap)` but also moved the **_select_** statement which was listening to context done. That made this [pipeline](https://github.com/keploy/keploy/actions/runs/9034964803/job/24829105617?pr=1867) stuck.
You just need to move this `r.telemetry.RecordedTestSuite(newTestSetID, testCount, mockCountMap)` at the bottom outside the **_select_** scope but inside the **_defer_** function scope. And that should work.

dxtym:
@gouravkrosx Thank you for your guidance. Made some changes according to the feedback.

pkg/service/record/record.go
```diff
@@ -72,13 +71,14 @@ func (r *Recorder) Start(ctx context.Context) error {
 	defer func() {
 		select {
 		case <-ctx.Done():
-			r.telemetry.RecordedTestSuite(newTestSetID, testCount, mockCountMap)
+			break
```
gouravkrosx, line 74: This break is not required as the select statement is not in a for loop.

gouravkrosx:
@dxtym Please resolve the addressed comments and please fix the lint pipeline.

pkg/service/record/record.go
dxtym: Can I open another PR for this issue and close this one? It looks like my local copy became a bit of a mess. I really apologize for that.

gouravkrosx:
Closed as contributor made new PR: https://github.com/keploy/keploy/pull/1914

Issue #1839, opened 2024-04-19
feat: `fallBackOnMiss` should be implemented in all the parsers


gouravkrosx:
- `fallBackOnMiss` is a flag used to enable passthrough( connecting to the actual destination service) if no mocks are found for a particular request during test mode.
- Since connecting to the real service in test mode may cause weird behaviors in matching for further requests, so to avoid this it should be by default false, currently only the HTTP parser has this option. 
- It should be implemented across all the parsers.
- Ref: https://github.com/keploy/keploy/blob/5645ab586c5f707dcc11f3025d265f569141d455/config/config.go#L61

gouravkrosx:
The issue is not relevant, in the current approach we should not even fallback.

PR #1889, opened 2024-05-22
feat: fallBackOnMiss for parsers

khareyash05:
Closes: #1839 

 Add FallBackOnMiss for Parser(missing for grpc parser)

## Type of change


pkg/core/proxy/integrations/generic/decode.go
```diff
@@ -69,18 +69,19 @@ func decodeGeneric(ctx context.Context, logger *zap.Logger, reqBuf []byte, clien
...
+					}
+					genericRequests = [][]byte{}
+					logger.Debug("the request buffer after pass through in generic", zap.Any("buffer", string(reqBuffer)))
+					if len(reqBuffer) > 0 {
+						genericRequests = [][]byte{reqBuffer}
+					}
+					logger.Debug("the length of genericRequests after passThrough ", zap.Any("length", len(genericRequests)))
 				}
-				logger.Debug("the length of genericRequests after passThrough ", zap.Any("length", len(genericRequests)))
 				continue
```
gouravkrosx, line 85: I don't think we should continue if we can't find any mocks and the passthrough is disabled. Could you check the behavior? You can do this by recording some generic mocks, deleting a few, and running in test mode. Ideally, we should only continue if the passthrough is successful. Otherwise, we should return an error due to the absence of mocks. This condition is valid for all the parsers.

khareyash05: Yeah this should be done! On it!

Issue #1917, opened 2024-06-03
[bug]: read indices checks file too instead of just checking test set


shivamsouravjha:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

* If a file is present in keploy, beside a dir of test set ,keploy tries to read it 

### Steps to reproduce

* Add a gitignore in keploy folder

### Environment

Release

### Version

Cloud

### Repository

keploy

PR #1918, opened 2024-06-03
chore: check for file type for finding indices

shivamsouravjha:
## Related Issue
  - Read only dirs while calculating indices.

Closes: https://github.com/keploy/keploy/issues/1917

While calculating indices we must just check for the folder types.


gouravkrosx (approved):
LGTM!

Issue #1919, opened 2024-06-03
feat: add support for global passthrough


gouravkrosx:
- This will enable the user to test their API end-to-end with the actual external service using the recorded test cases.

PR #1920, opened 2024-06-03
feat: add support for global passthrough in test mode

gouravkrosx:
## Related Issue
  - Support for global passthrough of external calls

Closes: https://github.com/keploy/keploy/issues/1919

- Add a flag to disable mocking in test mode. You can use it like, `keploy test -c "<appCMD>" --mocking=false`
- Move the Rerecord flag to only test in the cmd package.

## Type of change

`NA`

## Checklist:

## Screenshots (if any)

### DIFF
--- a/pkg/core/proxy/util.go
+++ b/pkg/core/proxy/util.go
@@ -1,9 +1,18 @@
 package proxy
 
 import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net"
 	"os"
 
+	pUtil "go.keploy.io/server/v2/pkg/core/proxy/util"
+	"go.keploy.io/server/v2/pkg/models"
+	"go.keploy.io/server/v2/utils"
 	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
 )
 
 // writeNsswitchConfig writes the content to nsswitch.conf file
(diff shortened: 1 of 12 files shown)


pkg/core/proxy/util.go
```diff
@@ -16,3 +25,60 @@ func writeNsswitchConfig(logger *zap.Logger, nsSwitchConfig string, data []byte,
...
+
+	// read requests from client
+	g.Go(func() error {
+		defer pUtil.Recover(logger, client, nil)
+		defer close(clientBuffChan)
+		pUtil.ReadBuffConn(ctx, logger, client, clientBuffChan, errChan)
+		return nil
+	})
+	// read responses from destination
+	g.Go(func() error {
```
slayerjain, line 51: Can we make these method less repetitive? 

The entire block below is repeating: 
```
g.Go(func() error {
		defer pUtil.Recover(logger, client, nil)
		defer close(clientBuffChan)
		return nil
	})
```
Are these repeating elsewhere too?

slayerjain (changes requested):
Apart from the repeating code. All looks fine to me. 

Can we make the flag smaller and more intuitive? Maybe `--mocking: false` or something?

pkg/core/proxy/util.go
gouravkrosx: Yeah, we can make it less repetitive. This was referenced from the generic parser. I will make the changes in that file also.

slayerjain (approved):
LGTM!

PR #1950, opened 2024-06-10
fix: sudo access in Darwin installation

charankamarapu:
1. Added sudo in installation script
2. remove error msgs of docker from installation script


slayerjain (approved):
LGTM!

PR #1974, opened 2024-06-18
refactor: rerecord feature for better user experience

gouravkrosx:
## Related Issue
  - Move rerecord from flag to a new command

Closes: https://github.com/keploy/keploy/issues/1975

- Added a new command called `rerecord` and removed the old flag `rerecord`. Usage: `keploy rerecord -c "<appCmd>  -t "test-set-1,test-set-2"`
- Refactor record service.
- Added new orchestrator service.

## Type of change

`NA`

## Checklist:

## Screenshots (if any)


.gitignore
```diff
@@ -46,6 +46,9 @@ test-reports
 **/.DS_Store
 **/._.DS_Store
 
+#Ignore the zip files
+*.zip
```
slayerjain, line 50: nice to have this

cli/provider/cmd.go
```diff
@@ -208,42 +196,47 @@ func (c *CmdConfigurator) AddFlags(cmd *cobra.Command) error {
...
-			cmd.Flags().Bool("ignore-ordering", c.cfg.Test.IgnoreOrdering, "Ignore ordering of array in response")
-			cmd.Flags().Bool("coverage", c.cfg.Test.Coverage, "Enable coverage reporting for the testcases. for golang please set language flag to golang, ref https://keploy.io/docs/server/sdk-installation/go/")
-			cmd.Flags().Bool("remove-unused-mocks", c.cfg.Test.RemoveUnusedMocks, "Clear the unused mocks for the passed test-sets")
-			cmd.Flags().Bool("go-coverage", c.cfg.Test.GoCoverage, "Enable go coverage reporting for the testcases")
-			cmd.Flags().Bool("fallBack-on-miss", c.cfg.Test.FallBackOnMiss, "Enable connecting to actual service if mock not found during test mode")
-			cmd.Flags().String("base-path", c.cfg.Test.BasePath, "Custom api basePath/origin to replace the actual basePath/origin in the testcases; App flag is ignored and app will not be started & instrumented when this is set since the application running on a different machine")
-			cmd.Flags().Bool("mocking", true, "enable/disable mocking for the testcases")
-		} else {
+
+		switch cmd.Name() {
```
slayerjain, line 221: switch case inside switch case. maybe move this to another method?

pkg/service/orchestrator/rerecord.go
```diff
@@ -0,0 +1,252 @@
...
+		o.logger.Info("Re-recording testcases for the given testset", zap.String("testset", testSet))
+
+		errGrp, _ := errgroup.WithContext(ctx)
+		recordCtx := context.WithoutCancel(ctx)
+		recordCtx, recordCtxCancel := context.WithCancel(recordCtx)
+
+		var errCh = make(chan error, 1)
+		var replayErrCh = make(chan error, 1)
+
+		select {
```
slayerjain, line 71: why do we have 2 back-2-back selects?

pkg/service/orchestrator/rerecord.go
```diff
@@ -0,0 +1,252 @@
...
+	if err != nil {
+		errMsg = "failed to extract host and port"
+		utils.LogError(o.logger, err, "")
+		o.logger.Debug("", zap.String("curl", tcs[0].Curl))
+		return false, fmt.Errorf(errMsg)
+	}
+	cmdType := utils.CmdType(o.config.CommandType)
+
+	var userIP string
+	if utils.IsDockerKind(cmdType) {
```
slayerjain, line 199: maybe we should rename `IsDockerKind` to `isDockerCmd`

pkg/util.go
```diff
@@ -271,3 +273,54 @@ var (
...
+
+	timer := time.NewTimer(timeout)
+	defer timer.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+			conn, err := net.DialTimeout("tcp", net.JoinHostPort(host, port), 1*time.Second)
```
slayerjain, line 313: since your ticker is 1 second, your timeout here should be less than 1 sec

slayerjain (changes requested):
I have added comments as per my quick observations.

pkg/service/orchestrator/rerecord.go
gouravkrosx: To avoid calling `o.replayTests(recordCtx, testSet)` if the ctx is done just after the first go routine. Since  `replayTests` is a blocking call.

cli/provider/cmd.go
gouravkrosx: You mean same 2nd switch case in a different method?

pkg/service/orchestrator/rerecord.go
```diff
@@ -0,0 +1,252 @@
...
+			utils.LogError(o.logger, err, "failed to stop re-recording")
+		}
+
+		// Check if the global context is done after each iteration
+		if ctx.Err() != nil {
+			break
+		}
+	}
+
+	if stopReason == "" {
```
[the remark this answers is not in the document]
gouravkrosx: yes, I guess.

pkg/service/orchestrator/rerecord.go
gouravkrosx: sure

pkg/util.go
gouravkrosx: Didn't get this point. why should it be less than 1 sec?

pkg/service/orchestrator/rerecord.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+		return nil
+	}
+
+	utils.LogError(o.logger, err, stopReason)
+	return fmt.Errorf(stopReason)
+}
+
+func (o *Orchestrator) replayTests(ctx context.Context, testSet string) (bool, error) {
+	var err error
```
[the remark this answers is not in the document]
gouravkrosx: `err` can be moved near to `userIP` variable and is used to avoid shadowing of outside defined variables. `errMsg` is used from the starting log itself hence declared at top.

pkg/service/orchestrator/rerecord.go
gouravkrosx: Here we've used child context without cancel to avoid the cancellation of the parent context. When we use errgroup and get an error from any of the go routines spawned by errgroup, it cancels the parent context. We don't want to stop the execution if there is an error in any of the test-set recording sessions, it should just skip that test-set and continue with the next one.

Keeping two back-to-back selects is used to not do blocking operation if parent ctx is done

gouravkrosx:
@slayerjain Addressed all the comments.

slayerjain (approved):

slayerjain:
LGTM!

Issue #1868, opened 2024-05-08
[feature]: add test pipeline for testing `Postgres` parser using different applications


gouravkrosx:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

- Currently, there's a test pipeline for the gin-mongo application, which is a combination of Golang and MongoDB, aimed at testing the Mongo parser. ([1](https://github.com/keploy/keploy/blob/44109632096744c364a1ec081261c01d02e6d830/.github/workflows/build_stage.yml#L23)), ([2](https://github.com/keploy/keploy/blob/44109632096744c364a1ec081261c01d02e6d830/.github/workflows/build_stage.yml#L27))
- Likewise, we aim to establish pipelines for all supported languages using Postgres as the database to test the Postgres parser.
- You can contribute to this effort by referring to our documentation on how to extend coverage for Keploy or understand its self-testing approach.
- Docs Link: https://keploy.io/docs/keploy-explained/testing-guide/

### Why should this be worked on?

- This initiative will enhance Keploy's coverage and benefit all users by supporting its testing capabilities through the concept of Keploy testing Keploy.
- It will promote confidence in reviewing pull requests, particularly those involving changes to the Postgres parser, by ensuring thorough testing and validation.

### Repository

keploy

Redarcher9:
Hi @gouravkrosx , Can I pick this up?

gouravkrosx:
@Redarcher9 
Yes, you can try, if you face any problems while following the docs, please let us know.

Redarcher9:
Thanks @gouravkrosx ,
I can see that there is a sample for fasthttp-postgres in go-samples repo, can i use this one?

gouravkrosx:
@Redarcher9 You can use [echo-sql](https://github.com/keploy/samples-go/tree/main/echo-sql) application.

Redarcher9:
Hi @gouravkrosx ,
I think I'm very close to fix this,
I am getting an error in my PR([link](https://github.com/Redarcher9/keploy/pull/1)) for postgres database connectivity in github actions tab([link](https://github.com/Redarcher9/keploy/actions/runs/9041498051/job/24846931347)).



![Screenshot 2024-05-11 at 12 19 50 PM](https://github.com/keploy/keploy/assets/19843336/a6469fdb-f793-4d4d-ade2-216f59bdc287)



[...]

PR #1980, opened 2024-06-20
feat: postgres parser workflows

khareyash05:
## Related Issue
  - Added fasthttp-postgres application into postgres workflows

Closes: #1868


khareyash05:
the workflows will fail due to potsgres url not being passed correctly working on the same

PR #1992, opened 2024-06-25
fix: denoise bugs

charankamarapu:
1. Noise in the reports was wrong because we are explicitly converting noise into lower case. Removed this.
2. Remove auto-detection of noise while updating test case because when we remove noise, still the noise is getting appended.


Sarthak160 (approved):
LGTM

PR #1994, opened 2024-06-26
fix: remove user docker compose file after use

khareyash05:
Removes the user docker compose file after usage

## Type of change

### DIFF
--- a/main.go
+++ b/main.go
@@ -72,6 +72,7 @@ func start(ctx context.Context) {
 	}
 	defer utils.DeleteLogs(logger)
 	defer utils.Recover(logger)
+	defer utils.DeleteComposeFile(logger)
 
 	// The 'umask' command is commonly used in various operating systems to regulate the permissions of newly created files.
 	// These 'umask' values subtract from the permissions assigned by the process, effectively lowering the permissions.
--- a/main.go
+++ b/main.go
@@ -71,10 +71,15 @@ func start(ctx context.Context) {
 		return
 	}
 	defer func() {
-		if err := utils.DeleteKeployMetaData(logger); err != nil {
-			fmt.Println("Failed to delete Keploy meta data", err)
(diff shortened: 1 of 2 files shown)


utils/utils.go
```diff
@@ -164,6 +164,22 @@ func DeleteLogs(logger *zap.Logger) {
 	}
 }
 
+func DeleteComposeFile(logger *zap.Logger) {
```
gouravkrosx, line 167: This is a duplicate code, the same code is there for deleteLogs, please pass the fileName and change the method name. Throw an error in the function and handle it at the caller function.

utils/utils.go
```diff
@@ -148,20 +148,21 @@ func LogError(logger *zap.Logger, err error, msg string, fields ...zap.Field) {
...
-func DeleteLogs(logger *zap.Logger) {
-
-	//Check if keploy-log.txt exists
-	_, err := os.Stat("keploy-logs.txt")
+func DeleteFileIfNotExists(logger *zap.Logger, name string) (err error) {
+	//Check if file exists
+	_, err = os.Stat(name)
 	if os.IsNotExist(err) {
-		return
+		return err
```
gouravkrosx, line 155: return nil here.

main.go
```diff
@@ -70,7 +70,17 @@ func start(ctx context.Context) {
 		fmt.Println("Failed to start the logger for the CLI", err)
 		return
 	}
-	defer utils.DeleteLogs(logger)
+	defer func() {
+		if err := utils.DeleteFileIfNotExists(logger, "keploy_logs.txt"); err != nil {
+			fmt.Println("Failed to delete Keploy Logs", err)
```
gouravkrosx, line 75: use utils logger (LogError) instead of print statements.

gouravkrosx (approved):
LGTM!

PR #1999, opened 2024-06-28
fix: kebabCase bug for nested flag

charankamarapu:
Fixed kebab error for nested flags and nested structs in config


gouravkrosx (approved):
LGTM!

PR #2037, opened 2024-07-05
fix: fixing coverage command

shivamsouravjha:
## Related Issue
 * CoverageCommand isn't needed as it is just overwritten by coverage flag.

### DIFF
--- a/cli/provider/cmd.go
+++ b/cli/provider/cmd.go
@@ -529,8 +529,6 @@ func (c *CmdConfigurator) ValidateFlags(ctx context.Context, cmd *cobra.Command)
 				return nil
 			}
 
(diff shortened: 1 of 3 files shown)


gouravkrosx (approved):
LGTM!

charankamarapu:
what if you want to log the original command. We generally do that before running the application. Now the complete command used for coverage will get printed, which doesn't look like good experience to user.

shivamsouravjha:
This command is overwritten in the cloud scenario. If we decide to log the original command, we can introduce a parameter similar to `CoverageCommand`. However, logging the original command is not a current feature request, so having unused code that breaks existing functionality is not advisable.

Moreover, logging the command is debatable since the user already enters it and it remains visible to them, even in the configuration file.

PR #2050, opened 2024-07-08
feat: in CI flag

khareyash05:
Closes: #2051 

#### Describe the changes you've made
Added `in-ci` flag which checks if the scripts are being used in CI or not

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Code style update (formatting, local variables)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

PR #2055, opened 2024-07-10
feat: rerecord workflows for keploy

khareyash05:

Rerecord workflows for OSS
## Type of change

### DIFF
--- a/.github/workflows/golang_docker.yml
+++ b/.github/workflows/golang_docker.yml
@@ -13,7 +13,7 @@ jobs:
 
     - name: Build binary
       run: |
-        go build  -tags=viper_bind_struct -o keployv2
+        go build -race -tags=viper_bind_struct -o keployv2
 
     - name: Checkout the samples-go repository
       uses: actions/checkout@v2
--- a/.github/workflows/golang_docker.yml
+++ b/.github/workflows/golang_docker.yml
@@ -13,7 +13,7 @@ jobs:
 
     - name: Build binary
       run: |
-        go build -race -tags=viper_bind_struct -o keployv2
+        go build -tags=viper_bind_struct -o keployv2
 
     - name: Checkout the samples-go repository
       uses: actions/checkout@v2
--- a/.github/workflows/test_workflow_scripts/golang-docker.sh
+++ b/.github/workflows/test_workflow_scripts/golang-docker.sh
@@ -96,6 +96,35 @@ if grep "WARNING: DATA RACE" "${test_container}.txt"; then
     cat "${test_container}.txt"
     exit 1
 fi
+sleep 10
+
+sudo -E env PATH=$PATH ./../../keployv2 rerecord -c 'docker run -p8080:8080 --net keploy-network --name ginApp_test gin-mongo' --containerName "$test_container" --inCi=true &> "${test_container}.txt"
+
+if grep "ERROR" "${test_container}.txt"; then
+    echo "Error found in pipeline..."
+    cat "${test_container}.txt"
+    exit 1
+fi
+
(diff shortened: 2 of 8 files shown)


khareyash05:
These CI will be successful upon successful merging of https://github.com/keploy/keploy/pull/2050

.github/workflows/golang_docker.yml
```diff
@@ -13,7 +13,7 @@ jobs:
 
     - name: Build binary
       run: |
-        go build  -tags=viper_bind_struct -o keployv2
+        go build -race -tags=viper_bind_struct -o keployv2
```
shivamsouravjha, line 17: incase of docker we run via docker image of keploy, binary is just needed to start the said docker image thus I didn't put the flag in pipeline as it wasn't needed.

.github/workflows/test_workflow_scripts/golang-docker.sh
```diff
@@ -61,8 +61,8 @@ send_request(){
     wait
 }
 
+container_name="ginApp_${i}"
 for i in {1..2}; do
-    container_name="ginApp_${i}"
     send_request &
     sudo -E env PATH=$PATH ./../../keployv2 record -c "docker run -p8080:8080 --net keploy-network --rm --name $container_name gin-mongo" --container-name "$container_name" --generateGithubActions=false &> "${container_name}.txt"
```
shivamsouravjha, line 67: wouldn't this move it out of loop ? like for different container name  it was loke ginApp_1,ginApp_2..now it would be ginApp_ as I isn't defined

.github/workflows/test_workflow_scripts/golang-docker.sh
```diff
@@ -80,10 +80,27 @@ for i in {1..2}; do
 
     echo "Recorded test case and mocks for iteration ${i}"
 done
+sleep 10
```
shivamsouravjha, line 84: this sleep isn't needed as send_request only exits when keploy record is stopped, so this would add unnecessary delay of 10 seconds

.github/workflows/test_workflow_scripts/golang-docker.sh
```diff
@@ -80,10 +80,27 @@ for i in {1..2}; do
...
+fi
+# Monitor Docker logs for race conditions during testing.
+if grep "WARNING: DATA RACE" "${test_container}.txt"; then
+    echo "Race condition detected in test, stopping pipeline..."
+    cat "${test_container}.txt"
+    exit 1
+fi
+
+sleep 10
+
```
shivamsouravjha, line 101: not needed adds additional unwanted delay

.github/workflows/test_workflow_scripts/golang-docker.sh
```diff
@@ -80,10 +80,27 @@ for i in {1..2}; do
...
+rerecord_container="ginApp_rerecord"
+sudo -E env PATH=$PATH ./../../keployv2 rerecord -c "docker run -p8000:8000 --rm --name $rerecord_container --network keploy-network gin-mongo" --containerName "$rerecord_container" --inCi=true &> "${rerecord_container}.txt"
 
 if grep "ERROR" "${test_container}.txt"; then
```
shivamsouravjha, line 105: we're writing logs into rerecord_container.txt but checking test_container.txt for error or race, needs to be fixed.

shivamsouravjha:
also incase of rerecord we should also consider checking the report files generated.
In current flow we won't be checking the report generated by running test sets generated via rerecord.
Thus if those testsets are wrongly generated we would still be passing them , as it is unchecked.

PR #2057, opened 2024-07-10
fix: ignore testCase bug

charankamarapu:
1. Ignore testCase is ignoring complete testSet.
2. Need add one more condition for skipping testSet.


Sarthak160 (approved):
LGTM

PR #2058, opened 2024-07-10
fix: add ignored field in test run summary

charankamarapu:
Added ignored field in the test reports and test run summary.


Sarthak160 (approved):
Ship it !

Issue #2076, opened 2024-07-16
[bug]: internal config files are added in keploy config


shivamsouravjha:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

* when a keploy config is generated the internal config files are appended to it as well that shouldn't be the case.

### Steps to reproduce

* run keploy config --generate

the flags
```
enableTesting: false
keployContainer: "keploy-v2"
keployNetwork: "keploy-network"
inDocker: false
cmdType: "native"
fallbackOnMiss: false
```
shouldn't be present.

### Environment

None

### Version

Cloud

### Repository

keploy

ayush3160:
I am working on this

gouravkrosx:
@shivamsouravjha Please look into this.

yaten2302:
@gouravkrosx @shivamsouravjha , I've a doubt regarding this issue, when we run - `keploy config --generate`, then `keploy.yml` is generated, but what are these internal config files? Are these files which are already present (maybe generated from some other pkg) Could you please share some steps to reproduce this? When I do `keploy config --generate`, then only `keploy.yml` is generated.

PR #2078, opened 2024-07-16
fix: Removed Internal Config from keploy config

ayush3160:
## Related Issue

Closes: #2076 

I have removed the internal config merging with default config because of which now keploy config only shows default config

## Type of change

NA

## Checklist:

## Screenshots (if any)
NA


ayush3160:

shivamsouravjha (changes requested):
the build is failing, please look into it. Seems like the changes made are breaking the workflow.

ayush3160:
@shivamsouravjha , I have checked locally and it is building successfully. Can you please help to find what could be the possible solution to this.

shivamsouravjha:
* I tried this and had an issue. Our normal experience is  when we run keploy test a keploy config is created. But the issue is  when we run keploy test the config is that it has params~
```
generateGithubActions: true
keployContainer: keploy-v2
keployNetwork: keploy-network
cmdType: native
inCi: false
```

We don't want these to be present in the keploy.yml file as well. That part of solution doesn't seem to be implemented in the current PR.

ayush3160:
@shivamsouravjha , I have done the changes for other commands also. Please have a look and let me know if any changes are needed.

shivamsouravjha (changes requested):
I tried running your changes locally with docker container and here is what I found ~ 
image
image

Please test the case of docker image as well.

shivamsouravjha (approved):
LGTM!

cli/provider/cmd.go
```diff
@@ -616,7 +617,11 @@ func (c *CmdConfigurator) ValidateFlags(ctx context.Context, cmd *cobra.Command)
 func (c *CmdConfigurator) CreateConfigFile(ctx context.Context, defaultCfg config.Config) error {
 	defaultCfg = c.UpdateConfigData(defaultCfg)
 	toolSvc := tools.NewTools(c.logger, nil)
-	configData := defaultCfg
+	configData, err := c.RemoveInternalParameters(defaultCfg)
```
charankamarapu, line 620: you can combine RemoveInternalParameters and yaml marshal into one function.

cli/provider/cmd.go
```diff
@@ -644,3 +649,21 @@ func (c *CmdConfigurator) UpdateConfigData(defaultCfg config.Config) config.Conf
...
+func (c *CmdConfigurator) RemoveInternalParameters(defaultCfg config.Config) (map[string]interface{}, error) {
+	var cfgMap map[string]interface{}
+	err := mapstructure.Decode(defaultCfg, &cfgMap)
+
+	if err != nil {
+		utils.LogError(c.logger, err, "error decoding struct")
+		return nil, errors.New("error decoding struct")
+	}
+
+	fieldsToRemove := []string{"generateGithubActions", "keployNetwork", "keployContainer", "inCi", "cmdType"}
```
charankamarapu, line 662: For this you can create a function in config package itself. Once user adds something in internal config it should automatically be deleted from the config. We should not mention it again in another place

charankamarapu (changes requested):
Please address the comments. Rest looks fine to me.

cli/provider/cmd.go
ayush3160: Done

config/default.go
```diff
@@ -64,6 +64,8 @@ inDocker: false
 cmdType: "native"
 `
 
+var InternalConfigFields = []string{"generateGithubActions", "keployNetwork", "keployContainer", "inCi", "cmdType"}
+
 var config = &Config{}
```
shivamsouravjha, line 69: why make it global when we have a function to fetch it as a string? You can make this a local variable.

ayush3160: Done.

ayush3160:
@shivamsouravjha , have done the changes please have a look.

config/default.go
```diff
@@ -110,3 +110,10 @@ func mergeStrings(srcStr, destStr string, infer bool, mergeOptions yaml.MergeOpt
 
 	return result.String()
 }
+
+func GetInternalConfigFields() []string {
+
+	InternalConfigFields := []string{"generateGithubActions", "keployNetwork", "keployContainer", "inCi", "cmdType"}
+
+	return InternalConfigFields
```
shivamsouravjha, line 118: lets keep it `internalConfigFields` as its the standard across repo

ayush3160: Done

shivamsouravjha (approved):
LGTM!

shivamsouravjha (commented):
If we've a case where in config we add a param
```	
Random        string       `json:"random" yaml:"random" mapstructure:"-"`
```

here in this case if we run this the value random won't be written in the keploy.yaml file because we're dependent on mapstructure. This can act as a blocker.

Since this value still has its yaml tag it should be written in the yaml file which currently isn't.

gouravkrosx:
Closing due to activity.

Issue #2071, opened 2024-07-15
[feature]: new version is available should be prompted


Sonichigo:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

Similar to zsh shell or any other CLI, once keploy has new release a user should be prompted just once that a new version is available and whether they want to update the version or not. In case the user doesn't want to update and enter [N] or [No], they shouldn't be prompt again in next turn.

### Why should this be worked on?

Usually, people are not aware that a new keploy version is released, and continue to work on old version till they see the release note or community call.

### Repository

keploy

Adity20:
Can I take this up @Sonichigo.

yaten2302:
Hey @Adity20 , can we collaborate on this issue please?

Adity20:
Sure @yaten2302

yaten2302:
@Sonichigo @Adity20 , actually, I've a doubt that whenever an update is available and if the user enters [N], then a `.keploy` file should be created in the root dir of the user's project to store that the user has opted not to update?

Sonichigo:
So a `.keploy` is already created in root home folder whenever a person installs keploy, since this file contains keploy installation ID. In the same file, the user's preferrence should be updated if he has opted not to update keploy.

yaten2302:
@Sonichigo sir, for this, I just thought of an approach, could you please guide that is it good or needs some improvement?

What we can do is that, in all the commands like - `keploy`, `keploy record`, `keploy test` etc... (except in `keploy update`) we can add a function that if a new version is available, then the user will be prompted with a message that a new version is available, click Y for yes and N for no. If the user prompts N, then the response be saved in the `.keploy` file and the message will not be shown again.

yaten2302:
Got it @Sonichigo sir👍 , also, could you please guide that how will we know that an update has been released by Keploy? Like, whenever a release is published it's published on github, then how we'll recognize that?

Adity20:
Same question @Sonichigo

Sonichigo:
> Got it @Sonichigo @Swpn0neel sir👍 , also, could you please guide that how will we know that an update has been released by Keploy? Like, whenever a release is published it's published on github, then how we'll recognize that?

The releases are available on GitHub, you can fetch from there keploy already has cmd named `update` which you can refer.

yaten2302:
Hey @Adity20 , ig I got that. As explained above by @Sonichigo sir, what we can do is that:
[...]

Issue #2047, opened 2024-07-08
[feature]: Add New Version Available Message to keploy


Swpn0neel:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

Notify users of a new version of keploy being available everytime keploy runs, the message would look like this:
 
 
![image](https://github.com/keploy/keploy/assets/114267538/e04d3663-3f10-4a06-a902-b7a805229a9f)

### Why should this be worked on?

Helps users be updated with the latest keploy updates.

### Repository

keploy

jaiakash:
![image](https://github.com/user-attachments/assets/67476856-d414-40bc-a105-8b37bfa2e641)
I think the update prompt should be something like this as the above image, right?

I will start working on that. Excited for KROC ✨

2nd, isn't this issue the same as #2071 ?

Swpn0neel:
@shivamsouravjha @Sonichigo Kindly assist @jaiakash on this one!

Sonichigo:
@jaiakash the issue #2071 is a different ux for the user, both are linked and split into two.

Sonichigo:
hey @jaiakash any progress?

jaiakash:
Hi @Sonichigo  i have added the PR which works 'fine'. 
There are some doubt from my side to handle edge cases. I have commented that in PR. Please check

Sarthak160:
@jaiakash Please resolve the linter issue. Also I ran it locally, ideally when user is using local dev version this functionality is not needed. It should have a check if version has `-dev` as prefix it should not work. Also If user enters 'y' for the update, all other process like keploy test or record should wait until the newer version is installed successfully .

PR #2098, opened 2024-07-19
Added new version available message to keploy whenever there is update

jaiakash:
## Related Issue
  - This PR added a feature that displays a new version available message (if any) whenever their keploy is run.

Closes: #2047 
Closes: #2071 

Add function in `main.go` that check `currentVersion` with `latestVersion` from github API and displays message accodingly.

## Type of change
- [x] This change requires a documentation update

## Checklist:

## Screenshots (if any)

| No SS | 
Screenshot 2024-07-19 at 1 36 51 PM
 |


jaiakash:

jaiakash:
Hi @Sonichigo @Swpn0neel I have used https://github.com/keploy/keploy/blob/55df9dc18bb1f1208ab7e57d5b7afea0900709a5/utils/utils.go#L313 function to check for the latest version and added the newly created [`checkForUpdate`](https://github.com/jaiakash/keploy/blob/676c9820f749d69b8041c3c6bf95d0db8bc91f93/main.go#L59) in `main.go` to check for new update on every run of Kelpoy.

There are some queries from my side.
1. For now the text is printed as plain text. If we want the design as outline in issue #2047 , we have to use external go library like [term](https://github.com/buger/goterm).  Shall I use that or is there a better way?
2. This new `checkForUpdate` is run everytime keploy is run. But shouldn't we hide that message when user directly run `keploy update` to hide repetitive message? Lmk what you think.

Sarthak160 (commented):
@jaiakash Please move this method to utils itself. Its better to implement less function definations in main.

Swpn0neel (changes requested):
If the user is using the latest version, we shouldn't print it (to avoid repetition)

jaiakash:
> If the user is using the latest version, we shouldn't print it (to avoid repetition)

Is user is using the latest version it wont be printed.

Swpn0neel (commented):
@Sarthak160 Currently they are printing the statement in plain text without any decorations, so should we keep that? Otherwise, they can use an external dependency to decorate it...

jaiakash:
Hi @Sarthak160 @Swpn0neel any update on the formating of the text? As of now this is now its printed

Screenshot 2024-07-19 at 1 36 51 PM

We can use libraries like [term](https://github.com/buger/goterm) to decorate it. Let me know whats the final decision.

Sarthak160:
@jaiakash We can simply add a progress bar which will pop if user accepts to download the newer version.

jaiakash:
> @jaiakash We can simply add a progress bar which will pop if user accepts to download the newer version.

@Sarthak160 Do you mean the progress bar when a user downloads the update? This PR only add the message that shows the availability of update whenever keploy is run. User have to manually type `keploy update` to update keploy. 

That progress bar makes more sense if we add the progress bar in that command. Let me know what you think and hence I can update the PR accodingly.

Swpn0neel:
So, if the user is on an outdated version, we'll show them that (which are already doing right now). Then, we'll prompt the user if they want to update the latest version or not... If yes, then we'll update it and it should've a progress bar for that!

jaiakash:
Understood, so i have to add progress bar for update operation (it can be when user press `y` during prompt of new update or user directly runs `keploy update` command). 

What I am doing right now, I would edit the logic at https://github.com/keploy/keploy/blob/e0139bb35646549441670355c51f36615cece39e/cli/update.go#L19

to show progress bar. 

Can I use this library - https://github.com/schollz/progressbar for progress bar.
@Swpn0neel please feel free to give your suggestions.

Swpn0neel:
Could you please check if we're already using a library for progress bar or not? If not, then we can surely proceed with the one you suggested!

Sonichigo:
@jaiakash any update?

jaiakash:
Hello I need a bit of help. 
My approach for now is, in case user press `y` during prompt, it should execute same code of `keploy update`, this https://github.com/jaiakash/keploy/blob/7a71cbe1bb3eff31799448b90298401fd1386993/pkg/service/tools/tools.go#L44 
 function. And I will add the progress bar logic there so we don't have to rewrite update logic.
 
Right not the problem I am facing is that if I import that function in https://github.com/jaiakash/keploy/blob/7a71cbe1bb3eff31799448b90298401fd1386993/utils/ctx.go#L38 

go is giving me `import cycle not allowed` error. How can I run the cli command `keploy update` directly form my https://github.com/jaiakash/keploy/blob/7a71cbe1bb3eff31799448b90298401fd1386993/utils/ctx.go#L38 function.

Sarthak160:
@jaiakash Ideally it should not prompt when the user is created its own -dev version. Also If we are not updating the keploy version automatically or any action after taking the user input y or n, Then what is the need to take the user's opininion.

Issue #2101, opened 2024-07-19
[feature]: Add Github Login 


charankamarapu:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

Need to add github login mechanism in keploy CLI which can be used in the future for feature like mock upload etc.

### Why should this be worked on?

There should be a basic login mechanism for other advanced features to come

### Repository

keploy

PR #2102, opened 2024-07-19
feat: Add Github login

charankamarapu:
## Related Issue

Add GitHub Login Mechanism

Closes: #2101

Changes:
1. Added GitHub Login flow
2. Add new mechanism to identify a unique device (installationID)
3. Removed pkg/platform/yaml/configdb/db.go as pkg/platform/yaml/configdb/user/db.go is already being used and both are exactly same

TODO (before):
1. Before pushing need to add build args like apiServerURL, GithubClient

## Type of change

## Checklist:

## Screenshots (if any)


pkg/platform/yaml/configdb/db.go
```diff
@@ -12,6 +12,8 @@ import (
 	"runtime"
 
 	"go.keploy.io/server/v2/pkg/platform/yaml"
+	yamlV "gopkg.in/yaml.v2"
```
slayerjain, line 15: why is this called yamlV?

pkg/platform/yaml/configdb/db.go
```diff
@@ -97,3 +98,62 @@ func (cdb *ConfigDb) setInstallationID(ctx context.Context, id string) error {
 
 	return nil
 }
+
+func (cdb *ConfigDb) ReadAccessToken(_ context.Context) (string, error) {
+	// read from cred.yaml
```
slayerjain, line 103: is this a typo?

pkg/platform/yaml/configdb/db.go
```diff
@@ -97,3 +98,62 @@ func (cdb *ConfigDb) setInstallationID(ctx context.Context, id string) error {
...
+	} else if err != nil {
+		fileName := filepath.Base(filePath)
+		return fmt.Errorf("unable to access the keploy file %s. error: %s", fileName, err.Error())
+
+	}
+	d, err := yamlV.Marshal(&token)
+	if err != nil {
+		return fmt.Errorf("failed to marshal document to yaml. error: %s", err.Error())
+	}
+	err = os.WriteFile(filePath, d, os.ModePerm)
```
slayerjain, line 154: I'm doubtful if this is even needed. We already create installation id, and this does not look more secure. 

How about we maintain a mapping of the installation ids to github ids on the backend? So for the same installation id, we could maintain the user and then only remove if someone wants to logout.

slayerjain: If we do want to make it even more secure, we could as well just use machine unique id (instead of auto generated uuid stored in yaml) and maintain the mapping in the cloud service when someone tries to login. Maybe something like this -https://unix.stackexchange.com/questions/144812/generate-consistent-machine-unique-id

utils/utils.go
```diff
@@ -757,3 +772,44 @@ func getHomeDir() (string, error) {
 func IsDockerCmd(kind CmdType) bool {
 	return (kind == DockerRun || kind == DockerStart || kind == DockerCompose)
 }
+
+func CheckAuth(ctx context.Context, host, token string, hardReset bool, logger *zap.Logger) (string, bool, string, string, error) {
```
slayerjain, line 776: I think these login related structs and methods below should go into a separate package in platform and ideally even interfaced.

utils/utils.go
```diff
@@ -34,6 +35,20 @@ import (
 )
 
 var WarningSign = "\U000026A0"
+var JwtToken = ""
+var APIServerURL string
+
+type GHAuthReq struct {
```
slayerjain, line 41: I think these login related structs and methods below should go into a separate package in platform and ideally even interfaced.

pkg/service/tools/tools.go
```diff
@@ -294,3 +319,143 @@ func (t *Tools) IgnoreTests(_ context.Context, _ string, _ []string) error {
...
+				utils.LogError(logger, err, "failed to close response body")
+			}
+		}()
+
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, err
+		}
+
+		if resp.StatusCode == http.StatusOK {
```
slayerjain, line 434: should invert this

pkg/service/tools/tools.go
```diff
@@ -294,3 +319,143 @@ func (t *Tools) IgnoreTests(_ context.Context, _ string, _ []string) error {
...
+
+func pollForAccessToken(ctx context.Context, logger *zap.Logger, deviceCode string, interval int) (*AccessTokenResponse, error) {
+	data := url.Values{}
+	data.Set("client_id", clientID)
+	data.Set("device_code", deviceCode)
+	data.Set("grant_type", "urn:ietf:params:oauth:grant-type:device_code")
+
+	fmt.Println("waiting for approval...")
+
+	for {
```
slayerjain, line 417: Looks like your polling for the access token in this for loop. Why dont we just raise a single request rather than polling in a for loop?

slayerjain (approved):
I think we can make design and stuctural improvements. Have mentioned some of them below - lets discuss more tomorrow

pkg/service/tools/tools.go
charankamarapu: Because this is a device registration through a gitHub app. Unless user allows create a access token, the device will not be registered, We will know if the device is registered or not from the gitHub app only. So we need to poll it. We can discuss more if there is another way get the access token.

pkg/service/tools/tools.go
charankamarapu: are saying about early return if there is non success code..?

pkg/platform/yaml/configdb/db.go
charankamarapu: my bad will change it.

pkg/platform/yaml/configdb/db.go
charankamarapu: yes

utils/utils.go
charankamarapu: yep can be done.

pkg/platform/yaml/configdb/db.go
charankamarapu: done

pkg/platform/yaml/configdb/db.go
charankamarapu: done

utils/utils.go
charankamarapu: done

pkg/service/tools/tools.go
charankamarapu: done

pkg/service/tools/tools.go
charankamarapu: done

main.go
```diff
@@ -23,9 +22,10 @@ import (
 
 var version string
 var dsn string
+var apiServerURI = "http://localhost:8080"
+var gitHubClientID = "Iv23liFBvIVhL29i9BAp"
```
slayerjain, line 26: are we sure this is not confidential?

slayerjain (approved):
LGTM!

pkg/platform/yaml/configdb/user/db.go
```diff
@@ -35,65 +29,27 @@ func HomeDir() string {
...
-	)
-
-	file, err := os.OpenFile(filepath.Join(path, "installation-id.yaml"), os.O_RDONLY, fs.ModePerm)
-	if err != nil {
-		return id
-	}
-	defer func() {
-		if err := file.Close(); err != nil {
-			utils.LogError(logger, err, "failed to close file")
+			return "", errors.New("failed to get machine id")
```
shivamsouravjha, line 47: the machine id is different in mac and lima. Hopefully this is an expected behaviour.

charankamarapu: yes

charankamarapu:
TODO:
1. add cress like api-server, GitHub ids in the release process

Issue #1775, opened 2024-04-03
[feature]: automatically add test reports folder to gitignore 


slayerjain:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

We should maybe create a gitignore if possible in the keploy folder that adds the test report folder and prevents it from accidentally being commited. 

More ideas to accomplishing this is also welcome. 

### Why should this be worked on?

Some users have been accientially commiting the reports folder present in the keploy folder to git. This can unnecessarily polute the git history. 

### Repository

keploy

ethanwater:
I assume we can just create a gitignore file, and write the report path to it? much like how it's done in Rust using cargo

Akash-Singh04:
@slayerjain Kindly review this and let me know if this implementation works!

ayush3160:
@Swpn0neel , Can you please assign this to me.
Thanks

Swpn0neel:
Go ahead! @ayush3160

PR #2116, opened 2024-07-22
feat: Add test reports folder to gitignore

ayush3160:
## Related Issue
Closes: #1775 

I have implemented a functionality in this pr to automatically add test reports folder to .gitignore file. When we will run test command then it creates a .gitignore file in keploy folder if not already present and add test report folder in it. 

## Type of change

NIL
NA

## Checklist:

## Screenshots (if any)
NA


utils/utils.go
```diff
@@ -760,3 +760,37 @@ func getHomeDir() (string, error) {
 func IsDockerCmd(kind CmdType) bool {
 	return (kind == DockerRun || kind == DockerStart || kind == DockerCompose)
 }
+
+func CreateGitIgnore(logger *zap.Logger, path string) error {
+	gitignorePath := path + "/.gitignore"
+	reportEntry := "/reports/"
```
charankamarapu, line 766: Be specific with the entry I think keploy/reports/ works better I guess. Please have a look.

charankamarapu (changes requested):
Please address the comments, rest looks good. Let me get back to you about why the pipeline is failing

utils/utils.go
ayush3160: I am storing .gitignore file in keploy folder itself as it was mentioned in the issue , So keploy/reports/ entry will not work.

charankamarapu: cool got it

charankamarapu:
I think the action is failing because the ,gitignore file is also considered to be test set in the action . You can add the gitignore file like how report folder is added and I guess then it should work - https://github.com/keploy/keploy/blob/c306030272404fc19b2a1a8783ea1ceefd123154/.github/actions/tester/install.sh#L55C77-L55C86.

ayush3160:
@charankamarapu , I think this is not the problem because when I ran the same command in terminal it is giving directories of test folder only as it should return. Please refer to the below Image. Let me know If I am doing something wrong here.

Screenshot from 2024-07-27 15-35-08

charankamarapu:
hmm looks like thats not the problem. Let me deep dive more.

charankamarapu:
Got the issue. We should consider only directories here in the readSession function. 
https://github.com/keploy/keploy/blob/c306030272404fc19b2a1a8783ea1ceefd123154/pkg/util.go#L199
the pilot which we are using for testing is using this function, due to this there is a mismatch in no of test sets. 
https://github.com/keploy/pilot/blob/c8c1d1a671385f4401c3f30f71b0888832787cac/main.go#L70

please raise separate PR for readSession fix and I review and merge it then and release, we can upgrade the keploy version in the pilot repo and that should fix it.

ayush3160:
Thanks @charankamarapu , I will create separate pr for this.

charankamarapu:
create PR on this repo by changing keploy version in the mod file - https://github.com/keploy/pilot

charankamarapu (approved):
LGTM

PR #2167, opened 2024-08-01
fix: debugfs deletion issue on running keploy with docker

khareyash05:

DebugFs volumes if being used in more than one applications in docker can cause issues as it will try to delete the debugfs volume even if its in use in another application.
Added a check to see if debugfs volume is present check for options 
If options same - dont try to delete the volume. Else delete

## Type of change


khareyash05:
Will be solved in https://github.com/keploy/keploy/pull/2172

PR #2224, opened 2024-08-20
fix: ignore machine id if not present

charankamarapu:
Issue :  If machineId is not present in machine(docker containers/ Kubernetes) currently keploy stops execution
solution: Ignore machineId if not present.


Sarthak160 (approved):
LGTM

PR #2287, opened 2024-09-13
update:  ebpf object files

charankamarapu:
Recent changes to ebpf required more kernel version. Fixed it so that the minimum kernel required will stay at 5.15


gouravkrosx:
@charankamarapu 
How about if we add verifier logs as well when the logger level is `debug` or just `info`?

charankamarapu:
cool let me add them

charankamarapu:
added the debug log

gouravkrosx (approved):
LGTM!

Issue #2282, opened 2024-09-11
[bug]: cancelling the test generate doesn't stop keploy 


shivamsouravjha:
### 👀 Is there an existing issue for this?

- [X] I have searched and didn't find similar issue

### 👍 Current behavior

* When I generate tests and hit ctrl + C it doesn't stop the utg 

### 👟 Steps to Replicate

* keploy gen --source-file-path="<path to source file>" \
    --test-file-path="<path to test file for above source file>" \
    --test-command="npm test" \
    --coverage-report-path="<path to coverage.xml>"
    
    

### 💻 Operating system

Linux

### 🎲 Version

I'm using the latest version of keploy

### 💻 Repository

keploy

PR #2291, opened 2024-09-17
fix: context cancellation and check if test works before for keploy utgen

khareyash05:
## What does this PR do?

Provide Context Cancellation for keploy generate
Check if test works before runnning utgen

## Related PRs and Issues

Closes: #2282 , #2292 

## Type of change

### DIFF
--- a/pkg/service/utgen/ai.go
+++ b/pkg/service/utgen/ai.go
@@ -302,7 +302,7 @@ func (ai *AIClient) Call(ctx context.Context, prompt *Prompt, maxTokens int) (st
 			if err == io.EOF {
 				break
 			}
-			time.Sleep(10 * time.Millisecond) 
+			time.Sleep(10 * time.Millisecond)
 		}
 	}
 	if ai.Logger.Level() == zap.DebugLevel {
--- a/pkg/service/utgen/ai.go
+++ b/pkg/service/utgen/ai.go
@@ -109,24 +109,9 @@ func (ai *AIClient) Call(ctx context.Context, prompt *Prompt, maxTokens int) (st
 
 	var apiBaseURL string
 
-	// Signal handling for Ctrl+C (SIGINT)
-	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
-	defer signal.Stop(sigChan)
-
-	// Create a cancellable context for the request
-	ctx, cancel := context.WithCancel(ctx)
+	// Create a context that cancels on SIGINT (Ctrl+C) or SIGTERM
+	ctx, cancel := signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)
 	defer cancel()
-
-	// Start a goroutine to listen for Ctrl+C and cancel the context
-	go func() {
-		select {
-		case <-sigChan:
-			cancel() // Cancel the context if Ctrl+C is pressed
-		case <-ctx.Done():
-			// Context already cancelled, do nothing
-		}
-	}()
 
 	if prompt.System == "" && prompt.User == "" {
 		return "", 0, 0, errors.New("the prompt must contain 'system' and 'user' keys")
@@ -162,7 +147,6 @@ func (ai *AIClient) Call(ctx context.Context, prompt *Prompt, maxTokens int) (st
 
 		ai.Logger.Debug("Making AI request to API server", zap.String("api_server_url", ai.APIServerURL), zap.String("token", token))
 		httpClient := &http.Client{}
-		// make AI request as request body to the API server
 		aiRequest := AIRequest{
 			MaxTokens: maxTokens,
 			Prompt:    *prompt,
@@ -185,7 +169,6 @@ func (ai *AIClient) Call(ctx context.Context, prompt *Prompt, maxTokens int) (st
 			return "", 0, 0, fmt.Errorf("error making request: %v", err)
 		}
 
-		// read the response body AIResponse
 		bodyBytes, _ := io.ReadAll(resp.Body)
 		var aiResponse AIResponse
 		err = json.Unmarshal(bodyBytes, &aiResponse)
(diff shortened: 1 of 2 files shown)


pkg/service/utgen/ai.go
```diff
@@ -107,6 +109,26 @@ func (ai *AIClient) Call(ctx context.Context, prompt *Prompt, maxTokens int) (st
...
+	// Create a cancellable context for the request
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Start a goroutine to listen for Ctrl+C and cancel the context
+	go func() {
+		select {
+		case <-sigChan:
+			fmt.Println("Received interrupt signal. Cancelling AI request...")
+			cancel() // Cancel the context if Ctrl+C is pressed
```
shivamsouravjha, line 126: are we sure we want to print this log?for other use cases do we print?

pkg/service/utgen/gen.go
```diff
@@ -116,9 +155,62 @@ func (g *UnitTestGenerator) Start(ctx context.Context) error {
...
+			_, err := cmd.CombinedOutput()
+			if err != nil {
+				g.logger.Error("Error running npm test")
+				return fmt.Errorf("npm test failed: %v", err)
+			}
+			g.logger.Info("npm test succeeded")
+		}
+
+		if g.lang == "python" {
+			// Run 'pytest'
```
shivamsouravjha, line 176:
```
	if g.srcPath == "" {
		if err := g.runCoverage(); err != nil {
			return err
		}
		if len(g.Files) == 0 {
			return fmt.Errorf("couldn't identify the source files. Please mention source file and test file using flags")
		}
	}

```

Since we already have this why do we need to run this command ?

And as shown in the image we would now that test is already broken, why don't we use this instead of handling each case which is repititive
image

pkg/service/utgen/ai.go
khareyash05: Not necessary cause Ctrl+c we have a log, maybe we can omit that. your take?

pkg/service/utgen/gen.go
khareyash05: Oh okay got it

shivamsouravjha:
also this fails with go utg

pkg/service/utgen/ai.go
shivamsouravjha: I don't generally prefer print statements and we've better logger at place. so we can omit this

pkg/service/utgen/gen.go
shivamsouravjha: this is still the same we should have used the pre-defined run coverage

pkg/service/utgen/gen.go
```diff
@@ -69,23 +72,58 @@ func NewUnitTestGenerator(srcPath, testPath, reportPath, cmd, dir, coverageForma
 }
 
 func (g *UnitTestGenerator) Start(ctx context.Context) error {
+	// Signal handling for Ctrl+C (SIGINT)
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
```
shivamsouravjha, line 76: rather than redefining this why don't we use the pre-defined context?


Like in main.go we've a sigchan handling this why not use that context?in get service 

```
		utgen.NewUnitTestGenerator
```
here we can use the pre-defined context.


thus we won't need this signal handling which is a duplicate code

khareyash05: Such as in like this? we have this in record
```
errGrp, _ := errgroup.WithContext(ctx)
```

pkg/service/utgen/ai.go
```diff
@@ -107,6 +109,25 @@ func (ai *AIClient) Call(ctx context.Context, prompt *Prompt, maxTokens int) (st
...
+	// Signal handling for Ctrl+C (SIGINT)
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
+	defer signal.Stop(sigChan)
+
+	// Create a cancellable context for the request
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Start a goroutine to listen for Ctrl+C and cancel the context
```
shivamsouravjha, line 121: creating a fresh context won't be a good idea..we try to use same context to different level so that same context cancellation and handler can stop entire thing at a point

pkg/service/utgen/gen.go
```diff
@@ -69,23 +71,42 @@ func NewUnitTestGenerator(srcPath, testPath, reportPath, cmd, dir, coverageForma
 }
 
 func (g *UnitTestGenerator) Start(ctx context.Context) error {
+	// Create a context that cancels on SIGINT (Ctrl+C) or SIGTERM
+	ctx, cancel := signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)
+	defer cancel()
```
shivamsouravjha, line 76: this isn't requite as the parent context already has it , please look at the code

pkg/service/utgen/ai.go
```diff
@@ -107,6 +109,10 @@ func (ai *AIClient) Call(ctx context.Context, prompt *Prompt, maxTokens int) (st
 
 	var apiBaseURL string
 
+	// Create a context that cancels on SIGINT (Ctrl+C) or SIGTERM
+	ctx, cancel := signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)
+	defer cancel()
+
```
shivamsouravjha, line 115: the parent context is already handling this pls refer to the code of ctx.go in utils. Thus by passing the parent context cancel we won't have to look at any particular context

khareyash05:
Attaching some screenshots for reference 
Screenshot 2024-09-24 at 22 53 46
Screenshot 2024-09-24 at 22 54 17
Screenshot 2024-09-24 at 22 54 41
Screenshot 2024-09-24 at 22 55 32

shivamsouravjha (approved):
LGTM!

PR #2403, opened 2024-11-08
refactor: prompts for UTG

khareyash05:
Update Prompts by giving more in depth clarity for functions
Give out sample test fnction format for each language while tackling corner cases 

Handle when refactored code is empty

Will need to get `app_modify` branch merged to incorporate full changes

### DIFF
--- a/pkg/service/utgen/assets/test_generation.toml
+++ b/pkg/service/utgen/assets/test_generation.toml
@@ -212,6 +212,7 @@ new_tests:
 {%- endif %}
   test_code: |
 {%- if language in ["python"] %}
+    import pytest
     @pytest.mark.asyncio
     def ...
 {%- else %}
@@ -235,7 +236,7 @@ new_tests:
     {%- endif %}
   library_installation_code: |
     {%- if language == "python" %}
-    pip install pytest pytest-asyncio
+    pip install pytest
     {%- elif language == "java" %}
     # Add the following to your Maven pom.xml:
     # <dependency>
@@ -268,7 +269,7 @@ Use block scalar('|') to format each YAML output.
 
 
 In Python, since this can be an issue Unclosed client session client_session: <aiohttp.client.ClientSession object at 0x102fed130>, please ensure using pytest-aiohttp to solve this issue
-In Python, Since  async def functions are not natively supported for testing and can be skipped, please add the decorator @pytest.mark.asyncio to the test functions that are async def functions, also if you are using the decorator, please import pytest as well.
+In Python, Since  async def functions are not natively supported for testing and can be skipped, please add the decorator @pytest.mark.asyncio to the test functions that are async def functions, also if you are using the decorator, please import pytest and pytest-asyncio as well.
 
 Response (should be a valid YAML, and nothing else):
 ```yaml
--- a/pkg/service/utgen/assets/test_generation.toml
+++ b/pkg/service/utgen/assets/test_generation.toml
@@ -11,26 +11,70 @@ Your task is to generate additional unit tests to complement the existing test s
 
 - **Analyze the Provided Code:**
   - Understand its purpose, inputs, outputs, and key logic or calculations.
+  - **Identify Return Types:**
+    - Determine the data types of return values for each function or method.
+    - Use return type information to guide the creation of relevant test cases.
+
+- **Refactor for Testability:**
+  - **Refactor the provided source code to improve testability**, including making external dependencies easily mockable, especially for asynchronous interactions.
+  - Ensure refactoring enhances testability without altering functionality or breaking existing behavior.
+  - Provide refactored code in the `refactored_source_code` field if changes are made.
+  - **Refactoring Techniques:**
+    - Use dependency injection to manage dependencies.
+    - Separate concerns to isolate different parts of the code.
+    - Implement interfaces or abstract classes to make components easily mockable.
 
 - **Utilize the Code Coverage Report:**
   - Identify specific parts of the code not yet covered by tests.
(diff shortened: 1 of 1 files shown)


pkg/service/utgen/assets/test_generation.toml
```diff
@@ -128,30 +212,30 @@ new_tests:
...
-    "import { assert } from 'assert';"
-    "import { myFunction } from 'my_module';"
+    import { assert } from 'assert';
+    import { myFunction } from 'my_module';
     {%- endif %}
   library_installation_code: |
     {%- if language == "python" %}
-    pip install pytest
+    pip install pytest pytest-asyncio
     {%- elif language == "java" %}
```
shivamsouravjha, line 239: this was a sampler any reason of adding asyncio, if we remove it would it cause any issue?

pkg/service/utgen/assets/test_generation.toml
```diff
@@ -180,6 +267,9 @@ Use block scalar('|') to format each YAML output.
 {% endif %}
 
 
+In Python, since this can be an issue Unclosed client session client_session: <aiohttp.client.ClientSession object at 0x102fed130>, please ensure using pytest-aiohttp to solve this issue
```
shivamsouravjha, line 270: I don't think this is correct

pkg/service/utgen/assets/test_generation.toml
khareyash05: Ya that was just for the decorator, I handled it below, will remove it

pkg/service/utgen/assets/test_generation.toml
khareyash05: Tests are failing due to unclosed session, I isolated it

shivamsouravjha: we can do a better thing, please check if we send this as error, print the prompt probably it should have had the error (this seems bit focused with address)

pkg/service/utgen/assets/test_generation.toml
```diff
@@ -128,26 +212,27 @@ new_tests:
...
     {%- if language == "python" %}
-    "import pytest"
-    "from my_module import my_function"
+    import pytest
     {%- elif language == "java" %}
-    "import org.junit.jupiter.api.Test;"
-    "import my.package.MyFunction;"
+    import org.junit.jupiter.api.Test;
+    import my.package.MyFunction;
     {%- elif language == "golang" %}
```
shivamsouravjha, line 227: any reason for chaning these ?

pkg/service/utgen/assets/test_generation.toml
```diff
@@ -128,14 +212,15 @@ new_tests:
 {%- endif %}
   test_code: |
 {%- if language in ["python"] %}
+    import pytest
+    @pytest.mark.asyncio
     def ...
```
shivamsouravjha, line 217: why are we making this change?

pkg/service/utgen/assets/test_generation.toml
```diff
@@ -128,14 +212,15 @@ new_tests:
...
 {%- else %}
     ...
 {%- endif %}
   new_imports_code: |
     {%- if language == "python" %}
     "import pytest"
-    "from my_module import my_function"
     {%- elif language == "java" %}
     "import org.junit.jupiter.api.Test;"
```
shivamsouravjha, line 225: any reason for removing this? i put it there so that it can get idea of both type of import formats

pkg/service/utgen/assets/test_generation.toml
```diff
@@ -168,9 +253,79 @@ new_tests:
...
   test_tags: happy path
-    ...
+
+refactored_source_code: |
+  # Here is the modified source code that retains original functionality but improves testability.
+  ...
 ```
 
+additions:
+  additional_instructions_text: |
```
shivamsouravjha, line 263: we already have something in the name of additional_instructions_text , might create confusion , also instead of sending all languages for all ,let's find out the language and send text accordingly

pkg/service/utgen/assets/test_generation.toml
khareyash05: Thanks, wasnt able to find one issue just now, i guess this will solve it

pkg/service/utgen/assets/test_generation.toml
```diff
@@ -180,6 +335,11 @@ Use block scalar('|') to format each YAML output.
 {% endif %}
 
 
+In Python, Since  async def functions are not natively supported for testing and can be skipped, please add the decorator @pytest.mark.asyncio to the test functions that are async def functions, also if you are using the decorator, please import pytest and pytest-asyncio as well.
+In Python, use absolute imports instead of relative imports to avoid issues with module resolution.
+
+In Golang, please import testing library and use the testing.T object to run the tests.
+
```
shivamsouravjha, line 342: any reason of keep this here in this format?

hopefully we don't give it a lot of texts, lets try to keep it crisp and concise, as a lot of instruction might not be the actual help

khareyash05: no not specific reason, could you please tell me where can i keep this, wasnt able to figure out

pkg/service/utgen/assets/test_generation.toml
khareyash05: Pytest wasnt. being added, i guess the comment below will solve the problem of async def functions

khareyash05:
Closing this PR as it will be merged with https://github.com/keploy/keploy/pull/2411

PR #2445, opened 2024-11-21
feat: [UTG] flaky flag and prioritize coverage check over flaky check

khareyash05:
Adds a flaky flag 

1. If test fails, return with no coverage and no more flaky check,
2. If test pass and coverage increases, check flakiness flag, 5 iteration to check flakiness


Sarthak160 (approved):
LGTM

PR #2481, opened 2024-12-27
fix: ensure installation script download to tmp directory

khareyash05:
if there is an existing keploy folder in the working directory , it moves it to the place, however to ensure that not happens, have to add a check on the folder where it is to be installed


khareyash05:

PR #2735, opened 2025-05-29
fix: exit CLI if logger initialization fails during startup

Shreya2619:
### What does this PR do?
This PR fixes a bug in the CLI where the logger initialization failure during startup causes undefined behavior. Now, the CLI exits early if the logger fails to initialize.

### Changes Made
- Added error check for logger initialization.
- Called os.Exit(1) to terminate if logger setup fails.
- Updated inline comments for clarity.

### Issue Reference
Fixes #7234

### Additional Notes
No breaking changes. Verified through local runs.

### DIFF
--- a/main.go
+++ b/main.go
@@ -51,11 +51,16 @@ func setVersion() {
 }
 
 func start(ctx context.Context) {
+	// logger, err := log.New()
+	// if err != nil {
+	// 	fmt.Println("Failed to start the logger for the CLI", err)
+	// 	return
+	// }
 	logger, err := log.New()
-	if err != nil {
-		fmt.Println("Failed to start the logger for the CLI", err)
-		return
-	}
+if err != nil {
+	fmt.Fprintln(os.Stderr, "ERROR: Failed to initialize logger:", err)
+	os.Exit(1)
+}
 	defer func() {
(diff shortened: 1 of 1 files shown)


Shreya2619:

Achanandhi-M:
Thank you for raising the PR! @Shreya2619 🎉 Our team will review it and get back to you shortly.

Shreya2619:
Hi @khareyash05, thanks for pointing that out!

The idea was to treat logger initialization as critical, since many parts of the CLI rely on it. Exiting early helps prevent silent issues and gives clearer signals in CI/CD environments. That said, I’m open to refining this if we expect cases where running without logs is acceptable.

Would love to hear your thoughts!

khareyash05:
> Hi @khareyash05, thanks for pointing that out!
> 
> 
> 
> The idea was to treat logger initialization as critical, since many parts of the CLI rely on it. Exiting early helps prevent silent issues and gives clearer signals in CI/CD environments. That said, I’m open to refining this if we expect cases where running without logs is acceptable.
> 
> 
> 
> Would love to hear your thoughts!

Hey,giving a fatal error isn't ideal for all envs, we shouldn't fail the execution because of that. Will be closing this one

PR #3309, opened 2025-11-25
chore: ignore content-length and message length mismatches when body content is identical

gouravkrosx:
This pull request introduces improvements to the matching logic for both gRPC and HTTP responses, specifically handling cases where the message or content length differs but the actual content matches. These changes help reduce false negatives during response validation by ignoring length mismatches when the decoded data or body content is identical.

Improvements to response matching logic:

* gRPC matching: If the decoded data matches but the message length differs, the message length mismatch is ignored, and the result is updated to reflect a successful match for message length. This also removes the length difference from the differences map.
* HTTP matching: If the body matches but the `Content-Length` header differs, the mismatch for `Content-Length` is ignored and marked as normal. The code then checks for any remaining header mismatches to determine the final match result.

### DIFF
--- a/pkg/matcher/grpc/match.go
+++ b/pkg/matcher/grpc/match.go
@@ -282,6 +282,22 @@ func Match(tc *models.TestCase, actualResp *models.GrpcResp, noiseConfig map[str
 		Actual:   actualDecodedData,
 	})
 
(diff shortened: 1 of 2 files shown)


kapishupadhyay22 (approved):
LGTM

PR #3388, opened 2025-12-17
feat: exclude delay from test execution time in summary

### LINKED ISSUE #3351: [feature]: remove inclusion of delay time in the test run summary
Currently in the summary it shows like - 

```
 <=========================================> 
  COMPLETE TESTRUN SUMMARY. 
	Total tests: 146
	Total test passed: 122
	Total test failed: 24
	Total time taken: "40.71 s"
```
It will be better if we don't include the application delay provided by the user and just show the test execution time taken by keploy. It will be good if renamed this field as `Total test execution time: `

WillyEverGreen:
Currently, the test summary’s reported execution time includes the user-provided application delay (`--delay` flag), which makes the metric misleading as it also accounts for startup wait time rather than actual test execution.

This PR corrects the timing calculation so that only the real test execution duration is reported.

Closes #3351

- Moved the test execution timer start point to after the application delay
- Renamed the summary label from `Total time taken` to `Total test execution time`
- Updated per-testset summary labels for naming consistency

## Links & References
**Closes:** #3351

- [x] Bug Fix
- [x] Code Refactor

## Added e2e test pipeline?
- [x] No, because the change affects reporting logic only and is covered by existing execution flow

- [x] No, the code is self-explanatory

## Added to documentation?
- [x] No documentation needed

- [x] No, manual verification through existing CLI execution is sufficient

## Self Review done?
- [x] Yes

- NA


WillyEverGreen:

WillyEverGreen:
recheck

WillyEverGreen:
recheck

pkg/service/replay/replay.go
```diff
@@ -898,6 +896,9 @@ func (r *Replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
 		}
 	}
 
+	// Start timing test execution after application delay
+	startTime := time.Now()
```
officialasishkumar: why this is added?

officialasishkumar: Please remove it

pkg/service/replay/replay.go
```diff
@@ -558,8 +558,6 @@ func (r *Replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
 	runTestSetCtx = context.WithValue(runTestSetCtx, models.ErrGroupKey, runTestSetErrGrp)
 	runTestSetCtx, runTestSetCtxCancel := context.WithCancel(runTestSetCtx)
 
-	startTime := time.Now()
```
officialasishkumar, line 561: why was this removed? 

please revert this

officialasishkumar:
the pipelines are failing. 

did you test this PR locally?

WillyEverGreen:
Thanks for pointing this out — you're right.

By moving startTime I ended up affecting the execution flow instead of just the reporting logic, which wasn’t the intention. That explains the pipeline failures as well.

I’ll revert the timer change back to its original place and update the summary calculation so the reported execution time excludes the --delay duration without touching the core execution lifecycle. I’ll also run the tests locally before pushing the fix.

PR #3401, opened 2025-12-20
fix: close TOCTOU race in MockManager and add race test

remo-lab:
## Summary of issue
While working on MockManager, I noticed a TOCTOU (time-of-check-time-of-use) race between the getter and setter paths under concurrent access.

GetFilteredMocksByKind and GetUnFilteredMocksByKind read a tree pointer from the internal map under a read lock, release the lock, and then iterate over the tree. At the same time, SetFilteredMocks / SetUnFilteredMocks can replace the entire map, which allows readers to continue using a stale tree pointer. This leads to data races reported by the Go race detector and can cause inconsistent mock reads under concurrency.
## Fix
Added a re-validation step after ensureKindTrees() so that getters prefer the current tree from the map if it was replaced concurrently.

Protected map length reads in SetFilteredMocks and SetUnFilteredMocks with a read lock to avoid concurrent map access races.

This closes the TOCTOU race window while keeping reader concurrency intact.
## Test

Added a small concurrent test that reproduces the race under the Go race detector.
To verify:
`go test ./pkg/agent/proxy -race -run TestMockManager_GetFilteredMocks_Race
`
screenshot of terminal before fix:
Screenshot 2025-12-20 182722
Screenshot 2025-12-20 182614
Screenshot 2025-12-20 182629

screenshot of terminal after fix:
Screenshot 2025-12-20 192220

The test consistently triggers race warnings before the fix and passes cleanly after.

### DIFF
--- a/pkg/agent/proxy/mockmanager.go
+++ b/pkg/agent/proxy/mockmanager.go
@@ -144,12 +144,11 @@ func (m *MockManager) GetFilteredMocksByKind(kind models.Kind) ([]*models.Mock, 
 	
 	if flt == nil {
 		// Tree doesn't exist, create it
-		flt, _ = m.ensureKindTrees(kind)
-		// Re-acquire lock to verify it's still in the map (another goroutine might have replaced the map)
+		_, _ = m.ensureKindTrees(kind)
+		// Unconditionally re-read from map to ensure we use the tree from current map state
+		// (another goroutine might have replaced the map or created the tree)
 		m.treesMu.RLock()
-		if existing := m.filteredByKind[kind]; existing != nil {
-			flt = existing // Use the existing tree if map was replaced
-		}
+		flt = m.filteredByKind[kind]
 		m.treesMu.RUnlock()
 	}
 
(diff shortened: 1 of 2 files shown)


remo-lab:
Hi @gouravkrosx @Sarthak160 @AkashKumar7902 ,
 I’ve opened a PR that fixes a race condition in MockManager around concurrent reads/writes and adds a small race repro test. I’ve also incorporated the feedback from the inline review (simplified the locking and added coverage for the unfiltered path).
Whenever you have some time, I’d really appreciate a look. Thanks a lot!

remo-lab:
Hi @gouravkrosx @Sarthak160 👋 — just a quick note that this PR has been verified by the reviewer,
Please let me know if anything else changes are needed before merging. Thanks!

remo-lab:
Hi @gouravkrosx @Sarthak160 👋 — just a quick note that this PR has been verified by the reviewer,
Please let me know if anything else changes are needed before merging. Thanks!
\

PR #3407, opened 2025-12-23
Add Tcp timeout and Protocol Based MYSQL detection

harshitashankar:
- When running an application with Keploy in record mode, if the application attempts to connect to an external service that is unreachable (e.g., a configuration service, API endpoint, or database), the application crashes or hangs instead of gracefully handling the connection failure. In cases when the unreachable endpoint is not crucial for app startup, app would crash even when it does not need to. 

## Links & References

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [ ] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

- [ ] 👍 yes

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

- [ ] 👍 yes, mentioned below

## Self Review done?
- [ ] ✅ yes

- NA

**Examples:**

---

## Additional checklist:

### DIFF
--- a/pkg/core/proxy/proxy.go
+++ b/pkg/core/proxy/proxy.go
@@ -24,6 +24,7 @@ import (
 	"go.keploy.io/server/v2/config"
 	"go.keploy.io/server/v2/pkg/core"
 	"go.keploy.io/server/v2/pkg/core/proxy/integrations"
+	mysqlUtils "go.keploy.io/server/v2/pkg/core/proxy/integrations/mysql/utils"
 	pTls "go.keploy.io/server/v2/pkg/core/proxy/tls"
 	"go.keploy.io/server/v2/pkg/core/proxy/util"
 	"go.keploy.io/server/v2/pkg/models"
@@ -65,16 +66,9 @@ type Proxy struct {
 	UDPDNSServer      *dns.Server
 	TCPDNSServer      *dns.Server
 	GlobalPassthrough bool
-	DatabasePorts     []uint32 // ports to treat as MySQL-compatible databases
 }
 
 func New(logger *zap.Logger, info core.DestInfo, opts *config.Config) *Proxy {
-	// Default database ports if not configured
-	databasePorts := opts.DatabasePorts
-	if len(databasePorts) == 0 {
-		databasePorts = []uint32{3306, 4000} // MySQL default port and TiDB default port
-	}
-
 	return &Proxy{
 		logger:            logger,
 		Port:              opts.ProxyPort, // default: 16789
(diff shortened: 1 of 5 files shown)


pkg/core/proxy/proxy.go
```diff
@@ -561,9 +603,11 @@ func (p *Proxy) handleConnection(ctx context.Context, srcConn net.Conn) error {
 
 	} else {
 		if rule.Mode != models.MODE_TEST {
-			dstConn, err = net.Dial("tcp", dstAddr)
+			dstConn, err = net.DialTimeout("tcp", dstAddr, 10*time.Second)
 			if err != nil {
 				utils.LogError(logger, err, "failed to dial the conn to destination server", zap.Uint32("proxy port", p.Port), zap.String("server address", dstAddr))
+				// Send HTTP error response if we've already read an HTTP request
+				p.sendHTTPErrorResponse(srcConn, initialBuf, logger, 500, "Internal Server Error")
```
gouravkrosx: We can’t simply send an HTTP response to the srcConn without knowing whether the client is speaking HTTP, as this can cause an immediate protocol violation. A non-HTTP client (like SSH or a database client) will not understand the response, leading to a connection failure with a cryptic error like “Connection reset by peer” or “Protocol error.”

gouravkrosx:
@harshitashankar Instead of adding an extra port-based check, we should refactor the logic so that we start listening to both the client and the server concurrently. Whichever side sends the first initial buffer can then be used to detect the protocol. Since in MySQL the server sends the first packet, we can inspect that buffer — if it’s MySQL, proceed with the MySQL flow; otherwise, fall back to the generic parser.

pkg/core/proxy/proxy.go
harshitashankar: added a check to only return http error resp if its an http request. can you please check?

harshitashankar:
> @harshitashankar Instead of adding an extra port-based check, we should refactor the logic so that we start listening to both the client and the server concurrently. Whichever side sends the first initial buffer can then be used to detect the protocol. Since in MySQL the server sends the first packet, we can inspect that buffer — if it’s MySQL, proceed with the MySQL flow; otherwise, fall back to the generic parser.

@gouravkrosx implemented. valid for mysql version >4.0

PR #3416, opened 2025-12-25
Prevent double-close panics in mock correlation by making channel closure idempotent

remo-lab:
## Summarry of Issue
This PR addresses a potential double-close of channels in the mock correlation manager during concurrent test unregister and shutdown paths.
Both UnregisterTest() and closeAllChannels() can attempt to close the same Done and mock channels.
Today, this is implicitly prevented by a mutex, which serializes execution and avoids concurrent closes. However, this safety relies entirely on the invariant that all channel closes always happen under the same lock.

This is fragile over time:
- future refactors could introduce new shutdown paths
- async calls or reordered logic could bypass the mutex
- channel close() is not idempotent and panics on second call

If that invariant is ever broken, the code can panic with close of closed channel.

## Test to reproduce:
I added targeted concurrency tests around the mock correlation manager to validate channel lifecycle behavior:
- Tests that confirm the current mutex-based implementation does not panic
- A demonstration test that explicitly shows the close of closed channel panic when synchronization is bypassed
- Stress and multi-test scenarios to exercise unregister and shutdown paths concurrently
These tests document both the existing behavior and the underlying risk.

# screenshot of terminal before fix:
Screenshot 2025-12-25 112108

# screenshot of terminal after fix:
Screenshot 2025-12-25 113917
Screenshot 2025-12-25 113938

## Solution
This PR makes channel closure explicitly safe and idempotent by using sync.Once per test ID:
- Each Done channel and mock channel has its own sync.Once
- Both UnregisterTest() and closeAllChannels() use Once.Do(close)
- All map access remains protected by the existing mutex
- Channel closure happens before map cleanup

This preserves current behavior while removing reliance on implicit locking guarantees.

## Impact
- Prevents close of closed channel panics under concurrent lifecycle events
- Makes channel ownership and lifetime explicit
- Future-proofs the code against refactors and new shutdown paths
- No functional behavior change; minimal overhead

### DIFF
--- a/pkg/service/orchestrator/mock_correlation.go
+++ b/pkg/service/orchestrator/mock_correlation.go
@@ -213,12 +213,18 @@ func (mcm *MockCorrelationManager) closeAllChannels() {
 			doneOnce.Do(func() {
 				close(testCtx.Done)
 			})
+		} else {
+			mcm.logger.Warn("Missing sync.Once for Done channel",
+				zap.String("testID", testID))
 		}
 
 		if mockChOnceExists && mockChExists {
 			mockChOnce.Do(func() {
 				close(mockCh)
 			})
+		} else if mockChExists && !mockChOnceExists {
+			mcm.logger.Warn("Missing sync.Once for mock channel",
+				zap.String("testID", testID))
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


remo-lab:
Hi @Achanandhi-M @amaan-bhati @gouravkrosx  👋
This PR hardens mock channel lifecycle handling by making channel closure idempotent using sync.Once and adds concurrency tests around unregister/shutdown paths to prevent potential double-close panics.
Happy to address any feedback. Thanks!

PR #3424, opened 2025-12-29
fix: redirect docker-compose logs to a file when output is a TTY

gouravkrosx:
This pull request enhances how the application handles output from docker-compose commands, especially when running in a terminal (TTY). It introduces logic to redirect docker-compose logs to a temporary file when appropriate, improving log accessibility and user experience. Additionally, it updates dependencies to support this functionality.

**Improvements to docker-compose command output handling:**

* When executing a docker-compose command and the output is a TTY, the output is now redirected to a temporary log file in the OS-specific temp directory, and the user is informed about how to view these logs live.
* If the output is not a TTY, docker-compose output continues to be sent directly to stdout/stderr, with appropriate debug logging.

**Dependency updates:**

* Added `golang.org/x/term v0.32.0` to `go.mod` to enable TTY detection.

**Codebase enhancements:**

* Imported `fmt`, `filepath`, and `golang.org/x/term` in `utils/signal_others.go` to support log file creation and TTY detection.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -113,6 +113,7 @@ require (
 	github.com/wI2L/jsondiff v0.5.0
 	github.com/zricethezav/gitleaks/v8 v8.28.0
 	golang.org/x/sync v0.15.0
+	golang.org/x/term v0.32.0
 	google.golang.org/grpc v1.71.0
 	gopkg.in/yaml.v2 v2.4.0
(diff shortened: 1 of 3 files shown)


Sarthak160 (approved):
LGTM!

Issue #3485, opened 2026-01-05
[bug]: Hardcoded 3-second sleep in process shutdown causes slow test runs and unverified termination


Gagan202005:
### 👀 Is there an existing issue for this?

- [x] I have searched and didn't find similar issue

### 👍 Current behavior

- Process termination in Keploy uses a hardcoded 3-second sleep between graceful (SIGTERM) and forceful (SIGKILL) shutdown, causing unnecessary delays even when processes exit immediately.
- Sends SIGTERM to process/group
- **Blindly waits 3 seconds** regardless of actual process state
- Then force kills with SIGKILL
- Never verifies if graceful shutdown actually worked

**Problematic Code:**
```go
// Graceful
err = cmd.Process.Signal(syscall.SIGTERM)
if err != nil {
    logger.Warn("failed to send SIGTERM to process; falling back to kill")
}
time.Sleep(3 * time.Second)  // ⚠️ Always waits 3 seconds
// Force
return cmd.Process.Kill()
```

**Root Cause** :
The function uses an arbitrary 3-second sleep instead of:

- Checking if the process actually exited
- Waiting only as long as necessary
- Verifying termination before returning

### 👟 Steps to Replicate

- Create a simple app that exits quickly (< 500ms)
- Run keploy test with multiple test cases
- Observe 3-second pause after each test completes
- Notice total test time is dominated by these pauses

### Example : time keploy test
- 5 tests, each takes 200ms to run
- Actual test time: 1 second
- Shutdown time: 15 seconds (3s × 5)
- Total: 16 seconds (94% wasted on shutdown!)

### 💻 Operating system

Linux

### 🎲 Version

Keploy 3.1.1

### 📦 Repository

keploy

Gagan202005:
Hi @Achanandhi-M @gouravkrosx @Sarthak160,
I’ve opened PR https://github.com/keploy/keploy/pull/3486 to address this issue.
Please let me know if any changes or improvements are required.

PR #3486, opened 2026-01-05
fix: replace hardcoded 3-second sleep with smart process termination

Gagan202005:
## Description

This PR fixes slow shutdown times by replacing a hardcoded 3-second wait with a smart polling approach that checks if the process actually stopped.

## Links & References

**Closes:** #3485 

## Problem:
When Keploy stops a process, it currently:
1. Sends a stop signal (SIGTERM)
2. Waits exactly 3 seconds (no matter what)
3. Force kills the process

This means even if your app stops in 100ms, Keploy still waits 3 full seconds.

## Solution:
Now Keploy:
1. Sends stop signal
2. Checks every 100ms if process stopped
3. Returns immediately when process exits
4. Only force kills after 5 seconds if needed

## Changes Made

**Modified:** `pkg/platform/http/utils/proc_unix.go`

- Removed hardcoded 3-second sleep
- Added polling loop that checks process status every 100ms
- Process termination is now verified before returning
- Added `stopProcessDirect()` helper function for single processes
- Increased timeout from 3s to 5s (but returns early when process exits)

## Performance Improvement

**Before:**
- Fast app (exits in 200ms): Waits 3000ms = **15x slower than needed**
- 5 test runs = 15 seconds wasted on shutdowns

**After:**
- Fast app (exits in 200ms): Waits 200ms = **optimal**
- 5 test runs = ~1 second for shutdowns

## Testing

- Code compiles successfully
-  Verified polling logic works correctly
-  Timeout still kicks in for slow processes
-  Both single process and process group termination handled

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

- [x] 👍 yes

## Self Review done?
- [x] ✅ yes


Gagan202005:
hi @Achanandhi-M @Sarthak160 @gouravkrosx
please review it

Gagan202005:
Hi @officialasishkumar @amaan-bhati @gouravkrosx @Sarthak160 
please review this pr whenever you get a chance

Issue #3487, opened 2026-01-05
[bug]: HTTP Response Body Resource Leak in AI Service


Gagan202005:
### 👀 Is there an existing issue for this?

- [x] I have searched and didn't find similar issue

### 👍 Current behavior

The AI service in `pkg/service/utgen/ai.go` has a resource leak where HTTP response bodies are not properly closed when errors occur during JSON unmarshalling or when receiving non-200 status codes.

### Root Cause
The `defer resp.Body.Close()` statement is placed **after** early return statements, meaning it never executes when errors occur:
```go
resp, err := httpClient.Do(req)
if err != nil {
    return "", fmt.Errorf("error making request: %v", err)
}

bodyBytes, _ := io.ReadAll(resp.Body)
err = json.Unmarshal(bodyBytes, &aiResponse)
if err != nil {
    return "", fmt.Errorf("...")  // ❌ Body never closed
}

if resp.StatusCode != http.StatusOK {
    return "", fmt.Errorf("...")  // ❌ Body never closed
}

defer resp.Body.Close()  // ⚠️ Too late - already returned!

```

### 👟 Steps to Replicate

- Run keploy gen with AI service configured
- Trigger API errors (rate limits, malformed responses, network issues)
- Monitor file descriptors: lsof -p <keploy-pid> | grep TCP
- Observe growing number of CLOSE_WAIT connections

### Impact :

- Connection pool exhaustion: Connections are not returned to pool for reuse
- File descriptor leaks: Each leaked response consumes a file descriptor
- Memory growth: Unclosed resources accumulate
- Production failures: After many AI requests during keploy gen operations, users may encounter "too many open files" errors

### 💻 Operating system

Linux

### 🎲 Version

Keploy 3.1.1

### 📦 Repository

keploy

Gagan202005:
Hi @Achanandhi-M @gouravkrosx @Sarthak160,
I’ve opened PR https://github.com/keploy/keploy/pull/3488 to address this issue.
Please let me know if any changes or improvements are required.

Achanandhi-M:
Hey @Gagan202005 , thank you for your efforts. The unit testing product is currently not maintained, so could you please pick up any other issue instead? Thanks for your understanding, and we really appreciate your contribution.

PR #3488, opened 2026-01-05
Fix: Move defer resp.Body.Close() to prevent resource leak

Gagan202005:
### Summary

- Fixed HTTP response body resource leak in AI service by moving `defer resp.Body.Close()` immediately after `httpClient.Do()`. This ensures the response body is always closed, even when early returns occur due to JSON unmarshalling errors or non-200 status codes.

## Links & References

**Fixes:** #3487 

### Changes

**File:** `pkg/service/utgen/ai.go`

Moved the defer statement from after error handling logic to immediately after the HTTP request:

```go
resp, err := httpClient.Do(req)
if err != nil {
    return "", fmt.Errorf("error making request: %v", err)
}
defer func() {
    err := resp.Body.Close()
    if err != nil {
        utils.LogError(ai.Logger, err, "failed to close response body")
    }
}()

// Safe to have early returns now
bodyBytes, _ := io.ReadAll(resp.Body)
err = json.Unmarshal(bodyBytes, &aiResponse)
if err != nil {
    return "", fmt.Errorf("...")  // ✅ Body properly closed
}
```

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

- [x] 👍 yes

## Self Review done?
- [x] ✅ yes

### DIFF
--- a/pkg/service/utgen/ai.go
+++ b/pkg/service/utgen/ai.go
@@ -176,6 +176,12 @@ func (ai *AIClient) Call(ctx context.Context, completionParams CompletionParams,
 		if err != nil {
 			return "", fmt.Errorf("error making request: %v", err)
 		}
+		defer func() {
+			err := resp.Body.Close()
+			if err != nil {
+				utils.LogError(ai.Logger, err, "failed to close response body for authentication")
+			}
+		}()
 
 		bodyBytes, _ := io.ReadAll(resp.Body)
 		var aiResponse AIResponse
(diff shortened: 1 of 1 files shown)


Achanandhi-M:
Hey @Gagan202005 , thank you for your efforts. The unit testing product is currently not maintained, so could you please pick up any other issue instead? Thanks for your understanding, and we really appreciate your contribution.

Issue #3509, opened 2026-01-09
[bug]: sync.Map passed by value causing lock copy violations detected by go vet


pigeio:
### 👀 Is there an existing issue for this?

- [x] I have searched and didn't find similar issue

### 👍 Current behavior

Running `go vet ./...` reveals that `sync.Map` is being passed by value in the telemetry and CLI provider code. This copies the internal mutex (`sync.noCopy`) and can lead to race conditions, deadlocks, and undefined behavior in concurrent operations.

**Expected behavior:** `sync.Map` should be passed by pointer (`*sync.Map`) to avoid copying the internal lock state.

**Actual behavior:** `sync.Map` is passed by value in struct fields and function parameters, triggering `go vet` warnings about lock copying.

### 👟 Steps to Replicate

1. Clone the Keploy repository
2. Navigate to the project root directory
3. Run the following command:
```bash
go vet ./...


### 📜 Logs (if any)

# go.keploy.io/server/v3/pkg/platform/telemetry
pkg/platform/telemetry/telemetry.go:34:43: NewTelemetry passes lock by value: go.keploy.io/server/v3/pkg/platform/telemetry.Options contains sync.Map contains sync.noCopy
pkg/platform/telemetry/telemetry.go:39:19: literal copies lock value from opt.GlobalMap: sync.Map contains sync.noCopy

# go.keploy.io/server/v3/cli/provider
cli/provider/service.go:38:19: literal copies lock value from TeleGlobalMap: sync.Map contains sync.noCopy


### 💻 Operating system

Linux

### 🧾 System Info (`uname -a`)

Linux Anand 5.15.167.4-microsoft-standard-WSL2 #1 SMP Tue Nov 5 00:21:55 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux


### 📦 OS Release Info (`cat /etc/os-release`)

NAME="Ubuntu"
VERSION="22.04.5 LTS (Jammy Jellyfish)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 22.04.5 LTS"
VERSION_ID="22.04"


### 🐳 Docker Info (if applicable)

Not used


### 🧱 Your Environment

Built from source (main branch)
Running on Ubuntu 22.04 via WSL2


### 🎲 Version

keploy server v3 (from go.mod: go.keploy.io/server/v3)
Go version: go1.22.1 linux/amd64


### 📦 Repository

keploy

### 🤔 What use case were you trying? (optional)

[...]

PR #3511, opened 2026-01-09
fix: avoid copying sync.Map in telemetry and CLI provider

Syedowais312:
## Summary

This PR fixes `go vet` copylocks warnings caused by passing `sync.Map` by value
in telemetry and CLI provider code.

`sync.Map` contains an internal mutex (`sync.noCopy`) and must not be copied
after first use. Passing it by value can lead to race conditions, deadlocks,
and undefined behavior under concurrent access.

## Changes

- Updated telemetry `Options` and `Telemetry` structs to use `*sync.Map`
- Updated `NewTelemetry` to accept and assign the pointer
- Passed `&TeleGlobalMap` instead of copying the map in the CLI provider

## Verification

- `go vet ./...` passes successfully
- No functional behavior changes

## Context

This was discovered during routine static analysis using `go vet`, which
reported lock-copying issues in:
- `pkg/platform/telemetry`
- `cli/provider`

Fixing this ensures safe concurrent usage and aligns with Go best practices.

Closes:  https://github.com/keploy/keploy/issues/3509

### DIFF
--- a/cli/provider/service.go
+++ b/cli/provider/service.go
@@ -14,7 +14,7 @@ import (
 	"go.uber.org/zap"
 )
 
-var TeleGlobalMap sync.Map
+var TeleGlobalMap = &sync.Map{}
 
 type ServiceProvider struct {
 	logger *zap.Logger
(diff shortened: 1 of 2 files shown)


Syedowais312:
Hi @gouravkrosx @Sarthak160 
This PR resolves `go vet` warnings related to copying `sync.Map` by value by switching to pointer usage.
`go vet ./...` passes cleanly after the change.
Happy to adjust if needed.

Issue #3575, opened 2026-01-19
[feature]: automatically set delay based on recording time


slayerjain:
### 👀 Is there an existing feature request for this?

- [x] I have searched the existing issues

### 🔖 Enhancement description

It would be really cool if the delay during test mode could be automatically set based on how long it took during record + some buffer (maybe 10 seconds). We could use the time it took between starting the application and the first test case to be hit by default. We could write this value to the keploy config and use it during keplot test/replay mode. 

### 🎤 Why should this be worked on?

It would reduce failures for first-time users trying `keploy test`. Most users may not be aware that there needs to be an application-specific delay configured for test mode. An automatic value would improve onboarding. 

### 💻 Repository

keploy

SaketV8:
Hi! @slayerjain I would like to work on this issue. Could you please assign this issue to me? Thanks!

Gagan202005:
Hey @slayerjain 
I’m interested in working on this enhancement. Auto-setting the test delay based on record time + buffer would greatly improve first-time onboarding.
If no one is working on it, I’d like to take it up

Gagan202005:
Hey @slayerjain @Achanandhi-M @gouravkrosx 

I’ve raised a PR to auto-calculate test.delay during keploy record (based on app startup time + buffer) and persist it into keploy.yml so users don’t need to manually guess delay during keploy test.

✅ Tested locally with record + test and confirmed the delay calculation works correctly and replay runs smoothly.

PR Link - https://github.com/keploy/keploy/pull/3587

Please review and share feedback

AuraReaper:
will pick this up today

justin212407:
@slayerjain I took a look at the open prs and saw that none of them are being actively worked upon.. so if this is open to work upon can i take this up?

slayerjain:
@justin212407 I have pinged another contributor on their PR to check if they are active. meanwhile it would be helpful to incorporate another mode that delays the app start until the app port is ready (+ 10 secs or something as buffer), and test it thoroughly.

justin212407:
> [@justin212407](https://github.com/justin212407) I have pinged another contributor on their PR to check if they are active. meanwhile it would be helpful to incorporate another mode that delays the app start until the app port is ready (+ 10 secs or something as buffer), and test it thoroughly.

Shall I open an issue and start working on that while this one get confirmed?

PR #3587, opened 2026-01-20
fix : Auto-calculate test delay during record and save it in keploy.yml

Gagan202005:
## Description
- Right now users manually set --delay in test mode when their app takes time to start.

- In this PR, Keploy will automatically calculate a good delay value while recording and then write it into keploy.yml so users don’t need to guess the delay for replay.

## Links & References

**Closes:** #3575 

---

## ✅ What Changed
- Auto-calculates `test.delay` based on:
  - Time taken from record start → first testcase captured
  - Adds a small buffer time
- Writes the final delay into `keploy.yml` under `test.delay`
- Does not override delay if the user explicitly provides `--delay`

---
## 🧩 Problem
Users currently need to manually set `--delay` during `keploy test` because different applications take different startup times.  
This is hard to guess and may cause replay failures if the delay is too low.

---

## 💡 Solution
Keploy now automatically:
1. Calculates the delay during recording
2. Adds buffer time
3. Clamps delay (min/max safe values)
4. Saves it in `keploy.yml`

---

## 🔧 Implementation Details
- Delay formula:
  - `delay = startup_time_to_first_testcase + buffer`
- Buffer time: **+10 seconds**
- Min delay: **5 sec**
- Max delay: **120 sec**
- Saved at:
  ```yaml
  test:
    delay: <calculated_value>


- [ ] 📦 Chore
- [x] 🍕 Feature
- [ ] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert


- [x] 👍 yes


## Self Review done?
- [x] ✅ yes

### calculation of test delay and persist to keploy.yml during record 

<img width="1440" height="722" alt="image" src="https://github.com/user-attachments/assets/937571c9-7827-4d28-b3f2-3334c751a4a2" />


### delay is persisted in keploy.yml

<img width="808" height="71" alt="image" src="https://github.com/user-attachments/assets/f9c43a35-c612-48ca-8d09-1df1e808c208" />
 

### testing (total time taken for testing = 17.07 seconds (delay -> 17 seconds) )

<img width="1439" height="798" alt="image" src="https://github.com/user-attachments/assets/7f034364-8e10-41c5-b1f5-6be3c6c38f50" />


### testing when delay is set manually (total time taken = 5.05 seconds(manually set delay -> 5 seconds))

<img width="1434" height="796" alt="image" src="https://github.com/user-attachments/assets/fbee24c2-923a-41bc-b1fd-883bcffca4b7" />

### DIFF
--- a/pkg/service/replay/replay.go
+++ b/pkg/service/replay/replay.go
@@ -767,7 +767,7 @@ func (r *Replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
 			return nil
 		})
 
-		agentCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		agentCtx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
 		defer cancel()
 
 		agentReadyCh := make(chan bool, 1)


gouravkrosx:
@Gagan202005 Thanks for the contribution.
We need one more change before I can review this. For interactive terminals, we should display a countdown timer (e.g., Replaying tests in 10…9…8…3…2…1 seconds) to indicate when the replay starts. For non-interactive terminals, there should be a simple fallback (e.g., a single log message) instead of an interactive ticker.

Gagan202005:
Hey @gouravkrosx , 

I’ve added the countdown timer for interactive terminals before replay starts, and added a simple fallback log message for non-interactive terminals as suggested.

Could you please review it now? 
Thanks!


https://github.com/user-attachments/assets/66734457-46e3-44a4-9916-8af8861e5b6a

PR #3687, opened 2026-01-31
Fix silent failures caused by nil error returns in proxy connection handling

### LINKED ISSUE #3685: Silent failures in proxy connection handling due to nil error returns
# Bug: Silent failures in proxy connection handling due to nil error returns

## Description
In [pkg/agent/proxy/proxy.go](cci:7://file:///d:/anujs/Documents/GitHub/keploy/pkg/agent/proxy/proxy.go:0:0-0:0), the [handleConnection](cci:1://file:///d:/anujs/Documents/GitHub/keploy/pkg/agent/proxy/proxy.go:308:0-695:1) function frequently reuses the `err` variable. In several critical logic blocks (session retrieval, mock manager loading, TLS destination lookup), if the map lookup fails (`!ok`), the code logs the failure but returns `err`.

Since `err` is often `nil` from successful preceding operations, the function returns `nil` (indicating success) despite having failed to perform the necessary work. This leads to connections being closed silently without proper propagation of the error state.

## Steps to Reproduce
1. Run the proxy agent.
2. Trigger a condition where `p.sessions.Get(0)` or `p.MockManagers.Load(0)` would return `false` (fail).
3. Observe that `utils.LogError` logs a message with a `nil` error, and the function returns clean (`nil`), preventing the caller from knowing something went wrong.

## Expected Behavior
The function should return a non-nil error describing exactly what failed (e.g., "failed to fetch session rule"), ensuring the system handles the failure mode correctly.

etk18:
## What does this PR do?

Fixes silent failures in `handleConnection` where missing session or mock manager entries
previously resulted in `nil` being returned, causing errors to be swallowed.

## What changed?

- Return explicit, non-nil errors when:
  - `p.sessions.Get(0)` fails
  - `p.MockManagers.Load(0)` fails
  - TLS destination lookup fails

This ensures callers are properly informed of failure states and prevents execution
from continuing in an invalid state.

## Additional Improvements (small but related)

- Added explicit errors when context values (`ClientConnectionID`, `DestConnectionID`)
  are missing, instead of logging unrelated or stale `err` values.
- This follows the same principle as the primary fix and improves debuggability
  with minimal behavioral impact.

## Related Issue

Closes #3685

## Notes

I considered adding a unit test, but this code path depends on active network
connections and proxy state. Happy to add a test or refactor for testability
if maintainers would prefer that.

### DIFF
--- a/pkg/agent/proxy/proxy.go
+++ b/pkg/agent/proxy/proxy.go
@@ -347,7 +347,8 @@ func (p *Proxy) handleConnection(ctx context.Context, srcConn net.Conn) error {
 	// TODO: to remove this sessions concept because it was meant for multiple clients-apps.
 	rule, ok := p.sessions.Get(uint64(0))
 	if !ok {
-		utils.LogError(p.logger, nil, "failed to fetch the session rule")
+		err := errors.New("failed to fetch the session rule")
+		utils.LogError(p.logger, err, "session lookup failed")
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


etk18:

etk18:
recheck

PR #3691, opened 2026-01-31
fix: eliminate data race in concurrent mock matching

remo-lab:
Prevent mutation of shared mock pointers during concurrent test replay. Create independent local copies before updating TestModeInfo fields.

I discovered a pretty critical race condition that was triggering during concurrent test replays. Basically, multiple goroutines were grabbing the same mock pointer from the `TreeDb` and trying to mutate it (specifically the `TestModeInfo` fields) at the same time without any locking.

To fix this, I updated the matching logic in the **HTTP**, **MySQL**, and **Generic** integrations. Now, instead of writing directly to the shared pointer, I’m creating a local copy of the mock on the stack first. We modify that local copy and then pass *that* to the update function. This keeps the shared state safe and stops the race detector from freaking out.

## Links & References

**Closes:** # (Not linked to an existing issue, just something I caught while debugging)

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑💻 Code Refactor
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes
  - The existing tests cover the functionality, and this is purely a concurrency fix. I verified it by checking that `go test -race` runs clean now.

- [x] 👍 yes
  - I didn't add too many, but the code changes themselves are pretty straightforward. The main "trick" is just the local copy, which I think is clear enough from the diff.

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki
  - This is an internal implementation detail, so no docs updates required.

- [x] 👍 yes, mentioned below
  - You can verify this by running the test suite with the race detector enabled: `go test -race ./pkg/...`.
  - Alternatively, if you have a test case that spams parallel requests (like a connection pool scenario), that's where this bug used to show up.

## Self Review done?
- [x] ✅ yes

- NA

**PR Title**: `fix: eliminate data race in concurrent mock matching`
**Branch Name**: `fix/mock-pointer-race`

---

## Additional checklist:


remo-lab:
Hi @officialasishkumar @gouravkrosx  ,I ran into a data race during concurrent test replays and fixed it by avoiding mutation of shared mock pointers.
The change makes local copies before updating state, which resolves the race cleanly.
Would really appreciate a review when you get a chance. Thanks!

remo-lab:
@officialasishkumar , is there any update on this pr?

PR #3761, opened 2026-02-10
feat: improve domain name extraction

slayerjain:
Improvements to domain parsing

### DIFF
--- a/pkg/platform/telemetry/telemetry.go
+++ b/pkg/platform/telemetry/telemetry.go
@@ -12,7 +12,6 @@ import (
 	"go.uber.org/zap"
 )
 
-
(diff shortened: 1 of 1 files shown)


pkg/service/replay/service.go
```diff
@@ -95,7 +95,7 @@ type TestSetConfig interface {
 
 type Telemetry interface {
 	TestSetRun(success int, failure int, testSet string, runStatus string)
-	TestRun(success int, failure int, testSets int, runStatus string)
+	TestRun(success int, failure int, testSets int, runStatus string, domains []string)
```
Sarthak160: Can we make it generic like metadata, incase we need more information in future other than domain ?

PR #3856, opened 2026-02-25
feat(matcher): add semantic HTML response validation with canonicalization pipeline

ShresthSamyak:
## Describe the Changes

This PR introduces semantic HTML response validation to Keploy’s matcher engine.

Previously, non-JSON HTTP response bodies were compared using raw string equality. This caused false negatives when two HTML responses were semantically identical but differed in attribute ordering, insignificant whitespace, or dynamically injected content.

This change adds a canonicalization-based HTML comparison pipeline while leaving the existing JSON logic completely untouched.

---

### 1. New: `pkg/matcher/html_compare.go`

Implements `CanonicalizeHTML()` using `golang.org/x/net/html`.

Pipeline:

1. **Size Guard**  
   Bodies larger than 1MB are rejected and fall back to raw string comparison to prevent excessive memory use in CI environments.

2. **Parse**  
   HTML is parsed into a DOM tree using the HTML5-compliant, error-correcting parser.

3. **Filter**  
   - Removes `<script>` and `<style>` elements  
   - Removes HTML comments  
   - Strips noisy attributes via regex patterns derived from `bodyNoise`

4. **Normalize**  
   - Sorts attributes deterministically (namespace-aware)
   - Collapses whitespace except inside `<pre>` and `<textarea>`
   - Removes empty text nodes

5. **Serialize**  
   Uses `html.Render()` to produce deterministic canonical output.

Traversal is fully iterative (explicit stack-based DFS). No recursion is used anywhere in the canonicalization pipeline, eliminating stack overflow risk for deeply nested HTML.

---

### 2. Modified: `pkg/matcher/http/match.go`

- Integrates semantic HTML comparison into the non-JSON branch of `Match()`.
- Uses `GuessContentType()` to sub-classify non-JSON bodies.
- Validates that both expected and actual bodies are HTML before canonicalization.
- Falls back to raw string comparison if:
  - Body types differ
  - Canonicalization fails
  - Size guard triggers
- Adds `case models.HTML` to the diff-logging switch.
- Introduces `buildHTMLConfig()` to translate `bodyNoise` keys into attribute-stripping regex patterns using `regexp.QuoteMeta` for safe glob handling.

JSON comparison logic is completely unchanged.

---

### 3. New: `pkg/matcher/html_compare_test.go`

Adds 16 unit tests covering:

- Whitespace normalization
- Attribute ordering
- Dynamic attribute stripping
- Script/style/comment removal
- Whitespace preservation rules
- Size guard enforcement
- Empty input handling
- Malformed HTML
- Content difference detection

---

### 4. Modified: `pkg/matcher/http/match_test.go`

Adds integration tests validating:

- Semantic HTML matches
- Content mismatches
- Type mismatch fallback
- Size guard fallback
- JSON regression safety

All existing tests pass without modification.

---

## What Is Not Changed

- JSON body detection and comparison
- JSON diff logic
- JSON noise handling
- `BodyResult.Type` behavior
- Non-HTML raw string comparison behavior

Backward compatibility is preserved.

---

## Known Limitations

- `GuessContentType()` prioritizes XML detection before HTML. Well-formed XHTML may be classified as XML and fall back to raw string comparison.
- HTML noise applies only to attribute names. Dot-separated JSON path keys are not applicable and are skipped.
- HTML diff output is whole-body canonical diff (not per-element like JSON).
- `<script>` and `<style>` stripping is unconditional.

Follow-up issues will address detection priority and richer HTML noise semantics.

---

## How to Test

```bash
# Canonicalization unit tests
go test ./pkg/matcher/... -v -run "TestCanonicalizeHTML" -count=1

# Match() HTML integration tests
go test ./pkg/matcher/http/... -v -run "TestMatch_HTML" -count=1

# JSON regression tests
go test ./pkg/matcher/http/... -v -run "TestMatch_JSON_HTMLIntegration_Regression" -count=1

# Full matcher suite
go test ./pkg/matcher/... ./pkg/matcher/http/... -cover -count=1
```

### DIFF
--- a/pkg/matcher/html_compare.go
+++ b/pkg/matcher/html_compare.go
@@ -1,0 +1,166 @@
+package matcher
+
+import (
+	"bytes"
+	"fmt"
+	"regexp"
+	"sort"
+	"strings"
+
+	"golang.org/x/net/html"
+)
+
+// MaxHTMLBodySize is the upper limit on HTML body size that will be canonicalized.
+// Bodies larger than this are compared as raw strings to avoid OOM in CI.
+const MaxHTMLBodySize = 1 << 20 // 1 MB
+
+type HTMLCompareConfig struct {
+	StripTags      map[string]bool
+	StripAttrRegex []*regexp.Regexp
+}
+
+// CanonicalizeHTML parses, cleans, normalizes, and deterministically serializes HTML content.
+// Returns an error if the input exceeds MaxHTMLBodySize.
+func CanonicalizeHTML(input string, config HTMLCompareConfig) (string, error) {
+	if input == "" {
+		return "", nil
+	}
+
+	// Fix 5: 1MB size guard — avoids OOM on large responses.
+	if len(input) > MaxHTMLBodySize {
+		return "", fmt.Errorf("HTML body exceeds maximum size for canonicalization (%d bytes)", MaxHTMLBodySize)
+	}
+
+	doc, err := html.Parse(strings.NewReader(input))
+	if err != nil {
+		return "", err
+	}
+
+	// Fix 1: No cloneNode — operate directly on the freshly parsed tree.
+	// The tree is not shared, so mutation is safe.
+	filterTree(doc, config)
+	normalizeTree(doc)
+
+	var buf bytes.Buffer
+	if err := html.Render(&buf, doc); err != nil {
+		return "", err
+	}
+
+	return buf.String(), nil
+}
+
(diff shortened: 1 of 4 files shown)


ShresthSamyak:

ShresthSamyak:
recheck

ShresthSamyak:
Hi @Sarthak160  and @gouravkrosx ,

All checks are now passing. Addressed bot review comments and fixed CLA/DCO issues.

Would appreciate a review when convenient.

PR #3914, opened 2026-03-11
fix: pkg/client/app silent errors

alrs:
This picks up two silent errors in `pkg/client/app`. The code reads like returning an error early before exhausting other options is not desirable, so both of these `err` variables are logged as warnings instead of returned.


alrs:

PR #3954, opened 2026-03-22
fix: repair release pipeline, self-hosted runner cleanup, and add CI gates

slayerjain:
### Release pipeline (commit 1-2)
- **build-docker failure**: `sigstore/cosign-installer` was pinned to a commit SHA that no longer exists upstream. Updated all five third-party action references to exact semver tags (`cosign-installer@v3.10.1`, `setup-buildx-action@v3.12.0`, `login-action@v3.7.0`, `metadata-action@v5.10.0`, `build-push-action@v6.19.2`)

### Windows self-hosted runner (commits 2-3)
- **Error 128 — stale git SSH config**: End-of-job cleanup used `--unset` (first match only) and only covered one key, leaving `core.sshCommand` and the `ssh://` variant behind. Fixed all cleanup steps to use `--unset-all` across all three SSH-related keys. Added pre-checkout cleanup to `pull-docker-image` and `cleanup` jobs.
- **Concurrency**: Added workflow-level concurrency group (matching macOS) to prevent concurrent runs on the same self-hosted runner, which caused config races between jobs.

### macOS self-hosted runner (commit 3)
- **Stale Docker artifacts**: Test jobs left containers (`ginApp`, `mongoDb`, etc.) and the `keploy-network` Docker network behind. Next run's `docker network create keploy-network` would fail with "already exists", causing cascading connection failures. Added container/network/process cleanup to both end-of-job steps in called workflows and the `clean_up_docker_macos` job.

### CI gate jobs — prevent selective re-runs (commit 3)
- Added a `CI Gate` job to all three main CI workflows (Linux, macOS, Windows)
- Each gate depends on **all** upstream jobs and checks their results
- Re-running only the gate is pointless: it re-evaluates upstream results and fails if any didn't succeed
- **Action required**: Configure `CI Gate (Linux)`, `CI Gate (macOS)`, and `CI Gate (Windows)` as the required status checks in branch protection settings, replacing the individual job names

## Test plan
- [ ] Push a new tag and verify the release pipeline completes
- [ ] Verify Windows self-hosted CI no longer hits error 128 on checkout
- [ ] Verify macOS self-hosted CI no longer hits "keploy-network already exists"
- [ ] Verify that re-running only the gate job fails when upstream jobs failed
- [ ] Update branch protection rules to require the three `CI Gate` checks

### DIFF
--- a/.github/workflows/golang_docker_macos.yml
+++ b/.github/workflows/golang_docker_macos.yml
@@ -48,32 +48,17 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
-      - name: Ensure Docker Desktop is running (macOS)
-        shell: bash
+      - name: Verify Docker Desktop is running (macOS)
         run: |
-          set -e
-          if docker info >/dev/null 2>&1; then
-            echo "Docker daemon is already running."
-          else
-            echo "Docker daemon not responding. Starting via launchctl..."
-            # Start Docker Desktop via launchctl so it's owned by launchd,
-            # not the CI job. The Actions runner kills child processes as
-            # "orphans" during post-job cleanup — launchctl avoids this.
-            # Use 'open' to start Docker Desktop — it's owned by
-            # launchd (WindowServer), not the CI job process tree.
-            open -g -a Docker
-            for i in $(seq 1 90); do
-              if docker info >/dev/null 2>&1; then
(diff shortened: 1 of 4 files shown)


slayerjain:
I will need to merge this since the macOS runners are getting killed from other pipelines.

PR #3984, opened 2026-03-27
fix: correct HTTP status check logic in auth

srivilliamsai:
## Summary
Fixes incorrect HTTP status code validation in the authentication flow.

## Problem
The condition `res.StatusCode != 200 || res.StatusCode >= 300` has a logic error:

| Status Code | `!= 200` | `>= 300` | Result (OR) | Correct? |
|-------------|----------|----------|-------------|----------|
| 200 | false | false | false ✓ | Yes |
| 201 | true | false | true ✗ | No - 201 is valid |
| 404 | true | true | true ✓ | Yes |

The OR operator causes any non-200 success code (like 201 Created) to be incorrectly rejected.

## Solution
Changed to proper range check using http package constants:

```go
// Before (buggy)
if res.StatusCode != 200 || res.StatusCode >= 300 {

// After (correct)  
if res.StatusCode < http.StatusOK || res.StatusCode >= http.StatusMultipleChoices {


charankamarapu (approved):
LGTM

PR #4013, opened 2026-04-05
feat: derive OpenAPI contracts from recorded traffic

mvanhorn:
## Summary

Adds `keploy contract generate` subcommand that reads recorded test cases and infers an OpenAPI 3.0 schema from the captured traffic.

## Demo

keploy contract generate demo

## Why this matters

Keploy records real API traffic and replays it as tests, but doesn't extract the implicit API contract from that traffic. When Keploy captures 50 requests to `POST /users` with a JSON body, it should be able to derive the expected schema. No open-source traffic-replay tool does automatic contract derivation from recordings.

The existing `keploy contract` command validates traffic against a provided spec. This adds the reverse: inferring the spec FROM traffic.

## Changes

- New `pkg/service/contract/infer.go` with `InferSchema()` that:
  - Groups operations by method+path
  - Infers JSON schema types recursively (string, number, boolean, array, object)
  - Builds a valid `openapi3.T` document
- `keploy contract generate` subcommand reads test cases, calls inference, validates result, writes `keploy/contract.yaml`
- 4 unit tests: single GET, POST with body, multiple endpoints, nested type inference

## Testing

- `go test ./pkg/service/contract/...` passes (4 new tests)
- `go build ./...` succeeds

This contribution was developed with AI assistance (Codex).


mvanhorn:
Addressed bot findings in f40d4fa:

- [P1] Fixed response description pointer aliasing - all responses were sharing the last iteration's description value
- [P1] Changed RequestBody.Required to false - single-sample inference can't determine if body is required
- [P2] Removed automatic Required marking on all object fields - one observation can't determine optionality
- [P2] Sorted testSetIDs for deterministic contract output across runs
- Added 2 regression tests covering pointer aliasing and Required behavior

All 6 contract inference tests pass.

officialasishkumar:
hi @mvanhorn , can you please check the copilot comments?

mvanhorn:
Addressed the copilot findings in be29470:

- **BodyRef hydration**: `GenerateFromTests` now reads offloaded request bodies from `BodyRef.Path` when `Body` is empty, so large payloads (>1MB) are included in schema inference (contract.go:431-443)
- **CLI backward compat**: added `--infer` flag (default `true`) to `contract generate` so the old service-mapping generation is still accessible via `--infer=false` (cli/contract.go:56-60)
- **Stale flags**: `generate` no longer inherits `--services`/`--tests` from `AddFlags` since `GenerateFromTests` doesn't use them (cli/contract.go:29-32)
- **Error guidance**: error logs in `GenerateFromTests` now include the searched path and a hint about expected directory structure (contract.go:419,428)
- **Validation in tests**: all test functions now call `doc.Validate(context.Background())`, which caught a real bug: empty JSON arrays produced an array schema with nil `Items`, violating the OpenAPI spec. Fixed by defaulting empty array items to an object schema (infer.go:142-144)

The `RequestBody.Required`, pointer aliasing, `sort.Strings`, and stale comment fixes were in earlier commits.

mvanhorn:
Addressed the remaining Copilot findings in 5c55c9a:

1. `--infer` default: flipped from true to false. Copilot's call on backward compat is correct: `contract generate` without flags has always meant service-mapping based generation, so inferring-by-default silently changes behaviour for anyone already scripting the command. Opt-in via `--infer` is the safer surface. Users who want the new path just add the flag.

2. `logger.Warn` on body hydration failure: downgraded to `Debug` and added the troubleshooting hint per the suggestion block. The failure is non-fatal (schema inference continues without the body), so Warn was too loud for project logging guidelines.

Findings 2 and 3 were pointing at the same log call on adjacent lines; one change covers both.

Verified: `go build ./cli/... ./pkg/service/contract/...` clean, `go test ./pkg/service/contract/...` passes, `go vet` clean, `gofmt` clean.

PR #4023, opened 2026-04-08
fix: prioritize replay mocks by request timestamp and fix build errors

### LINKED ISSUE #4021: [bug]: replay mock filtering can select stale dependency mocks when trailing calls cross the testcase response boundary
## 👀 Is there an existing issue for this?

## 👍 Current behavior
When replaying a testcase window that contains multiple compatible MySQL prepared-statement executions across a test set, a dependency mock can be dropped from the prioritized mock set if its request starts inside the testcase window but its response finishes slightly after the recorded testcase response timestamp. Once that happens, replay can select an older stale mock with the same `COM_STMT_EXECUTE` shape and parameters, which produces the wrong response for the later testcase. The expected behavior is to keep prioritizing mocks whose outgoing request starts within the testcase window so the correct later mock is replayed.

## 👟 Steps to Replicate
1. Record a test set where two different testcases issue the same MySQL prepared statement shape and parameter values but expect different snapshots/results.
2. Make the later testcase's final dependency call start inside the testcase window and finish a few microseconds after the recorded testcase response timestamp.
3. Replay the test set on Keploy v3 before the fix.
4. Observe that the trailing dependency mock is excluded from the prioritized mock set and an older compatible mock can be selected instead.

## 📜 Logs (if any)
A minimal timing pattern that reproduces the bug is:
```
testcase request timestamp   = T0
trailing mock request        = T1 (T0 <= T1 <= testcase response time)
trailing mock response       = T2 (slightly after testcase response time)
stale compatible mock        = older than T0, but same request shape
result                       = stale mock is selected during replay
```

## 💻 Operating system
MacOS

## 🧾 System Info (`uname -a`)
```text
Darwin Asishs-MacBook-Air.local 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:51 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T8112 arm64
```

## 📦 OS Release Info (`cat /etc/os-release`)
N/A on macOS

## 🐳 Docker Info (if applicable)
N/A

## 🧱 Your Environment
- Local checkout of `keploy/keploy`
- Replay logic bug; appears OS-agnostic

## 🎲 Version
Affected before fix on `v3.3.66` lineage. Verified fixed locally on `v3.3.66-1-g530066893`.

## 📦 Repository
keploy

## 🤔 What use case were you trying? (optional)
Replaying recorded gRPC and database traffic where multiple testcases reuse the same prepared-statement shape but depend on different recorded snapshots.

vaani1127:
Fixed Mock Filtering Logic: Updated pkg/util.go to prioritize replay mocks based solely on the outgoing request start time. This ensures that "trailing" mocks (where the request starts within the testcase window but the response finishes slightly after) are correctly prioritized instead of being replaced by older stale mocks.
Resolved Build Errors: Fixed compilation issues in pkg/models/const.go where variadic arguments were missing the ... expansion operator.
Added Regression Test: Implemented TestFilterTcsMocks_TrailingMockAcrossBoundary_4021 in pkg/util_test.go to verify the fix and prevent regression of issue #4021.

Closes: #4021


vaani1127:

PR #4049, opened 2026-04-14
 fix: cgroup path for daemonset

khareyash05:
Signed-off-by: Yash Khare <khareyash05@gmail.com>## Describe the changes that are made
- 

## Links & References

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [ ] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

- [ ] 👍 yes

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

- [ ] 👍 yes, mentioned below

## Self Review done?
- [ ] ✅ yes

- NA

**Examples:**

---

## Additional checklist:


Mrsandeep27 (commented):
Two new agent flags (\`--cgroup-path\`, \`--skip-hooks\`) plus the conditional path-resolution in \`pkg/agent/hooks/linux/hooks.go\` — clean and well-scoped for DaemonSet mode.

A few observations:

**1. Input validation on `--cgroup-path`.** If a user passes a non-existent path or one without the right magic marker (\`cgroup.controllers\` typically), the eBPF attach will fail with a kernel-level error that's hard to diagnose. Worth doing a \`stat\`/\`Access\` check right after \`opts.CgroupPath != \"\"\` and returning a clear error like \"cgroup path <x> does not exist or is not a cgroupv2 hierarchy\" so operators debug at configure time, not at runtime.

**2. `--skip-hooks` is advertised here but I don't see the consumer in this diff.** The cmd.go hunk adds the flag and config plumbing; where's the actual \`if cfg.Agent.SkipHooks { return nil }\` branch? If it's in a separate PR, worth linking it; if it's in this PR but trimmed from the first 60 diff lines I saw, ignore this one.

**3. Flag docs.** The help text says "used by DaemonSet agent for targeting specific pods" / "used by DaemonSet orchestrator agent" — these are useful hints, but as plain strings they won't surface in \`--help\` context. Consider adding a short section to the DaemonSet docs about when to use which flag, because a user who runs \`keploy agent --help\` and sees these will most likely just ignore them.

**4. Log line hygiene.** \`h.logger.Info(\"Using explicit cgroup path\", zap.String(\"path\", cGroupPath))\` fires unconditionally on every \`load()\`. In a DaemonSet with many restarts this adds noise. Consider \`Debug\` level, or adding \`once.Do\` so it only logs the first time per process.

**5. `DetectCgroupPath` return value is now ignored in the explicit-path branch.** If \`opts.CgroupPath\` ever contains a stale path (e.g. the pod restarted and cgroup moved), you silently use the bad path instead of falling back to detection. Worth deciding: is the explicit override authoritative (current behavior — fail loud if wrong) or a hint (fall back to detection on failure)? Lean toward the current behavior but call it out in the flag help.

Nothing blocking — direction is right. #4 is the smallest to fix before merge.

PR #3975, opened 2026-03-26
fix: add failure_reason and logs fields to TestReport for APP_HALTED status

BillionClaw:
## Describe the changes that are made
- Added `failure_reason` and `logs` fields to the `TestReport` struct in `pkg/models/testrun.go`
- When a test set halts due to app errors (`APP_HALTED`, `INTERNAL_ERR`, etc.), the error message from `AppError` is now captured and stored in the report instead of being discarded
- Updated `pkg/service/replay/replay.go` to populate `FailureReason` when any non-canceled app error occurs
- Updated `pkg/service/report/report.go` to display a warning line showing the failure reason when a test set halted, before the FAILED TEST CASES section

## Links & References
- **Closes:** #3968

## What type of PR is this? (check all applicable)
- [ ] 📦 Chore
- [ ] 🍕 Feature
- [ ] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 🙅 no, because they aren't needed

## Added comments for hard-to-understand areas?
- [ ] 🙅 no, because the code is self-explanatory

## Added to documentation?
- [ ] 🙅 no documentation needed

## Are there any sample code or steps to test the changes?
- [ ] 👍 yes, mentioned below

## Self Review done?
- [ ] ✅ yes

## Any relevant screenshots, recordings or logs?
- NA

PR #4062, opened 2026-04-15
fix: add Logs field to TestReport

yangluxin613:
Add Logs field to the TestReport struct. This field is needed for better test failure reporting.

Fixes #3975.

### DIFF
--- a/pkg/models/testrun.go
+++ b/pkg/models/testrun.go
@@ -23,6 +23,7 @@ type TestReport struct {
 	TimeTaken     string       `json:"time_taken" yaml:"time_taken"`
 	CmdUsed       string       `json:"cmdUsed,omitempty" yaml:"cmdUsed,omitempty"`
 	AppLogs       string       `json:"appLogs,omitempty" yaml:"app_logs,omitempty"`
+	Logs          string       `json:"logs,omitempty" yaml:"logs,omitempty"`
 }
 
 type TestCoverage struct {


Mrsandeep27 (commented):
Quick observation — \`TestReport\` now carries both:
\`\`\`go
AppLogs string \`json:\"appLogs,omitempty\" yaml:\"app_logs,omitempty\"\`
Logs    string \`json:\"logs,omitempty\"   yaml:\"logs,omitempty\"\`
\`\`\`

Two thoughts:

**1. Semantic overlap.** What distinguishes \`Logs\` from the existing \`AppLogs\`? If \`AppLogs\` is specifically for user-application stdout/stderr and \`Logs\` is for keploy-agent logs (or vice versa), a field-level comment would make the distinction obvious to downstream consumers that serialize \`TestReport\` into reports, dashboards, or CLI output.

**2. JSON/YAML naming inconsistency.** \`AppLogs\` uses \`appLogs\` in JSON but \`app_logs\` in YAML. \`Logs\` uses \`logs\` for both. The existing inconsistency is unfortunate; if a contributor is touching this struct anyway, consider unifying to one style (snake_case or camelCase, not both) in a follow-up. Not blocking this PR — just surfacing it.

**3. Population.** Is anything actually writing \`Logs\` yet? If this PR just adds the field without a producer, it'll always serialize as empty and consumers can't rely on it. Worth linking to the PR/branch that populates it, so reviewers know the surface area.

The diff itself is safe (additive, \`omitempty\`), so non-blocking. Mostly about clarifying intent.

PR #4063, opened 2026-04-15
feat: add JSON serialization format for testcases, mocks, reports

ayush3160:
## Summary

Adds an opt-in **JSON** serialization format alongside the existing YAML format for all persisted artifacts (testcases, mocks, reports, mappings, testset configs). Default remains `yaml` so existing recordings continue to work unchanged.

**Motivation:** YAML encoding/decoding is a measurable bottleneck on large test suites — JSON is significantly faster on `Time/op`, `Memory/op`, and `Allocs/op` using the stdlib `encoding/json`.

## Usage

Three ways to enable (precedence: flag > env > config file > default):

```bash
# CLI flag (persistent, works on record/test/rerecord/postman/export/import)
keploy record --storage-format=json -c "./myapp"

# Environment variable
KEPLOY_STORAGE_FORMAT=json keploy test -c "./myapp"

# Config file (keploy.yml)
storageFormat: json
```

Invalid values are rejected with a clear error:
```
invalid --storage-format value "xml": allowed values are 'yaml' and 'json'
```

## File-layout impact

| Artifact         | YAML (default)                           | JSON                                     |
|------------------|------------------------------------------|------------------------------------------|
| Test cases       | `<set>/tests/test-1.yaml`                | `<set>/tests/test-1.json`                |
| Mocks            | `<set>/mocks.yaml` (multi-doc, `---`)    | `<set>/mocks.json` (NDJSON, one per line)|
| Reports          | `reports/<run>/<set>-report.yaml`        | `reports/<run>/<set>-report.json`        |
| Mappings         | `<set>/mappings.yaml`                    | `<set>/mappings.json`                    |
| Testset config   | `<set>/config.yaml`                      | `<set>/config.json`                      |

NDJSON is chosen for mock files so the cheap streaming-append pattern used by YAML (`---` separator) is preserved — no need to read/rewrite the whole file on every `InsertMock`.

Secrets are always stored as `secret.yaml` regardless of format (unchanged).

## Design notes

- `pkg/platform/yaml/codec.go` introduces `Format`, `NetworkTrafficDocJSON` (which uses `json.RawMessage` for the polymorphic `Spec` field — the existing path uses `yaml.Node`), and `MarshalDoc` / `UnmarshalDoc` that dispatch on format. For JSON, the `yaml.Node` spec is decoded to `interface{}` then marshalled to `json.RawMessage`; on the way back, JSON is unmarshalled to `interface{}` then encoded into a `yaml.Node` so the rest of the code (protocol-specific `Spec.Decode(&httpSpec)` calls, etc.) keeps working unchanged.
- Every DB layer (`testdb`, `mockdb`, `reportdb`, `mapdb`, `configdb/testset`) gets a `Format` field and a `NewWithFormat(...)` constructor. The existing `New(...)` constructors remain and default to YAML for backwards compatibility.
- New format-aware I/O helpers (`WriteFileF`, `ReadFileF`, `CreateFileF`, `DeleteFileF`, `FileExistsF`, `FindLastIndexF`, `ReadSessionIndicesF`, `NewMockReaderF`) sit alongside the existing YAML-defaulted wrappers.
- `export` service now auto-detects `.yaml` and `.json` test files so users who switch formats can still export to Postman without migration.
- `import` service (Postman) uses the configured format when writing new tests.

## Test plan

- [x] `go build ./...` — clean
- [x] `go test ./...` — all pass (previously-passing tests stay passing; 14 new tests in `pkg/platform/yaml/codec_test.go`)
- [x] Covered by new tests:
  - `TestParseFormat` / `TestFileExtension`
  - `TestMarshalDocJSON` / `TestMarshalDocYAML`
  - `TestUnmarshalDocJSON`
  - `TestJSONRoundTripHTTPSpec` (HTTP request/response/created round-trip through JSON)
  - `TestMarshalGenericJSON` / `TestMarshalGenericYAML` (reports, mappings)
  - `TestMockReaderNDJSON` (streams multiple JSON docs)
  - `TestMockReaderNDJSONEmptyLines` (empty-line tolerance)
  - `TestWriteFileReadFileJSON`
  - `TestWriteFileAppendNDJSON` (NDJSON separator behaviour)
  - `TestFileExistsF`
  - `TestReadSessionIndicesF` (ignores `.yaml` when format=json and vice-versa)
- [x] Manually verified: `keploy config --generate` writes `storageFormat: "yaml"` default; `keploy record --storage-format=invalid` rejects with the expected error; `--storage-format=json` is accepted and threaded through.

## Backwards compatibility

- Default behaviour is unchanged — users who don't set the flag get YAML files identical to today.
- Existing YAML testcases/mocks are not auto-migrated; switching formats is a fresh-start operation for new recordings. (A follow-up migration helper could be added if needed.)
- `ReadSessionIndices` was split into a format-aware variant so if both `.yaml` and `.json` exist in the same directory, each format sees only its own files instead of getting duplicates.


ayush3160:
### Heads-up on `proxy-stress-test` red CI

All three `proxy-stress-test` configs (`record_build_replay_build`, `record_build_replay_latest`, `record_latest_replay_build`) now go red on this branch — but **not because of any change in this PR**.

**Root cause: keploy/integrations#188** — postgres-v3 recorder drops query mocks for concurrent simple-Query (lib/pq) workloads. The 20 concurrent goroutines in `proxy-stress-test`'s `/api/batch-transfer` each fire the same literal-LIMIT SELECT through lib/pq simple-Query; the recorder captures every connection-startup mock for the pool fanout (19× `type:session/config`) but **zero** of the per-invocation query mocks. At replay the matcher reports `candidates: 0` for the SQL hash and rejects with "no recorded invocation shares this SQL hash".

I reproduced it in a 120-line Go service ([repro in #188](https://github.com/keploy/integrations/issues/188)) — it's independent of this PR's JSON-storage work and reproduces on a clean lib/pq + literal-LIMIT app.

**Why CI was green before this PR:**

The proxy-stress-test has actually been failing 3-of-5 cases per test-set on **every config since the test was added** ([keploy#4028](https://github.com/keploy/keploy/pull/4028), 2026-04-09). The yaml-replay loop in `proxy_stress_test/golang-docker.sh` only `echo`'d the per-test-set `status:` without enforcing it, so the script exited 0 regardless of FAILED test-sets and the job was marked green. From the CI artifacts (now uploaded via `fdfc51e4`):

```
test-set-0-report.yaml   status: FAILED  success: 2  failure: 3  medium-risk: 3
test-set-1-report.yaml   status: FAILED  success: 2  failure: 3  medium-risk: 3
```

— same on all 3 configs (rbrb, rbrl, rlrb). Only `record_build_replay_build` flagged the failure because the json-replay branch added by this PR uses `json_scan_reports`, which actually enforces `status: PASSED`.

**What changed on this PR (`e7d15303`):**

Made the yaml-replay loop in `golang-docker.sh` enforce `status: PASSED` (matching what the json branch already did). The json branch is untouched — the goal is yaml getting as strict as json, not the other way around. Result: real CI signal for the latent integrations#188 bug, instead of silently green pipelines.

**Expected behaviour going forward:**

- All 3 proxy-stress-test configs stay red until integrations#188 ships.
- Once #188 lands, the postgres mock-capture under concurrency works and the pipeline goes green automatically — no further changes needed here.
- Other CI scripts likely have similar silent-pass yaml-replay loops (`grep -rln 'echo.*status:' .github/workflows/test_workflow_scripts/` shows ~10 candidates), but those are out of scope for this PR — only proxy-stress was hardened here.

cc @slayerjain for visibility — the test you added in #4028 is now correctly enforcing what it was always meant to.

charankamarapu (approved):
LGTM

gouravkrosx (approved):

Issue #4018, opened 2026-04-07
[bug]: filterOutDeleted not applied to unfiltered mocks in v3 agent UpdateMockParams


officialasishkumar:
### 👀 Is there an existing issue for this?

- [x] I have searched and didn't find similar issue

### 👍 Current behavior

When multiple mocks share identical request signatures (e.g. two MySQL `COM_STMT_EXECUTE` mocks with the same query and parameters but different responses), the wrong mock can be matched during test replay. After an earlier test case consumes mock A, a later test case that should match mock B instead matches mock A again, returning incorrect data.

**Root cause:** In v2, `FilterAndSetMocks` applied `filterOutDeleted` to **both** the filtered and unfiltered mock sets:

https://github.com/keploy/keploy/blob/25779c88/pkg/service/replay/replay.go#L1566-L1567

```go
// v2 — pkg/service/replay/replay.go
filtered = filterOutDeleted(filtered)
unfiltered = filterOutDeleted(unfiltered)
```

In v3, `UpdateMockParams` only applies `filterOutDeleted` to `filteredMocks`, skipping `unfilteredMocks`:

https://github.com/keploy/keploy/blob/d0206cc7/pkg/service/agent/agent.go#L308-L309

```go
// v3 — pkg/service/agent/agent.go
if params.TotalConsumedMocks != nil {
    filteredMocks = a.filterOutDeleted(filteredMocks, params.TotalConsumedMocks)
    // unfilteredMocks is never passed through filterOutDeleted
}
```

`filterOutDeleted` serves two purposes:
1. Removes mocks with `Usage: Deleted`
2. Updates `SortOrder` for consumed mocks (`Usage: Updated`) to a higher value from `totalConsumedMocks`

The mock tree is ordered by `SortOrder` ascending (lower = matched first). When `filterOutDeleted` is skipped for unfilteredMocks, consumed mocks retain their original low `SortOrder` and get matched before the correct mock. This affects any protocol matcher that reads from the unfiltered tree (e.g. MySQL via `GetUnFilteredMocks()`).

### 👟 Steps to Replicate

1. Record a test set where two or more test cases trigger outgoing calls with identical request signatures (same query, same parameters) but different expected responses
2. The mocks for these calls will have no `reqTimestampMock`/`resTimestampMock` fields (common for MySQL mocks)
3. Run the test set with keploy v3 (feat/sse-support branch)
[...]

PR #4072, opened 2026-04-16
fix: apply consumed state to unfiltered mocks

VivekArgSharma:
- apply `filterOutDeleted` to both filtered and unfiltered mock sets in `UpdateMockParams`
- add regression tests covering deleted mocks and consumed state propagation for unfiltered mocks

## Links & References

**Closes:** #4018

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- #4018
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

- [ ] 👍 yes

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

- [x] 👍 yes, mentioned below

## Self Review done?
- [x] ✅ yes

- NA

## Sample steps to test
- run `go test ./pkg/service/agent`
- verify `UpdateMockParams` removes deleted mocks from both filtered and unfiltered sets
- verify consumed state updates (`IsFiltered`, `SortOrder`) propagate to unfiltered mocks

## Additional checklist:

### DIFF
--- a/pkg/service/agent/agent.go
+++ b/pkg/service/agent/agent.go
@@ -319,6 +319,7 @@ func (a *Agent) UpdateMockParams(ctx context.Context, params models.MockFilterPa
 	// Filter out deleted mocks if totalConsumedMocks is provided
 	if params.TotalConsumedMocks != nil {
 		filteredMocks = a.filterOutDeleted(filteredMocks, params.TotalConsumedMocks)
+		unfilteredMocks = a.filterOutDeleted(unfilteredMocks, params.TotalConsumedMocks)
 	}
 
 	// Set the filtered mocks to the proxy
--- a/pkg/service/agent/agent.go
+++ b/pkg/service/agent/agent.go
@@ -1110,6 +1110,7 @@ func (a *Agent) UpdateMockParams(ctx context.Context, params models.MockFilterPa
 	// Filter out deleted mocks if totalConsumedMocks is provided
 	if params.TotalConsumedMocks != nil {
 		filteredMocks = a.filterOutDeleted(filteredMocks, params.TotalConsumedMocks)
+		unfilteredMocks = a.filterOutDeleted(unfilteredMocks, params.TotalConsumedMocks)
 	}
 
 	// Atomically update mocks AND the active test window when the proxy
--- a/pkg/service/agent/agent_test.go
+++ b/pkg/service/agent/agent_test.go
@@ -1,0 +1,136 @@
+package agent
+
+import (
+	"context"
+	"testing"
+
+	coreAgent "go.keploy.io/server/v3/pkg/agent"
+	"go.keploy.io/server/v3/pkg/agent/proxy/integrations"
+	"go.keploy.io/server/v3/pkg/models"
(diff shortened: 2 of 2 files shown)


VivekArgSharma:

VivekArgSharma:
recheck

charankamarapu:
Hi @VivekArgSharma were you able to replicate the problem provided in the issue...? If yes, can you give the sample in which you have replicated it and steps to replicate the issue.

Mrsandeep27 (commented):
Good catch — unfiltered mocks getting stale \"deleted\" entries while the filtered set stays clean was exactly the kind of divergence that silently breaks fallthrough behavior. The fix is minimal and surgical.

A few thoughts before merge:

**1. Nice test.** \`proxySpy\` cleanly verifies both slices. Worth extending one case to assert the **order** is preserved after filtering — deletion-by-name works today, but if the agent ever switches to index-based state tracking, silently reordering the unfiltered slice would break downstream consumers.

**2. `filterOutDeleted` is now called twice** with the same \`totalConsumedMocks\` slice. If \`filterOutDeleted\` builds a lookup map internally, you're rebuilding it twice per call. Tiny perf thing — for a hot path, caching the lookup once at the top of the function and passing it into both calls would be cleaner. Not a blocker.

**3. Semantics question.** Before this PR, \`unfilteredMocks\` was intentionally allowed to contain \"deleted\" entries — was that a deliberate design (so consumers could see what *was* there and compare) or just an oversight? If it was deliberate anywhere (logging, debug UI), this PR becomes a behavior change for those consumers too. A line in the PR description confirming you audited the \`unfilteredMocks\` consumers for this would help reviewers sleep better.

**4. Commit trailer.** The PR title is \"fix: apply consumed state to unfiltered mocks\" but the task list shows \"10 of 25 tasks\" — is that the contribution checklist, or is there more work remaining for this PR? If the core fix is done, worth flipping those remaining items off or calling out which ones were intentionally skipped.

Direction looks right — small diff, clear intent, has a regression test.

slayerjain:
Maintainer note — verifying #4430, which just landed on `main`.

That PR replaced the CLA workflow's exact-equality gate with `contains()`. The old gate required the comment body to equal `recheck` byte-for-byte, so a signature comment carrying the trailing period our own bot asked for never matched, and a typed `recheck` carrying a CRLF matched only intermittently. When the gate didn't match, the step was skipped, the action never ran, and contributors who had signed correctly sat on a red check until their next push.

This comment deliberately wraps `recheck` in surrounding text — a form the previous gate would have rejected outright. If the fix works, the CLA Assistant step now runs instead of being skipped. Sorry for the noise on your PR, @VivekArgSharma; this is the least disruptive place to prove it.

Issue #4083, opened 2026-04-20
[bug]: mappings.yaml backfill attributes mocks to wrong test; self-heal warning spam; agent mode unset on record/test


slayerjain:
## 👀 Is there an existing issue for this?

- [x] I have searched and didn't find similar issue

## 👍 Current behavior

Three distinct defects surface on the mappings.yaml pipeline in `keploy record` → `keploy test`, all reproducible end-to-end against a single sample app that exercises Mongo + Redis + Postgres + HTTP in the same request.

**1. Cross-test mock leak during mappings.yaml backfill.**
When a mapping file is not written at record time (the default, since top-level `disableMapping: true`) and is backfilled on first replay via `test.updateTestMapping: true`, the generated file attributes mocks to the wrong test case. Example observed in a reproducer: `mock-5` (a Generic/Redis `HELLO` captured at `T-14s` during app startup, before any test request arrived) ends up inside `test-3`'s `mock_entries`, because a Redis client reconnect during `test-3`'s replay consumed that mock.

Root cause: `upsertActualTestMockMapping` (`pkg/service/replay/utils.go:250`) unconditionally appends every mock returned by `GetConsumedMocks()` to the current test case. Attribution tracks **replay-time consumption order**, so any reconnect, retry, or test flake moves a mock to a different owner. Record-time timestamps on the mocks are ignored even though they're available (`ReqTimestampMock` / `ResTimestampMock` on `MockState`).

**2. Noisy "self-healing per-kind tree: global update succeeded but per-kind missed" warning.**
Fires 9-12 times per replay for legitimate mock promotions (e.g. Postgres matcher moving a mock from the filtered pool to consumed state via `UpdateUnFilteredMock`). Root cause is two-part: (a) `UpdateUnFilteredMock`/`DeleteUnFilteredMock`/`DeleteFilteredMock` in `pkg/agent/proxy/mockmanager.go` update the global tree outside `m.treesMu` and then the per-kind tree inside, so a concurrent `SetUnFilteredMocks` swap or parallel update on the same mock can interleave and break the per-kind exact-key lookup; (b) the message is emitted at WARN, but the self-heal `insert` below it recovers correctly, making the log line noise.

**3. Enterprise `agent` subprocess starts with `mode=""`, times out on readiness.**
[...]

PR #4084, opened 2026-04-20
fix: correct mappings.yaml backfill, silence self-heal spam, restore agent mode wiring

slayerjain:

Three correctness fixes on the `mappings.yaml` backfill pipeline, all traced to a single investigation against a combined sample app that uses HTTP + Mongo + Postgres + Redis multiple times in the same request.

1. **`fix(replay)`** — `upsertActualTestMockMapping` (`pkg/service/replay/utils.go`) now scopes the mocks it assigns to each test case by the test's record-time request/response window (`HTTPReq.Timestamp` / `HTTPResp.Timestamp`, with gRPC + `Created` fallbacks via a new `recordReqResTimestamps` helper). Previously it appended every mock returned by `GetConsumedMocks()` in replay-order, so an app-startup mock that happened to be consumed during a later test's replay (for example a Redis `HELLO` after a client reconnect) got attributed to the wrong test in the backfilled mapping. DNS mocks are exempt because they have no stable record timestamps; when both timestamps are zero the legacy append-all behavior is kept for old fixtures. Call sites in `replay.go` updated accordingly for both the HTTP/gRPC and streaming paths.

2. **`fix(proxy)`** — `UpdateUnFilteredMock` / `DeleteUnFilteredMock` / `DeleteFilteredMock` in `pkg/agent/proxy/mockmanager.go` now perform the global and per-kind tree mutations inside one `m.treesMu.Lock()` scope, with the per-kind tree lazy-init inlined under the same lock. Previously the global update ran outside `treesMu`, opening a window where a concurrent `SetUnFilteredMocks` swap or a parallel update on the same mock could leave the global and per-kind trees disagreeing on whether the old key was still present — the "self-healing per-kind tree: global update succeeded but per-kind missed" warning fired 9–12 times per replay (PostgresV2 path) on current `main`. The message is also demoted from `Warn` to `Debug` and reworded to reflect the legitimate promotion-from-filtered case, since the recovery `insert` below it is correct and the log line is just noise on the hot path.

3. **`fix(cli)`** — `CmdConfigurator.ValidateFlags` (`cli/provider/cmd.go`) now checks `Lookup("json")` before calling `GetBool`, instead of always returning an error when the flag isn't registered on the subcommand. With the prior behavior, the enterprise build (which only registers `--json` on UI commands) short-circuited `ValidateFlags` before reaching the `case "agent":` branch further down that wires `c.cfg.Agent.Mode` from `--mode`. The enterprise agent process therefore started with `mode=""`, bound its HTTP server on port 0, and the parent timed out with `keploy-agent did not become ready in time`. With the fix, the enterprise agent initializes correctly on native record/test.

**End-to-end validation** (sample app: Express + `mongodb@6.8` + `redis@4.7` + `pg@8.12` + local HTTP upstream, five endpoints including a `/combo` that touches every dependency):

| Metric                                                     | Before (current `main` / v3.3.75) | After (this branch) |
|-----------------------------------------------------------|-----------------------------------|---------------------|
| App-startup mock leaked into later test's `mock_entries`   | yes (e.g. `mock-5` → `test-3`)    | **no** — every mock's `reqTimestampMock` falls inside its test's recorded window |
| `self-healing per-kind tree` warnings per replay           | 9–12 (PostgresV2)                 | **0**               |
| Record→replay pass rate (OSS, Redis falls back to Generic) | 2/3 on baseline reproducer        | 4/5 (test-5 is Redis-only; unrelated, Redis parser is enterprise) |
| Record→replay pass rate (enterprise, Redis parser active)  | agent hung at port 0               | **5/5** — including the Redis-heavy combo endpoint |
| First replay generates `mappings.yaml`                     | yes                               | yes                 |
| Second replay loads generated file                         | `hasMeaningfulMappings: true`     | `hasMeaningfulMappings: true` |

## Links & References

**Closes:** #4083

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

The existing e2e harnesses cover record/replay per-sample-app but don't exercise the mapping backfill flow across a combined multi-dependency request. Happy to extend the fixture matrix in a follow-up if the team has a preferred landing spot (e.g. `samples-typescript`).

- [x] 👍 yes

In particular: the filter rationale on `upsertActualTestMockMapping`, the lock-scope contract on `UpdateUnFilteredMock`, and the rewritten self-heal comment explaining the legitimate filtered→unfiltered promotion case.

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

- [x] 👍 yes, mentioned below

**Reproducer**

```bash
# sample-app/server.js (abridged)
app.post('/multi-redis/:k', ...)  // 2 SET + 2 INCR + 3 GET
app.post('/multi-mongo/:u', ...)  // upsert + $inc + findOne + countDocuments
app.post('/multi-pg/:u',    ...)  // 2 INSERT + COUNT + SUM
app.get ('/multi-http',     ...)  // 3 axios.get to http://localhost:4000
app.post('/combo/:u',       ...)  // all four deps interleaved
```

```bash
# 1. Record with defaults (disableMapping:true, updateTestMapping:false)
keploy config --generate
sudo keploy record -c "node server.js" --path . --disable-ansi
# curl every endpoint, SIGINT, confirm no keploy/test-set-0/mappings.yaml

# 2. Flip config for backfill
sed -i 's/^disableMapping: true/disableMapping: false/'   keploy.yml
sed -i 's/    updateTestMapping: false/    updateTestMapping: true/' keploy.yml

# 3. First replay — verify mappings.yaml is generated and every
#    mock in each test's mock_entries has reqTimestampMock inside
#    the test's HTTPReq/HTTPResp window (not before first test started)
sudo keploy test -c "node server.js" --path . --disable-ansi --delay 10
grep -c "self-healing per-kind" replay.log   # expect 0

# 4. Second replay — verify hasMeaningfulMappings: true is logged
sudo keploy test -c "node server.js" --path . --disable-ansi --delay 10
```

## Self Review done?
- [x] ✅ yes

Before (current `main`):
```
WARN  self-healing per-kind tree: global update succeeded but per-kind missed
      {"kind": "PostgresV2", "mockName": "mock-24", "testModeInfo": {"Id":20,"sortOrder":71}}
... × 12 for this one replay

mappings.yaml:
  - id: test-3
    mock_entries:
      - name: mock-5            # captured T-14s during app init, leaked in
        kind: Generic
        reqTimestampMock: "2026-04-20T21:04:47.914..."
      - name: mock-16           # test-3's own PostgresV2 mocks follow
        ...
```

After:
```
INFO  Successfully saved test-mock mappings  {"testSetID": "test-set-0", "numTests": 5}

# grep -c "self-healing per-kind" — 0

mappings.yaml:
  - id: test-3
    mock_entries:
      - name: mock-17          # every reqTimestampMock is inside
        kind: PostgresV2       # test-3's recorded HTTPReq/HTTPResp window
        reqTimestampMock: "2026-04-20T21:34:33.702995..."
        ...
```

## Additional checklist:

### DIFF
--- a/cli/provider/cmd.go
+++ b/cli/provider/cmd.go
@@ -626,11 +626,18 @@ func (c *CmdConfigurator) PreProcessFlags(cmd *cobra.Command) error {
 }
 
 func (c *CmdConfigurator) ValidateFlags(ctx context.Context, cmd *cobra.Command) error {
-	jsonOutput, err := cmd.Flags().GetBool("json")
-	if err != nil {
-		errMsg := "failed to get the json flag"
-		utils.LogError(c.logger, err, errMsg)
-		return errors.New(errMsg)
+	// The --json flag isn't registered on every subcommand (record / agent
+	// don't define it in enterprise builds), so Lookup + fallback avoids
+	// an early-return that would skip every subsequent validation step
+	// below (including the agent-mode wiring that the agent subprocess
+	// needs — without it, the agent starts up with mode="").
+	jsonOutput := false
+	if cmd.Flags().Lookup("json") != nil {
+		if v, err := cmd.Flags().GetBool("json"); err == nil {
+			jsonOutput = v
+		} else {
+			utils.LogError(c.logger, err, "failed to get the json flag")
+		}
 	}
 	c.cfg.JSONOutput = jsonOutput
 
--- a/pkg/agent/proxy/mockmanager.go
+++ b/pkg/agent/proxy/mockmanager.go
@@ -1173,10 +1173,20 @@ func (m *MockManager) UpdateUnFilteredMock(old *models.Mock, new *models.Mock) b
 		m.treesMu.Lock()
 		updatedNewKind = unf.update(old.TestModeInfo, new.TestModeInfo, new)
 
-		// Self-heal if global updated but per-kind missed (e.g., not present yet)
+		// Self-heal if global updated but per-kind missed.
+		//
+		// This is the legitimate filtered→unfiltered promotion case: a
+		// parser (Postgres v2 matcher is the load-bearing caller) calls
+		// UpdateUnFilteredMock on a mock that lived only in the filtered
+		// pool, so the unfiltered per-kind tree was never seeded with
+		// it. The insert below brings per-kind into sync with global so
+		// future per-kind fast-path reads (GetMySQLCounts,
+		// GetPostgresSessionMocks, etc.) see the consumed mock. Kept at
+		// Debug because it fires dozens of times per normal replay and
+		// is not user-actionable.
 		if updatedGlobal && !updatedNewKind {
-			if m.logger != nil {
-				m.logger.Warn("self-healing per-kind tree: global update succeeded but per-kind missed",
+			if m.logger != nil && m.logger.Core().Enabled(zap.DebugLevel) {
+				m.logger.Debug("seeding per-kind unfiltered tree from global on first update",
 					zap.String("kind", string(newK)),
 					zap.String("mockName", new.Name),
 					zap.Any("testModeInfo", new.TestModeInfo),
--- a/pkg/service/replay/replay.go
+++ b/pkg/service/replay/replay.go
@@ -1594,7 +1594,8 @@ func (r *Replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
 				testPass, testResult = r.CompareGRPCResp(testCase, &respCopy, testSetID, emitFailureLogs)
 			}
 
-			upsertActualTestMockMapping(actualTestMockMappings, testCase.Name, consumedMocks)
+			tcReqTime, tcRespTime := recordReqResTimestamps(testCase)
+			upsertActualTestMockMapping(actualTestMockMappings, testCase.Name, consumedMocks, tcReqTime, tcRespTime)
 
 			// log the consumed mocks during the test run of the test case for test set
 			r.logger.Debug("consumed mocks for test case",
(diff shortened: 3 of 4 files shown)


slayerjain:
Revised: the first push of this branch reshuffled tree-mutation locking in UpdateUnFilteredMock/Delete{Un,}FilteredMock under a single m.treesMu.Lock scope. That version stayed green on all Linux fuzzer jobs (Mongo/Postgres/gRPC/MySQL) but regressed run_golang_native_windows / gin-mongo — the Windows mongo v2 matcher started reporting mongo mock miss — closing connection for test-set-1's update/find operations after the test-set transition. Could not reproduce on Linux locally (7/7 pass against samples-go/gin-mongo with private integrations linked), so the lock reshuffle looks platform-sensitive.

This revision narrows the mockmanager change to the log demotion alone (WARN → Debug). The self-heal insert below it already kept per-kind consistent with global on first touch — that's what actually mattered for customers. The lock-ordering cleanup can be revisited in a follow-up once the Windows reproducer is understood.

Also added DCO sign-offs on all three commits and rebased onto latest origin/main.

slayerjain:
Pushed an empty chore re-trigger. Previous Windows gin-mongo job (id 72176310203) ran 30 minutes on 'Run gin-mongo application' and was cancelled by the runner's timeout. Cross-referencing the three commits on this branch:

- `fix(replay)`: replay-only code path in `pkg/service/replay/{utils,replay}.go`. No effect on record mode.
- `fix(proxy)`: mockmanager diff is log-level only (Warn → Debug) inside `UpdateUnFilteredMock`, which parsers only call during replay-time matching. No effect on record mode.
- `fix(cli)`: semantic no-op on OSS, because OSS registers `--json` as a `PersistentFlag` on the root `keploy` command (cli/provider/cmd.go:295), so `cmd.Flags().GetBool("json")` already returned `(false, nil)` — the `Lookup` guard only changes behaviour on enterprise builds where the flag is NOT registered on record/test/agent.

Linux gin-mongo (all three variants) passes on this branch, and locally I reran the exact same two-iteration record + replay flow against samples-go/gin-mongo with the same OSS binary + private integrations linked → 7/7 green. The Windows record-iteration-2 hang at "Sending traffic..." looks like a runner-level infra flake rather than a code defect on this PR.

PR #4091, opened 2026-04-21
feat: log pruned mock details during UpdateMocks

ayush3160:
## Summary
- Collect each mock dropped by `UpdateMocks` (name, kind, metadata) into a single slice and log it on the existing `pruned mocks successfully` debug line.
- Applied to both the YAML and gob prune paths in `pkg/platform/yaml/mockdb/db.go` so behavior stays consistent across formats.
- Kept as one structured log per prune call (not one per mock) to avoid log pollution on large test sets.

## Test plan
- [x] `go build ./pkg/platform/yaml/mockdb/...`
- [x] `go vet ./pkg/platform/yaml/mockdb/...`
- [ ] Run replay with `--debug` on a test set that triggers pruning and confirm the `prunedMocks` array shows name/kind/metadata for each dropped mock.

### DIFF
--- a/pkg/platform/yaml/mockdb/db.go
+++ b/pkg/platform/yaml/mockdb/db.go
@@ -122,6 +122,16 @@ type MockYaml struct {
 	gobFlushErr error
 }
 
+// prunedMockInfo is the structured per-mock entry logged when
+// UpdateMocks drops a mock. Collected into a single slice and emitted
+// once per prune call so operators can see exactly which mocks were
+// dropped without flooding the log with one line per mock.
+type prunedMockInfo struct {
+	Name     string            `json:"name"`
+	Kind     string            `json:"kind"`
+	Metadata map[string]string `json:"metadata"`
+}
+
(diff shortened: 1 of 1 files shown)


gouravkrosx (approved):

ayush3160:
Addressed Copilot's feedback in 4614c54a:

- **Gate per-mock collection on debug level.** Introduced `debugEnabled := ys.Logger.Core().Enabled(zap.DebugLevel)` in both prune paths and fall back to a plain `prunedCount` integer when debug is off — no more allocation / reflection cost on large test sets when the log would be dropped.
- **Cap the logged slice.** New `maxPrunedMocksLogged = 100` constant. The slice is pre-sized to the cap, `append` short-circuits once full, and the log now emits `prunedMocksTruncated=true` whenever the real pruned count exceeds what's logged — so `--debug` on ~10^5-mock test sets won't produce multi-MB log lines.

Applied symmetrically to both YAML and gob prune paths. `go build` + `go vet` pass.

PR #4106, opened 2026-04-22
fix(runner): align UpdateMockParams contract with replay + agent mock-lifetime rewrite

charankamarapu:
## Motivation

Commit 2196aba3 (#4073 — "mock-lifetime unification — session / connection / per-test") substantially rewrote the agent-side contract for `UpdateMockParams`:

- **`TotalConsumedMocks` is now load-bearing.** `Agent.UpdateMockParams` calls `filterOutDeleted` on every invocation — mocks whose cumulative `Usage == Deleted` are stripped from the re-pushed pool, and mocks with non-Deleted state have their `IsFiltered` + `SortOrder` carried forward. Without the client sending `TotalConsumedMocks`, per-test mocks consumed in step N *resurrect* in step N+1 and stateful-protocol `SortOrder` never advances across steps.
- **Per-test window is pushed into `MockManager` via `SetMocksWithWindow`.** MockManager uses it for (a) startup-init promotion — mocks with `req < firstWindowStart` (HikariCP pool warm-up, JDBC handshake, Hibernate init) are promoted to session lifetime rather than dropped; (b) strict-window containment — out-of-window per-test mocks are dropped under strict mode.
- **Strict-window is now default** (#4096 — `feat(test): enable strictMockWindow by default`). The effective flag is `params.StrictMockWindow OR KEPLOY_STRICT_MOCK_WINDOW=1`, with an explicit `0` winning over both.

The runner had already been refactored to the once-per-`TestSetID` cached setup pattern, but had drifted from the shape the post-#4073 agent actually consumes. This PR brings runner back in line with `replay.go`'s `RunTestSet` (which is the reference implementation that feeds the same agent API).

Independently, I discovered that runner had a pre-existing agent-readiness race in docker-compose / sandbox mode: `MockOutgoing` fires immediately after `Setup`, but the keploy-agent container's host-port forward hadn't finished binding. Every test case in the set fails with `dial tcp localhost:<port>: connect: connection refused`. Replay gates this at `replay.go:939-952` via `AgentHealthTicker`; runner did not. Fix included here as a second commit on the branch.

### 1. Track cumulative consumed mocks per TestSetID
- Added `totalConsumed map[string]models.MockState` + `consumedMu sync.Mutex` to `testSetSetup`, with `mergeConsumed` / `snapshotConsumed` helpers.
- After every step's `executeAndCompare`, call `GetConsumedMocks` and merge the result into `setup.totalConsumed`.
- `sendPerTestParams` now passes `TotalConsumedMocks: setup.snapshotConsumed()` on every `UpdateMockParams` call.
- Effect: per-test mocks deleted in step N stay deleted in step N+1 (no resurrection on pool re-push), and `SortOrder` / `IsFiltered` state for session/connection mocks survives the re-push.

Mirrors replay's local `totalConsumedMocks` map (`replay.go:855`, updated at `:1470-1486`) and its per-step `SendMockFilterParamsToAgent` call at `replay.go:1433`.

### 2. Initial wide-window `UpdateMockParams` in setup
- `setupTestSet` now emits `UpdateMockParams{AfterTime: BaseTime, BeforeTime: now, MockMapping: []string{}}` after `MakeAgentReadyForDockerCompose`, before returning.
- Purpose: establishes a valid filter window on MockManager before the app's startup-init traffic lands. Without it, any traffic the app emits between `MakeAgentReadyForDockerCompose` and the first per-step `UpdateMockParams` has no window to match against — under strict mode (now default) those mocks get dropped.

Mirrors `replay.go:1047` (compose path) and `replay.go:1156` (non-compose path).

### 3. Respect `config.DisableMapping`
- `useMappingBased` is now derived as `len(mappings) > 0 && !config.DisableMapping`. Previously hardcoded `true`.
- Stored on `testSetSetup` and passed on every `UpdateMockParams`.

Mirrors `replay.determineMockingStrategy` at `replay.go:3443`.

### 4. Filter DNS entries from mismatch report
- `checkMockMismatches` now drops any expected or consumed mock whose kind is `models.DNS`.
- Uses a name→Kind registry (`mockKindByName`) built from the loaded pool — mapping entries frequently have empty `Kind`, so the registry is the authoritative source.
- Effect: tests that resolve a hostname no longer produce false-positive "DNS mock not consumed" entries in the `MockMismatches` struct. DNS resolution order is non-deterministic and cannot be meaningfully asserted.

Mirrors `replay.go:1499-1517`.

### 5. Seed process-wide sort counter
- After loading mocks, `setupTestSet` calls `keployPkg.InitSortCounter(max(len(filtered), len(unfiltered)))`.
- Prevents collision between recorded mock sort-order and any live-generated sort numbers during replay.

Mirrors `replay.go:1128`.

### 6. Reorder: `waitForApp` before per-test `UpdateMockParams`
- `RunTest` now calls `waitForApp` *before* `sendPerTestParams`.
- Keeps the initial wide-window filter (from item 2) active during app startup traffic, rather than immediately narrowing to the first step's window.

Matches replay's ordering: initial broad `SendMockFilterParamsToAgent` precedes the first per-test one (`replay.go:1047/1156 → 1433`).

### 7. Remove discarded pre-step `GetConsumedMocks`
- Old code called `GetConsumedMocks` and threw away the result ("to rule out config mocks from consumed list" comment, but nothing consumed the value). Dead code — replaced by the post-step call in item 1, which actually uses the result.

### 8. Agent-health gate before `MockOutgoing`
- `setupTestSet` now waits on `keployPkg.AgentHealthTicker` before firing `MockOutgoing` when `config.Agent.AgentURI` is configured.
- 120-second timeout, 1-second poll interval.
- Skipped when `AgentURI` is empty (in-process embedders where the agent lives in the same process).

Fixes the docker-compose / sandbox startup race where `MockOutgoing` fires before the keploy-agent container's host-port forward is bound, producing `connect: connection refused` on the agent URL and failing every test case in the set with "failed to load mocks".

Mirrors `replay.go:939-952`.

## Struct changes

```go
type testSetSetup struct {
    id              string
    mappings        map[string][]models.MockEntry
    mockKindByName  map[string]models.Kind       // new: name→Kind lookup for DNS filter
    useMappingBased bool                          // new: honours DisableMapping
    cleanup         func()

    consumedMu    sync.Mutex                      // new
    totalConsumed map[string]models.MockState     // new: cumulative across steps
}
```

New helpers: `mergeConsumed`, `snapshotConsumed`, `(*Runner).strictMockWindow`.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [ ] Sandbox integration test against an app with real DB mocks — confirms agent-health gate resolves the `connection refused` failure mode observed on `orderflow` / `docker compose up` runs.
- [ ] Multi-step test-set with a Postgres mock used by two steps — confirms step 2 does not re-match the mock step 1 consumed.
- [ ] Test-set with a test that resolves a hostname — confirms the `MockMismatches` output no longer contains DNS entries.
- [ ] Run under `test.strictMockWindow: true` (current default) — confirms HikariCP-style startup-init traffic still replays.
- [ ] Single-step HTTP-only test-set — confirms no regression in the simple path.
- [ ] Config with `disableMapping: true` — confirms runner falls back to timestamp-only filtering.

## Rollback notes

Changes are confined to `pkg/service/runner/runner.go`. No API/struct changes are exported beyond the existing `Service`/`RunTestOpts`/`TestResult` surface. Reverting this PR returns the runner to hardcoded `UseMappingBased: true`, no `TotalConsumedMocks`, no DNS filtering, and no agent-health gate — which is the pre-#4073-alignment behaviour.

## Related

- #4073 — Agent mock-lifetime unification that introduced the new `UpdateMockParams` contract this PR aligns against.
- #4096 — Strict mock window enabled by default, which makes item 2 (initial wide window) load-bearing rather than cosmetic.


gouravkrosx (approved):
LGTM!

PR #4117, opened 2026-04-23
fix(record/http): persist chunked captures instead of skipping on partial reassembly

slayerjain:
## Summary
- `handleHttp1Connection` (synchronous + sampling-tracked ingress path in `pkg/agent/proxy/incoming/http.go`) was dropping every HTTP exchange with `Transfer-Encoding: chunked` — either direction — with the log line *"Skipping testcase capture for streaming exchange"*. Because Go's `http.ResponseWriter` defaults to chunked framing whenever the handler omits `Content-Length`, this hit basically every modern JSON REST endpoint and produced the "1-of-25 tests persisted, rest orphaned as mocks" symptom on stress runs.
- The skip was gated purely on the chunked flag; there was no framing / integrity / ErrUnexpectedEOF check behind it, so it was always dropping valid captures. The `reqCapture/respCapture.Truncated()` guard already rejects oversized streams, so the chunked-only bail was redundant as well as incorrect.
- Fix removes the `chunked -> captureEnabled=false` + `chunked -> skip` branches. The `tee` attached to `req.Body` / `resp.Body` reads the already-decoded body bytes (net/http chunk-decodes transparently) and the existing capture-budget check downstream handles genuinely oversized streams. Early `releaseLock()` on streaming exchanges is preserved so a slow upload still can't wedge the sync semaphore.

## Root cause (short version)
`if forceCloseMode && (req.ContentLength == -1 || isChunked(req.TransferEncoding)) { captureEnabled = false }` and the mirrored response-side check + post-exchange `if chunked { shouldCapture = false }` were short-circuiting the happy path for any chunked exchange.

## Test plan
- [x] `go build ./...` — clean
- [x] `go vet ./pkg/agent/proxy/incoming/...` — clean
- [x] `go test ./pkg/agent/proxy/...` — all green including the new regressions
- [x] New `TestHandleHttp1Connection_ChunkedExchangeIsCaptured` — httptest upstream returns an implicitly-chunked JSON response; the test asserts `CaptureHook` fires exactly once with the fully-decoded body.
- [x] New `TestHandleHttp1Connection_ChunkedRequestIsCaptured` — client sends a manually chunk-framed POST; the test asserts the upstream receives the decoded body and `CaptureHook` fires with the captured request body.
- [x] Verified both tests FAIL on the pre-fix code (stashed `http.go`, ran the new tests: `CaptureHook was never invoked for chunked exchange — the skip bug has regressed`).

## Scope
Surgical — only the chunked-skip decisions in `pkg/agent/proxy/incoming/http.go`. No changes to:
- the V2/streaming zero-copy path (`handleHttp1ZeroCopy` / `parseStreamingHTTP`) — that path never had the bug
- the HTTP/1 parser in `pkg/agent/proxy/integrations/http/` — that's PR #4113 territory and the bug does not live there

### DIFF
--- a/pkg/agent/proxy/incoming/http.go
+++ b/pkg/agent/proxy/incoming/http.go
@@ -350,9 +350,14 @@ func (pm *IngressProxyManager) handleHttp1Connection(ctx context.Context, client
 		// Request modifications for sync/sampling modes.
 		if forceCloseMode {
 			if req.ContentLength == -1 || isChunked(req.TransferEncoding) {
+				// Release the lock early for streaming exchanges so a
+				// single slow/streaming upload doesn't wedge the sync
+				// semaphore. Capture stays enabled: tee reads the
+				// already-decoded body bytes, and the downstream
+				// capture-budget check (respCapture.Truncated()) handles
+				// genuinely oversized streams.
 				releaseLock()
 				chunked = true
-				captureEnabled = false
 			} else if captureEnabled && pm.synchronous && acquiredLock {
 				mgr := syncMock.Get()
 				if !mgr.GetFirstReqSeen() {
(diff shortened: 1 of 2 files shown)


slayerjain:
Superseded by #4120 (v3 post-merge-suite consolidation, merged). Closing to reduce queue noise.

PR #4118, opened 2026-04-23
feat(replay): poll --health-url instead of fixed delay; --delay stays as fallback

slayerjain:
## Summary
- Replace the unconditional pre-test `time.After(Delay)` sleep in `pkg/service/replay/replay.go` (both sites, ~L1057 and ~L1202) with an active HTTP health poll when the new `Test.HealthURL` is set. First 2xx = app up, proceed immediately.
- Two new flags on `keploy test` / `keploy rerecord`: `--health-url` and `--health-poll-timeout` (default 60s), plumbed via the existing viper/mapstructure config (`Test.HealthURL`, `Test.HealthPollTimeout`).
- Default behavior is byte-for-byte unchanged: empty `HealthURL` keeps the historical fixed-delay path. If polling the configured URL never sees a 2xx within `HealthPollTimeout`, we log WARN `health probe timed out, falling back to fixed delay` and fall through to the existing `time.After(Delay)` sleep - so users are never worse off than today.

## Poll loop shape
500ms ticker, 1s per-request GET timeout, first 2xx wins, ceiling = `--health-poll-timeout`. Ctx cancel unblocks immediately. Timeout = WARN + fall through to `time.After(Delay)`.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./pkg/service/replay/...` (new `health_poll_test.go`, all pass):
  - [x] 200 on first try proceeds <1s
  - [x] 503x3 then 200 proceeds after >=3*500ms (~1.5s)
  - [x] Never-2xx falls back to `--delay` with WARN logged (verified via zap observer)
  - [x] Empty `HealthURL` never hits the network (poller stubbed to fail the test)
  - [x] Ctx cancel during poll returns false (triggers `TestSetStatusUserAbort` in caller)
- [ ] Manual smoke on a real app with a slow-starting container + `--health-url=http://localhost:8080/healthz`

Built on main @ fc849d99.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cli/provider/cmd.go
+++ b/cli/provider/cmd.go
@@ -332,6 +332,8 @@ func (c *CmdConfigurator) AddUncommonFlags(cmd *cobra.Command) {
 		cmd.Flags().Uint32("grpc-port", c.cfg.Test.GRPCPort, "Custom grpc port to replace the actual port in the testcases")
 		cmd.Flags().Uint32("sse-port", c.cfg.Test.SSEPort, "Custom SSE port to replace the actual port in the SSE testcases")
 		cmd.Flags().Uint64P("delay", "d", 5, "User provided time to run its application")
+		cmd.Flags().String("health-url", c.cfg.Test.HealthURL, "HTTP(S) URL polled before the first test is fired; first 2xx response proceeds immediately. Empty (default) preserves the fixed --delay behavior.")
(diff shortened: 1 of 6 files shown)


slayerjain:
Superseded by #4120 (v3 post-merge-suite consolidation, merged). Closing to reduce queue noise.

PR #4147, opened 2026-04-27
fix(http-recorder): stamp reqTimestamp from wire-arrival, not post-parse

slayerjain:
## Summary

The sync-path HTTP capture loop in `pkg/agent/proxy/incoming/http.go` sampled `reqTimestamp := time.Now()` **after** `http.ReadRequest` returned. Under HTTP/1.1 keepalive with a `bufio.Reader` over the client conn, that lags the actual wire-arrival of the request by parser-iteration time + bufio fill jitter — tens to hundreds of microseconds.

Downstream parser recorders (postgres v3 `query_capture.go`, etc.) stamp their own captures at decode time, which closely tracks wire-read time. So the recorder ends up writing `tc.HTTPReq.Timestamp` values a few µs **later** than the postgres `reqTimestampMock` for queries the SUT issued in direct response to that HTTP request.

At replay time, `mockmanager.GetFilteredMocksInWindow` does **strict containment** of recorded postgres `reqTimestampMock` against `[HTTPReq.Timestamp, HTTPResp.Timestamp]`. A µs-scale negative delta is enough to push a perfectly-correct postgres mock outside the window's left edge, leaving the per-test cohort empty for that SQL hash and surfacing as `transactional: no invocation matched ... candidates: 0` even though the mock exists in mocks.yaml.

This was the **residual pgtype-tour-postgres flake** after #4141 fixed the TLS read-timestamp drift: customer-tag-cohort and listmonk landed clean, but pgtype-tour still produced 5 of 174 spurious failures across stress iterations of `record-build-replay-build` with the SAME binary on both ends.

## Root cause evidence

Local stress repro on `pgtype-tour-postgres` with `keploy@d9997aea` (which already includes #4141):
- iteration 1: 174/174 PASS
- iteration 2: 169/174 (5 fails) — same binary, same exerciser, just a different scheduling pattern

Failing tests' HTTP request timestamp vs the matching postgres invocation timestamp:

| test | hash short | HTTP req start | nearest pg reqTs | delta |
|---|---|---|---|---|
| 7  | INSERT measurements   | 13:46:02.964441843 | 13:46:02.964429143 | **−12.7 µs** |
| 21 | SELECT network_assets | 03.231123109       | 03.231106429       | **−16.7 µs** |
| 43 | SELECT measurements   | 03.522456461       | 03.522361052       | **−95 µs**   |
| 65 | SELECT ranges         | 03.725984432       | 03.725399151       | **−585 µs**  |

The pg invocation arrived a few µs **before** its triggering test's HTTP request was stamped — not because pg arrived early, but because the HTTP recorder stamped late.

## Fix

Wrap `clientConn` in a `wireTimeConn` that stamps the time of every non-empty `Read` on an atomic field. The sync-loop samples `wireConn.LastReadTime()` after `http.ReadRequest`, getting the timestamp of the socket Read whose buffer produced the parsed bytes — i.e. wire-arrival time, not post-parse time.

On HTTP/1.1 keepalive where bufio prefetches the next request's bytes during a prior Read, `LastReadTime` is still on or before THIS request's actual arrival (never later), preserving the strict window containment invariant downstream parsers depend on.

The streaming HTTP path (`parseStreamingHTTP`, used for chunked / unknown-length / non-sync exchanges) already uses `asyncPipeFeeder.LastReadTime()` and was unaffected; this brings the sync path up to the same wire-time semantics.

## Validation

- `go test ./pkg/agent/proxy/...` — all green.
- New unit test `TestWireTimeConn_LastReadTimePrecedesBufioParseCompletion` locks in the contract: `LastReadTime` must be sampled inside the Read syscall, never later than the call site that consumes the buffered bytes.
- CI's `pgtype-tour-postgres` matrix is the integration-level proof point — once this lands, integrations main bumps `go.mod` to pick up the fix and the `record-build-replay-build` cell flips deterministically green.

## Test plan

- [x] Unit test for the wrapper's read-time semantics
- [x] Full proxy package test suite
- [ ] Woodpecker CI on this PR
- [ ] After merge: bump `integrations/go.mod` and confirm pgtype-tour-postgres + sap-demo-java-postgres land clean

### DIFF
--- a/pkg/agent/proxy/incoming/http.go
+++ b/pkg/agent/proxy/incoming/http.go
@@ -99,6 +99,51 @@ type timedChunk struct {
 	readAt time.Time
 }
 
+// wireTimeConn wraps a net.Conn and stamps the time of the most recent
+// non-empty Read on an atomic field. The sync-path HTTP capture loop
+// uses this to sample the wire-arrival time of an HTTP request: by the
+// time http.ReadRequest returns the parsed request, the bufio.Reader
+// behind it has already pulled the request's bytes off the socket via
+// one or more Read calls on the wrapped conn, each of which updates
+// lastReadNano. Sampling lastReadNano AFTER ReadRequest returns gives
+// the timestamp of when those bytes arrived on the wire, not the
+// (later) time at which parsing completed.
+//
+// This matters because per-test mock windowing (mockmanager
+// GetFilteredMocksInWindow) does strict containment of recorded
+// invocation reqTimestampMock against [HTTPReq.Timestamp,
+// HTTPResp.Timestamp]. If the HTTP recorder stamps reqTimestamp from
+// time.Now() AFTER parse, downstream parser recorders (postgres v3)
+// that stamp their own captures at decode time can produce
+// reqTimestampMock values a few microseconds EARLIER than the HTTP
+// reqTimestamp, falling outside the window's left edge and causing
+// otherwise-correct mocks to be filtered out at replay. See
+// https://github.com/keploy/integrations/pull/151 for the related
+// pgtype-tour-postgres flake symptom (`candidates: 0` for SQL hashes
+// that exist in the recorded mocks.yaml).
+type wireTimeConn struct {
+	net.Conn
+	lastReadNano atomic.Int64
+}
+
+func (c *wireTimeConn) Read(p []byte) (int, error) {
+	n, err := c.Conn.Read(p)
+	if n > 0 {
+		c.lastReadNano.Store(time.Now().UnixNano())
+	}
+	return n, err
+}
+
+// LastReadTime returns the time of the most recent non-empty Read on
+// the wrapped conn, or the zero time if no bytes have been read yet.
+func (c *wireTimeConn) LastReadTime() time.Time {
+	n := c.lastReadNano.Load()
+	if n == 0 {
+		return time.Time{}
+	}
+	return time.Unix(0, n)
+}
+
 // Global cap on the bytes held in flight across every asyncPipeFeeder's
 // channel. The per-feeder channel is sized to tolerate large response
 // bodies (5 MB needs ~160 slots at 32 KB/chunk), so under stress workloads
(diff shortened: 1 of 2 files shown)


slayerjain:
## RCA: gin-mongo `record_build_replay_build` regression on this PR

The first revision over-corrected on HTTP/1.1 keepalive. Pushed a follow-up commit (\`6674d3d\`) clamping \`reqTimestamp\` at iteration-start.

### Failure mode

On the first revision, gin-mongo `test-3` (POST /url) failed with `connection(localhost:27017) socket was unexpectedly closed: EOF`, while `test-1` and `test-2` on the same recording passed.

`test-1` and `test-2` both fired POST /url against the same keepalive connection. Each handler INSERTs into mongo. The keploy proxy intercepts the mongo wire and stamps each mongo request mock with its capture timestamp. The dispatcher then filters mocks per-test by strict containment in `[HTTPReq.Timestamp, HTTPResp.Timestamp]`.

`bufio.Reader` ahead of `http.ReadRequest` can serve request N entirely from bytes prefetched during the prior iteration's underlying socket Read (which was consuming request N-1's tail). When that happens:

- `wireConn.LastReadTime()` is from request N-1's read window — **DURING** the previous test's handler.
- Using it raw as request N's `reqTimestamp` pushes the per-test mock-window left edge backwards **across the previous test boundary**.
- Mongo mocks captured during request N-1's handler now fall inside BOTH windows.
- Dispatcher's window-then-FIFO selection returns a stale mock whose request bytes don't match the live wire bytes → keploy closes the conn on the mismatch → mongo client sees EOF.

`test-1` was unaffected because `lastReadNano` is initially zero (falls back to `time.Now`). `test-2` was unaffected because its window left edge falls before `test-1`'s mocks fired (no prior-test contamination possible).

### Fix

Capture `iterStart := time.Now()` BEFORE `ReadRequest`, then clamp:

```go
reqTimestamp := wireConn.LastReadTime()
if !reqTimestamp.After(iterStart) {
    reqTimestamp = iterStart
}
```

- When `ReadRequest` drives a fresh socket Read this iteration (the common keepalive case where bufio's buffer was empty), `lastReadNano > iterStart` → use `lastReadNano`. Tightest pre-parse upper bound on actual wire arrival.
- When bufio serves entirely from buffer (no fresh Read this iter), `lastReadNano ≤ iterStart` → fall back to `iterStart`. Always after the previous response was written, so prior-test mocks are guaranteed to be outside this test's window.

### Regression test

`TestReqTimestampClampedAtIterStartUnderBufioPrefetch` simulates the prefetch case (primes bufio's buffer via Peek before `iterStart`, then drives `ReadRequest` entirely from buffer) and asserts the clamp falls back to `iterStart`, not the stale prior-iter `lastReadNano`.

The original `TestWireTimeConn_LastReadTimePrecedesBufioParseCompletion` still passes — the wrapper's internal contract is unchanged; only the call-site interpretation tightened.

### Note on the spring-petclinic failure

That one's a runner-cleanup timeout — Spring Boot didn't reach "Started PetClinicApplication" before GH Actions terminated orphan PIDs (4068 sh, 4069 java). It's not coupled to this code change. Will confirm once CI re-runs against the new commit.

PR #4150, opened 2026-04-28
feat(postgres-v3): cross-window autopromote at agent before window-partition

slayerjain:
Integration-Ref: feat/postgres-v3-lifetime-autopromote

## Summary
- Wires the new integrations-side `EvaluateCrossWindowAutopromote` helper (companion change in keploy/integrations#160) into `ApplyMockFilters` so postgres-v3 cohorts that span multiple test windows are visible to the autopromote heuristic in their entirety.
- Mutates `mock.TestModeInfo.Lifetime` in place to `LifetimeSession` for promoted mocks and re-routes them from the per-test pool into the session pool BEFORE the per-test window-partition runs.
- Env off-switch: `KEPLOY_PG_V3_CROSS_WINDOW_AUTOPROMOTE=0`.

## Why this exists

The integrations-side `ApplyAutopromote` runs inside `PerTestProvider` AFTER the per-test pool has been window-filtered. A SQL hash that fires once per test across N tests presents to that pass as N separate `cohort_size=1` evaluations and never trips the heuristic's `len(candidates) >= 2` gate.

The listmonk auth-middleware session-lookup is the canonical case: 33 recorded invocations of `SELECT data FROM sessions WHERE id=$1 AND created_at >= NOW() - INTERVAL $3 * $2`, one per test, with bind-invariant inputs (single session cookie reused across tests). The strict per-test window filter drops them as out-of-window before the per-window autopromote pass can see them - every test after the first one fails with HTTP 403 `invalid session`.

`ApplyMockFilters` owns the full `originalFiltered` slice immediately upstream of the per-test partition. Calling the heuristic there gives it the un-windowed view it needs and lets the agent re-route promoted mocks from the per-test pool into the session pool in a single in-place pass. The session pool is not subject to per-test window filtering, so promoted mocks survive and the auth-middleware lookup matches.

## OSS build compatibility

`postgres_autopromote.go` is committed but does NOT import integrations. It exposes `RegisterCrossWindowAutopromoteHook` and a nil-safe applier. A gitignored generated companion file (`postgres_autopromote_wire.go`, produced by `setup-private-parsers/action.yaml` alongside `extraparsers.go`) does the actual integrations import and registration. Public `Go on Linux` build (no setup-private-parsers) sees an inert hook and behaves identically to the pre-fix path.

## Companion

- keploy/integrations#160 adds `EvaluateCrossWindowAutopromote` (the function this PR wires up) and the underlying autopromote heuristic. Both PRs must merge together for the listmonk fix to land. This PR's CI uses `Integration-Ref: feat/postgres-v3-lifetime-autopromote` (above) to pair with that branch during review.

## Test plan
- [x] go test ./pkg/service/agent/ -run TestApplyCrossWindowAutopromote (7 unit tests added; all pass)
- [x] go build ./... succeeds (workspace mode)
- [x] GOWORK=off go build ./... succeeds (public OSS build, no integrations dep)
- [ ] After both PRs merge: re-run keploy/integrations CI listmonk-postgres pipeline; expect the auth-middleware 403 failures to clear
- [ ] Verify the env off-switch with KEPLOY_PG_V3_CROSS_WINDOW_AUTOPROMOTE=0 returns to the prior behaviour

## Risk
Heuristic itself is well-tested in the integrations side. The new code path is gated by an env flag (default ON, with off-switch). If the cross-window pass ever produces a regression in the wild, operators can flip `KEPLOY_PG_V3_CROSS_WINDOW_AUTOPROMOTE=0` to revert without rebuilding the binary.


slayerjain:
Closing in favor of recorder-side fix: classify session-shared postgres traffic correctly at record-time so the agent-layer cross-window autopromote band-aid is unnecessary. Tracking the new work in a separate branch.

PR #4156, opened 2026-04-28
fix(replay): skip TLS verification on HTTPS test cases

khareyash05:
## Summary

Replay HTTPS test cases unconditionally fail today because the replay HTTP client verifies the server cert. With this change, the replay HTTP client skips TLS verification across all three transport branches in `prepareHTTPRequest` (keep-alive / close / default).

## Why this is needed

Captured HTTPS test cases keep their original URL (e.g. `https://app.svc.cluster.local:8443/...`). At replay time the hostname/port get rewritten to a local target — typically `localhost:<port>` through a `kubectl port-forward` to a fresh replay pod, or some other dev/CI tunnel — so the cert's CN cannot match the dial target.

In the common case the replay target also serves a self-signed cert (sample apps, short-lived replay pods that regenerate the keystore on every image build), and that cert is not in any system CA bundle.

Without `InsecureSkipVerify`, Go's HTTP client fails the TLS handshake with x509 errors **before** the request is sent — so the response body never reaches the comparison engine and every HTTPS test case becomes a guaranteed false negative.

### Why skipping verify is safe here

- Mock comparison happens at the **HTTP payload layer** (status, headers, body), not at the TLS layer.
- The wire-level decryption of TLS for capture happens on the recording side via uprobes / JSSE bytecode instrumentation, not through this client.
- Verifying the cert on the replay client adds no correctness guarantee for the test outcome — it only gates whether the test gets to run at all.

## Reproduction prior to this fix

1. Deploy a Spring Boot sample app with a self-signed PKCS12 keystore and an HTTPS connector on `:8443`.
2. Record a session that hits `https://<svc>:8443/<endpoint>`.
3. Trigger replay (any path that goes through `prepareHTTPRequest` against a port-forwarded backend).
4. Every HTTPS test case fails with an `x509: certificate signed by unknown authority` (or `cert is for X, not Y`) before the request is dispatched. Plaintext HTTP test cases pass.

After this PR, the same scenario passes — verified end-to-end on Java 8 + Java 17 sample apps with HTTP, MySQL, and TLS-via-JSSE captures.

## Files

- `pkg/util.go` — add `tls.Config{InsecureSkipVerify: true}` to all three http.Transport branches in `prepareHTTPRequest`.

## Test plan

- [ ] Existing replay tests that hit plaintext HTTP backends still pass (no behavior change for non-HTTPS).
- [ ] Replay against an HTTPS backend with a self-signed cert and a hostname mismatch (e.g. CN=hello-java, dial target localhost:8443) succeeds and the test outcome reflects the response payload comparison.

### DIFF
--- a/pkg/util.go
+++ b/pkg/util.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
 	"encoding/xml"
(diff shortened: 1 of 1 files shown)


khareyash05:
Closing — landing the right fix in keploy/k8s-proxy instead. Cluster-mode replay should explicitly trust the replay pod's cert (option 1 in the discussion above), not blanket-skip verification across every OSS replay path.

PR #4157, opened 2026-04-29
feat(replay): TLSConfig hook on SimulationConfig + Hooks for HTTPS test cases

khareyash05:
## Summary

Adds an opt-in extension point so non-OSS callers (e.g. cluster-mode auto-replay in `keploy/k8s-proxy`) can inject a `*tls.Config` into the replay HTTP client without forking `pkg/util` or shipping a separate transport. **Default `nil` preserves the stdlib system-pool behaviour for every existing OSS user; no behaviour change unless a caller opts in.**

## Why

Replay HTTPS test cases captured against an in-cluster service keep their original URL (e.g. `https://app.svc.cluster.local:8443/...`). At replay time the dial target is rewritten to a local target — typically `localhost:<port>` through a kubectl port-forward to a short-lived replay pod, or some other dev/CI tunnel — so the cert the server presents cannot match the dial hostname.

The replay pod commonly serves a self-signed cert generated at image build (PKCS12 keystore baked into the image, or a cert-manager Issuer signing per pod). That cert is not in any system CA bundle, and modern Go's x509 verifier additionally demands a SAN matching the dial host.

Without a hook, every HTTPS test case fails the TLS handshake (\`x509: certificate is not valid for any names\` or \`signed by unknown authority\`) **before** the request is dispatched, so the response body never reaches the comparison engine.

### Why not just \`InsecureSkipVerify\`?

A blanket \`InsecureSkipVerify\` in OSS would weaken TLS verification for every other replay path — CI lanes replaying against a real staging backend with a properly-issued cert, for example. This PR keeps strict TLS the default and lets specific callers install a curated \`*tls.Config\` — for instance one that pins the replay pod's leaf cert via \`VerifyConnection\` (TOFU), without softening the system-trust path for anyone else.

## What changes

- \`pkg/util.go\` — \`SimulationConfig\` gains \`TLSConfig *tls.Config\`; all 3 \`http.Transport\` branches in \`prepareHTTPRequest\` pass it through as \`TLSClientConfig\`. \`nil\` leaves the transport at stdlib defaults (behaviour unchanged).
- \`pkg/service/replay/hooks.go\` — \`Hooks.tlsConfig\` field + \`(*Hooks).SetReplayTLSConfig\` setter; \`SimulateRequest\` forwards it into \`pkg.SimulationConfig.TLSConfig\` for HTTP testcases.

## Consumer

The cluster-mode auto-replay launcher in \`keploy/k8s-proxy\` (separate change, not in this PR) installs a TOFU-pinning \`*tls.Config\` via the new \`SetReplayTLSConfig\` setter:

\`\`\`go
&tls.Config{
    InsecureSkipVerify: true, // replaced by VerifyConnection
    VerifyConnection: func(cs tls.ConnectionState) error {
        // pin the leaf cert observed on the first handshake;
        // reject any subsequent handshake whose leaf doesn't match
    },
}
\`\`\`

The replay HTTP client then accepts the replay pod's self-signed cert (whatever it is — TOFU) but rejects any mid-session cert substitution. Strict-TLS callers are unaffected.

## Test plan

- [ ] Existing replay tests against plaintext HTTP backends still pass (no behaviour change for non-HTTPS).
- [ ] Existing replay tests against HTTPS backends with system-trust certs still pass (\`TLSConfig == nil\` path unchanged).
- [ ] Cluster-mode replay against a self-signed HTTPS backend passes when the consumer installs a pinning \`*tls.Config\` via \`SetReplayTLSConfig\`.

Verified end-to-end against an 8-app matrix (Java 17 / Java 8 × MySQL+TLS / Kafka / HBase / Pulsar) running through the k8s-proxy cluster-mode dispatcher with TOFU pinning enabled — zero \`x509\` errors when TLS test cases were captured.


khareyash05:
Pushed an additional fix on this branch (c5ed2c72) for a recorder-side race that was making Java 8 + MySQL TLS replays hang indefinitely.

## Why this was needed

`asyncMySQLDecode` in `pkg/agent/proxy/integrations/mysql/recorder/query.go` reads client and server bytes from `clientBuffChan` and `destBuffChan` and feeds both into a single FIFO `decodeChan` via a non-deterministic `select` in `handleClientQueries`. On a fast loopback (typical of `kind` clusters and our DS-mode capture pipeline) the next client command can be enqueued ahead of the trailing server bytes of the prior result set — additional column-def packets, the rows, or the `EOF` / `OK_with_EOF` terminator.

When that happens, the existing force-flush at line 346 produces a partially-captured result set whose declared `ColumnCount` exceeds `len(Columns)` and/or whose `FinalResponse` is `nil`:

```
column_count_pkt | column_def_1..M | <end>           // M < ColumnCount
or
column_count_pkt | column_defs | row_pkt | <end>     // missing terminator
```

The replay encoder emits these bytes verbatim. Drivers that strictly require the trailing terminator block in `socketRead0` forever:

| | Behavior |
|---|---|
| Connector/J 8.x on Java 17 | Tolerates a missing terminator on otherwise-complete result sets, so the bug only surfaced sporadically. |
| Connector/J 8.x on Java 8 | Strictly requires it — every `loadServerVariables` call hung 10 s+ until the OSS replayer's per-test deadline tripped. |

## The fix

Single file (`pkg/agent/proxy/integrations/mysql/recorder/query.go`, +227 lines). `flushMock` calls a new `closeIncompleteResultSetForFlush` helper that, when the decoder state is `stateExpectColumns` / `stateExpectEOFAfterColumns` / `stateExpectRows`:

1. Truncates `rs.ColumnCount` to `len(rs.Columns)` so the encoded column-count packet matches the column-def packets actually emitted.
2. Synthesizes the trailing terminator on `FinalResponse` using the negotiated capabilities — `0xFE`-prefixed `OK_with_EOF` if `CLIENT_DEPRECATE_EOF` is set (with a trailing lenenc info byte if `CLIENT_SESSION_TRACK` is also set), legacy 5-byte EOF otherwise.
3. Picks `seqID = max(observed seqs in columns / EOFAfterColumns / rows) + 1`.
4. Rewrites `bundle.Header.Type` to `TextResultSet` / `BinaryProtocolResultSet` so `wire.EncodeToBinary` dispatches through the result-set encoder.

The encoder (`wire/phase/query/ResultSetPacket.go`) and replayer (`replayer/query.go`, `replayer/conn.go`) are byte-identical to before — the fix produces structurally-complete mocks at record time so replay just works.

## Verification

`kind keploy-ds → keploy-replay`, Spring Boot + Connector/J 8.4.0 against `mysql:8.0`. Five scenarios:

| Scenario | total_tests | failed_count | noisy_count | duration | hangs |
|---|---|---|---|---|---|
| Java 8 + TLS run #1 | 9 | **0** | 4 | 11.42 s | 0 |
| Java 8 + TLS run #2 | 4 | **0** | 2 | 11.40 s | 0 |
| Java 17 + TLS regression | 8 | **0** | 3 | 11.24 s | 0 |
| Java 8 + non-TLS (`sslMode=DISABLED`) | 3 (app) | **0** (app) | 1 (app) | 13.19 s | 0 |
| Java 17 + non-TLS | 9 | **0** | 4 | ~11 s | 0 |

`failed_count` is `autoReplayMetrics.failed_count` — the system's authoritative verdict. `noisy_count` is auto-detected data drift (auto-increment IDs, list ordering, `created_at` timestamps), same shape as the pre-fix Java 17 baseline. Zero `context deadline exceeded`, zero `CommunicationsException` across all runs.

officialasishkumar (approved):

gouravkrosx (approved):

PR #4162, opened 2026-04-30
fix(testdb): correct YAML noise assertion decode type assertion (#3872)

### LINKED ISSUE #3872: [bug]: Noise assertions silently dropped during YAML decode due to incorrect type assertion
## Bug Description

When decoding test cases from YAML format, noise assertions are silently dropped due to an incorrect type assertion in the `Decode()` function.

## Root Cause

In `pkg/platform/yaml/testdb/util.go`:

**Encoding (line 75):**
```go
if len(noise) > 0 {
    a[models.NoiseAssertion] = noise  // noise is map[string][]string
}
```

**Decoding (line 345):**
```go
if key == models.NoiseAssertion {
    noiseMap, ok := raw.(map[models.AssertionType]interface{})  // Wrong type!
    if !ok {
        logger.Warn("noise assertion not in expected map[AssertionType]interface{}", zap.Any("raw", raw))
        continue
    }
    // ... rest of processing
}
```

The type assertion expects `map[models.AssertionType]interface{}`, but the actual type is `map[string][]string` (since `noise` variable is declared as `map[string][]string`).

## Impact

- Noise assertions are lost when test cases are loaded from YAML files
- Users' noise configurations are ignored during test replay
- No error is thrown, making this a silent data loss bug
- Affects both HTTP and gRPC test cases (same bug exists in both branches)

## Evidence

The bug manifests in two places:
1. HTTP branch: Lines 344-363
2. gRPC branch: Lines 384-403

Both use identical incorrect type assertion logic.

## Reproduction Steps

1. Create a test case with noise assertions
2. Encode it to YAML using `EncodeTestcase()`
3. Decode it back using `Decode()`
4. Check `tc.Noise` - it will be empty despite noise being present in YAML

## Suggested Fix

Change the type assertion from:
```go
noiseMap, ok := raw.(map[models.AssertionType]interface{})
```

To:
```go
noiseMap, ok := raw.(map[string]interface{})
```

Then iterate through `noiseMap` properly:
```go
for field, inner := range noiseMap {
    tc.Noise[field] = []string{}
    arr, ok := inner.([]interface{})
    if !ok {
        continue
    }
    for _, item := range arr {
        if s, ok2 := item.(string); ok2 && s != "" {
            tc.Noise[field] = append(tc.Noise[field], s)
        }
    }
}
```

## Related Issues

This may be related to:
- #3492 (Global noise configuration not applied)
- #2456 (Robust denoise functionality)
- #1098 (Detecting accurate noisy parameters)

## Environment

- Go version: (check with `go version`)
- Keploy version: Latest main branch (2026-02-27)
- File: `pkg/platform/yaml/testdb/util.go`

SAY-5:
Fixes #3872. `pkg/platform/yaml/testdb/util.go` decoded the noise assertion sub-map by asserting:

```go
noiseMap, ok := raw.(map[models.AssertionType]interface{})
```

YAML never decodes a sub-map into that custom-typed key, yaml.v3 produces `map[string]interface{}`, yaml.v2 produces `map[interface{}]interface{}`. The assertion always failed, the `Debug` log fired, and `tc.Noise` stayed empty. Users' noise configurations were silently dropped during test replay.

## Change

Extracted a `populateNoise(raw, noise)` helper that:

- Accepts `map[string]interface{}` (yaml.v3), `map[interface{}]interface{}` (yaml.v2), and `map[string][]string` (programmatic round-trips).
- Iterates entries, initialising `noise[field] = []string{}` (preserves the existing 'empty key still present' behaviour) and accepting both `[]interface{}` and `[]string` for the inner values.

Both the HTTP and gRPC branches in `DecodeTestcase` now delegate to this helper, eliminating the duplicate buggy block.

## Test plan

- [x] `go build ./pkg/platform/yaml/testdb/...`, clean
- [x] `go vet ./pkg/platform/yaml/testdb/...`, clean
- [x] `go test ./pkg/platform/yaml/testdb/...`, all existing tests pass
- [x] New `noise_decode_test.go` covers all three input shapes (yaml.v3 map, yaml.v2 map, concrete `map[string][]string`)
- [x] Pre-fix the existing pattern `raw.(map[models.AssertionType]interface{})` returns `ok == false` for every yaml-decoded shape (asserted by tracing the test inputs)

## Customer-data hygiene

No real recordings or customer data introduced. Test inputs use synthetic header names (`X-Request-ID`, `Cookie`) and field paths (`body.timestamp`).

### DIFF
--- a/pkg/platform/yaml/testdb/noise_decode_test.go
+++ b/pkg/platform/yaml/testdb/noise_decode_test.go
@@ -1,0 +1,54 @@
+package testdb
+
+import (
+	"testing"
+)
+
+// TestPopulateNoise_AcceptsYamlV3MapShape regression-tests #3872: the previous
+// implementation type-asserted the noise sub-map as
+// map[models.AssertionType]interface{} which YAML never produces, so noise
+// assertions were silently dropped on load.
+func TestPopulateNoise_AcceptsYamlV3MapShape(t *testing.T) {
+	noise := map[string][]string{}
+	raw := map[string]interface{}{
+		"header.X-Request-ID": []interface{}{"foo", "bar"},
+		"body.timestamp":      []interface{}{},
+	}
+	populateNoise(raw, noise)
+
+	if got, want := len(noise["header.X-Request-ID"]), 2; got != want {
+		t.Errorf("header.X-Request-ID len = %d, want %d", got, want)
+	}
+	if noise["header.X-Request-ID"][0] != "foo" {
+		t.Errorf("header.X-Request-ID[0] = %q, want %q", noise["header.X-Request-ID"][0], "foo")
+	}
+	if _, ok := noise["body.timestamp"]; !ok {
+		t.Error("body.timestamp key should be present (initialised even when empty)")
+	}
+}
+
+// TestPopulateNoise_AcceptsYamlV2MapShape covers the older library output
+// where keys arrive as interface{}.
+func TestPopulateNoise_AcceptsYamlV2MapShape(t *testing.T) {
+	noise := map[string][]string{}
+	raw := map[interface{}]interface{}{
+		"header.X-Request-ID": []interface{}{"foo"},
+	}
+	populateNoise(raw, noise)
+	if got, want := noise["header.X-Request-ID"][0], "foo"; got != want {
+		t.Errorf("got %q, want %q", got, want)
+	}
+}
+
+// TestPopulateNoise_AcceptsConcreteSliceShape covers programmatic
+// constructions where the noise map round-trips without YAML in between.
+func TestPopulateNoise_AcceptsConcreteSliceShape(t *testing.T) {
+	noise := map[string][]string{}
+	raw := map[string][]string{
+		"header.Cookie": {"session", "tracking"},
+	}
+	populateNoise(raw, noise)
+	if got, want := len(noise["header.Cookie"]), 2; got != want {
+		t.Errorf("got %d entries, want %d", got, want)
+	}
+}
--- a/pkg/platform/yaml/testdb/util.go
+++ b/pkg/platform/yaml/testdb/util.go
@@ -322,25 +322,12 @@ func Decode(yamlTestcase *yaml.NetworkTrafficDoc, logger *zap.Logger) (*models.T
 		for key, raw := range httpSpec.Assertions {
 			tc.Assertions[key] = raw
 			if key == models.NoiseAssertion {
-				noiseMap, ok := raw.(map[models.AssertionType]interface{})
-				if !ok {
-					logger.Debug("noise assertion not in expected map[AssertionType]interface{}", zap.Any("raw", raw))
-					continue
-				}
-				for kt, inner := range noiseMap {
-					field := string(kt)
-					// initialize slice
-					tc.Noise[field] = []string{}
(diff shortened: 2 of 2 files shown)


slayerjain:
Hi @SAY-5 — thank you for the care you put into this one. The write-up was clear, the helper is cleanly factored, and you traced the code path properly. Before merging I ran the round-trip end to end, and it turns out the premise in #3872 doesn't hold — so I want to walk through what we found, because the reasoning is genuinely subtle and the mistake is the issue's, not yours.

**The type assertion actually succeeds today.** `yaml.v3` propagates the enclosing map's type down into nested `interface{}` mappings. In `decode.go` (v3.0.1, ~L806):

```go
if outt.Elem() == ifaceType {
    if outt.Key().Kind() == reflect.String {
        d.stringMapType = outt
    }
```

`httpSpec.Assertions` is declared as `map[models.AssertionType]interface{}` — string-kind key, `interface{}` element — so `d.stringMapType` becomes that type, and the nested `noise:` mapping is materialised as `map[models.AssertionType]interface{}` rather than the `map[string]interface{}` you'd expect. A type probe on the decoded value confirms it:

```
concrete type of raw noise assertion = map[models.AssertionType]interface {}
asserts as map[models.AssertionType]interface{} => true
asserts as map[string]interface{}               => false
```

So the existing assertion at `pkg/platform/yaml/testdb/util.go:429` and `:465` matches on every real decode, and noise config is not being dropped on `main`.

**Which means the switch in `populateNoise` inverts the behaviour.** It covers `map[string]interface{}`, `map[interface{}]interface{}` and `map[string][]string`, but not `map[models.AssertionType]interface{}` — and since `AssertionType` is a distinct named type, no case matches and the function silently no-ops. There's a test on `main` that catches it, `TestMatch_DecodedFromYAML_SectionedNoise` in `pkg/matcher/http/decode_test.go`, which decodes a real YAML fixture end to end. It passes on `main` and fails on this branch:

```
--- FAIL: TestMatch_DecodedFromYAML_SectionedNoise
    decode_test.go:76: decoded noise[body] = [], want the sectioned path list
```

One thing worth flagging for your own benefit: that test landed only a couple of days ago, and the three tests here call `populateNoise` directly with hand-constructed maps rather than going through YAML — so they couldn't have surfaced this. That's a trap the codebase set for you, not an oversight on your part. Worth reaching for a round-trip test whenever the thing under test is a decode path.

I'm going to close this along with #3872, which we've confirmed is a false positive — its stated reproduction steps pass on `main`. Sorry you spent effort against a bad issue; that's on us for leaving it open.

If you'd like to keep going, there is a real improvement buried in here. `Decode` and `DecodeJSON` each carry their own copy of this normalisation logic, and `expandAssertionsJSON` handles a different set of shapes again. Unifying them onto one helper that covers all four shapes (with `map[models.AssertionType]interface{}` first), keeps a `logger.Debug` for anything unmatched instead of silently ignoring it, and is gated by an end-to-end round-trip test would be a genuinely valuable change. Happy to review that if you want to take it on.

Thanks again for the contribution.

slayerjain:
Closing per the analysis above — noise decode is working correctly on `main`, and this change would regress it. No reflection on the work itself; the issue it was written against was wrong.

Thanks again @SAY-5, and the unification refactor mentioned above is still very much open if you'd like to pick it up.

PR #4163, opened 2026-05-01
fix(replay): fold comma-separated request headers into a single wire field

ayush3160:

- `ToHTTPHeader` previously called `strings.Split(value, \",\")` on every recorded header, fanning a single comma-separated field into multiple slice elements. Go's `http.Client` then emits one wire header per element, so an axios-style `Accept: application/json, text/plain, */*` mock replays as **three duplicate `Accept` fields** on the wire.
- Strict server middlewares that build a response via `Response(headers=dict(**request.headers), ...)` raise `TypeError: dict() got multiple values for keyword argument 'accept'` on duplicate field names — a known FastAPI / Starlette failure mode that bites any service recorded against such middleware.
- The YAML serialisation already collapses repeated same-name headers to a single comma-folded string (`map[string]string` cannot represent duplicate keys), so the recorder cannot losslessly distinguish \"one header with internal commas\" from \"N repeated headers\". Keeping that string intact at replay (one wire header per YAML key) mirrors what most HTTP clients (axios, requests, fetch) actually emit and is semantically equivalent at the receiver for RFC 7230 §3.2.2 list-valued headers (Accept, Accept-Encoding, Baggage, Forwarded, Via, ...). `Set-Cookie` is the well-known multi-row exception and is a response header, never built by this helper.
- Updated `TestToHTTPHeader_WithTimeValue_909` to assert the single-element slice (the previous expectation pinned the buggy split behaviour) and added `TestToHTTPHeader_AcceptListValue_FoldedSingleHeader` to lock in the wire shape this fix depends on.

## Links & References

**Closes:** NA

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- Reproducible duplicate-Accept TypeError observed when replaying tests recorded against FastAPI services that use a `dict(**request.headers)` metrics middleware.
### 📄 Related Documents
- RFC 7230 §3.2.2 — comma-folded list-valued header field equivalence.

- [x] 🐞 Bug Fix

## Added e2e test pipeline?

The fix is exercised by two unit tests at `pkg/util_test.go`. An end-to-end test would require recording a request whose target server runs `dict(**request.headers)` middleware and asserting replay no longer 500s; happy to add one if reviewers prefer.

- [x] 👍 yes

The new behaviour and trade-off (rare repeated same-name fields fold into one comma-separated field) are documented inline in `ToHTTPHeader`.

## Added to documentation?

- [x] 👍 yes, mentioned below

```bash
go test ./pkg/ -run TestToHTTPHeader -v
```

Both tests pass. To reproduce the original bug pre-patch: spin up a FastAPI app whose middleware calls `Response(headers=dict(**request.headers), ...)`, send a request with multiple `Accept` fields, and observe the `TypeError: dict() got multiple values for keyword argument 'accept'`.

## Self Review done?
- [x] ✅ yes

- NA — see commit body for the wire-level analysis.


ayush3160:
Closes #4164

gouravkrosx (approved):

PR #4187, opened 2026-05-08
feat(log): add debug-level file sink helper to utils/log

ayush3160:

Adds `log.AddDebugFileSink(logger, file, capBytes)` to `utils/log` — returns a new `*zap.Logger` that, in addition to whatever sinks the input logger already had, writes every debug-level-or-above entry to `file`. Returns a `*log.DebugFileSink` handle exposing `Flush()`, `Capped()`, and `Written()` so the caller can flush before close and read the cap-tripped state for downstream metadata.

Mechanics:
- The new sink is composed alongside the existing core via `zapcore.NewTee` so each branch keeps its own level filter — the existing console core honors `LogCfg.Level`; the new file core is locked at `DebugLevel`.
- The debug branch is wrapped in its own `redactingCore` so field-level redaction runs before bytes hit the file. The writer is wrapped in `redactingWriter` for the post-encode pass.
- The file path is buffered (256 KiB `zapcore.BufferedWriteSyncer`) to amortize syscalls, and wrapped in a new `cappedWriteSyncer` that silently drops bytes past a configurable cap (default 100 MiB). Hitting the cap is non-fatal; `DebugFileSink.Capped()` reports it for downstream metadata.
- Exposes `redactingCore.Inner()` so callers can compose new cores around the same redaction-aware structure (used internally by `AddDebugFileSink`, available externally for parity).

Bench (in this PR):
```
BenchmarkBaseline_Write-8             3,973,989    654.8 ns/op
BenchmarkAddDebugFileSink_Write-8     3,153,037    765.0 ns/op   (+17%)
```
For a 50k-record run that is ~5 ms of added logger time. Bundle compression and upload (in the enterprise PR) dominate the user-visible latency.

This is the foundation PR for the `keploy cloud replay` debug-bundle feature. Standalone change — does not affect any existing call site.

## Links & References

**Closes:** N/A (foundation; closed by the umbrella issue once all three PRs merge)

### 🔗 Related PRs
- keploy/api-server: `feat/cloud-debug-bundle` (api-server endpoint, independent)
- keploy/enterprise: `feat/cloud-debug-bundle` (consumes this helper)
### 🐞 Related Issues
- keploy/enterprise#2003 (umbrella feature request)
### 📄 Related Documents
- NA

- [x] 🍕 Feature

## Added e2e test pipeline?

This is a library-level change. End-to-end verification happens in the enterprise PR (manual smoke against staging api-server). Unit tests + benchmark in this PR cover the new code paths.

- [x] 👍 yes

Doc comments cover the cap semantics, buffering rationale, and the redaction-once invariant the tee preserves.

## Added to documentation?

Internal helper; not a user-facing CLI change.

- [x] 👍 yes, mentioned below

```bash
go test ./utils/log/...
go test -bench=. -run=^$ -benchtime=2s ./utils/log/...
```

Tests cover:
- Entries emitted before attach do not appear in the file.
- Entries emitted after attach land in the file regardless of console level.
- 256 KiB buffer requires `Flush()` to materialize entries on disk.
- Soft cap stops writes at the configured byte count without errors.
- Single outer `redactingCore` ensures `RedactEntry` fires once per entry, not 2× per tee branch.

## Self Review done?
- [x] ✅ yes

- NA — library change, no UI surface.

### DIFF
--- a/main.go
+++ b/main.go
@@ -263,6 +263,10 @@ func maybeAttachDebugFileSink(logger *zap.Logger) (*os.File, *log.DebugFileSink)
 		return nil, nil
 	}
 	*logger = *wrapped
+	// Publish the sink so cross-package helpers (e.g. the agent's
+	// per-test-set rotation in pkg/agent/routes) can reach it without
+	// threading the sink through every constructor.
+	log.SetDebugFileSink(sink)
 	logger.Debug("KEPLOY_DEBUG_FILE attached", zap.String("path", path))
 	return f, sink
 }
--- a/main.go
+++ b/main.go
@@ -55,6 +55,28 @@ func start(ctx context.Context) {
 	}
 	utils.LogFile = logFile
 
+	// If KEPLOY_DEBUG_FILE is set, tee debug-level log records into that
+	// file in addition to whatever sinks log.New configured. Used by
+	// `keploy cloud replay` to capture the agent container's debug
+	// stream into the user's keploy folder via a bind mount; the
+	// support-bundle pipeline picks the file up automatically since it
+	// lives under cfg.Path. Also useful as a generic "give me the full
+	// debug log without --debug printing it to my terminal" knob for
+	// any keploy invocation.
+	debugFile, debugSink := maybeAttachDebugFileSink(logger)
+	if debugFile != nil {
+		defer func() {
+			if debugSink != nil {
+				if ferr := debugSink.Flush(); ferr != nil {
+					logger.Warn("KEPLOY_DEBUG_FILE: flush failed", zap.Error(ferr))
+				}
+			}
+			if cerr := debugFile.Close(); cerr != nil {
+				logger.Warn("KEPLOY_DEBUG_FILE: close failed", zap.Error(cerr))
+			}
+		}()
+	}
+
 	// Start pprof HTTP server for live profiling if PPROF_PORT is set
 	if pprofPort := os.Getenv("PPROF_PORT"); pprofPort != "" {
 		go func() {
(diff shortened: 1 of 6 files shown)


ayush3160:
Cross-links (couldn't update the body — `gh pr edit` is currently blocked by the projects-classic GraphQL deprecation):

- 🐞 keploy/enterprise#2003 — umbrella feature request
- 🔗 keploy/api-server#1648 — server-side `/cloud/debug-bundles/upload` endpoint (independent)
- 🔗 keploy/enterprise#2004 — CLI wire-up (consumes this helper)

This PR is the foundation; the enterprise PR's `go.mod` is bumped to this PR's HEAD commit (`a557ec0a013b`).

ayush3160:
**Update — agent log capture.** Added a second commit (c92344f0) that:

1. Honors a new `KEPLOY_DEBUG_FILE` env var in `main.go` — when set, the entrypoint attaches `log.AddDebugFileSink` onto the live logger via the in-place `*logger = *wrapped` pattern. Reuses the buffered + capped pipeline from the first commit. Generic enough to be useful outside `cloud replay` too (any keploy invocation can opt into a debug log file).

2. Bind-mounts `conf.Path` into the keploy-agent container at `/keploy-host` and sets `KEPLOY_DEBUG_FILE=/keploy-host/agent-debug.log` in `GenerateKeployAgentService`. Skipped when `conf.Path` is empty or relative.

Net effect for `keploy cloud replay`: the agent writes its debug log onto the host inside `<cwd>/keploy/agent-debug.log`. The cloud-CLI bundle pipeline walks `cfg.Path` already, so the agent log rides along automatically — no docker-logs tailing, no subprocess management, no race with `compose down`.

Tests cover the happy path (volume + env emitted), empty-path skip, and relative-path skip.

ayush3160:
**Update — per-test-set rotation of the agent debug log.** Added commit `c690356c`:

- `cappedWriteSyncer` gains a mutex so writes serialize against `Swap`.
- `DebugFileSink.RotateForScope(scope)` flushes the buffered writer to the previous file and atomically swaps in `<dir>/<scope>/<basename>`. Repeated calls with the same scope are no-ops; passing `""` rotates back to the unscoped origin.
- New package globals `SetDebugFileSink` / `GetDebugFileSink` so the agent's `BeforeSimulate` route handler can reach the sink without DI.
- `pkg/agent/routes.HandleBeforeSimulate` calls `log.RotateDebugFileForTestSet(req.TestSetID)` on every simulate; the helper short-circuits when `TestSetID == currentScope`. Failures are logged at warn — the existing single-file capture keeps receiving bytes.

Net bundle layout once this lands and the agent image is rebuilt:
\`\`\`
<cwd>/keploy/agent-debug.log                   (boot, hooks)
<cwd>/keploy/<test-set-1>/agent-debug.log
<cwd>/keploy/<test-set-2>/agent-debug.log
\`\`\`

Four new tests cover the rotation happy path, rotation back to origin, the nil-sink safety path, and the global accessor.

ayush3160:
**Fix — empty per-test-set files.** Empirically caught: the per-test-set `agent-debug.log` files in the bundle were 0 bytes even though the rotation primitive in `BeforeSimulate` was firing correctly.

Root cause: `AddMode("agent")` in the agent CLI's `Validate` path rebuilds the logger's core from scratch and returns a brand-new `*zap.Logger`. The codebase consumes it via in-place pointee mutation (`*c.logger = *new`), which silently discards any tee branches attached to the previous core — including the debug-file sink that `AddDebugFileSink` installed at boot. The unscoped `agent-debug.log` captured ~50 lines (boot up through "config file not found") and stopped abruptly at the moment `AddMode` fired; everything after that flowed into the void. Same hazard exists for `ChangeLogLevel`, `RedirectToStderr`, and `ChangeColorEncoding`.

Commit `797a7fa9` adds a private `reattachDebugFileSink` helper that looks up the active sink (via `SetDebugFileSink`) and rewraps the rebuilt logger with a tee branch reusing the existing buffered + capped chain. All four core-rebuilding helpers call it before returning, so future callers get the file sink preserved automatically.

Three regression tests added:
- `SurvivesAddMode` — the exact scenario the cloud-replay agent hits.
- `RotateAfterAddMode` — confirms `RotateForScope` still works once the logger is rebuilt.
- `SurvivesChangeLogLevel` — covers the `--debug` path.

@user — to verify, rebuild the agent image once more (against keploy `797a7fa9` + enterprise HEAD) and re-run.

charankamarapu (approved):
LGTM

gouravkrosx (approved):

PR #4192, opened 2026-05-11
feat(matcher): noise-aware exact-body match for form-encoded bodies

ayush3160:
## Summary

`ExactBodyMatch`'s noise-aware second pass only ran on JSON — `application/x-www-form-urlencoded` bodies dropped out unread, even when `Mock.Noise` carried relevant patterns. Practical consequence: any rotating field in an AWS SigV4 form request (boto3 IRSA `WebIdentityToken=…`, OAuth2 `client_assertion=…`, Vault kubernetes-auth `token=…`) couldn't be relaxed via noise. Exact-match failed because the byte rotation diverged from the recording; the fuzzy fallback then picked whichever unrelated mock had the closest absolute Levenshtein distance — typically an SNS `Publish`/`PublishBatch` payload of similar length — and handed the wrong XML back to the SDK, which crashed at parse time with errors like `KeyError: 'AssumeRoleWithWebIdentityResult'`.

Companion change: [keploy/enterprise#2007](https://github.com/keploy/enterprise/pull/2007) extends the obfuscator to emit a key-anchored `^<key>=[^&]+$` noise regex alongside the existing value-anchored one whenever a form-encoded field carries a secret-shaped value. This PR is the matcher half that consumes those rules.

Root cause and forensic walkthrough: [keploy/k8s-proxy#394](https://github.com/keploy/k8s-proxy/issues/394).

## What changes

- **`pkg/agent/proxy/integrations/http/match.go`** — new branch inside the noise-aware second pass of `ExactBodyMatch`. When both the request and mock bodies look form-encoded, tokenises both by splitting on `&` (no URL decoding — so segments match the obfuscator's `^<rawKey>=[^&]+$` regex verbatim), tests each segment against the mock's noise patterns, wildcards any matching segment, and requires the surviving key sets on both sides to be equal with each surviving key's value(s) byte-equal.
- Two helpers: `looksLikeFormEncoded` (mirror of the same heuristic in `enterprise/pkg/secret`) and `formBodiesMatchModuloNoise`.
- The JSON / generic / fuzzy paths are untouched and remain green.

## Why this is the right shape

- **No hardcoded credential field names.** The matcher just consults whatever the obfuscator emitted; the obfuscator decides which keys to noise based on its existing `Detector.ContainsSecretPattern` value-shape classifier (already covers JWTs, base64 blobs, AWS-style keys, generic API tokens). Adding a new auth scheme means adding nothing — both halves work on shape, not on a list.
- **Symmetric checks.** Mock-side extra keys must appear in the request (or be noise); request-side extra keys must appear in the mock (or be noise). Otherwise a permissive mock (e.g. SNS PublishBatch with a wildcarded `Message`) could absorb arbitrary unrelated requests by accident.
- **Multi-valued keys** (`a=1&a=2`) handled positionally.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./pkg/agent/proxy/integrations/http/... -count=1` passes — including three new tests:
  - `TestExactBodyMatch_FormEncodedKeyNoise` — the customer IRSA happy path: live token differs from recorded, match still succeeds via `^WebIdentityToken=[^&]+$`
  - `TestExactBodyMatch_FormEncodedMismatchOnNonNoisyField` — a divergent non-noisy field (RoleArn) still fails the match
  - `TestExactBodyMatch_FormEncodedExtraKeyOnRequest` — a request-side extra non-noisy key (the SNS PublishBatch absorption scenario) fails the match
- [x] End-to-end against the customer's testset: previously matched the wrong mock at 12.99% fuzzy_levenshtein → now matches the correct STS mock at 100% exact_body_form_noise_aware on the first try; fuzzy never runs.

Refs keploy/k8s-proxy#394

### DIFF
--- a/pkg/agent/proxy/integrations/http/match.go
+++ b/pkg/agent/proxy/integrations/http/match.go
@@ -420,13 +420,13 @@ func (h *HTTP) SchemaMatch(ctx context.Context, input *req, unfilteredMocks []*m
 // First pass: fast string equality.
 // Second pass: noise-aware comparison that skips obfuscated fields identified
 // by Mock.Noise patterns. Three body shapes are handled in the second pass:
-//   1. A mock body that is itself entirely a noise value — auto-match.
-//   2. application/x-www-form-urlencoded — per-segment noise check (see
-//      formBodiesMatchModuloNoise). Lets a noise regex of the shape
-//      ^<key>=[^&]+$ wildcard a rotating field (IRSA WebIdentityToken=…,
-//      OAuth client_assertion=…, etc.) without depending on the SDK to
-//      produce byte-identical request bodies at replay.
-//   3. JSON — field-by-field comparison via JSONBodyMatchScore.
+//  1. A mock body that is itself entirely a noise value — auto-match.
+//  2. application/x-www-form-urlencoded — per-segment noise check (see
+//     formBodiesMatchModuloNoise). Lets a noise regex of the shape
+//     ^<key>=[^&]+$ wildcard a rotating field (IRSA WebIdentityToken=…,
+//     OAuth client_assertion=…, etc.) without depending on the SDK to
+//     produce byte-identical request bodies at replay.
+//  3. JSON — field-by-field comparison via JSONBodyMatchScore.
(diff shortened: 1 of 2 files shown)


charankamarapu (approved):
LGTM

gouravkrosx (approved):

PR #4193, opened 2026-05-11
fix(log): rotate debug file per test set, not per test case

ayush3160:
## Summary

* Move the agent-side debug-file rotation out of `BeforeSimulate` (per test case) into `BeforeTestSetCompose` (per test set, fires once after agent ready in DockerCompose mode).
* Add CLI-side rotation in `Hooks.BeforeTestSetRun` / `Hooks.AfterTestSetRun` so cloud-replay's debug stream also gets a per-set folder — **including test sets that resolve to NO_TESTS_TO_RUN**, which the per-test-case hook never reached.
* Add `TestSetID` to `models.BeforeTestSetCompose` and plumb it through `Instrumentation` / `Hooks` / `AgentClient` so the agent route handler can see the test-set identity.

## Why

The original placement in #4187 hooked `RotateDebugFileForTestSet` into `HandleBeforeSimulate`. That callback fires N times per test set (once per test case) and **never fires for test sets with zero test cases**, so:

1. `NO_TESTS_TO_RUN` test sets never produced a per-set folder.
2. Test sets with cases got rotated late — all pre-simulate work (`StoreMocks`, `MockOutgoing`, mock filtering) landed in the unscoped pre-rotation log, not the per-set file. Empirically, this meant ~500 KB of agent work pre-rotation in a real cloud-replay bundle.

Rotating at `BeforeTestSetCompose` fires exactly once per test set right after agent ready and before any test-case work, capturing the full agent timeline in the per-set file. The CLI-side hooks (`BeforeTestSetRun` / `AfterTestSetRun`) fire on the cloud-replay CLI process for every test set, so even empty test sets get their own CLI log folder.

## Breaking change

`replay.TestHooks.BeforeTestSetCompose` gains a `testSetID` argument. Enterprise build update lands in a coordinated follow-up PR after this one is tagged.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./pkg/service/replay/... ./utils/log/...` pass
- [ ] Re-run `keploy cloud replay` against a multi-test-set app (mix of populated and empty sets) and verify:
  - [ ] Each populated test set has a non-empty `<test-set>/agent-debug.log`
  - [ ] Each test set (populated AND empty) has a non-empty `<test-set>/cloud-debug.log`
  - [ ] Top-level `<keploy>/agent-debug.log` contains only agent boot bytes (not StoreMocks / MockOutgoing)
  - [ ] Top-level `<keploy>/cloud-debug.log` contains pre-loop CLI work + post-loop summary

## Related

- Follows up on #4187 (`feat(log): add debug-level file sink helper to utils/log`)
- Companion enterprise PR: opened after this is tagged

### DIFF
--- a/pkg/agent/routes/record.go
+++ b/pkg/agent/routes/record.go
@@ -127,7 +127,22 @@ func (a *Agent) HandleBeforeTestSetCompose(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
-	if err := agent.ActiveHooks.BeforeTestSetCompose(r.Context(), req.TestRunID); err != nil {
+	// Rotate the debug-file sink (if attached via KEPLOY_DEBUG_FILE) to
+	// a per-test-set scope. This is the per-test-set boundary — fires
+	// exactly once for the test set when the agent comes up in
+	// DockerCompose mode, before any test case runs. Rotating here
+	// (rather than per test case via BeforeSimulate) means each test
+	// set gets exactly one rotation and the per-set file captures the
+	// agent's pre-simulate work (mock load, store-mocks, etc.), not
+	// just the simulate path.
+	if req.TestSetID != "" {
+		if err := log.RotateDebugFileForTestSet(req.TestSetID); err != nil {
+			a.logger.Warn("debug file rotation for test set failed; continuing without rotation",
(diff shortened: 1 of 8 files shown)


charankamarapu (approved):
LGTM

gouravkrosx (approved):

PR #203, opened 2022-10-06
Disable test export by default

murugesanp:
Signed-off-by: murugesanp <murugesan148@gmail.com>

## Related Issue
  -https://github.com/keploy/keploy/issues/167

Closes: [#167](https://github.com/keploy/keploy/issues/167)

#### Describe the changes you've made
A clear and concise description of what you have done to successfully close your assigned issue. Any new files? or anything you feel to let us know!

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Code style update (formatting, local variables)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Please let us know if any test cases are added

Please describe the tests(if any). Provide instructions how its affecting the coverage.

#### Describe if there is any unusual behaviour of your code(Write `NA` if there isn't)
A clear and concise description of it.

## Checklist:

- [x ] My code follows the style guidelines of this project.
- [ x] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [x ] My changes generate no new warnings.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes.

## Screenshots (if any)

|        Original         |          Updated           |
|:-----------------------:|:--------------------------:|
| **original screenshot** | <b>updated screenshot </b> |

PR #4203, opened 2026-05-18
fix(proxy): reuse MockManager across test-sets so long-lived parsers see new mocks

ayush3160:
## Summary

`Proxy.Mock()` is invoked per test-set during replay (`Agent.MockOutgoing` → `Proxy.Mock`). It currently allocates a fresh `MockManager` on every call:

```go
p.setMockManager(NewMockManager(NewTreeDb(customComparator), NewTreeDb(customComparator), p.logger))
```

That strands long-lived parser goroutines on the previous instance. The canonical case is the postgres-v3 replayer's `runLoop` — its `PerTestProvider` / `SessionProvider` / `StartupProvider` capture the `*MockManager` handed to `MockOutgoing` and keep polling it for the connection's lifetime. With a per-test-set replacement:

- the **new** MockManager receives the new test-set's mocks via `StoreMocks` / `UpdateMockParams` and bumps its revision (which is exactly what the revision-gated providers from [keploy/integrations#203](https://github.com/keploy/integrations/pull/203) rely on)
- the **parsers** keep reading off the orphaned manager whose revision never advances
- result: every test from test-set N (N≥2) misses with `transactional: no invocation matched` — the parser sees test-set N-1's last cohort, never the new one

The fix: only allocate a `MockManager` when one doesn't exist yet. Once allocated it persists across test-sets; `SetMocksWithWindow` already swaps the filtered / unfiltered / startup trees atomically and bumps revisions, so per-test-set mock content stays correct while parsers see a stable handle.

## Reproduction

The autoreplay path doesn't normally trigger this (each test-set spawns a fresh pod → fresh asyncpg connections → fresh `ReplayV3` call → fresh `PerTestProvider` bound to the current MockManager). The bug surfaces whenever a user-app connection survives a test-set boundary — e.g. the docker-compose "lift-app" debug path keeps one asyncpg pool warm across test-sets, the v3 `runLoop` survives, and `PerTestProvider.Current()` keeps polling the orphaned MockManager.

A full e2e repro (FastAPI + SQLAlchemy + asyncpg matching the provider-engagement-service stack, 2-test-set traffic driver, docker-compose with the lift-app patch) made the failure deterministic. The agent debug log shows two distinct MockManager pointers across the test-set boundary:

| Event | `mockManagerPtr` | revision sequence |
|---|---|---|
| test-set-0's 5 tests | `0xc002184960` | 1, 3, 5, 7, 9, 11, 13 |
| test-set-1's 5 tests | `0xc002417a40` (new) | 1, 3, 5, 7, 9, 11 |

The v3 parser's `mockDbPtr` is pinned to `0xc002184960` throughout. With this PR the same `*MockManager` is observed across both test-sets and the revision counter advances continuously.

## Pairs with

- keploy/integrations#203 (`feat/postgres-v3-live-tier-providers`) — the integrations side of the same workstream. PR #203's revision-gated rebuilds are necessary for cross-test-set mock visibility but cannot fire if the parser is reading off an orphaned MockManager; this PR keeps the MockManager stable so PR #203's gating works as designed. Integrations branch updated with required compile fixes in keploy/integrations@d24142c50fe3dc7705dc0623185892d63bb72e27.

## Commit

- `677b6ac8` — `fix(proxy): reuse MockManager across test-sets so long-lived parsers see new mocks`

## Test plan

- [x] `go test ./pkg/agent/proxy/...` passes locally
- [x] Enterprise build (`CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build`) succeeds with `replace go.keploy.io/server/v3 => ../keploy`
- [ ] e2e: rebuild enterprise image with this fix + PR #203, replay the 2-test-set leak repro under docker-compose with lift-app — every test in test-set-1 should now resolve against test-set-1's own mocks (previously: all 5 failed with `candidates: 0`)
- [ ] e2e: confirm the existing single-test-set autoreplay flow is unchanged (one MockManager observed for the run, revisions advance per `UpdateMockParams`, no regression)
- [ ] Soak: long-running replay with many test-sets — no MockManager leak (only one is held)


charankamarapu (approved):
LGTM

gouravkrosx (approved):

PR #4204, opened 2026-05-19
feat(test): --keep-app-alive — start user app once across all test-sets

ayush3160:
## Summary

Adds a `--keep-app-alive` flag on the `test` subcommand that starts the user app ONCE for the whole replay run instead of restarting it per test-set. When the flag is on, `--delay` is honoured only on the first test-set (subsequent ones inherit the warm app). Currently wired for the docker-compose command path — that's the path production globality autoreplay uses.

Default OFF preserves the historical per-test-set restart behaviour.

## Why

Cross-test-set bugs that depend on a long-lived TCP connection — asyncpg pools, JDBC HikariCP pools, anything that hangs onto a socket across the test-set boundary — cannot reproduce in `keploy test` today. Every test-set spins up a fresh app container, which gives every test-set its own fresh connections, and the parser-side session/startup cohort-staleness bugs that [keploy/integrations#203](https://github.com/keploy/integrations/pull/203) fixes never surface in the e2e gate.

Production globality autoreplay runs a single app process across the whole replay run. This flag is what lets the docker-compose replay path match that shape for local repro and CI gating against PR #203.

## Reusing existing infra

`Replayer.RunTestSet` already had a `serveTest bool` parameter that gated the three lifecycle points we need gated:

| Line | What `!serveTest` gates |
|---|---|
| `replay.go:762` | per-test-set `NotifyGracefulShutdown` |
| `replay.go:919` | per-test-set `RunApplication` spawn (docker-compose) |
| `replay.go:1181` | per-test-set `RunApplication` spawn (non-compose) |

But the parameter was never reachable from any user-facing knob — the only call site hardcoded `false`, so the feature was committed (see `git log -S 'serveTest bool'`: *"feat: skip app launch is testSets"*, *"feat: skip app run multiple times"*) but unreachable. This PR wires it up rather than duplicating the gates.

## Changes

- `config/config.go`: `Test.KeepAppAlive bool` (yaml/json/mapstructure: `keepAppAlive`)
- `cli/provider/cmd.go`: `--keep-app-alive` flag on the `test` subcommand + alias entry `"keepAppAlive": "keep-app-alive"`
- `pkg/service/replay/replay.go`:
  - Hoists `cmdType := utils.CmdType(r.config.CommandType)` above the test-set loop so the one-shot spawn can read it (the previous in-loop declaration is removed).
  - **One-shot `RunApplication`** on the outer errgroup in `Start()`, gated by `r.config.Test.KeepAppAlive && r.instrument && cmdType == utils.DockerCompose`.
  - Call site at `Start()` now passes `r.config.Test.KeepAppAlive` as the `serveTest` argument instead of `false` — reuses the existing gates verbatim.
  - New `Replayer.isFirstTestSet` field (mirror of `isLastTestSet`), set inside the test-set loop before `RunTestSet` is invoked.
  - Both `waitForAppReady` call sites (compose and non-compose branches) skip the wait when `KeepAppAlive && !isFirstTestSet`.

## Scope notes

- Restricted to `cmdType == utils.DockerCompose`. The native / single-binary paths follow a different `RunApplication` lifecycle that hasn't been audited for this mode. The flag is a no-op on those paths today (the one-shot spawn simply doesn't fire) so it's harmless to leave it as a future-expansion seam.
- When `KeepAppAlive` is **off**, the historical per-test-set restart behaviour is preserved bit-for-bit. The new code paths are gated behind the flag; `serveTest` at the call site evaluates to `false` and every existing call to `RunApplication` / `NotifyGracefulShutdown` / `waitForAppReady` fires exactly as before.

## Verification

- `go build ./...`, `go vet ./...`, `golangci-lint v2.12.2 run` — all clean
- `go test ./pkg/service/replay/...` and `go test ./pkg/agent/proxy/...` pass
- Enterprise build (`CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build`) succeeds against this branch via `replace go.keploy.io/server/v3 => ../keploy`
- `keploy test --help` shows the new flag

## Commit

`431bb22f` — `feat(test): --keep-app-alive — start user app once across all test-sets`

## Pairs with

- [keploy/integrations#203](https://github.com/keploy/integrations/pull/203) — adds the postgres-v3 SessionProvider / StartupProvider that revision-gate the session/startup cohort indexes. PR #203 fixes the bug on the parser side; this PR makes the bug **observable in e2e CI** by keeping the asyncpg pool's TCP connection warm across the test-set boundary. The new `.woodpecker/pg-v3-session-tier-staleness.yml` pipeline in PR #203 currently shows both legs green because the released keploy binary doesn't have this flag yet — once this PR ships, the expected-red leg (record-build-replay-latest) on PR #203 will start failing as designed.

## Test plan

- [x] Unit tests in `pkg/service/replay/` still pass
- [x] Build/vet/lint clean
- [ ] Manual e2e: run the docker-compose lift-app repro (or PR #203's `.woodpecker/pg-v3-session-tier-staleness.yml` fixture) locally with this binary — expect Bug 1 to surface on test-set-2's `SHOW work_mem` when PR #203 is NOT applied to integrations; expect 15/15 pass when PR #203 IS applied.
- [ ] Regression: existing replay flows with `--keep-app-alive` UNSET should behave identically to before this PR (per-test-set app restart preserved).


charankamarapu (approved):
LGTM

gouravkrosx (approved):

PR #4220, opened 2026-05-26
feat(ci): memory-load e2e pipelines (MySQL/Mongo/gRPC) + fixes for orphan-TC mock loss

pathakharshit:
### TL;DR
This PR adds **end-to-end "memory-load" stress pipelines for the MySQL, Mongo, and gRPC parsers**, and fixes the recording bugs those pipelines exposed — bugs where, under heavy load + tight memory, the recorder lost some mocks and left "orphan" test cases that then failed at replay with `socket was unexpectedly closed: EOF`.

### Background (so the rest makes sense)
Keploy records an app twice-over: it captures each incoming HTTP request as a **test case (TC)** and the app's outgoing DB/queue calls as **mocks**, pairing them by time. At replay it serves the recorded mocks back to the app. If a TC ends up on disk but its mock does **not** (an *orphan TC*), replay has nothing to answer the app's DB call, the connection is closed, and the test fails with `EOF`. Nothing in CI exercised the recorder under **memory pressure + high throughput**, so these losses went uncaught.

### 1. New feature — memory-load e2e pipelines (the main addition)
Added three CI lanes in `golang_docker.yml` (each with its own `golang-docker.sh` driver):
- **`go-memory-load-mysql`**, **`go-memory-load-mongo`**, **`go-memory-load-grpc`**

Each lane records a sample app (from `samples-go@feat/all-memory-load-apps`) while a **k6 load test** hammers it and Keploy runs under a **deliberately tight memory limit**, then replays. This stress-tests the parser *and* the recorder under exactly the pressure that used to silently drop mocks. Lanes run in the record/replay matrix (record-build/replay-build, record-build/replay-latest, record-latest/replay-build).

Because these lanes run under intentional memory pressure on shared 2-vCPU runners — where a few non-deterministic replay divergences are normal and **not** real regressions — `check_test_report` now uses a **configurable failure-tolerance gate** (`MAX_FAIL_PERCENT`, default **2%**): the lane fails only if **more than 2%** of test cases fail, and it **logs exactly what it tolerated** (e.g. `Aggregate: 6/877 failed (0.68%). Tolerance: 2%`) so nothing is hidden.

> Note: the plain `go-memory-load` (postgres) lane is intentionally left **record-only** (replay commented out, matching `main`).

### 2. Bug fixes these pipelines surfaced

**a) Mocks for already-captured TCs were dropped under memory pressure → orphan TCs** (`syncMock.go`)
When memory got tight, the agent wiped buffered mocks to free RAM — including mocks whose TC had **already** been captured during a calm period. Fix: decide by the **request timestamp (when the call happened)**, not the decode timestamp. Only mocks whose request happened *during* a pressure window are dropped (those never had a TC, so dropping them orphans nothing); calm-captured mocks and once-per-boot **startup mocks** are always kept.

**b) The teardown tail of mocks was lost at shutdown → orphan TCs** (`agent.go`, `record.go`)
On Ctrl+C the agent flushes its last mocks onto the agent→host stream, but the host-side reader used to exit the instant the context cancelled — severing the connection and discarding everything still in flight. Fix:
- **Buffer** the host receive channel so the reader pulls the whole tail off the wire into memory at network speed instead of being throttled to disk speed.
- A **bounded 60s drain** that keeps reading past cancellation until the stream ends.
- A **drain-gate handshake** plus `docker stop --time 30` so the agent container isn't hard-killed mid-flush.

**c) The disk writer couldn't keep up, so the tail piled up** (`mockdb/db.go`, `record.go`)
The old path wrote each mock to YAML synchronously (open+encode+flush per mock, ~160/sec). Reworked into a **single background batching writer**: encode + append to a buffer, and flush to disk **once per batch** (when the queue momentarily empties) instead of once per mock — far faster, no data loss, file writes stay serialized (no corruption).

**d) MySQL dropped DB bytes mid-connection under pressure** (`mysql/recorder/query.go`)
The per-chunk "is recording paused?" gate could drop bytes for an already-captured TC in the middle of a connection, corrupting the recording. Fix: bytes always flow through to the app; mock atomicity is enforced at the mock layer (the pressure logic above) instead of by dropping wire bytes.

**e) Clearer replay diagnostics** (`mysql/replayer/query.go`)
When a TC matches **zero** mocks at replay, the error now hints that its mock was dropped at record time (teardown lag / memory pressure) instead of a generic miss.

## Links & References

**Closes:** #

### 🔗 Related PRs
- NA

### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [x] 🍕 Feature  
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] ✅ Test  
- [x] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [x] 👍 yes

- [x] 👍 yes

## Added to documentation?

- [x] 👍 yes, mentioned below

**How to test locally**
1. Check out `samples-go@feat/all-memory-load-apps`.
2. From a sample dir (e.g. `samples-go/go-memory-load-mysql`), run the lane driver:
   `source .github/workflows/test_workflow_scripts/golang/go_memory_load_mysql/golang-docker.sh`
   (it builds the app, records under k6 load + a tight `--memory-limit`, SIGINTs, then replays).
3. Or watch the CI lanes `go-memory-load-{mysql,mongo,grpc}` on this PR.
**Expected:** lanes pass within the 2% tolerance; teardown test cases (`get-orders-*`, `get-analytics-*`, large-payload CRUD) no longer fail with `socket was unexpectedly closed: EOF`.

## Self Review done?
- [x] ✅ yes

- NA


pkg/agent/proxy/supervisor/supervisor.go
```diff
@@ -242,14 +242,6 @@ func (s *Supervisor) Run(ctx context.Context, fn ParserFunc, sess *Session) Resu
 		return s.classifyReturn(ctx, r.panicked, r.panicVal, r.stack, r.err)
 
 	case <-s.hungCh:
-		// Debug-level: hang abort is a designed control-flow path —
-		// the dispatcher's FallthroughToPassthrough handling picks it
-		// up and the relay keeps forwarding bytes. Operators who want
-		// to tune behaviour have explicit knobs (see next_step).
-		s.cfg.Logger.Debug("parser hang detected; aborting",
```
kapishupadhyay22, line 249: please add this log again

pkg/agent/proxy/proxy.go
```diff
@@ -94,20 +58,13 @@ func isMysqlPort(port uint32, configured []uint32) bool {
...
-		zap.String("phase", phase),
-		zap.Int64("connID", connID),
-		zap.String("addr", addr),
-		zap.Int64("dur_ns", durNs),
-		zap.Int64("ts_ns", time.Now().UnixNano()),
-	}
-	logger.Info("[PROBE/dial]", append(base, fields...)...)
+// destCacheEntry holds all recently-seen destinations from a single source IP.
+// Keyed by destination port so the fallback can be used safely when the app
+// only connects to ONE backend on a given port (unambiguous) but not when
```
kapishupadhyay22: this should be removed, because of the following reasons : 
1. if redirect proxy map is failing we should improve the logic there
2. we should not consider caching a single dest to a corresponding single source (not a ideal behaviour for a real world API)

pkg/agent/proxy/proxy.go
```diff
@@ -44,41 +43,6 @@ type ParserPriority struct {
 	ParserType integrations.IntegrationType
 }
 
-// probeFanoutEnabled toggles [PROBE/proxy] / [PROBE/dial] structured
-// logs via KEPLOY_PROBE_FANOUT=1. The flag is read once at first use
-// and cached in an atomic so hot-path branches are a single load.
-// Used to diagnose the parallel-TLS-fanout stall (SAP + Postgres
-// concurrent handshakes) without polluting normal record runs.
-var (
```
kapishupadhyay22, line 52: don't remove this if not required

pkg/agent/routes/record.go
```diff
@@ -339,15 +340,59 @@ func (a *Agent) HandleIncoming(w http.ResponseWriter, r *http.Request) {
...
+					zap.Int("tcs_sent_to_cli", tcsSentSoFar),
+					zap.Int("tcs_suppressed_pressure_overlap", tcsSuppressedSoFar),
+					zap.Int("pressure_ranges_total", syncmgr.Get().PressureRangeCount()),
+					zap.Int64("mocks_dropped_by_pressure", finalDropped),
+					zap.Int64("mocks_added_successfully", finalAdded),
+				)
 				return
 			}
+
+			// Bug 0 fix: suppress this TC if memory pressure was active at any
```
kapishupadhyay22: improve this comment, and make it more understandable

pkg/agent/routes/record.go
```diff
@@ -339,15 +340,59 @@ func (a *Agent) HandleIncoming(w http.ResponseWriter, r *http.Request) {
...
+			// We use pressure-range overlap (not per-mock-drop timestamps) because
+			// the Mongo-mock goroutine and the HTTP-TC goroutine race: a paired
+			// mock can be dropped AFTER the TC has already reached this handler,
+			// in which case a per-mock ledger would still be empty. Pressure
+			// ranges, however, are recorded the instant SetMemoryPressure(true)
+			// flips memoryPause — happens-before any mock drop it causes — so an
+			// overlap check is race-free.
+			tcRespTime := t.HTTPResp.Timestamp
+			if tcRespTime.IsZero() {
+				// Defensive: HTTPResp.Timestamp is normally set by the proxy
```
kapishupadhyay22: Test case response timestamp should not have a zero value, we should not take 30 seconds as the response time

pkg/agent/routes/record.go
```diff
@@ -339,15 +340,59 @@ func (a *Agent) HandleIncoming(w http.ResponseWriter, r *http.Request) {
...
+				// hook before the TC enters this channel. If it's missing for
+				// some reason, widen the window with a 30s ceiling so we still
+				// catch most concurrent pressure events without going unbounded.
+				tcRespTime = t.HTTPReq.Timestamp.Add(30 * time.Second)
+			}
+			hasOverlap, overlapCount := syncmgr.Get().WasPressureActiveInWindow(t.HTTPReq.Timestamp, tcRespTime)
+
+			if hasOverlap {
+				tcsSuppressedSoFar++
+				a.logger.Info("agent: TC suppressed — memory pressure overlapped TC window, not sent to CLI",
```
kapishupadhyay22: this should be a debug log

pkg/agent/routes/record.go
```diff
@@ -653,7 +698,6 @@ func (a *Agent) MakeAgentReady(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	a.logger.Debug("Agent marked as ready", zap.String("file", agentReadyFilePath))
```
kapishupadhyay22, line 657: this is an important log, please keep it

pkg/platform/http/agent.go
```diff
@@ -362,6 +450,37 @@ func (a *AgentClient) GetMappings(ctx context.Context, opts models.IncomingOptio
 
 		decoder := json.NewDecoder(res.Body)
 
+		// SAME FIX AS GetOutgoing — drain the mappings stream past
```
kapishupadhyay22: improve this comment

pkg/platform/yaml/mockdb/db.go
```diff
@@ -69,41 +70,26 @@ type MockYaml struct {
...
-	gobOverflows   atomic.Uint64
-	// gobStopClosed is true after Close() has invoked close(gobStop).
-	// A subsequent Close() that arrives after the first one timed out
-	// waiting for gobDone must not close the channel a second time
-	// (that would panic). Cleared when the writer finally exits and
-	// we transition back to gobRunning=false.
-	gobStopClosed bool
-	// gobFlushErr holds the terminal flush/close error from
-	// gobFlushAndClose so that Close() can surface it to its caller
+	// Synchronous mock writer. The recording consumer calls InsertMock
```
kapishupadhyay22: need to benchmark the performance - speed vs memory with the previous approach, can move to this if clear improvements are there

kapishupadhyay22 (commented):
please address the review comments

pkg/agent/proxy/syncMock/syncMock.go
```diff
@@ -10,6 +11,92 @@ import (
...
+
+// tcCorrHint returns a short string suggesting which TC(s) the dropped
+// mock is likely associated with. Since we don't track TC→mock mapping at
+// AddMock time, we use req_ts of the mock relative to the current clock to
+// give the log reader a timestamp anchor for cross-referencing the TC yaml.
+func tcCorrHint(mock *models.Mock) string {
+	if mock == nil {
+		return ""
+	}
+	if !mock.Spec.ReqTimestampMock.IsZero() {
```
kapishupadhyay22: this should not be zero, if it is found to be zero then we have to fix the corresponding bug

pkg/agent/proxy/syncMock/syncMock.go
```diff
@@ -10,6 +11,92 @@ import (
...
 )
 
+// mockSummary returns a short human-readable description of the mock's
+// operation, used in pressure-drop and teardown logs so you can identify
+// which SQL query / mongo operation was lost without reading a full YAML.
+//
+//	MySQL  → "MySQL/mocks COM_QUERY: SELECT * FROM orders LIMIT 10"
+//	Mongo  → "Mongo/mocks kind=Mongo ts=18:43:16.357"
+//	Other  → "Http/mock-42"
+func mockSummary(mock *models.Mock) string {
```
kapishupadhyay22: why are we creating a special function for mysql and mongo specifically ?

pkg/agent/proxy/syncMock/syncMock.go
```diff
@@ -253,9 +425,35 @@ func (m *SyncMockManager) AddMock(mock *models.Mock) {
...
+		// request time, not now: drop only if the request ITSELF happened
+		// during pressure (the ingress never captured it, so there is no TC).
+		// Unknown timestamp → keep (safer than orphaning a captured TC).
+		reqTime := mock.Spec.ReqTimestampMock
+		if !reqTime.IsZero() && m.pressureActiveAtLocked(reqTime) {
+			m.mu.Unlock()
+			totalDropped := m.pressureDropped.Add(1)
+			totalAdded := m.totalAdded.Load()
+			if logger := m.dropLogger(); logger != nil {
+				logger.Info("agent: mock dropped by memory pressure",
```
kapishupadhyay22: this can flood the logs, please make it a debug log

pkg/agent/proxy/syncMock/syncMock.go
```diff
@@ -480,37 +707,191 @@ func (m *SyncMockManager) GetFirstReqSeen() bool {
...
+		}
+	}
+	m.mu.Unlock() // NEVER hold mu while logging — logging inside a lock causes a deadlock under I/O pressure (see BUG 5: 70-minute CI hang)
+
+	// Only log on state TRANSITIONS to avoid flooding logs on every 500ms memoryguard tick.
+	// Pressure can fire hundreds of times per second — we only want to see the moment it starts and clears.
+	if logger := m.dropLogger(); logger != nil {
+		if enabled && !wasEnabled {
+			// Pressure just turned ON (false→true transition)
+			logger.Info("agent: memory pressure activated — pressure-request mocks dropped, calm-captured kept",
```
kapishupadhyay22: improve the logs in this file

pkg/agent/proxy/syncMock/syncMock.go
```diff
@@ -89,13 +176,48 @@ type SyncMockManager struct {
...
+	// entry has end == zero while pressure is still active. Guarded by mu.
+	//
+	// This is the join key for the Bug 0 TC-suppression fix:
+	// WasPressureActiveInWindow checks whether any range overlaps a TC's
+	// [HTTPReq.Timestamp, HTTPResp.Timestamp] window. Using pressure ranges
+	// instead of per-mock drop timestamps is RACE-FREE — memoryguard records
+	// the start the instant it flips memoryPause, BEFORE any mock-parser
+	// goroutine sees the result; so even if the paired AddMock fires AFTER
+	// the TC has reached routes/record.go, the overlap is already visible.
+	pressureRanges []pressureRange
```
kapishupadhyay22, line 148: is there any horizon after which we are not tracking the pressure interval ranges.
this will be used as continuous recording, which means that if we don't cleanup ranges, it will stored thousands or millions of state changes => leading to memory bottleneck and a forever increasing loop complexity.

kapishupadhyay22 (approved):
Please ensure that pipelines pass before merging.

kapishupadhyay22:
I have validated the behaviour of pipelines.

gouravkrosx (approved):

PR #4222, opened 2026-05-27
fix: pay --keep-app-alive startup delay when leading test-set is empty

ayush3160:
Under `--keep-app-alive`, the startup readiness wait (`--delay`, or the `--health-url` poll) is supposed to be paid **once** — before the first request — so the user application has time to boot. The skip was gated on the test-set loop index:

```go
r.isFirstTestSet = (i == 0)
...
if serveTest && !r.isFirstTestSet { /* skip waitForAppReady */ }
```

But a test-set with **no test cases** (or one that is fully ignored) returns early from `RunTestSet` *before* `waitForAppReady` ever runs. When such a set sorts first, the next test-set — the first one that actually fires requests — sees `isFirstTestSet == false` and **skips the wait**, hammering an app that hasn't finished starting. Every request then fails with `connection reset by peer` / `EOF` and `ACTUAL STATUS 0`.

**Fix:** replace the index-based flag with an `appReadyWaitDone` latch that is set **only after `waitForAppReady` actually completes**. Empty/ignored leading test-sets return before the wait and never set the latch, so the first test-set that reaches the wait still pays the startup delay; subsequent sets skip it as intended (warm app). The latch is reset at the start of each run. Non `--keep-app-alive` runs (`serveTest == false`) are unchanged and still wait per test-set.

Touched: `pkg/service/replay/replay.go` (struct field + reset + both `waitForAppReady` gate sites — docker-compose and native/docker-run/docker-start).

## Links & References

**Closes:** NA
- NA (no tracking issue)

### 🔗 Related PRs
- Follow-up to #4204 (`feat(test): --keep-app-alive`)
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

- [x] 👍 yes

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

- [x] 👍 yes, mentioned below

Repro with a multi-test-set cloud replay where the **first (sorted) test-set is empty**:

1. `keploy cloud replay --app <app> --keep-app-alive --delay 40` against an app slow to boot (e.g. Express + Prisma + Postgres), with one empty leading test-set and others that have cases.
2. **Before:** first request fires ~2s after the container starts (the 40s delay is skipped); all tests fail with `connection reset by peer` / `EOF` (status 0).
3. **After:** ~40s elapse between `Starting Application` and the first `starting test for …`; tests run against a booted app.

## Self Review done?
- [x] ✅ yes

Symptom (status 0 on every test because the app wasn't up yet):

```
INFO  Using stored delay from auto-replay config  {"delay": 40}
...
INFO  Starting Application
INFO  starting test for {"test case": "...", "test set": "...-whst9"}   # ~2s later, not 40s
ERROR failed to send testcase request to app {"error": "... read: connection reset by peer"}
EXPECT STATUS 201  |  ACTUAL STATUS 0
```

## Additional checklist:

### DIFF
--- a/pkg/service/replay/replay.go
+++ b/pkg/service/replay/replay.go
@@ -158,13 +158,21 @@ type Replayer struct {
 	instrument      bool
 	isLastTestSet   bool
 	isLastTestCase  bool
-	// isFirstTestSet mirrors isLastTestSet — true while the test-set
-	// about to run is the first one of the current replay run. Read
-	// by the waitForAppReady gate so --delay applies only to the first
-	// test-set when --keep-app-alive is set (subsequent test-sets
-	// inherit a warm app and would otherwise pay --delay seconds of
-	// dead time per boundary).
-	isFirstTestSet     bool
+	// appReadyWaitDone latches true the first time waitForAppReady
+	// actually completes a readiness wait during the current replay
+	// run. Read by the waitForAppReady gate so that, under
+	// --keep-app-alive, --delay (or a HealthURL poll) is paid exactly
+	// once — on the first test-set that reaches the wait — and skipped
+	// thereafter (subsequent test-sets inherit a warm app).
+	//
+	// Keyed on "did the wait actually run" rather than the test-set
+	// loop index: a leading empty/ignored test-set returns before the
+	// wait (see the len(testCases)==0 / ignored early-returns in
+	// RunTestSet), so an index-based "first test-set" flag would let
+	// that empty set consume the first-set status and the next set —
+	// the first one that actually fires requests — would skip the wait
+	// against an app that has not finished booting.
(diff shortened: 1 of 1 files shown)


gouravkrosx (approved):

gouravkrosx:
@ayush3160 Please check the failing pipelines before merging.

PR #4226, opened 2026-05-29
fix(mysql/replay): synthesize PREPARE_OK for JDBC cachePrepStmts orphaned mocks

khareyash05:
## Summary

Fixes auto-replay failure on TiDB + JDBC with `useServerPrepStmts=true&cachePrepStmts=true` (observed in production for a Flipkart customer running `global-shipment-master`). The connection drops with `no matching mock found. Re-record mocks if the SQL query has changed.` on `COM_STMT_PREPARE`, with `closest_mock: ""`.

### Root cause

JDBC's client-side prepared-statement cache reuses server-allocated `statement_id`s across requests without re-issuing `COM_STMT_PREPARE`. When a statement ID was allocated **before** the recording window opened (e.g. during pool warmup), the recorder captures the subsequent `COM_STMT_EXECUTE` but never sees a PREPARE for that ID. The trace ends up with EXECUTE mocks whose PREPARE counterparts simply don't exist.

Concrete evidence from one customer trace (`global-shipment-master` testset):

| Statement ID | PREPARE mock? | EXECUTE mocks |
|---|---|---|
| 1–15 | ❌ none | 18 across stmtIDs 1–15 |
| 16 | ✅ 1 (`update shipments...`) | 1 |

15 of 16 statement IDs have no PREPARE mock — they were allocated by TiDB before keploy attached.

On replay, the JDBC client's cache is cold so it *does* issue `COM_STMT_PREPARE`, but for those 15 SQL strings there's no PREPARE mock to match against, so `matchCommand` returns `closest_mock: ""` and `query.go` drops the connection.

### Fix

Two complementary fallbacks in `pkg/agent/proxy/integrations/mysql/replayer/match.go`:

1. **`matchCommand` — synthesize `COM_STMT_PREPARE_OK` for unmocked PREPAREs.** Fresh `statement_id` from `decodeCtx.NextStmtID`. One placeholder `ColumnDefinition41` per `?` in the request SQL (type=252/`MYSQL_TYPE_BLOB`, the wire convention for unknown placeholder types; drivers send their own type bytes in EXECUTE). Synthetic stmtID + query stashed in `decodeCtx.PreparedStatements` / `StmtIDToQuery` so the subsequent EXECUTE can resolve `actualQuery`.

2. **`matchStmtExecutePacketQueryAware` — accept EXECUTE on params alone when the mock has no recorded PREPARE.** When `expectedQuery==""` (i.e. `buildRecordedPrepIndex` returned no entry for the mock's stmtID), require only `allParamsMatched`. Per-test mocks are consumed on match, so ordered EXECUTEs with identical param signatures still pick up distinct mocks in recorded order.

Follows the same policy as 9e16ae46 (`fix(mysql/replay): synthesize OK for unmocked COM_STMT_RESET`) and the existing synthetic OK for unmocked control/DDL in `matchCommand` — tolerate protocol traffic the record window didn't capture but doesn't actually require a recorded response for correctness.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./pkg/agent/proxy/integrations/mysql/...` passes (replayer, recorder, wire, preparedstmt, rowscols)
- [ ] Re-run customer's auto-replay against `global-shipment-master` testset; expect 2 obsolete tests to now produce real pass/fail results instead of dropping on PREPARE
- [ ] Verify no regression for traces with full PREPARE+EXECUTE pairs (existing strict matching path is unchanged when `eq != ""`)
- [ ] Verify EXECUTE param-signature collisions across different SQLs are disambiguated by per-test mock consumption order (recorded EXECUTEs are consumed in capture order)


Aditya-eddy (approved):
lgtm

gouravkrosx (approved):
Do we need to add any regression tests for it??

khareyash05:
> Do we need to add any regression tests for it??

we will require a new app, which i have added in sampless-java, this will not cause any regression isues

khareyash05:
also bypassing due to flaky pipelines, and for customer issue

PR #4231, opened 2026-06-01
feat(proxy): SCRAM-SHA-256-PLUS support across TLS-MITM via LD_PRELOAD shim

ayush3160:
## Summary

Resolves the long-standing `FATAL: SCRAM channel binding check failed` that any Postgres client using SCRAM-SHA-256-PLUS hits when its connection is recorded through keploy's TLS-MITM proxy. The fix lands the cert-hash swap **inside the app process** via an LD_PRELOAD shim, paired with a deferred-publish (mitm, real) hash map written by the proxy to a shared volume.

Today this issue is documented as ["unfixable at the proxy layer without the upstream server's private key"](https://github.com/keploy/integrations/blob/main/pkg/postgres/v3/replayer/session/startup.go) and every CI sample works around it with `channel_binding=disable`. This PR makes it work transparently, without app-side configuration changes.

## Why it works

SCRAM channel binding RFC 5929 `tls-server-end-point` mixes the **hash of the TLS server cert** into the proof. Under keploy's MITM the app sees keploy's cert, while postgres sees its own — hashes diverge, FATAL fires by design. Replacing the hash on the wire doesn't work (the SCRAM proof depends on it), and keploy doesn't have the password to recompute the proof.

The trick: **swap the hash inside libpq before the proof is computed.** The shim hooks OpenSSL's `X509_digest`, calls the real implementation, then overwrites the buffer with the real upstream's hash (which keploy knows — it terminates upstream TLS itself). libpq then computes a real proof using the real password against an auth-message containing the real hash. The server verifies it normally.

```
   app                          keploy                       real postgres
    │                              │                                │
    │──TLS(keploy_cert)───────────▶│                                │
    │                              │──TLS(real_cert)───────────────▶│
    │ CertForClient → RegisterMITM │  UpgradeDestTLS → RegisterReal │
    │             both halves ──▶ cbmap.Publish writes cbmap.txt    │
    │                              │                                │
    │ libpq calls X509_digest      │                                │
    │  └ shim swaps to real_hash   │                                │
    │ SCRAM client-final ─────────▶│───────────────────────────────▶│
    │                              │              proof verifies ✓  │
    │◀──────────────── AuthenticationOk ─────────────────────────────│
```

## What's in this PR

**New package** — `pkg/agent/proxy/cbmap/`
- Deferred-publish hash map. `RegisterMITM` and `RegisterReal` are symmetric — whichever half arrives second triggers atomic write of `cbmap.txt`. Hash algorithm follows RFC 5929 (SHA-256 default, SHA-384/512 for matching cert sigs, MD5/SHA-1 normalized to SHA-256). `CleanupConnection` reaps orphan halves on connection teardown.
- 9 unit tests covering both orderings, empty inputs, dedup, multi-upstream, and 50-pair concurrent registration.

**LD_PRELOAD shim** — `pkg/agent/proxy/tls/asset/cbshim.{c,so}`
- ~150 LoC of C. Hooks `X509_digest` via `dlsym(RTLD_NEXT, ...)`. Reads `cbmap.txt` (path overridable via `CBSHIM_HASHMAP` env). No-op when the computed hash isn't in the table. Stateless / thread-safe.
- Compiled artifact embedded via `//go:embed asset/cbshim.so` and written to `/tmp/keploy-tls/` by `setupSharedVolume`.

**Three TLS-handshake hook points register their respective halves**
- `tls/ca.go::CertForClient` → `RegisterMITM` (every client TLS handshake — cache-hit and mint paths both)
- `util/tls_upgrader.go::UpgradeDestTLS` → `RegisterReal` (V1 / legacy parser flow, e.g. dialPostgresSSLUpstream)
- `relay/directive_proc.go::handleUpgradeTLS` → `RegisterReal` (V2 / supervisor parser flow — Postgres v3, MySQL, Mongo)
- `util.Conn` and relay's `read_time` wrappers gain `NetConn()`/`Unwrap()` so the cbmap walker can reach the underlying `*tls.Conn` through them.
- Connection ID = source TCP port (same key `SrcPortToDstURL` already uses).

**End-to-end test** — `tests/e2e/scram-channel-binding/`
- Docker-compose: postgres (`hostssl scram-sha-256`) + MITM proxy linked against real `cbmap.Publish` + libpq client with the shim compiled inside the app image.
- `./run.sh` asserts 3 phases:
  - **Phase 1**: no shim, `channel_binding=require` → must FATAL (proves CB is actually in effect; without this the rest is meaningless)
  - **Phase 2**: shim, `channel_binding=prefer` → must succeed (libpq picks PLUS, shim swaps, auth verifies)
  - **Phase 3**: shim, `channel_binding=require` → must succeed (no possibility of silent downgrade — proves a real PLUS proof was sent)
- Exits 0 on all phases pass.

## Verified end-to-end

Against an actual `keploy record` invocation with the v3 Postgres parser linked:

```
🐰 Keploy(agent): cbmap-relay: calling RegisterReal     src_port=55694  subject=CN=postgres
🐰 Keploy(agent): cbmap: RegisterReal                   conn_id=55694   ready=false
🐰 Keploy(agent): cbmap: RegisterMITM                   conn_id=55694   ready=true
[cbshim] loaded 1 entries from /tmp/cbmap-test.txt
[cbshim] swap: digest of len=32 matched a known MITM cert — substituting real hash
 one | who | ts            ← psql queries succeed
```

Mocks produced are structured `kind: PostgresV3` (not generic byte passthrough):

```yaml
kind: PostgresV3
postgresV3:
  type: query
  query:
    sqlNormalized: SELECT $1 AS one
    bindValues:
      - !!binary MQ==
```

## Design properties

- **Order-independent.** Deferred publish handles upstream-first (`KEPLOY_INTERCEPT_PG`/dialPostgresSSLUpstream) AND client-first (TLSUpgrader-driven parser flows) without per-path special-casing.
- **Cryptographically sound.** TLS encryption unchanged. Real SCRAM proof from real password against the real auth-message. The shim only adjusts a single 32-byte input.
- **Safe to inject globally.** No-op when there's no matching cbmap entry. Non-OpenSSL apps (Java JSSE, Go-native, .NET, rustls) never invoke `X509_digest`. Static-linked apps silently ignore LD_PRELOAD. setuid binaries skip LD_PRELOAD by kernel design.
- **Multi-connection / multi-upstream safe.** Lookup keyed by `H(mitm_cert)` which is per-upstream stable. No per-connection state in the shim. 50-pair concurrent unit test passes.

## Companion PR

Webhook side (LD_PRELOAD env-var injection): https://github.com/keploy/k8s-proxy/pull/465 — replace with k8s-proxy PR # once both are open. This keploy PR is independently mergeable; the webhook PR completes the kubernetes rollout (the shim file gets staged on the shared volume by `setupSharedVolume` here, so without webhook injection the file is present but no app loads it).

## Limitations / follow-ups

- `cbshim.so` is currently `linux/amd64 + glibc + OpenSSL 3.x` only. arm64 / musl variants need separate embedded blobs + runtime selection. Tracked as follow-up.
- No CLI flag to disable globally yet — `chmod 000 /tmp/keploy-tls/cbshim.so` or un-mounting suffices to opt out.
- Real Postgres cert rotation (extremely rare in practice but possible) leaves stale (mitm, real) pairs — would need invalidation on next handshake. Acceptable for v1.

## Test plan

- [x] All unit tests pass: `go test ./pkg/agent/proxy/cbmap/`
- [x] Self-contained build (without local extraparsers.go): `go build ./...`
- [x] Docker-compose e2e: `cd tests/e2e/scram-channel-binding && ./run.sh` exits 0
- [x] Real `keploy record` against vanilla PG + libpq client with shim produces structured PostgresV3 mocks
- [x] Concurrent multi-upstream (160 queries × 2 PG instances) in PoC harness: 0 failures
- [ ] CI green
- [ ] Manual verification against sphinx-equivalent k8s deployment (depends on k8s-proxy PR landing first)

## References

- RFC 5802 — SCRAM
- RFC 5929 §4.1 — `tls-server-end-point` channel binding
- libpq `pgtls_get_peer_certificate_hash` in `fe-secure-openssl.c`


ayush3160:
## Issues we hit + current limitations

Posting this as a comment for visibility — also lives in the PR at [`pkg/agent/proxy/cbmap/LIMITATIONS.md`](../blob/feat/scram-channel-binding-shim/pkg/agent/proxy/cbmap/LIMITATIONS.md).

---

# Issues and Limitations — Channel-Binding Shim

Companion to `README.md`. This file collects everything we learned
during the implementation. Some of these are bugs we fixed mid-flight
and want to keep visible so they don't regress; others are
fundamental limitations that constrain who can use the feature.

## 1. Implementation bugs we hit (now fixed)

Recorded here so they don't regress and so reviewers understand the
shape of the problem space.

### 1.1 Wrong cbmap hook call site

**Symptom**: `cbmap.txt` was published in OSS `--cmd-type native` mode (where
`KEPLOY_INTERCEPT_PG=1` enabled the proxy-side path) but stayed empty in
the production-shaped flow with the integrations Postgres v3 parser linked.

**Root cause**: my first wiring put `cbmap.RegisterReal` only inside
`proxy.go::dialPostgresSSLUpstream`. The integrations v3 parser uses
the parser-driven TLSUpgrader path, which never calls that helper.

**Fix**: register the real cert at the single chokepoint that ALL
upstream-TLS-establishing paths hit:

- V1 / legacy parser flow: `util/tls_upgrader.go::UpgradeDestTLS`
- V2 / supervisor parser flow: `relay/directive_proc.go::handleUpgradeTLS`
- Raw `InterceptPostgresSSLRequest` flow: `proxy.go::dialPostgresSSLUpstream`

### 1.2 Cert-cache lookup order bug

**Symptom**: `cbmap-relay: no peer certs on upgraded conn` even when the
upstream TLS handshake clearly produced a peer cert.

**Root cause**: the upgraded `net.Conn` returned from `UpgradeDestTLS`
is wrapped in `*relay.readTimeReportingConn` (and sometimes
`*relay.readTrackingConn`). My `unwrapToTLSConn` walker only knew
about `*util.SafeConn` and `*util.Conn`, so the type assertion to
`*tls.Conn` failed silently.

**Fix**: add `Unwrap() net.Conn` methods to the relay wrappers
(`pkg/agent/proxy/relay/read_time.go`). The walker is now generic —
any wrapper that exposes `Unwrap()` or `NetConn()` is drilled through
until a `*tls.Conn` is found.

### 1.3 RegisterMITM / RegisterReal ordering depended on the flow

**Symptom**: in the `KEPLOY_INTERCEPT_PG=1` flow the upstream dial
fires BEFORE the app-side TLS handshake completes — so when
`registerRealForCBMap` ran, `CertForClient` hadn't minted the MITM
cert yet, and the cert cache was empty.

**Root cause**: my first design used a synchronous cache lookup
(`CachedLeafDER(host)`) at register time. Different proxy flows have
different orderings between client-side and server-side handshakes.

**Fix**: replace the synchronous lookup with a **deferred-publish**
state machine. `RegisterMITM` and `RegisterReal` are symmetric — each
stashes one half against a `connID` (the app-side source port). The
SECOND half to arrive triggers `cbmap.Publish` regardless of ordering.
`CleanupConnection` reaps half-arrived state on connection teardown.

### 1.4 LD_PRELOAD propagation in native mode

**Symptom**: in `--cmd-type native`, the user app process (gunicorn)
had every other CA-bundle env var set correctly, but **no
`LD_PRELOAD`** in `/proc/$pid/environ`.

**Root cause**: `tls/ca.go::SetEnvForPath` runs inside the
keploy-AGENT process. Its `os.Setenv` only affects the agent's env.
The user app is `exec`'d by the keploy-CLI process — a sibling of the
agent, not its child. The agent and the CLI share a parent (the
`sudo` wrapper) but not an environment.

The existing CA-bundle env vars happened to "work" despite this bug
because native mode also installs the CA into the system trust store
— apps pick it up without the env var. `LD_PRELOAD` has no such
fallback.

**Fix**: set `LD_PRELOAD` (and `CBSHIM_HASHMAP`) in
`pkg/client/app/App.Run`, which runs in the CLI process right before
`os.Exec`. The user app inherits the just-`Setenv`'d value.

### 1.5 Shim's RTLD_NEXT can't see psycopg2-binary's bundled libcrypto

**Symptom**: with `psycopg2-binary` (which bundles libpq + libcrypto
into a single .so), the shim logged `cannot resolve X509_digest via
RTLD_NEXT` and libpq surfaced `could not generate peer certificate
hash`.

**Root cause**: `psycopg2-binary` loads `_psycopg.cpython.so` with
`RTLD_LOCAL` (Python's default). The bundled libcrypto's symbols
aren't exported globally, so `dlsym(RTLD_NEXT, "X509_digest")` finds
no candidate. The shim returns 0; libpq treats that as failure.

**First attempt that backfired**: I added a `dlopen("libcrypto.so.3")`
unconditional fallback. That worked for plain `psycopg2` but
**segfaulted** for `psycopg2-binary`: the foreign X509 struct's
internal layout differs between the bundled libcrypto and the system
one. Calling system libcrypto's `X509_digest` on a foreign X509
crashes in `X509_get_signature_nid()` or similar internal accessors.

**Fix**: walk `/proc/self/maps` BEFORE the dlopen fallback. Only
dlopen a libcrypto that is ALREADY mapped in the process (the
RTLD_LOCAL-Python-`import` case — the libcrypto IS in the process,
just not in the global symbol namespace). If no libcrypto is mapped
at all (the `psycopg2-binary` case), bail with a clear error pointing
the operator at the actionable workaround.

### 1.6 GLIBC forward-incompatibility

**Symptom**: `/usr/local/bin/python3.13: /lib/x86_64-linux-gnu/libc.so.6:
version 'GLIBC_2.38' not found (required by /tmp/keploy-tls/cbshim.so)`

**Root cause**: I built `cbshim.so` on the host (Debian 13 / glibc
2.41). GCC at that glibc level auto-substitutes `sscanf()` with
`__isoc23_sscanf()` — a C23-spec'd symbol added in glibc 2.38. The
resulting .so carries a `GLIBC_2.38` dependency. glibc is
backward-compatible (binary built on glibc N runs on N+x) but NOT
forward-compatible (binary built on N+x will NOT run on N).

**Fix**: rebuild the embedded `cbshim.so` inside a `debian:bookworm`
container (glibc 2.36). Max GLIBC requirement after rebuild: 2.34.
That works on `debian:bookworm` (2.36), `ubuntu:22.04` (2.35), and
everything newer.

### 1.7 Shim's default cbmap path was the PoC leftover

**Symptom**: shim logged `[cbshim] hashmap not found:
/tmp/scram-poc-hashmap` even though keploy was writing to
`/tmp/keploy-tls/cbmap.txt`.

**Root cause**: my early PoC harness hardcoded `/tmp/scram-poc-hashmap`
as the default; the shim's compiled-in fallback never got updated
when I changed the production default in `cbmap.DefaultPath`.

**Fix**: shim default is now `/tmp/keploy-tls/cbmap.txt` (matches
`cbmap.DefaultPath`), AND keploy explicitly sets `CBSHIM_HASHMAP=cbmap.Path()`
in the env it passes to the app — so the two agree even when
`KEPLOY_CBMAP_PATH` overrides the location.

---

## 2. Fundamental limitations of the current implementation

These are NOT bugs and won't be fixed by patching the shim. They're
architectural constraints worth documenting up front.

### 2.1 Embedded shim is `linux/amd64 + glibc + OpenSSL 3.x` only

The `go:embed`'d `cbshim.so` is a single binary built for one
target. It will not load on:

- **arm64** containers (Apple Silicon dev boxes, AWS Graviton, Raspberry Pi)
- **musl-based images** (Alpine, distroless static)
- **glibc < 2.34** (Ubuntu 20.04 with glibc 2.31, RHEL 8 with 2.28)
- **OpenSSL 1.x** (very old images — OpenSSL 1.1 sunsetted in
  Sep 2023; should be rare but exists in long-tail enterprise images)

Effect: the shim silently fails to load (ld.so error) and the
container goes back to today's behavior (FATAL on SCRAM-PLUS).

**Follow-up needed**: CI matrix that produces 4+ variants
(amd64/glibc, amd64/musl, arm64/glibc, arm64/musl) and embeds them
all, with a runtime selector picking the right one based on the app
image's libc / arch. See section 5.

### 2.2 Apps with statically-bundled libcrypto can't be helped

**Affected apps** (non-exhaustive):

- `psycopg2-binary` (PyPI wheel — bundles libpq + libcrypto)
- `cryptography` (PyPI wheel — bundles its own OpenSSL)
- Oracle Instant Client
- Some Go binaries using `cgo` with `CGO_ENABLED=1` and a vendored
  OpenSSL
- Some Rust binaries using `openssl-sys` with the `vendored` feature

For these, `RTLD_NEXT` from the shim returns NULL (no global
libcrypto in the chain) and `/proc/self/maps` shows no system
libcrypto mapping. The shim bails with a clear error. The fix is on
the app side:

- Python: switch from `psycopg2-binary` to `psycopg2` (source dist),
  or use the OS-packaged `python3-psycopg2`.
- Go cgo: build without `-tags=vendored` for openssl deps.
- Rust: use `openssl` crate with default features (dynamic) instead
  of `openssl/vendored`.

**Why we can't fix this at the shim level**: the bundled libcrypto's
X509 internal struct layout is private to that build. Calling a
DIFFERENT libcrypto's `X509_digest` on a foreign X509 segfaults inside
the cert accessors. There's no safe way for the shim to bridge.

### 2.3 Non-OpenSSL TLS stacks are unreachable

The shim hooks `X509_digest`, which only exists in OpenSSL / BoringSSL.
Languages that use a non-OpenSSL TLS implementation are invisible to
the shim:

| Stack | Default TLS | Shim helps? |
|---|---|---|
| Java | JSSE (Java's own crypto) | ❌ No `X509_digest` call |
| Go | crypto/tls (in-tree) | ❌ Pure Go SHA computation |
| .NET | SChannel (Windows) / System.Net | ❌ Managed runtime |
| Rust + rustls | rustls | ❌ Pure Rust |
| Python + ssl module alone | bundled into _ssl.cpython.so | ⚠️ Depends on build |

These need entirely different interception mechanisms (Java agent,
Go binary patching, .NET assembly rewriting). Out of scope for this
PR.

### 2.4 setuid / setgid binaries strip LD_PRELOAD

The kernel ignores `LD_PRELOAD` for setuid/setgid binaries as a
hardening measure. Rare in containerized apps (most are not setuid)
but worth flagging — `ping`, `mount`, etc. would skip the shim if
they ever opened a PG connection (they wouldn't, but the principle
holds).

### 2.5 No graceful disable per-deployment

Right now, the shim is on if the .so is staged and `LD_PRELOAD` is
set. To disable per-deployment you have two options:

- **K8s mode**: remove the `LD_PRELOAD` env var injection in the
  webhook (requires webhook config change or an annotation override
  — not implemented).
- **Docker / native mode**: `chmod 000 /tmp/keploy-tls/cbshim.so` or
  un-mount the volume.

**Follow-up**: add a `--enable-channel-binding-shim` config flag
gating both the agent's `extractCBShimToTemp` AND the env-var
injection. Default on.

### 2.6 Upstream cert rotation invalidates cached mappings

If the real Postgres rotates its server cert mid-recording-run, the
`(H_mitm, H_real)` pair in `cbmap.txt` becomes stale. The next SCRAM
proof will use the OLD `H_real` and PG will reject it.

In practice: cert rotation during a recording session is rare. But
it's a real failure mode for long-running keploy-agent processes
(months-long staging environments).

**Follow-up**: detect cert change on each upstream TLS handshake; if
the captured leaf differs from what's already cached, re-publish.
Simple addition to `RegisterReal`.

### 2.7 `cbmap.txt` grows unboundedly

Each new upstream pair adds an entry. There's no eviction; the file
only grows. For typical fleets this is fine (a handful of upstreams
per process). For dynamically-many-upstreams workloads (e.g., a
multi-tenant SaaS connecting to many customer DBs), the file could
bloat.

**Follow-up**: LRU eviction or TTL on entries.

---

## 3. Footguns and edge cases

Things that are "working as intended" but might confuse someone.

### 3.1 The shim's stderr is noisy at default verbosity

Even without `CBSHIM_DEBUG=1`, the shim prints to stderr on every
resolver failure (`FATAL: cannot resolve X509_digest`). For apps
that don't use OpenSSL at all (like pure-Java services), the shim
loads, fails to find X509_digest in any reasonable place, and prints
this on startup.

The message is alarming for ops teams who don't know what cbshim is.

**Follow-up**: gate the message behind `CBSHIM_DEBUG=1` (or check if
the process actually uses libcrypto before logging).

### 3.2 `apt-mark manual` is required when slim-purging libpq-dev

If a user's Dockerfile installs `libpq-dev` as a build dep and then
`apt-get purge --auto-remove libpq-dev` to slim down the image,
**apt also removes the runtime `libpq5`** (it's considered an orphan
once `libpq-dev` is gone). The result: `ImportError: libpq.so.5:
cannot open shared object file`.

The fix is `apt-mark manual libpq5 libssl3` before the purge. Not
keploy-specific, just sensible apt usage — but worth documenting
because the cascade is silent at build time.

### 3.3 Shim cannot distinguish "key rotation in flight" from "wrong cert"

If a stale entry is in `cbmap.txt` (e.g., from a previous run
against a different PG), the shim will substitute the wrong hash and
the SCRAM proof will fail with `password authentication failed`
(misleading — it's not actually a password problem).

**Workaround**: delete `cbmap.txt` between runs that target different
PG instances.

**Follow-up**: include a TTL on entries OR have the shim verify by
attempting both the swapped hash AND the original.

### 3.4 Native mode writes shim to `/tmp` — assumes writable, exec-allowed `/tmp`

Some hardened deployments mount `/tmp` as `noexec` or read-only. In
that case `extractCBShimToTemp` either fails (read-only — error
caught, shim disabled) or writes successfully but the .so is
non-executable so `ld.so` refuses it.

**Follow-up**: prefer a path that's exec-allowed (e.g., user's home
or `/var/lib/keploy`) when `/tmp` is `noexec`. Detect via `stat -f`
on `/tmp`.

### 3.5 Concurrent keploy runs share `/tmp/keploy-tls/cbmap.txt`

If two `keploy record` invocations run on the same host (different
apps, different PGs), they both write to
`/tmp/keploy-tls/cbmap.txt`. The atomic-rename mechanism keeps the
file consistent, but they'll see each other's entries. Mostly
benign — the shim's lookup-by-H(mitm_cert) discriminates by cert
identity — but unexpected if you're debugging.

**Workaround**: set `KEPLOY_CBMAP_PATH` per-run.

### 3.6 Replay mode and the shim

The shim is a record-time aid. During replay, keploy isn't doing
MITM (it's serving from recorded mocks), so the shim shouldn't be
needed. But `LD_PRELOAD` is currently injected unconditionally
(record AND replay).

In replay, the shim loads, reads `cbmap.txt`, but since no real
upstream is dialed there's no fresh `cbmap` entry. If a stale entry
matches by chance, the shim would substitute a hash that doesn't
correspond to any real cert — could cause subtle replay-mode bugs.

**Follow-up**: scope the LD_PRELOAD injection to record mode only,
or have the shim no-op when keploy isn't running.

---

## 4. Operational and security concerns

For the team's tracking, not blocking the PR.

### 4.1 Security audit posture

The shim is technically a "TLS-MITM-companion hook into a
cryptographic primitive inside the user app process." Any security
review will flag the LD_PRELOAD pattern. The defense:

- The trust boundary is unchanged: keploy is already doing TLS MITM
  with the CA injected into the app's trust store. The shim doesn't
  enable any new attacker capability.
- The shim only substitutes a 32-byte hash inside a function that's
  ALREADY being called as part of legitimate cert processing. It
  doesn't store credentials, doesn't write the network, doesn't
  bypass authentication. The auth itself (SCRAM proof) is still
  done by the real client using its real password.
- The cbmap.txt's contents are derived from certs that are already
  visible on the network — no secret material is added.

This should be documented in a security-review section of the cbmap
README or in a separate SECURITY.md, before this lands in any
SOC2/ISO-audited deployment.

### 4.2 Operational visibility

The cbmap.Publish path has Debug-level logs. Operators who don't
turn on Debug get no visibility into whether the shim is working.

**Follow-up**: a metric (counter of successful publishes, counter of
RTLD_NEXT-fallback hits, counter of bail-outs) or an Info-level
summary line per recording session.

### 4.3 OpenSSL major version bumps

The shim's `int X509_digest(const X509 *, const EVP_MD *, unsigned
char *, unsigned int *)` signature is the OpenSSL 3.x ABI. If a
future OpenSSL bump changes this (rare — `X509_digest` has been
stable since OpenSSL 1.0) the shim breaks for that target. Pin a
verification step in CI.

---

## 5. Tracked follow-ups (priority-ordered)

1. **Multi-arch / multi-libc shim build matrix** — produce
   `linux/amd64 + glibc`, `+ musl`, `arm64 + glibc`, `+ musl`. Embed
   all four; runtime selector picks the right one based on the
   target app image's libc.

2. **`--enable-channel-binding-shim` flag** with default on. Lets
   operators opt out without `chmod 000` games.

3. **Replay-mode scoping** — only inject LD_PRELOAD in record mode.

4. **Upstream cert rotation invalidation** — re-publish on cert
   change detected at next handshake.

5. **Cleaner stderr** — gate non-debug shim log lines behind a
   verbosity flag.

6. **`/tmp` noexec detection** — fall back to a writable+exec path
   for the native-mode shim staging.

7. **Operational metrics** — Prometheus-style counters for publish
   success, RTLD_NEXT fallback, bail-out reasons.

8. **Security writeup** — formal SECURITY.md section explaining the
   trust model. Should land alongside the first GA release.

9. **`cbmap.txt` LRU** — cap size or TTL for long-running agents.

10. **Static-libcrypto detection at config time** — keploy could
    inspect the target image's binaries during pod admission /
    docker-run rewriting and refuse-to-enable the shim if a
    statically-bundled libcrypto is detected (clearer failure mode
    than the runtime error).

ayush3160:
Companion enterprise-side PR opened: **https://github.com/keploy/enterprise/pull/2056**

It extends `.ci/scripts/prepare-dockerfile.sh` with the cross-compile RUN block (mirrors the existing time-freezer §4 block). Both PRs need to land together — landing only this OSS PR makes the embedded shim disappear from enterprise builds; landing only the enterprise PR is a no-op against current OSS (the OSS Dockerfile rewrite would target files that aren't yet declared embeddable).

ayush3160:
Closing as superseded — the merge of `origin/main` brought this branch's diff vs `main` to **zero**. All of this PR's content has either moved or been replaced:

| Original content | Where it lives now |
|---|---|
| OSS `cbshim.CBShim` interface plumbing in `pkg/agent/proxy/{proxy.go,tls/ca.go,relay/directive_proc.go}` | Already merged via #4239 (better factored — extracts the interface + factory hook, leaves the impl out of OSS) |
| `pkg/agent/proxy/cbmap/` package | Moved to keploy/enterprise#2091 as `pkg/agent/proxy/cbshim_ld/` with the rendezvous logic reshaped from package globals into struct fields |
| `pkg/agent/proxy/tls/asset/cbshim.c` + `build.sh` + asset README + `.gitignore` | Moved to keploy/enterprise#2091 as `pkg/agent/proxy/cbshim_ld/asset/` |
| `tests/e2e/scram-channel-binding/` | Will land in keploy/enterprise#2091 (or a follow-up enterprise CI PR) once the build-side is ready |
| `pkg/agent/proxy/util/tls_upgrader.go` / `util.go` / `pkg/client/app/app.go` / `relay/read_time.go` — direct `cbmap` wiring | Reverted to main; the equivalent functionality now flows through main's `mitmPublishHook` / `RealCertHook` indirection (added in #4239) which enterprise impls register against |
| `pkg/platform/docker/docker.go` LD_PRELOAD env injection for compose | Will move to enterprise's docker integration (the enterprise build is what carries the cbshim.so file in the first place) |

### Why this is the right outcome

The original PR's value was the LD_PRELOAD-based SCRAM-PLUS shim — that goal is **achieved** by keploy/enterprise#2091 (the LD_PRELOAD impl, registering via the `cbshim.CBShim` factory pattern), and #4253 (the OSS multi-factory upgrade that lets it coexist with the eBPF impl from #4239 / enterprise#2070).

The "factory + interface" pattern keeps the OSS surface free of binary-blob embeds, C source, and per-impl wiring. Each implementation (eBPF or LD_PRELOAD) is self-contained in its own enterprise package and registers via blank import.

### Related

- **keploy/keploy#4253** — OSS multi-factory upgrade so eBPF + LD_PRELOAD can coexist
- **keploy/enterprise#2091** — new enterprise LD_PRELOAD impl
- **#4239** — already-merged OSS interface extraction (parent abstraction)
- **keploy/enterprise#2070** — already-merged enterprise eBPF impl (sibling implementation)

PR #4239, opened 2026-06-02
refactor(cbshim): extract OSS-side interface; move SCRAM-PLUS impl to enterprise

ayush3160:
## Summary

Pulls the SCRAM-SHA-256-PLUS channel-binding shim implementation out of OSS keploy. What remains here is a thin interface (`cbshim.CBShim`) plus a factory-registration hook (`RegisterFactory` / `NewFromFactory`) consumed by `pkg/agent/proxy`. The actual BPF programs, uprobe attach code, and rendezvous map all move to keploy/enterprise alongside other paid features.

Companion PR: **keploy/enterprise#2070** — adds the cbshim package, BPF source, generated artifacts, sample app, and Woodpecker e2e workflow.

## Why this shape

The previous design shipped the channel-binding shim — a BPF-uprobe-backed workaround for SCRAM-SHA-256-PLUS failing across keploy's TLS MITM — in OSS. Moving it behind an interface here lets us:

1. **Keep OSS small** — no BPF artifacts in the OSS binary, no eBPF dependency
2. **Gate the feature** — `config.Record.ChannelBindingShim` (under `record:` in `keploy.yml`) defaults false; even on builds that have a factory registered, it must be explicitly enabled
3. **Avoid implementation lock-in** — anyone with a different cbshim implementation (e.g. an LD_PRELOAD variant for non-BPF kernels) can register their own factory without touching OSS

## What's in this PR

**Added** (`pkg/agent/proxy/cbshim/iface.go`):
```go
type CBShim interface {
    RegisterMITM(connID string, mitmDER []byte)
    RegisterReal(connID string, realDER []byte, sigAlgo x509.SignatureAlgorithm)
    CleanupConnection(connID string)
    AttachToProcessTree(rootPID int) error
    StartProcEventConsumer(ctx context.Context)
    Close() error
}

func RegisterFactory(f Factory)                                 // implementation's init() calls this
func NewFromFactory(logger *zap.Logger) (CBShim, error)         // proxy.New() consumes this
```

**Removed** from `pkg/agent/proxy/cbshim/`:
- `cbshim.go`, `attach_state.go`, `exec_helpers.go`, `cbshim_stub.go` — implementation
- `cbshim_{x86,arm64}_bpfel.{go,o}` — bpf2go generated artifacts
- `doc.go` — old go:generate directive

**Removed** workflow:
- `.github/workflows/scram_channel_binding_linux.yml` — moved to enterprise's Woodpecker CI alongside the impl
- `.github/workflows/test_workflow_scripts/python/scram-channel-binding/` — moved to `enterprise/.ci/samples/`

**Modified**:
- `config/config.go` — adds `Record.ChannelBindingShim bool` (default false). Settable via `record.channelBindingShim: true` in `keploy.yml`.
- `pkg/agent/proxy/proxy.go` — `cbshim` field/parameter/setter switches from `*cbshim.CBShim` (concrete) to `cbshim.CBShim` (interface); `cbshim.New(logger)` replaced by `cbshim.NewFromFactory(logger)` gated on `opts.Record.ChannelBindingShim || opts.Agent.ChannelBindingShim`.
- `pkg/models/instrument.go`, `pkg/service/record/record.go`, `pkg/platform/docker/{docker.go,util.go}`, `pkg/platform/http/agent.go` — propagation plumbing so the agent subprocess receives `--channel-binding-shim` on argv (covers native + docker + docker-compose cmd-types). The `--channel-binding-shim` flag is **hidden** on `keploy agent` (internal-only) and **absent from `keploy record`** in OSS — the user-facing CLI surface for this feature lives in the enterprise CLI provider, since OSS can't deliver the feature anyway (no factory registered).
- `cli/provider/cmd.go` — register and read the hidden `keploy agent` `--channel-binding-shim` flag for the orchestrator→agent propagation channel.

## Behavioral impact

- **OSS users**: SCRAM-SHA-256-PLUS auth through keploy's TLS MITM fails with `SCRAM channel binding check failed` — the same behaviour as before any cbshim shipped. This is the documented OSS limitation; users who need PLUS support run an enterprise build.
- **Enterprise users**: identical to today's behaviour when `record.channelBindingShim: true` is set in `keploy.yml` (or whichever user-facing CLI flag the enterprise side adds). With the flag at its default (false), the feature is inert (no BPF load, no uprobes) and SCRAM-PLUS regresses to the OSS behaviour.

## YAML key

```yaml
record:
  channelBindingShim: true
```

Lives under `record:` (not `test:`) because the shim is only active at record time — during replay keploy serves postgres traffic from mocks, no real SCRAM handshake completes against postgres, so the shim has nothing to do.

## Test plan

- [x] `go build ./...` clean
- [x] `golangci-lint run --timeout=5m` → 0 issues
- [x] `gofmt -l -d` → clean
- [x] Sample postgres app (`scram-sample-app`, psycopg2-binary → SCRAM-SHA-256-PLUS): non-PLUS endpoints (`/healthz`) → 200; PLUS endpoint (`/db/ping`) → 500 (factory nil = expected on OSS)
- [x] Enterprise binary against same sample with `record.channelBindingShim: true` → 6/6 endpoints 2xx (companion PR's local verification)

## Related

- **keploy/enterprise#2070** — companion PR (impl + BPF + Woodpecker e2e)
- **keploy/ebpf#99** — closed; cbshim BPF source moved into enterprise too


pkg/agent/proxy/cbshim/iface.go
```diff
@@ -0,0 +1,92 @@
...
+//
+// The proxy package consumes cbshim only via the CBShim interface
+// declared here, so the OSS build has no eBPF dependency, no BPF
+// artifacts to embed, and no cbshim implementation. When the cbshim
+// implementation is not registered (i.e. on an OSS-only build, or in
+// an enterprise build with the feature disabled), the proxy operates
+// exactly as it did before cbshim existed: SCRAM-SHA-256-PLUS clients
+// connecting through keploy's TLS MITM will fail with
+// "SCRAM channel binding check failed", which is the documented
+// limitation for non-enterprise builds.
```
ayush3160: Remove the enterprise references from the comments.

config/config.go
```diff
@@ -214,6 +214,7 @@ type Test struct {
...
+	ChannelBindingShim          bool                `json:"channelBindingShim" yaml:"channelBindingShim" mapstructure:"channelBindingShim"`                                                // Enable the SCRAM-SHA-256-PLUS channel-binding shim. The shim attaches eBPF uprobes to libcrypto's X509_digest and rewrites the cert-hash libpq folds into the SCRAM proof, so postgres clients running with channel_binding=require still authenticate through keploy's TLS MITM. OSS builds have no implementation registered and ignore this flag entirely. Enterprise builds register a factory at init() time and respect this flag — defaulted FALSE in both OSS and enterprise, so users who need SCRAM-PLUS support must opt in explicitly via keploy.yml (test.channelBindingShim: true). Requires CAP_BPF + a kernel that allows bpf_probe_write_user; without those the factory returns an error and the proxy keeps working for non-PLUS clients.
```
ayush3160: Why is this in the test struct ? this is used in the recording ?

pkg/agent/proxy/tls/ca.go
```diff
@@ -56,9 +70,6 @@ var caPKey []byte //private key
 //go:embed data/mozilla_roots.pem
 var embeddedFallbackRoots []byte
```
ayush3160, line 103: why did we removed , what purpose it was serving ?

//go:embed asset
var _ embed.FS

ayush3160:
Thanks for the review — addressed all three in `44615797`:

### 1. `pkg/agent/proxy/cbshim/iface.go` — "Remove the enterprise references from the comments."
Done. The package doc + every method comment now describes the interface contract without naming a specific implementor. `RegisterFactory` is described as "called from init() in the implementation package" rather than "from init() in the enterprise cbshim package". Same treatment for the lifecycle methods.

### 2. `config/config.go` — "Why is this in the test struct? this is used in the recording?"
You're right — cbshim activates during both **record** (real upstream postgres TLS) and **replay** (mock postgres still goes through the MITM proxy, the app still computes the channel-binding hash against keploy's cert and folds it into the SCRAM proof), so neither `Test` nor `Record` is a good home.

Moved to the top-level `Config` struct so it's just `test.channelBindingShim: true` → `channelBindingShim: true` in `keploy.yml`. `proxy.New` now reads `opts.ChannelBindingShim` directly.

### 3. `pkg/agent/proxy/tls/ca.go` — "why did we removed, what purpose it was serving?" (the `//go:embed asset/cbshim.so` directive)
That was from the **legacy LD_PRELOAD** approach (commit `904dfb0e` on this branch). Back then cbshim was a libpq-compatible `.so` we built from C and embedded into the keploy binary; at recording start, the proxy wrote `cbshim.so` to a shared volume and the admission webhook injected `LD_PRELOAD=…/cbshim.so` into the user app. The shim ran in-process inside the app and dlsym'd `X509_digest` to substitute the hash.

It got removed when we moved cbshim to the **eBPF uprobe approach** (the design this PR ships). The uprobe attaches to `X509_digest` at the inode level from the kernel side — no `.so` injection, no `LD_PRELOAD`, no shared volume staging. The same substitution happens but the kernel does it, which is why the embed went away and the volume-stage code in `SetupCAForApp` got dropped too.

The trigger for the move was specifically `psycopg2-binary`: auditwheel ships its bundled libcrypto under `RTLD_LOCAL` with a mangled SONAME, so `dlsym(RTLD_NEXT, "X509_digest")` from the LD_PRELOAD shim returned NULL and the substitution silently no-op'd. Uprobes don't care about RTLD scopes or SONAME mangling — they bind to the inode + symbol offset — so the new design reaches into auditwheel-bundled wheels that defeated the old one.

ayush3160:
Thanks Copilot — addressed all 9 in `4a161b93`. Also picked up the `gofmt` failure CI was flagging (a field-tab alignment in `pkg/models/instrument.go` after the `ChannelBindingShim` field I added).

### Code fixes

**`pkg/agent/proxy/proxy.go:2318` — MITMPublishHook not cleared before Close**
Fixed. `StopProxyServer` now calls `SetCBShim(nil)` first (which detaches the global `tls.MITMPublishHook`), saves the handle, then closes it. Removes the use-after-close window if an in-flight `CertForClient` races shutdown.

**`pkg/agent/proxy/proxy.go:955` — "BPF load failed" log misleads on the disabled path**
Fixed. Dropped the "BPF load failed at proxy.New" wording; lowered to Debug; reworded to "not attached (feature disabled or no implementation registered)" so it accurately covers both OSS-no-factory and flag-off cases.

**`pkg/agent/proxy/tls/ca.go:1284` — `publishMITM` missing the documented `sourcePort==0` guard**
Fixed. Added `|| sourcePort == 0` to the early-return guard. Implementation now matches the comment.

**`pkg/agent/proxy/cbshim/iface.go:13` — package comment uses the wrong config path**
Fixed. Updated to `config.Record.ChannelBindingShim` (the actual location after the move from `Test` to `Record`), with a note about the hidden agent-side `--channel-binding-shim` argv mirror.

**`pkg/agent/proxy/opportunistic_tls.go:301/306/310/314` — per-connection Info logs**
All four lowered to Debug with a single block-level comment explaining why. These were genuinely hot-path diagnostics and would have flooded operator logs in high-throughput recording.

### Description / not-code fix

**`config/config.go:83` — PR description says `Test`, code says `Record`**
Not a code change. The PR went through a few field-location iterations (Test → top-level → Record) and the description in the PR body lagged the moves. Updating the PR description to call out that the YAML key is `record.channelBindingShim: true` so reviewers land on the right place.

### CI

The previous run failed `gofmt -l -d .` because of the field-tab alignment Copilot didn't comment on but was the actual blocker. Pushed locally clean: `go build`, `golangci-lint run`, `gofmt -l -d .` all clean.

ayush3160:
Thanks Copilot, second round addressed in `a63451d8`:

### `pkg/agent/proxy/tls/ca.go:1284` — publishMITM read MITMPublishHook twice

You were right that the local-snapshot from round 1 wasn't enough — under `-race` the bare global read still races with concurrent writes from `SetCBShim(nil)`. Wrote a stress test (`TestMITMPublishHook_RaceWithToggling`) that confirmed it triggers immediately.

Proper fix:
- `MITMPublishHook` → `mitmPublishHook atomic.Pointer[func(string, []byte)]`
- New `SetMITMPublishHook(h)` setter is the only sanctioned way to mutate the hook
- `publishMITM` does a single `Load() + deref` instead of check-then-call
- `proxy.go`'s `SetCBShim` switched to the setter

Race test stays clean under `-race` now.

### `pkg/agent/proxy/opportunistic_tls.go:306` — never called CleanupConnection

Fair catch — half-rendezvous state could accumulate if either MITM or real-cert publish never arrived (or arrived after `Publish` had already moved on). `hijackAndMITM` now `defer p.cbshim.CleanupConnection(connID)` right after `RegisterReal`. If the rendezvous completed normally via `Publish`, the pending entry is already cleared and the defer is a no-op; otherwise the half-state drops before it can leak.

### `pkg/agent/proxy/tls/ca.go:52` — no unit test for the MITMPublishHook wiring

Added `pkg/agent/proxy/tls/mitm_publish_hook_test.go` with three tests:

- **FiresOnceWithCorrectArgs** — calls `CertForClient`, asserts the hook fires exactly once, with `connID == strconv.Itoa(<source port>)` and the same DER `CertForClient` returns. Pinned regression coverage for "someone refactored `publishMITM` away".
- **NoCallWhenHookNil** — calls `CertForClient` with the hook unset (the OSS default + the post-shutdown state). No panic, valid cert returned. Pinned regression coverage for the nil-safe guard.
- **RaceWithToggling** — one goroutine churns `SetMITMPublishHook(on)` / `SetMITMPublishHook(nil)` at max speed while 8 worker goroutines call `CertForClient` × 25 each. With the pre-refactor bare-global pattern, `-race` flags this immediately; with the `atomic.Pointer`, it stays quiet.

All three plus the existing `TestCertCache*` tests pass under `go test -race`.

ayush3160:
Third round addressed in `00bb4899`:

### `proxy.go:709 / :712 / :941` — operator-facing logs missing `next_step`
Added structured `next_step` fields to all three cbshim Info logs (factory error, no-implementation-registered, AttachToProcessTree error). Each one names the concrete checks (CAP_BPF, kernel ≥ 5.5 with bpf_probe_write_user permitted, enterprise vs OSS build) so operators don't have to grep source to know what to verify. Matches the convention SetupCA's error logs use.

### `proxy.go:2324` — shutdown race even with atomic hook clear
Genuinely real race. The atomic.Pointer from round 2 prevented nil-deref between Load and call, but didn't fence `cb.Close()` against a goroutine that had already loaded a non-nil ptr and was mid-call to `RegisterMITM`.

Switched to `sync.RWMutex` around the hook:
- `publishMITM` holds an RLock across the nil-check + call
- `SetMITMPublishHook` takes the write Lock — blocks until in-flight readers finish

So `StopProxyServer`'s `SetMITMPublishHook(nil) → cb.Close()` sequence now drains in-flight `RegisterMITM` calls before close. Tradeoff is RLock contention on the hot path, but it's in the noise next to the cert mint / TLS handshake on the same call.

### `opportunistic_tls.go:320` — Debug log eagerly evaluates Sprintf
Gated under `if ce := p.logger.Check(zap.DebugLevel, …); ce != nil { ce.Write(…) }` so `fmt.Sprintf("%T", ...) + RemoteAddr().String()` only run when debug is enabled. Matches the repo's pattern for expensive-to-format debug breadcrumbs.

### `cbshim/iface.go:15` — `record:channelBindingShim` shorthand was wrong
Fixed to the actual nested-YAML shape: `record.channelBindingShim` (or `record:\n  channelBindingShim: true`). No more copy-paste landmine.

### `directive_proc.go:414` — RealCertHook had no unit coverage
Added `pkg/agent/proxy/relay/real_cert_hook_test.go` with two tests:

- **FiresWithExpectedArgs** — stages a real loopback TLS handshake against a self-signed cert, wires the client-side `*tls.Conn` as the "upgraded" dest conn, asserts `RealCertHook` fires **exactly once** with `connID == strconv.Itoa(sourcePort)`, the actual server-leaf DER, and the right signature algorithm.
- **NoCallWhenHookNil** — calls `publishRealCertFromUpgraded` with the hook unset and with a non-TLS conn. No panic, no hook fire.

Catches the silent-regression class Copilot called out: a refactor that hashed the cert or keyed on a different addr field would compile + pass every existing relay test but break the SCRAM-PLUS rendezvous. With this test, that breakage now fails the build.

All tests pass under `go test -race`.

ayush3160:
Fourth round addressed in `d69515ee`:

### `proxy.go:2330` — use-after-close if drain times out
Real concern. `dialPostgresSSLUpstream` takes `p.cbshim` as a function parameter, so its captured local survives a concurrent `SetCBShim(nil)`. If `waitForConnDrain` times out and we call `cb.Close()` immediately afterward, an in-flight pg-SSL handshake goroutine can still race `RegisterReal` against the closed cbshim.

Fix: the actual `cb.Close()` is now deferred onto a goroutine that `activeConns.Wait()`s first. Two cases:

- Drain succeeded in the grace window → `Wait()` returns immediately, `Close` runs equivalently-synchronously.
- Drain timed out → the deferred `Close` keeps the BPF resources alive until the last in-flight goroutine returns. If `Wait` never returns (genuinely stuck goroutine), the kernel reclaims BPF resources at process exit anyway.

```go
go func() {
    p.activeConns.Wait()
    if err := cb.Close(); err != nil { ... }
}()
```

### `proxy.go:2036` — RegisterReal can leak without CleanupConnection
Right — `dialPostgresSSLUpstream` only fires `RegisterReal` on success, but if the matching MITM half never arrives (TLS handshake fails, postgres rejects auth pre-SCRAM, client disconnects mid-handshake), `Publish` never fires and the entry sticks in cbshim's pending map.

Fix: `defer p.cbshim.CleanupConnection(connID)` right after the `dialPostgresSSLUpstream` call site, scoped to `handleConnection`'s lifetime. Handle is snapshotted into a local so a concurrent `SetCBShim(nil)` during shutdown can't race the deferred call. No-op when `Publish` already drained the entry — same shape as the `opportunistic_tls.go` fix from round 2.

### `directive_proc.go:816` — Debug log evaluates `leaf.Subject.String()` eagerly
Same pattern as the `opportunistic_tls.go` debug logs from round 3 — `pkix.Name.String()` walks the cert subject's RDN sequence, non-trivial on every UpgradeTLS directive. Gated under `log.Check(zap.DebugLevel, …)`:

```go
if ce := log.Check(zap.DebugLevel, "relay: RealCertHook fired"); ce != nil {
    ce.Write(zap.String("conn_id", ...), zap.String("subject", leaf.Subject.String()))
}
```

All builds + lint + race tests still green.

ayush3160:
Fifth round addressed in `93613059`:

### `directive_proc.go:823` + `opportunistic_tls.go:296` — Port==0 collisions across cbshim publishers
Both fixed. `tls.publishMITM` already had a `sourcePort==0` guard from round 1; the symmetric guards on the two RealCertHook paths were missing. Wrapped conns that surface `*net.TCPAddr{Port: 0}` (sentinel / unspecified) would publish `connID="0"` and collide with every other Port==0 connection in cbshim's pending map. Both paths now skip in that case, matching the MITM publisher's semantics.

### `opportunistic_tls.go:304` — eager Debug log evaluation
Gated the success-branch debug log under `p.logger.Check(zap.DebugLevel, …)`. `RemoteAddr().String()` allocates each call and `SignatureAlgorithm.String()` walks the enum table — both pointless when debug is off. Same pattern the other three branches in this file (and `directive_proc.go:816`) already use.

### `cbshim/iface.go:94` — `(nil, nil)` ambiguity
Fair — the docstring promised `(nil, nil)` meant "no factory registered", but the implementation forwarded a registered factory's `(nil, nil)` verbatim. `proxy.New` would then take the "no implementation registered (OSS build)" branch even though the factory ran and silently produced nothing.

Tightened the contract: a registered factory returning `(nil, nil)` is now a hard error from `NewFromFactory`. The `(nil, nil)` signal is reserved for the actual unregistered case.

### `mitm_publish_hook_test.go:137` — tight toggler spinloop
Added `runtime.Gosched()` at the end of each toggle iteration. Without it, the toggler pinned a core on low-CPU CI runners and starved the worker goroutines doing cert mints, inflating the test wall-clock without changing what it covers.

All builds + lint + race tests still green.

ayush3160:
Sixth round addressed in `166c97db`:

### `proxy.go:2354` — `p.SetCBShim(nil)` races concurrent `p.cbshim` reads
Real race. `cbshim.CBShim` is an interface (2-word pointer); the nil-write isn't atomic, and the opportunistic-TLS path reads `p.cbshim` multiple times within a single connection (nil-check → `RegisterReal` → `defer CleanupConnection`). `-race` would flag this — and the multi-read sites could observe a torn nil between consecutive reads.

Two-part fix:

**StopProxyServer no longer mutates `p.cbshim`**. Just clears the global `tls.MITMPublishHook` (whose setter already drains in-flight `publishMITM` via the RWMutex from round 3). The deferred-close goroutine from round 4 still waits on `activeConns.Wait()` before calling `cb.Close()` — the close uses a captured local, so it doesn't need `p.cbshim` to have been nilled. Since shutdown is in progress (listener closed, parent ctx cancelled, no new accepts), leaving `p.cbshim` pointing at the now-quiesced handle is safe — its publisher-facing surface is already inert via the cleared global hook.

**`opportunistic_tls.go` snapshots `p.cbshim` once** at the top of the cbshim block:

```go
cb := p.cbshim
if cb != nil {
    ...
    cb.RegisterReal(...)
    defer cb.CleanupConnection(...)
}
```

Defence in depth — even if a future call site (or some debug-only override) mutates `p.cbshim`, the snapshot pins the handle for the whole hijack lifetime.

Build + lint + tests still green.

gouravkrosx (approved):

gouravkrosx:
@ayush3160 Please check the pipelines before merging.

PR #4273, opened 2026-06-12
feat(mismatch): unified mock-mismatch reporting + whole-mock side-by-side diff

Aditya-eddy:

Complete mock-mismatch reporting for the replay flow. This PR targets `main` and is **self-contained**: it carries the unified-reporting foundation (commits from #4271, authorship preserved) because the additions below build directly on it (the `mismatch.Builder`, `MockFieldDiff` model, `JSONFieldDiffs`). If #4271 merges first, this reduces to just the additions.

**Foundation (from #4271):** universal `mismatch.Builder`, noise-vocabulary field diffs, `MatchPhase`/`CandidateCount`, `--schema-noise-strict` flag, `mockdb.PersistMockNoise`, `keploy report` rendering of unmatched calls.

**Added on top:**

**1. Plumbing — the structured report now reaches the user, attributed to the right test, on every exit path.**
Previously `StartErrorDrain` discarded every mock-not-found error (nothing opened a capture window) and the replayer only fetched `GetMockErrors` when `!r.instrument`, so a normal local `keploy test -c …` built the report but never surfaced it.
- `proxy.go`: a **per-test capture window** — `BeginTestErrorCapture` opens a fresh accumulator before each test, the single `StartErrorDrain` goroutine routes each mock-miss into it, and `GetMockErrors` drains-and-closes it. A bounded `pendingMockErrors` FIFO retains pre-window / background misses. A `flushMarker` rendezvous between `GetMockErrors` and the drain guarantees no in-flight miss is lost when the window closes (race-tested under `-race`). `BeginTestErrorCapture` **discards** a never-closed prior window rather than carrying it forward, so a failed fetch can't misattribute one test's misses to the next.
- `service/replay`: `beginTestErrorCapture` before each `SimulateRequest`, and `attachMockErrors` (→ `GetMockErrors`) on **every** test-exit path — the normal result, the early simulation-error / invalid-response returns, the nil-matcher-result / nil-test-case-result branches (which now also persist a failed result), and the **streaming (Phase 2) path** (finalized *after* stream consumption, see below). A miss attaches to the test that caused it and never carries into the next, on all transports.
- **Single owner, single consumer.** The replayer reaches the agent over HTTP in every deployment (native loopback, k8s sidecar, enterprise), so `instrumentation.GetErrorChannel()` is always nil and the legacy `monitorProxyErrors` channel-consumer never fired anywhere. Removed it — and its dead `AddProxyErrorForTest` / `GetFailuresForTestCase` feeders — so the agent's drain goroutine is the **sole** consumer of the error channel and `attachMockErrors` is the **sole** finalizer.

**2. Whole-mock side-by-side diff.**
Render the closest mock vs the live request side-by-side (`EXPECTED MOCK | RECEIVED REQUEST`) with differing lines highlighted and long unchanged runs folded, instead of a one-line cell.
- `models`: `ClosestMockReq`/`ReceivedReq` on `MockMismatchReport` + `UnmatchedCall`.
- `mismatch.Builder.WithRenderedRequests`; HTTP request renderer `http/mismatch_render.go` (JSON pretty-printed, sensitive/obfuscated values redacted).
- `service/replay/utils.go`: `printSideBySideDiff` (LCS align + git-style folding).

**3. Dedicated request-body noise bucket.**
`test.globalNoise.requestbody` for HTTP request-body match noise, decoupled from the response-assertion `body` bucket, so response noise can't silently weaken request matching under `--schema-noise-strict`.

**4. Streaming (SSE / NDJSON / multipart / chunked / binary) mismatch reporting.**
The deferred Phase-2 streaming path now opens the capture window before `SimulateRequest` and finalizes it **after** `CompareHTTPStream` + the post-stream consumed-mock drain — because outgoing mock calls continue while the stream body is consumed, so closing right after headers would drop or misattribute in-stream misses. Streaming-test misses now reach `FailureInfo.UnmatchedCalls` with all fields, same as non-streaming.

Verified end-to-end: local `keploy test` against a sample app, and the **live k8s-proxy flow** (agent builds the report → proxy renders the side-by-side `MOCK MISMATCH` report).

## Links & References

**Closes:** NA

### 🔗 Related PRs
- Supersedes / incorporates #4271 (its commits are included here). Coordinate merge order — if #4271 lands first, rebase drops it to just the additions.
- Cross-repo consumers of the widened report: keploy/api-server#1751 (model), keploy/k8s-proxy#521 (forward to dashboard + `BeginTestErrorCapture` wiring), keploy/enterprise-ui `feat/mock-mismatch-rich-fields` (render).
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [x] 🍕 Feature
- [x] 🐞 Bug Fix
- [x] 🧑‍💻 Code Refactor

## Added e2e test pipeline?
- [x] ✅ yes
- Added a `mock-mismatch` suite (`test_workflow_scripts/golang/mock_mismatch/golang-linux.sh` + a matrix entry in `golang_linux.yml`): records the http-pokeapi sample, mutates the recorded request path so the live request no longer matches any mock on replay, and asserts the **specific** HTTP unmatched-call report surfaces — both the `Mock mismatch: [HTTP] … /api/v2/` console line and a single matching `unmatched_calls` entry in the test-set report YAML.
- Unit coverage also exists for the `mismatch` builder, `JSONFieldDiffs`, the HTTP field-diff/redaction paths, and the capture-window race (`capture_window_test.go`, run under `-race`).

- [x] 👍 yes

## Added to documentation?
- (`test.globalNoise.requestbody` + the side-by-side output are worth a short docs note as a follow-up.)

### DIFF
--- a/.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh
+++ b/.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh
@@ -125,17 +125,22 @@ if grep -qE "Mock mismatch: \[HTTP\][^]]*/api/v2/" test_logs.txt; then
     mismatch_reported=true
 fi
 
-# Equivalent specific check on the test-set report yaml: an unmatched_calls
-# entry whose protocol is HTTP and whose actual_summary references /api/v2/.
+# Equivalent specific check on the test-set report yaml: a SINGLE unmatched_calls
+# item whose protocol is HTTP *and* whose actual_summary references /api/v2/.
+# actual_summary appears only inside unmatched_calls items, and each item's
+# protocol line precedes its actual_summary, so tracking the most recent
+# protocol per item binds both fields to the same entry (independent greps
+# could otherwise match different items).
 shopt -s nullglob
 reports=( ./keploy/reports/test-run-*/test-set-*-report.yaml )
 shopt -u nullglob
 for rf in "${reports[@]}"; do
-    # Limit the scan to each report's unmatched_calls block and require both an
-    # HTTP protocol and an /api/v2/ actual_summary within it.
-    if awk '/unmatched_calls:/{f=1} f' "$rf" | grep -qE "protocol: ?\"?HTTP" \
-        && awk '/unmatched_calls:/{f=1} f' "$rf" | grep -qE "actual_summary:.*/api/v2/"; then
-        echo "✓ $(basename "$rf") has an HTTP /api/v2/ unmatched_calls entry"
+    if awk '
+        /protocol:/        { httpItem = ($0 ~ /HTTP/) }
+        /actual_summary:/  { if (httpItem && $0 ~ /\/api\/v2\//) { found = 1 } }
+        END                { exit found ? 0 : 1 }
+    ' "$rf"; then
+        echo "✓ $(basename "$rf") has a single HTTP /api/v2/ unmatched_calls entry"
         mismatch_reported=true
     fi
 done
--- a/.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh
+++ b/.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh
@@ -114,30 +114,40 @@ if grep "WARNING: DATA RACE" test_logs.txt; then
     exit 1
 fi
 
-# Assert the mismatch surfaced. Primary signal: the console "MOCK MISMATCH"
-# report / per-call "Mock mismatch:" heading printed by the replayer.
+# Assert the SPECIFIC forced mismatch surfaced — an HTTP unmatched outgoing
+# call on the mutated /api/v2/ path. A generic banner (e.g. a routine DNS miss)
+# must NOT satisfy this, otherwise the suite could pass while the HTTP mismatch
+# was actually dropped. The per-call heading is "Mock mismatch: [HTTP] <METHOD>
+# <path>" and the live path is the unmutated /api/v2/... the app requested.
 mismatch_reported=false
-if grep -qiE "MOCK MISMATCH|Mock mismatch:" test_logs.txt; then
-    echo "✓ replay surfaced the MOCK MISMATCH report"
+if grep -qE "Mock mismatch: \[HTTP\][^]]*/api/v2/" test_logs.txt; then
+    echo "✓ replay reported the forced HTTP /api/v2/ mock mismatch"
     mismatch_reported=true
 fi
 
-# Secondary signal: the test-set report YAML carries unmatched_calls / match_phase.
+# Equivalent specific check on the test-set report yaml: an unmatched_calls
+# entry whose protocol is HTTP and whose actual_summary references /api/v2/.
 shopt -s nullglob
 reports=( ./keploy/reports/test-run-*/test-set-*-report.yaml )
 shopt -u nullglob
 for rf in "${reports[@]}"; do
-    if grep -qE "unmatched_calls|match_phase|closest_mock" "$rf"; then
-        echo "✓ report $(basename "$rf") carries the structured mismatch fields"
+    # Limit the scan to each report's unmatched_calls block and require both an
+    # HTTP protocol and an /api/v2/ actual_summary within it.
+    if awk '/unmatched_calls:/{f=1} f' "$rf" | grep -qE "protocol: ?\"?HTTP" \
+        && awk '/unmatched_calls:/{f=1} f' "$rf" | grep -qE "actual_summary:.*/api/v2/"; then
+        echo "✓ $(basename "$rf") has an HTTP /api/v2/ unmatched_calls entry"
         mismatch_reported=true
     fi
 done
 
 if [ "$mismatch_reported" != true ]; then
-    echo "::error::replay did NOT surface the mock mismatch — report regressed"
-    cat test_logs.txt
+    echo "::error::replay did NOT report the forced HTTP /api/v2/ mock mismatch"
+    echo "--- mismatch/unmatched lines in test_logs.txt ---"
+    grep -iE "mismatch|unmatched|no matching" test_logs.txt | head -20
+    echo "--- tail of test_logs.txt ---"
+    tail -40 test_logs.txt
     exit 1
 fi
 
-echo "mock-mismatch e2e passed: the unmatched outgoing call was reported"
+echo "mock-mismatch e2e passed: the forced HTTP /api/v2/ unmatched call was reported"
 exit 0
--- a/.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh
+++ b/.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh
@@ -78,7 +78,11 @@ send_request() {
     pid=$(pgrep keploy)
     echo "$pid Keploy PID"
     echo "Killing Keploy"
-    sudo kill "$pid"
+    # Unquoted on purpose: pgrep can return BOTH the keploy CLI and the agent
+    # process, so $pid may be multiple newline-separated PIDs. Quoting it passes
+    # one mangled arg to kill ("failed to parse argument"), keploy survives, and
+    # `wait` on the record command hangs to the job timeout. Word-split instead.
+    sudo kill $pid
 }
 
 # Record one iteration of test cases + mocks.
--- a/.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh
+++ b/.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh
@@ -15,6 +15,22 @@ echo "iid.sh executed"
 
 git fetch origin
 
+# Mock-mismatch reporting (the per-test capture window + unmatched_calls in the
+# report) is new in this PR. On the cross-version matrix the 'latest' replay
+# binary is the published release that predates this feature, so it can never
+# emit the "Mock mismatch: [HTTP]" line / unmatched_calls entry this suite
+# asserts (success is inverted) — skip cleanly there rather than fail
+# deterministically. Local build binaries (*/build/keploy, */build-no-race/
+# keploy) have the feature. Mirrors the streaming sibling + risk_profile /
+# connect_tunnel capability gates.
+case "${REPLAY_BIN:-}" in
+    */build/keploy|*/build-no-race/keploy) ;;
+    *)
+        echo "REPLAY_BIN ($REPLAY_BIN) predates mock-mismatch reporting; skipping suite"
+        exit 0
+        ;;
+esac
+
 if [ -f "./keploy.yml" ]; then
     rm ./keploy.yml
 fi
--- a/.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh
+++ b/.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh
@@ -1,0 +1,131 @@
+#!/bin/bash
+
+# E2E for the mock-mismatch report. Reuses the http-pokeapi sample (it makes
+# mockable outgoing HTTP calls), records it, then MUTATES the recorded mocks so
+# the live outgoing request no longer matches any mock on replay. Unlike the
+# other suites, success here is INVERTED: we assert that replay SURFACES the
+# mismatch (the "MOCK MISMATCH" report with an unmatched outgoing call), not
+# that every test passed.
+
+echo "$RECORD_BIN"
+echo "$REPLAY_BIN"
+
+source ./../../.github/workflows/test_workflow_scripts/test-iid.sh
+echo "iid.sh executed"
+
+git fetch origin
+
+if [ -f "./keploy.yml" ]; then
+    rm ./keploy.yml
+fi
+rm -rf keploy/
+
+build_go_app() {
+  local attempt=1
+  local max_attempts=4
+  local sleep_sec=5
+  while [ "$attempt" -le "$max_attempts" ]; do
+    if GOPROXY="proxy.golang.org,direct" go build -o http-pokeapi; then
+      return 0
+    fi
+    if [ "$attempt" -ge "$max_attempts" ]; then
+      echo "::error::go build for http-pokeapi failed after ${max_attempts} attempts"
+      return 1
+    fi
+    echo "go build attempt ${attempt} failed; retrying in ${sleep_sec}s…"
+    sleep "$sleep_sec"
+    sleep_sec=$((sleep_sec * 2))
+    attempt=$((attempt + 1))
+  done
+}
+build_go_app
+echo "go binary built"
+
+sudo "$RECORD_BIN" config --generate
+config_file="./keploy.yml"
+sed -i 's/global: {}/global: {"body": {"updated_at":[]}}/' "$config_file"
+
+send_request() {
+    sleep 6
+    app_started=false
+    while [ "$app_started" = false ]; do
+        if curl -X GET http://localhost:8080/api/locations; then
+            app_started=true
+        fi
+        sleep 3
+    done
+    echo "App started"
+    response=$(curl -s -X GET http://localhost:8080/api/locations)
+    location=$(echo "$response" | jq -r ".location[0]")
+    curl -s -X GET "http://localhost:8080/api/locations/$location"
+    sleep 7
+    pid=$(pgrep keploy)
+    echo "$pid Keploy PID"
+    echo "Killing Keploy"
+    sudo kill "$pid"
+}
+
+# Record one iteration of test cases + mocks.
+send_request &
+"$RECORD_BIN" record -c "./http-pokeapi" --generateGithubActions=false 2>&1 | tee record_logs.txt
+if grep "WARNING: DATA RACE" record_logs.txt; then
+    echo "::error::Race condition detected in recording"
+    cat record_logs.txt
+    exit 1
+fi
+wait
+echo "Recorded test cases and mocks"
+
+# Force a mock mismatch: rewrite the recorded request host on the mock side so
+# the live outgoing request (to pokeapi.co) no longer matches any recorded
+# mock. Done on the mocks only — the test cases (inbound) are untouched.
+shopt -s nullglob
+mock_files=( ./keploy/test-set-*/mocks.yaml )
+shopt -u nullglob
+if [ ${#mock_files[@]} -eq 0 ]; then
+    echo "::error::No recorded mocks found to mutate under ./keploy/test-set-*/"
+    cat record_logs.txt
+    exit 1
+fi
+for mf in "${mock_files[@]}"; do
+    sed -i 's#pokeapi\.co#pokeapi-mismatch.invalid#g' "$mf"
+    echo "mutated recorded mocks: $mf"
+done
+
+# Replay. The mutated mocks no longer match → keploy must report the unmatched
+# outgoing call. The replay itself is EXPECTED to not all-pass here.
+"$REPLAY_BIN" test -c "./http-pokeapi" --delay 7 --debug --generateGithubActions=false 2>&1 | tee test_logs.txt
+
+if grep "WARNING: DATA RACE" test_logs.txt; then
+    echo "::error::Race condition detected in test"
+    cat test_logs.txt
+    exit 1
+fi
+
+# Assert the mismatch surfaced. Primary signal: the console "MOCK MISMATCH"
+# report / per-call "Mock mismatch:" heading printed by the replayer.
+mismatch_reported=false
+if grep -qiE "MOCK MISMATCH|Mock mismatch:" test_logs.txt; then
+    echo "✓ replay surfaced the MOCK MISMATCH report"
+    mismatch_reported=true
+fi
+
+# Secondary signal: the test-set report YAML carries unmatched_calls / match_phase.
+shopt -s nullglob
+reports=( ./keploy/reports/test-run-*/test-set-*-report.yaml )
+shopt -u nullglob
+for rf in "${reports[@]}"; do
+    if grep -qE "unmatched_calls|match_phase|closest_mock" "$rf"; then
+        echo "✓ report $(basename "$rf") carries the structured mismatch fields"
+        mismatch_reported=true
+    fi
+done
+
+if [ "$mismatch_reported" != true ]; then
+    echo "::error::replay did NOT surface the mock mismatch — report regressed"
+    cat test_logs.txt
+    exit 1
+fi
+
+echo "mock-mismatch e2e passed: the unmatched outgoing call was reported"
+exit 0
--- a/.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh
+++ b/.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh
@@ -76,9 +76,12 @@ fi
 wait
 echo "Recorded test cases and mocks"
 
-# Force a mock mismatch: rewrite the recorded request host on the mock side so
-# the live outgoing request (to pokeapi.co) no longer matches any recorded
-# mock. Done on the mocks only — the test cases (inbound) are untouched.
+# Force a mock mismatch: rewrite the recorded request PATH on the mock side.
+# HTTP schema matching compares the request URL path (not host), so changing
+# the recorded "/api/v2/..." path means the live outgoing request to
+# "/api/v2/..." no longer matches any recorded mock -> the matcher reports an
+# unmatched outgoing call. Only the mocks are touched; the test cases (inbound
+# requests) are untouched, so the inbound path still replays normally.
 shopt -s nullglob
 mock_files=( ./keploy/test-set-*/mocks.yaml )
 shopt -u nullglob
@@ -87,10 +90,19 @@ if [ ${#mock_files[@]} -eq 0 ]; then
     cat record_logs.txt
     exit 1
 fi
+mutated_any=false
 for mf in "${mock_files[@]}"; do
-    sed -i 's#pokeapi\.co#pokeapi-mismatch.invalid#g' "$mf"
-    echo "mutated recorded mocks: $mf"
+    sed -i 's#/api/v2/#/api/v0-mismatch/#g' "$mf"
+    if grep -q '/api/v0-mismatch/' "$mf"; then
+        echo "mutated recorded request path in: $mf"
+        mutated_any=true
+    fi
 done
+if [ "$mutated_any" != true ]; then
+    echo "::error::mutation changed no recorded request path — sample/mock layout changed; e2e can't force a mismatch"
+    head -50 "${mock_files[0]}"
+    exit 1
+fi
 
 # Replay. The mutated mocks no longer match → keploy must report the unmatched
 # outgoing call. The replay itself is EXPECTED to not all-pass here.
--- a/pkg/agent/proxy/integrations/http/match.go
+++ b/pkg/agent/proxy/integrations/http/match.go
@@ -1244,6 +1244,12 @@ func mergeReqBodyNoise(existing, detected map[string][]string) map[string][]stri
 // mockDb; otherwise it uses the pre-fetched slice to avoid a redundant read.
 func (h *HTTP) buildHTTPMismatchReport(request *http.Request, liveBody []byte, mockDb integrations.MockMemDb, httpMocks []*models.Mock, headerNoise, userBodyNoise map[string][]string, diag *matchDiag) *models.MockMismatchReport {
 	actualKey := request.Method + " " + request.URL.Path
+	// Destination identifies WHICH upstream this missed call targeted; the same
+	// method+path can hit several hosts. Host header first, URL authority next.
+	dest := request.Host
+	if dest == "" && request.URL != nil {
+		dest = request.URL.Host
+	}
 	if httpMocks == nil && (diag == nil || len(diag.schemaMatched) == 0) {
 		// Mirror match()'s pool-merging + ordering strategy so
 		// mismatch diagnostics see the same candidate set in the
@@ -1251,11 +1257,13 @@ func (h *HTTP) buildHTTPMismatchReport(request *http.Request, liveBody []byte, m
 		perTestMocks, err := mockDb.GetPerTestMocksInWindow()
 		if err != nil {
 			return mismatch.NewReport(mismatch.ProtocolHTTP, actualKey).
+				WithDestination(dest).
 				WithNextSteps("Failed to read mock database. Check logs for errors and retry.").Build()
 		}
 		sessionMocks, err := mockDb.GetSessionMocks()
 		if err != nil {
 			return mismatch.NewReport(mismatch.ProtocolHTTP, actualKey).
+				WithDestination(dest).
 				WithNextSteps("Failed to read mock database. Check logs for errors and retry.").Build()
 		}
 		mocks := make([]*models.Mock, 0, len(perTestMocks)+len(sessionMocks))
@@ -1279,6 +1287,7 @@ func (h *HTTP) buildHTTPMismatchReport(request *http.Request, liveBody []byte, m
 
 	if candidateCount == 0 && len(schemaSurvivors) == 0 {
 		return mismatch.NewReport(mismatch.ProtocolHTTP, actualKey).
+			WithDestination(dest).
 			WithPhase(models.MatchPhaseNoMocks, 0).Build()
 	}
 
@@ -1289,6 +1298,7 @@ func (h *HTTP) buildHTTPMismatchReport(request *http.Request, liveBody []byte, m
 	closestMock := pickClosestCandidate(request, schemaSurvivors, httpMocks)
 	if closestMock == nil || closestMock.Spec.HTTPReq == nil {
 		return mismatch.NewReport(mismatch.ProtocolHTTP, actualKey).
+			WithDestination(dest).
 			WithPhase(phase, candidateCount).Build()
 	}
 
@@ -1341,6 +1351,7 @@ func (h *HTTP) buildHTTPMismatchReport(request *http.Request, liveBody []byte, m
 	redactFieldDiffs(fieldDiffs, closestMock.Noise)
 
 	b := mismatch.NewReport(mismatch.ProtocolHTTP, actualKey).
+		WithDestination(dest).
 		WithPhase(phase, candidateCount).
 		WithClosest(closestMock.Name, fieldDiffs).
 		// Whole-request renders for the CLI side-by-side diff. Mock.Noise is
--- a/pkg/agent/proxy/integrations/http/match.go
+++ b/pkg/agent/proxy/integrations/http/match.go
@@ -1337,7 +1337,13 @@ func (h *HTTP) buildHTTPMismatchReport(request *http.Request, liveBody []byte, m
 
 	b := mismatch.NewReport(mismatch.ProtocolHTTP, actualKey).
 		WithPhase(phase, candidateCount).
-		WithClosest(closestMock.Name, fieldDiffs)
+		WithClosest(closestMock.Name, fieldDiffs).
+		// Whole-request renders for the CLI side-by-side diff. Mock.Noise is
+		// passed so obfuscated values stay redacted on both sides.
+		WithRenderedRequests(
+			renderMockRequest(mockReq, closestMock.Noise),
+			renderLiveRequest(request, liveBody, closestMock.Noise),
+		)
 	if len(fieldDiffs) == 0 {
 		// Nothing structurally differs vs the closest candidate, yet the
 		// matcher rejected everything — point the user at the phase instead
--- a/pkg/agent/proxy/integrations/http/match.go
+++ b/pkg/agent/proxy/integrations/http/match.go
@@ -1083,10 +1083,11 @@ func (h *HTTP) detectReqBodyNoise(enabled bool, mock *models.Mock, reqBody []byt
 }
 
 // mergeNoiseMaps combines two noise maps into a fresh map; entries in a win
-// on key collision. Either input may be nil.
+// on key collision. Either input may be nil. The result never aliases an
+// input map, so callers may hold or extend it without mutating shared state.
 func mergeNoiseMaps(a, b map[string][]string) map[string][]string {
-	if len(b) == 0 {
-		return a
+	if len(a) == 0 && len(b) == 0 {
+		return nil
 	}
 	out := make(map[string][]string, len(a)+len(b))
 	for k, v := range b {
--- a/pkg/agent/proxy/integrations/http/match.go
+++ b/pkg/agent/proxy/integrations/http/match.go
@@ -1335,6 +1335,11 @@ func (h *HTTP) buildHTTPMismatchReport(request *http.Request, liveBody []byte, m
 		fieldDiffs = append(fieldDiffs, mismatch.JSONBodyDiffs(mockReq.Body, string(liveBody), ignore)...)
 	}
 
+	// Redact secret / obfuscated values out of the structured diffs before they
+	// are attached — they (and the Diff string derived from them) are persisted
+	// into the report YAML and the platform API.
+	redactFieldDiffs(fieldDiffs, closestMock.Noise)
+
 	b := mismatch.NewReport(mismatch.ProtocolHTTP, actualKey).
 		WithPhase(phase, candidateCount).
 		WithClosest(closestMock.Name, fieldDiffs).
--- a/pkg/agent/proxy/integrations/http/match.go
+++ b/pkg/agent/proxy/integrations/http/match.go
@@ -13,6 +13,7 @@ import (
 	"github.com/agnivade/levenshtein"
 	"go.keploy.io/server/v3/pkg"
 	"go.keploy.io/server/v3/pkg/agent/proxy/integrations"
+	"go.keploy.io/server/v3/pkg/agent/proxy/integrations/mismatch"
 	"go.keploy.io/server/v3/pkg/agent/proxy/integrations/util"
 	"go.keploy.io/server/v3/pkg/matcher"
 	"go.keploy.io/server/v3/pkg/models"
@@ -115,10 +116,25 @@ type req struct {
 	raw    []byte
 }
 
-func (h *HTTP) match(ctx context.Context, input *req, mockDb integrations.MockMemDb, headerNoise map[string][]string, schemaNoiseDetection bool, schemaNoiseStrict bool) (bool, *models.Mock, error) {
+// matchDiag captures where the match cascade stopped when no mock matched, so
+// the mismatch report can tell the user WHICH stage ruled everything out and
+// diff the live request against the candidates that were still alive there —
+// instead of the canned "headers or body differ".
+type matchDiag struct {
+	phase         string         // models.MatchPhase* constant
+	candidates    int            // HTTP mocks considered
+	schemaMatched []*models.Mock // candidates alive after schema match (nil if none)
+}
+
+// match returns (matched, mock, diag, err). diag is non-nil only when
+// matched is false and no error occurred. userBodyNoise carries the user's
+// test.globalNoise body bucket (root-relative dotted paths, lowercased) so
+// manual noise config participates in mock matching with the same vocabulary
+// as response assertions.
+func (h *HTTP) match(ctx context.Context, input *req, mockDb integrations.MockMemDb, headerNoise map[string][]string, userBodyNoise map[string][]string, schemaNoiseDetection bool, schemaNoiseStrict bool) (bool, *models.Mock, *matchDiag, error) {
 	for {
 		if ctx.Err() != nil {
-			return false, nil, ctx.Err()
+			return false, nil, nil, ctx.Err()
 		}
 
 		// Fetch HTTP mocks from BOTH pools:
@@ -147,12 +163,12 @@ func (h *HTTP) match(ctx context.Context, input *req, mockDb integrations.MockMe
 		perTestMocks, err := mockDb.GetPerTestMocksInWindow()
 		if err != nil {
 			utils.LogError(h.Logger, err, "failed to get per-test mocks")
-			return false, nil, errors.New("error while matching the request with the mocks")
+			return false, nil, nil, errors.New("error while matching the request with the mocks")
 		}
 		sessionMocks, err := mockDb.GetSessionMocks()
 		if err != nil {
 			utils.LogError(h.Logger, err, "failed to get session mocks")
-			return false, nil, errors.New("error while matching the request with the mocks")
+			return false, nil, nil, errors.New("error while matching the request with the mocks")
 		}
 		combined := make([]*models.Mock, 0, len(perTestMocks)+len(sessionMocks))
 		combined = append(combined, perTestMocks...)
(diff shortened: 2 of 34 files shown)


Aditya-eddy:
## Codex review

Codex completed the review and posted a [formal pull-request review](https://github.com/keploy/keploy/pull/4273#pullrequestreview-4492514445).

- Model: `gpt-5.5`
- Reasoning: `high`
- Job: `17b0457d-7cfb-4597-82d9-fd3c5c97aaa3`

pkg/agent/proxy/integrations/http/mismatch_render.go
```diff
@@ -0,0 +1,142 @@
...
+	// values are still redacted.
+	for _, k := range keys {
+		v := header[k]
+		if isSensitiveHeader(k) || isObfuscated(v) {
+			v = valueRedacted
+		}
+		fmt.Fprintf(&b, "\n%s: %s", textproto.CanonicalMIMEHeaderKey(k), v)
+	}
+	if body != "" {
+		fmt.Fprintf(&b, "\n\n%s", prettyJSONBody(body))
```
Aditya-eddy: **Blocker — Full request previews expose sensitive and unbounded payloads**

The complete body is copied into `ReceivedReq` unchanged. Query parameters are likewise included unchanged, and `Mock.Noise` is only checked against whole header values. Consequently credentials in JSON/form bodies or query parameters are printed, serialized into report YAML/API payloads, and retained in up to 512 pending errors. Redact sensitive query keys and body fields before constructing the report, including applying recorded obfuscation paths to the corresponding live fields. Also impose a byte limit on previews and field values so the count-bounded queue is actually memory-bounded.

pkg/service/replay/replay.go
```diff
@@ -1979,19 +2007,33 @@ func (r *Replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
...
+						// live HTTP agent transport returns a nil
+						// GetErrorChannel, so the channel-fed store above stays
+						// empty in instrument mode (local `keploy test -c`) —
+						// GetMockErrors is the one source that works on all
+						// transports. Fetched calls are also pushed into
+						// mockMismatchFailures so the end-of-run MOCKS MISMATCH
+						// SUMMARY table shows them; the store was already read
+						// for THIS test above (filtered by test ID), so this
+						// cannot double-count.
+						if mockErrors, err := r.instrumentation.GetMockErrors(runTestSetCtx); err == nil {
```
Aditya-eddy: **Major — Passing-test mismatches are never fetched**

Despite the comment saying this fetch occurs in every mode, it remains inside `if testStatus == Failed || Obsolete`. A miss during a passing test therefore remains in `pendingMockErrors`, so the advertised green-run mismatch summary cannot show it; it may instead be attached to a later failed test. Fetch and add mock errors after every simulation, independently of status, then conditionally attach them to `FailureInfo` where appropriate.

pkg/agent/proxy/proxy.go
```diff
@@ -2605,6 +2641,10 @@ func (p *Proxy) StartErrorDrain(ctx context.Context) {
 					}
 					if acc := p.activeTestErrors.Load(); acc != nil {
 						acc.add(err)
+					} else if isMockNotFoundErr(err) && p.pendingMockErrors.addBounded(err, maxPendingMockErrors) {
```
Aditya-eddy: **Major — Global pending queue cannot attribute errors to a test**

All misses outside an active capture window enter one session-wide queue, and the code confirms no caller opens such a window. `GetMockErrors` then assigns everything drained to whichever test calls it next, including startup/background traffic and late errors from the previous test. Wire `BeginTestErrorCapture`/`EndTestErrorCapture` into test boundaries or tag queued entries with test identity; a global FIFO cannot produce reliable per-test reports.

pkg/agent/proxy/integrations/http/match.go
```diff
@@ -1170,57 +1234,142 @@ func mergeReqBodyNoise(existing, detected map[string][]string) map[string][]stri
...
+		recordedQuery = mockQuery
+	}
+	fieldDiffs = append(fieldDiffs, mismatch.QueryParamDiffs(recordedQuery, request.URL.Query())...)
+	fieldDiffs = append(fieldDiffs, mismatch.HeaderKeyDiffs(mockReq.Header, request.Header, headerNoise)...)
+
+	// Body diffs: JSON bodies get field-level diffs excluding everything the
+	// matcher itself ignores (learned req_body_noise + user body noise).
+	if len(liveBody) > 0 && pkg.IsJSON([]byte(mockReq.Body)) && pkg.IsJSON(liveBody) {
+		ignore := mergeNoiseMaps(stripBodyPrefix(mockReq.ReqBodyNoise), userBodyNoise)
+		fieldDiffs = append(fieldDiffs, mismatch.JSONBodyDiffs(mockReq.Body, string(liveBody), ignore)...)
```
Aditya-eddy, line 1353: **Major — Reported body-noise paths target the wrong bucket**

`JSONBodyDiffs` emits paths such as `body.user.id`, while HTTP request matching now reads only `globalNoise.requestbody` with root-relative keys. Following the report's copy-paste guidance therefore modifies response noise and does not resolve the request mismatch. Emit request-specific paths or render an explicit configuration suggestion such as `requestbody: {user.id: []}` instead of claiming `body.user.id` is directly reusable.

pkg/service/replay/utils.go
```diff
@@ -876,27 +875,164 @@ func (tfs *TestFailureStore) PrintFailuresTable() {
...
+	if mockTarget != liveTarget {
+		dp.PushHeaderDiff(mockTarget, liveTarget, "url", nil)
+	}
+
+	// Headers: push only the keys whose values differ, mirroring the response
+	// header diff. Identical (including both-redacted) headers are skipped.
+	seen := map[string]bool{}
+	for k, mv := range mockHeaders {
+		seen[k] = true
+		if lv := liveHeaders[k]; mv != lv {
```
Aditya-eddy: **Minor — Empty-valued header differences disappear from the renderer**

Map lookup conflates an absent header with a present header whose value is empty. A missing recorded `X-Flag: ` header compares as `"" == ""`, while an extra empty live header is suppressed by `lv != ""`; either can be the exact key-presence mismatch that rejected the mock, yet the rendered diff is blank. Use two-value map lookups and report unseen keys regardless of their value.

Aditya-eddy (commented):
## Codex review

Completed with `gpt-5.5` using `high` reasoning.
## Summary
The PR adds structured mismatch diagnostics and shared rendering, but the current implementation can expose request secrets and misattribute or retain mismatches across tests. The new request-body noise guidance is also incompatible with the bucket consumed by matching.

## Verdict
Request changes — Sensitive-data exposure and incorrect mismatch attribution must be fixed before merge.

## Cross-repository checks
No external repository contract was required. I inspected the local HTTP agent transport, proxy error route, report serialization models, matcher renderer, and replay lifecycle. Per host constraints, no project code or tests were executed.

## Nits & lint
No additional lint or formatting issues found.

Aditya-eddy:
## Codex review

Codex completed the review and posted a [formal pull-request review](https://github.com/keploy/keploy/pull/4273#pullrequestreview-4492551500).

- Model: `gpt-5.5`
- Reasoning: `high`
- Job: `4e5ca48c-0c35-4e6f-b3d5-a3dc50c01290`

Aditya-eddy (commented):
## Codex review

Completed with `gpt-5.5` using `high` reasoning.
## Summary
The latest commit fixes the brace structure, header lookup, and request-body noise guidance. However, secrets still leak through structured diffs and several preview paths, while the global error queue still cannot reliably attribute misses to tests.

## Verdict
Request changes — Sensitive request data can still be persisted, and mock misses remain vulnerable to cross-test attribution.

## Cross-repository checks
No external repository contract was required. I traced the local agent transport, proxy queue, replay lifecycle, report serialization, and platform-facing models. Per host constraints, no project code or tests were executed.

## Nits & lint
No additional lint or formatting issues found.

pkg/agent/proxy/integrations/http/match.go
```diff
@@ -1170,57 +1234,142 @@ func mergeReqBodyNoise(existing, detected map[string][]string) map[string][]stri
...
+	}
+
+	b := mismatch.NewReport(mismatch.ProtocolHTTP, actualKey).
+		WithPhase(phase, candidateCount).
+		WithClosest(closestMock.Name, fieldDiffs).
```
Aditya-eddy, line 1364: **Blocker — Structured field diffs bypass all redaction**

`fieldDiffs` contains raw query and JSON values from `QueryParamDiffs`/`JSONBodyDiffs`. `WithClosest` stores these values and `Build` copies them into the human-readable `Diff`; both are serialized into report YAML and platform API payloads. Thus `password`, token, and noise-matching values leak even though `ClosestMockReq`/`ReceivedReq` are sanitized. Redact `Expected` and `Actual` before attaching the diffs, using sensitive path keys and `Mock.Noise`, then derive `Diff` only from that sanitized copy.

pkg/agent/proxy/integrations/http/mismatch_render.go
```diff
@@ -0,0 +1,212 @@
...
+		return body
+	}
+	return string(out)
+}
+
+func redactJSONValue(v interface{}, nc *util.NoiseChecker) {
+	switch t := v.(type) {
+	case map[string]interface{}:
+		for k, val := range t {
+			if s, ok := val.(string); ok {
```
Aditya-eddy, line 308: **Blocker — Request preview redaction still exposes common secret forms**

A sensitive key is redacted only when its value is a string. Values such as `{"password":1234}`, `{"secret":{"value":"x"}}`, and `{"tokens":["x"]}` remain visible; scalar array elements matching noise patterns are also never checked. Additionally, non-JSON form bodies are returned verbatim unless the entire body matches a noise regex, and a query parse error returns the raw, uncapped query. Redact the complete value whenever its key is sensitive regardless of type, inspect scalar array values for noise, parse supported form bodies or fail closed by redacting them, and never return raw query text on parse failure.

pkg/service/replay/replay.go
```diff
@@ -1969,30 +1997,43 @@ func (r *Replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
...
+					}
+					// Fetch in EVERY mode, not just non-instrument. The live HTTP
+					// agent transport returns a nil GetErrorChannel, so the
+					// channel-fed store above stays empty in instrument mode
+					// (local `keploy test -c`) — GetMockErrors is the one source
+					// that works on all transports. Fetched calls are also pushed
+					// into mockMismatchFailures so the end-of-run mock mismatch
+					// report shows them; the store was already read for THIS test
+					// above (filtered by test ID), so this cannot double-count.
+					if mockErrors, err := r.instrumentation.GetMockErrors(runTestSetCtx); err == nil {
```
Aditya-eddy: **Major — Draining after a test does not create a reliable test boundary**

`GetMockErrors` still drains a process-global asynchronous queue. The drain goroutine may enqueue the current test's miss after this call completes, causing the next test to receive it; late/background traffic has the same problem. This block is also reached only for normal non-streaming results, so simulation-error and streaming paths leave their misses queued. Introduce explicit test-scoped capture or tag errors with test identity at production time, and close/drain that scope on every exit path, including streaming and errors.

pkg/agent/proxy/integrations/http/mismatch_render.go
```diff
@@ -0,0 +1,212 @@
...
+// mock's Mock.Noise so obfuscated values stay redacted.
+func renderMockRequest(mockReq *models.HTTPReq, mockNoise []string) string {
+	if mockReq == nil {
+		return ""
+	}
+	header := make(map[string]string, len(mockReq.Header))
+	for k, v := range mockReq.Header {
+		header[k] = v
+	}
+	return renderRequestPreview(string(mockReq.Method), mockURLPath(mockReq.URL), urlParamsQuery(mockReq.URLParams), header, mockReq.Body, mockNoise)
```
Aditya-eddy: **Minor — Mock query rendering differs from the matcher’s query source**

The matcher falls back to query parameters parsed from `mockReq.URL` when `URLParams` is empty, but the rendered mock uses only `URLParams`. Older or alternate recordings that store the query solely in `URL` therefore show a false URL difference against the live request. Use the same URLParams-then-parsed-URL fallback on the rendered mock side.

Aditya-eddy:
## Codex review

Codex completed the review and posted a [formal pull-request review](https://github.com/keploy/keploy/pull/4273#pullrequestreview-4492694234).

- Model: `gpt-5.5`
- Reasoning: `high`
- Job: `c5d22900-0ef2-496e-9824-1f7eabef856f`

pkg/agent/proxy/proxy.go
```diff
@@ -2645,37 +2690,65 @@ func (p *Proxy) GetErrorChannel() <-chan error {
...
+			rawErrs = append(rawErrs, err)
+		default:
+			break drainLoop
 		}
 	}
 
+	// Then the per-test accumulator (errors the drain goroutine already routed
+	// into the active window). Drained AFTER the sweep so an item the goroutine
+	// moved off errChannel concurrently is still captured here.
+	if acc := p.activeTestErrors.Load(); acc != nil {
```
Aditya-eddy: **Major — GetMockErrors races with the asynchronous drain**

The channel sweep and accumulator drain are not atomic with `StartErrorDrain`. The drain goroutine can receive an error, pause before `acc.add`, then resume after this `acc.drain()` returns. The miss remains in the old accumulator and is discarded by the next `BeginTestErrorCapture`, so reporting is still timing-dependent. Serialize channel routing and window closure through one owner/control channel, or atomically close/swap the active accumulator and rendezvous with the drain before returning its contents.

pkg/service/replay/replay.go
```diff
@@ -2121,6 +2164,8 @@ func (r *Replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
 			streamProxyErrCtx, streamProxyErrCancel := context.WithCancel(runTestSetCtx)
 			go r.monitorProxyErrors(streamProxyErrCtx, testSetID, tc.Name)
 
+			r.beginTestErrorCapture(runTestSetCtx)
```
Aditya-eddy, line 2181: **Major — Streaming and simulation-error paths never drain or close the window**

This opens a streaming capture window, but Phase 2 never calls `GetMockErrors`. Both normal and streaming simulation-error branches also `continue` without draining. Their mismatches are therefore discarded by the next `Begin`, and the final active accumulator remains installed after the last test; `StartErrorDrain` then appends every subsequent proxy error to an unbounded slice until shutdown. Add an explicit end/drain operation and invoke it on every exit path, attaching collected calls before persisting the result. For streaming tests, close it only after the stream body has been consumed.

.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh
```diff
@@ -0,0 +1,131 @@
...
+shopt -s nullglob
+mock_files=( ./keploy/test-set-*/mocks.yaml )
+shopt -u nullglob
+if [ ${#mock_files[@]} -eq 0 ]; then
+    echo "::error::No recorded mocks found to mutate under ./keploy/test-set-*/"
+    cat record_logs.txt
+    exit 1
+fi
+for mf in "${mock_files[@]}"; do
+    sed -i 's#pokeapi\.co#pokeapi-mismatch.invalid#g' "$mf"
```
Aditya-eddy: **Major — The e2e mutation does not create a mock mismatch**

The sample's recorded request URLs are path-only (`/api/v2/...`), while `pokeapi.co` appears in mocked response bodies. This replacement therefore changes responses, not outgoing request matching. HTTP `SchemaMatch` also compares URL path rather than host, so even a full recorded URL host change would not reject the mock. Mutate a matcher-visible request property such as the recorded path, method, or query-key set, verify at least one request field was changed, and assert a specific `unmatched_calls` entry rather than accepting an unrelated response failure.

Aditya-eddy (commented):
## Codex review

Completed with `gpt-5.5` using `high` reasoning.
## Summary
The prior redaction and request-body noise findings are resolved. However, capture windows still lose errors under concurrency, leak on streaming/error paths, and the new e2e does not create a mock mismatch.

## Verdict
Request changes — Mock mismatches can still be dropped, and the added e2e does not exercise the intended behavior.

## Cross-repository checks
Cloned public `keploy/samples-go` under `/tmp/pr-keploy-keploy-4273` and verified the `http-pokeapi` mock request URLs are path-only while `pokeapi.co` occurs in response payloads. No project code or tests were executed per host constraints.

## Nits & lint
No additional lint or formatting issues found.

Aditya-eddy:
## Codex review

Codex completed the review and posted a [formal pull-request review](https://github.com/keploy/keploy/pull/4273#pullrequestreview-4492708263).

- Model: `gpt-5.5`
- Reasoning: `high`
- Job: `d6e29944-497a-44f4-9923-aaaa0f535b90`

pkg/agent/proxy/proxy.go
```diff
@@ -2603,9 +2644,20 @@ func (p *Proxy) StartErrorDrain(ctx context.Context) {
...
 						return
 					}
-					if acc := p.activeTestErrors.Load(); acc != nil {
-						acc.add(err)
-					} else {
+					// Route under captureMu so the window can't be swapped
+					// between the Load and the add. The accumulator is bounded
+					// too, so even a window left open (e.g. an error path that
+					// skipped GetMockErrors) can't grow without limit.
+					p.captureMu.Lock()
```
Aditya-eddy, line 2676: **Major — In-flight received errors can still miss the closing window**

`captureMu` prevents adding to a swapped-away accumulator, but it does not synchronize an error already received from `errChannel`. The drain goroutine can receive the current test's error, pause before this lock, then let `GetMockErrors` acquire the lock, find the channel empty, swap the window to nil, and return. The goroutine subsequently sees no active window and puts the error in `pendingMockErrors`, which the next `BeginTestErrorCapture` discards. Use a single owner loop with a close/drain control message or another barrier that confirms all previously received errors have been routed before closing the window.

.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh
```diff
@@ -0,0 +1,143 @@
...
+if grep "WARNING: DATA RACE" test_logs.txt; then
+    echo "::error::Race condition detected in test"
+    cat test_logs.txt
+    exit 1
+fi
+
+# Assert the mismatch surfaced. Primary signal: the console "MOCK MISMATCH"
+# report / per-call "Mock mismatch:" heading printed by the replayer.
+mismatch_reported=false
+if grep -qiE "MOCK MISMATCH|Mock mismatch:" test_logs.txt; then
```
Aditya-eddy: **Major — E2E accepts unrelated mismatch reports**

The path mutation now correctly forces a matcher-visible change, but this predicate accepts the generic mismatch banner from any stored failure. On failed runs, routine DNS misses are not filtered and can print the same banner; the YAML fallback likewise accepts any `unmatched_calls`, `match_phase`, or `closest_mock`. The suite can therefore pass while the mutated HTTP mismatch is still dropped. Assert a specific entry, such as `Mock mismatch: [HTTP] GET /api/v2/` in logs, or parse an `unmatched_calls` item whose `protocol` is `HTTP` and whose `actual_summary` contains `/api/v2/`.

Aditya-eddy (commented):
## Codex review

Completed with `gpt-5.5` using `high` reasoning.
## Summary
The redaction, request-body noise, and accumulator growth findings remain resolved; capture storage is bounded and streaming no longer opens an abandoned window. However, the capture race can still drop an in-flight error, and the e2e assertion can pass on an unrelated mismatch.

## Verdict
Request changes — Timing-dependent mismatch loss remains possible, and the e2e does not specifically verify the forced HTTP mismatch.

## Cross-repository checks
No new external contract check was required. The previously verified `keploy/samples-go` request-path assumption is consistent with the local HTTP matcher. No project code or tests were executed per host constraints.

## Nits & lint
- Comments in `Proxy` and `beginTestErrorCapture` still claim no caller exists or that streaming calls the helper; both are stale after this fix.

Aditya-eddy:
## Codex review

Codex completed the review and posted a [formal pull-request review](https://github.com/keploy/keploy/pull/4273#pullrequestreview-4492721387).

- Model: `gpt-5.5`
- Reasoning: `high`
- Job: `68c22330-01b1-461c-855a-81d0e1e20258`

Aditya-eddy (commented):
## Codex review

Completed with `gpt-5.5` using `high` reasoning.
## Summary
The redaction, request-body noise, bounded accumulator, streaming leak fix, and requested comment updates remain resolved. The rendezvous works on its successful path without deadlock, and the console e2e assertion is specific. However, timeout/full-channel paths can still lose misses, and the YAML fallback does not bind protocol and path to the same entry.

## Verdict
Request changes — The capture-window race remains possible when the bounded rendezvous cannot complete.

## Cross-repository checks
No new external contract check was required. The local HTTP matcher confirms `/api/v2/` is matcher-visible. Review was static only; project code and tests were not executed per host constraints.

## Nits & lint
- `activeTestErrors` and `StartErrorDrain` comments still say errors are discarded outside a window, although mock misses are retained.
- `maxPendingMockErrors` says the replayer drains per failed test, but it now drains after every normal test.

pkg/agent/proxy/proxy.go
```diff
@@ -2645,37 +2726,93 @@ func (p *Proxy) GetErrorChannel() <-chan error {
...
+	// strand an in-flight miss (the remaining race the previous fix left open).
+	// The app has already responded by now, so this test's pushes precede the
+	// marker. A bounded wait keeps a stalled drain from blocking replay.
+	if p.errDrainActive.Load() && p.errChannel != nil {
+		marker := &flushMarker{done: make(chan struct{})}
+		select {
+		case p.errChannel <- marker:
+			select {
+			case <-marker.done:
+			case <-time.After(2 * time.Second):
```
Aditya-eddy, line 2757: **Major — Failed rendezvous still closes the capture window**

If the channel is full, the marker is never enqueued; if acknowledgment exceeds two seconds, the wait expires. In both cases execution still swaps `activeTestErrors` to nil. Errors queued before the attempted marker can then be routed into `pendingMockErrors` and discarded by the next `BeginTestErrorCapture`, preserving the original loss race. Only close the window after marker acknowledgment. On enqueue/ack timeout, return an error while retaining the window and ensure the next begin cannot replace it without completing the flush.

.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh
```diff
@@ -0,0 +1,153 @@
...
+
+# Equivalent specific check on the test-set report yaml: an unmatched_calls
+# entry whose protocol is HTTP and whose actual_summary references /api/v2/.
+shopt -s nullglob
+reports=( ./keploy/reports/test-run-*/test-set-*-report.yaml )
+shopt -u nullglob
+for rf in "${reports[@]}"; do
+    # Limit the scan to each report's unmatched_calls block and require both an
+    # HTTP protocol and an /api/v2/ actual_summary within it.
+    if awk '/unmatched_calls:/{f=1} f' "$rf" | grep -qE "protocol: ?\"?HTTP" \
```
Aditya-eddy: **Major — YAML fallback does not verify one specific unmatched call**

These independent greps can match `protocol: HTTP` in one unmatched-call item and `/api/v2/` in another item or later report content. The fallback can therefore pass without an HTTP `/api/v2/` unmatched call. Parse YAML or scope the scan to individual list items and require both fields in the same item.

Aditya-eddy:
## Codex review

Codex completed the review and posted a [formal pull-request review](https://github.com/keploy/keploy/pull/4273#pullrequestreview-4492732713).

- Model: `gpt-5.5`
- Reasoning: `high`
- Job: `ddaf18a9-bb40-4376-9c6b-f0db720b9057`

pkg/service/replay/replay.go
```diff
@@ -1599,6 +1627,8 @@ func (r *Replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
 			testCaseProxyErrCtx, testCaseProxyErrCancel := context.WithCancel(runTestSetCtx)
 			go r.monitorProxyErrors(testCaseProxyErrCtx, testSetID, testCase.Name)
 
+			r.beginTestErrorCapture(runTestSetCtx)
```
Aditya-eddy, line 1638: **Major — Early exits leave the capture window unfinalized**

This opens a capture window, but `GetMockErrors` is reached only after successful simulation, valid response type assertions, comparison, and result construction. The simulation-error path at line 1637 and the invalid-response paths `continue` without draining it. A subsequent `BeginTestErrorCapture` carries those misses forward and attributes them to the next test; if this is the final test, they are never surfaced. Finalize and attach captured errors on every exit, preferably by wrapping one test iteration in a helper with a guaranteed cleanup/finalization step.

Aditya-eddy (commented):
## Codex review

Completed with `gpt-5.5` using `high` reasoning.
## Summary
The rendezvous failure handling and e2e item-scoped assertions are fixed; redaction, request-body noise, bounded storage, streaming handling, and comments remain resolved. One non-streaming lifecycle gap still loses or misattributes mismatch reports on early exits.

## Verdict
Request changes — Capture windows are not finalized on every non-streaming test exit path.

## Cross-repository checks
No external repository contract was required. Static review only; project code and tests were not executed. The latest proxy rendezvous logic and e2e YAML/console assertions were inspected locally.

## Nits & lint
No additional lint or formatting issues found.

Aditya-eddy:
## Codex review

Codex completed the review and posted a [formal pull-request review](https://github.com/keploy/keploy/pull/4273#pullrequestreview-4492741144).

- Model: `gpt-5.5`
- Reasoning: `high`
- Job: `2224e30d-1bc9-4f1a-b18d-d25933e45083`

pkg/service/replay/replay.go
Aditya-eddy, line 1638: **Major — Nil matcher results still leave the capture window open**

The window is opened here, but HTTP matching has reachable `return false, nil` paths when template rendering or JSON marshaling fails. That reaches the existing `test result is nil` branch and breaks the iteration without calling `attachMockErrors`, leaving the window active and losing its mismatch from the test result. Put finalization in a guaranteed per-iteration cleanup path, and create/persist a failed result for matcher-internal failures so captured calls have somewhere to attach.

pkg/service/replay/replay.go
```diff
@@ -3636,6 +3702,42 @@ func (r *Replayer) copyDirContents(src, dst string) error {
...
+}
+
+// attachMockErrors drains the per-test mock-error capture window (closing it)
+// and records any unmatched outgoing calls onto the result and the end-of-run
+// summary store. It must run on EVERY test-iteration exit — including the early
+// simulation-error / invalid-response returns — so the window is finalized for
+// THIS test and a miss is never carried forward to the next test or lost.
+func (r *Replayer) attachMockErrors(ctx context.Context, testSetID, testCaseName string, result *models.TestResult) {
+	mockErrors, err := r.instrumentation.GetMockErrors(ctx)
+	if err != nil {
```
Aditya-eddy, line 3763: **Major — GetMockErrors failures silently skip window closure**

A transient HTTP/agent error causes this helper to return without closing the remote capture window or reporting the failure. The next `BeginTestErrorCapture` can then carry the prior window's late misses into the next test, while the current result is persisted without them. Add an explicit idempotent end/abort capture operation and invoke it even when draining fails, or make finalization retry and propagate failure before the result is persisted.

Aditya-eddy (commented):
## Codex review

Completed with `gpt-5.5` using `high` reasoning.
## Summary
The named early exits now invoke the new drain-and-attach helper, and the previously resolved redaction, noise hint, bounded storage, rendezvous, streaming, e2e, and comment fixes remain intact. Two lifecycle gaps still allow a capture window or mismatch result to escape finalization.

## Verdict
Request changes — Capture finalization is still not guaranteed on matcher failures or GetMockErrors transport errors.

## Cross-repository checks
No external repository contract was required. Statically inspected the local replay lifecycle, in-process and HTTP agent transports, proxy drain/rendezvous logic, mismatch redaction, requestbody noise plumbing, and e2e assertions. Project code and tests were not executed per host constraints.

## Nits & lint
No additional lint or formatting issues found.

pkg/service/replay/replay.go
```diff
@@ -1969,32 +2005,40 @@ func (r *Replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
...
+					}
+					// UnmatchedCalls is fetched/drained for EVERY test, not just
+					// failed/obsolete ones. Two reasons: (1) a miss during an
+					// otherwise-passing test must still surface; (2) GetMockErrors
+					// drains a process-global queue, so fetching only on
+					// failed/obsolete tests would let a passing test's miss linger
+					// and be misattributed to the next failed test. Draining per
+					// test keeps each test's misses with that test. The channel
+					// source below is already test-tagged.
+					for _, f := range r.mockMismatchFailures.GetFailuresForTestCase(testSetID, testCase.Name) {
```
Aditya-eddy: **Major — Concurrent channel monitor can bypass result attachment**

`monitorProxyErrors` and `StartErrorDrain` compete for the same proxy error channel. If the monitor receives a miss but adds it to `mockMismatchFailures` after this snapshot, `GetMockErrors` cannot return it because the monitor already consumed it. The summary may eventually contain the miss, but this test's persisted `FailureInfo.UnmatchedCalls` will not. Give the channel a single owner, or stop and join the monitor before taking this snapshot and draining the capture window.

Aditya-eddy:
## Codex review

Codex completed the review and posted a [formal pull-request review](https://github.com/keploy/keploy/pull/4273#pullrequestreview-4492835370).

- Model: `gpt-5.5`
- Reasoning: `high`
- Job: `c158419f-408d-4be7-a36c-556b9f240c95`

Aditya-eddy (commented):
## Codex review

Completed with `gpt-5.5` using `high` reasoning.
## Summary
The removed monitor path was redundant: production replay is wired only to AgentClient, whose GetErrorChannel returns nil. GetMockErrors preserves all ten rich fields, but capture finalization still has gaps that can lose or misattribute mismatches.

## Verdict
Request changes — Capture windows are not finalized reliably on all exits, and streaming tests drop mismatch reports entirely.

## Cross-repository checks
Locally confirmed replay production wiring uses only HTTP AgentClient, GetErrorChannel is nil, and GetMockErrors copies/decodes all rich UnmatchedCall fields. keploy/api-server and keploy/k8s-proxy were unavailable without authentication, so downstream storage compatibility remains unverified. Static review only; no project code or tests executed.

## Nits & lint
No additional lint or formatting issues found.

pkg/service/replay/replay.go
```diff
@@ -1596,19 +1623,19 @@ func (r *Replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
 
 			started := time.Now().UTC()
 
-			testCaseProxyErrCtx, testCaseProxyErrCancel := context.WithCancel(runTestSetCtx)
-			go r.monitorProxyErrors(testCaseProxyErrCtx, testSetID, testCase.Name)
+			r.beginTestErrorCapture(runTestSetCtx)
```
Aditya-eddy, line 1638: **Major — Nil matcher results leave the capture window open**

The window opens before comparison, but HTTP matching has reachable `return false, nil` paths. The subsequent `testResult == nil` branch breaks without calling `attachMockErrors`, so captured misses can be carried into another test or never persisted. Finalize capture in guaranteed per-iteration cleanup and persist a failed result for matcher-internal failures.

pkg/service/replay/replay.go
```diff
@@ -3636,50 +3666,39 @@ func (r *Replayer) copyDirContents(src, dst string) error {
...
-	errorChannel := r.instrumentation.GetErrorChannel()
-	if errorChannel == nil {
-		r.logger.Debug("Proxy error channel is nil, skipping error monitoring")
+// attachMockErrors drains the per-test mock-error capture window (closing it)
+// and records any unmatched outgoing calls onto the result and the end-of-run
+// summary store. It must run on EVERY test-iteration exit — including the early
+// simulation-error / invalid-response returns — so the window is finalized for
+// THIS test and a miss is never carried forward to the next test or lost.
+func (r *Replayer) attachMockErrors(ctx context.Context, testSetID, testCaseName string, result *models.TestResult) {
+	mockErrors, err := r.instrumentation.GetMockErrors(ctx)
```
Aditya-eddy, line 3762: **Major — GetMockErrors failures do not close capture**

When the HTTP request or response decoding fails, this helper returns without closing or aborting the agent-side window. The next begin can carry the previous test's errors forward, causing cross-test attribution. Add an idempotent end/abort endpoint, or retry finalization and propagate failure before persisting the result.

pkg/service/replay/replay.go
```diff
@@ -2117,9 +2132,9 @@ func (r *Replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
 				break
 			}
 
-			// Proxy Monitor: Start a per-test proxy error monitor.
-			streamProxyErrCtx, streamProxyErrCancel := context.WithCancel(runTestSetCtx)
-			go r.monitorProxyErrors(streamProxyErrCtx, testSetID, tc.Name)
+			// NOTE: no BeginTestErrorCapture here — the streaming (Phase 2) path
```
Aditya-eddy: **Major — Streaming tests drop every mismatch field**

Deferred streaming tests intentionally neither open a capture window nor call `GetMockErrors`. Any streaming-test mock miss remains pending and never reaches `FailureInfo.UnmatchedCalls`, dropping `Protocol`, `ActualSummary`, `ClosestMock`, `Diff`, `NextSteps`, `MatchPhase`, `CandidateCount`, `FieldDiffs`, `ClosestMockReq`, and `ReceivedReq`. Open capture before simulation and finalize it after stream consumption on every streaming exit path.

Aditya-eddy:
## Codex review

Codex completed the review and posted a [formal pull-request review](https://github.com/keploy/keploy/pull/4273#pullrequestreview-4492868668).

- Model: `gpt-5.5`
- Reasoning: `high`
- Job: `6983d51b-d089-47e6-ba35-8db213f676de`

pkg/service/replay/replay.go
```diff
@@ -2035,10 +2050,30 @@ func (r *Replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
...
 				utils.LogError(r.logger, nil, "test result is nil")
+				// Matcher returned no result (e.g. a comparison path returning
+				// (false, nil)). Same as above: finalize the window + persist a
+				// failed result so the miss surfaces and can't carry forward.
+				failedResult := r.CreateFailedTestResult(testCase, testSetID, started, "internal error: comparison returned no result")
+				r.attachMockErrors(runTestSetCtx, testSetID, testCase.Name, failedResult)
+				if insErr := r.reportDB.InsertTestCaseResult(runTestSetCtx, testRunID, testSetID, failedResult); insErr != nil {
+					utils.LogError(r.logger, insErr, "failed to insert failed test case result for nil comparison result")
+				}
+				currentFailures++
```
Aditya-eddy: **Major — Nil-result fallbacks double-count failures**

The outcome counters were already updated at lines 1850-1864 before reaching either nil fallback. A `(false, nil)` comparison increments `currentFailures` there and again here. The `testCaseResult == nil` branch similarly adds a failure without reversing the previously recorded success/obsolete/failure. Streaming repeats this pattern before its nil fallback. Persist the failed result, but replace the previously counted outcome rather than incrementing again.

pkg/service/replay/replay.go
```diff
@@ -2391,13 +2430,28 @@ func (r *Replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
...
 				if loopErr != nil {
 					utils.LogError(r.logger, loopErr, "failed to save streaming test result")
 					break
 				}
 			} else {
 				utils.LogError(r.logger, nil, "streaming test result is nil")
+				// Finalize the window + persist a failed result (same rationale as
+				// the non-streaming nil-result branch) so a streaming miss surfaces
+				// and can't carry forward.
+				failedResult := r.CreateFailedTestResult(tc, testSetID, started, "internal error: streaming comparison returned no result")
```
Aditya-eddy: **Major — Streaming mismatch can panic before the nil cleanup**

This fallback is not reached when `hadStreamingMismatch` is true and `CompareHTTPResp` returns `(false, nil)`, which is possible through the matcher's existing error paths. The code dereferences `testResult.BodyResult` at line 2331 first, causing a panic and leaving the capture window unfinalized. Check `testResult == nil` immediately after `CompareHTTPResp`, before any dereference, then finalize and persist the failed result.

Aditya-eddy (commented):
## Codex review

Completed with `gpt-5.5` using `high` reasoning.
## Summary
The capture-window reset and normal rendezvous behavior are correct, and GetMockErrors failures are now logged. Nil results are persisted once, but their counters are incremented twice, and one streaming nil-result path can panic before finalization.

## Verdict
Request changes — Nil-result handling still corrupts counts and can leave a streaming capture window unfinalized.

## Cross-repository checks
No cross-repository contract check was required for this confirmation pass. Static inspection only; project code and tests were not executed. The read-only partial clone prevented fetching missing Git objects for `git show`, so verification used the checked-out files and `origin/main...HEAD` diff.

## Nits & lint
- The comment at replay.go:3715-3717 still says streaming does not open or fetch a capture window.
- The proxy.go:2807 comment says the next begin carries leftovers forward, but BeginTestErrorCapture now discards them.

Aditya-eddy:
## Codex review

Codex completed the review and posted a [formal pull-request review](https://github.com/keploy/keploy/pull/4273#pullrequestreview-4492877140).

- Model: `gpt-5.5`
- Reasoning: `high`
- Job: `3f20a6ab-8986-4313-82c6-53a5b134ef00`

Aditya-eddy (commented):
## Codex review

Completed with `gpt-5.5` using `high` reasoning.
## Summary
Commit 628ffcc resolves the prior findings. Nil-result paths are counted once, streaming checks nil before dereferencing, and each explicit post-capture exit calls the finalizer exactly once. Static review only; no project code or tests were executed.

## Verdict
Approve — The requested counter, nil-safety, capture-finalization, and comment fixes are correct.

## Cross-repository checks
No cross-repository check was required. Reviewed the checked-out source and origin/main...HEAD diff; the partial read-only clone could not materialize missing objects for git show.

## Nits & lint
No additional lint or formatting issues found.

slayerjain (commented):
## Review — `feat(mismatch): unified mock-mismatch reporting + whole-mock side-by-side diff`

Reviewed the net delta on top of the #4271 foundation (`9ae829d..HEAD`), built it, ran the unit suite + the `-race` capture-window test (all green), and ran a multi-agent adversarial pass over the high-risk surfaces. This is strong, careful work — especially the concurrency. Approving in spirit; there's **one CI blocker** and a **redaction/security cluster** I'd want closed before merge, plus a few nits.

I personally verified the two scariest claims in the PR description and they both hold:
- **`monitorProxyErrors` really was dead** — `commonServices.Instrumentation` is always `*http.AgentClient`, whose `GetErrorChannel()` returns `nil`, so the consumer hit its nil guard everywhere. The removal (and the `AddProxyErrorForTest`/`GetFailuresForTestCase` feeders) is clean, no remnants.
- **The per-test capture window is sound.** The flush-marker FIFO rendezvous, the bounded accumulators, `captureMu` serialization, and the discard-on-reopen attribution choice are all correct, and the one-`begin`/one-`attach` invariant holds across every non-streaming exit path. Nicely done.

The `body`→dedicated-`requestbody` decoupling is the right call — it directly fixes the strict-mode "response noise silently weakens request matching" hole. 👍

---

### 🔴 Blocker (CI will go red)

**1. Non-streaming e2e has no replay-version gate → `record_build_replay_latest` cell fails deterministically.**
`.github/workflows/test_workflow_scripts/golang/mock_mismatch/golang-linux.sh` requires the `Mock mismatch: [HTTP] …` log + an `unmatched_calls` report entry (success is inverted: `exit 1` if absent). But the `latest` matrix config replays with the published release binary, which predates this whole feature, so it can never produce that output. Your **streaming sibling already gates this** (`case "${REPLAY_BIN:-}" in */build/keploy|*/build-no-race/keploy) ;; *) … exit 0`); the non-streaming script is missing the identical guard. Add the same case-gate at the top.

### 🟠 Should fix before merge

**2. Test-set-scoped `requestbody` noise is silently dropped.** `LeftJoinNoise` (`pkg/service/replay/utils.go:274-296`) hard-codes only `body` and `header`, so `test.globalNoise.test-sets.<id>.requestbody` never merges — only the global bucket works. Make `LeftJoinNoise` bucket-agnostic (iterate every section present in `tsNoise`) so `requestbody` (and any future bucket) joins like the others.

**3. Redaction completeness — secret leak into the persisted report + dashboard.** `mismatch_render.go` is the *sole* redaction layer before `ClosestMockReq`/`ReceivedReq` reach report YAML, the console, and the platform API. It's a key-name denylist + the mock's `Noise` regexes as backstop. Two gaps let a real credential through:
   - A secret under a **non-denylisted key** (e.g. `session_state`, `assertion`, a bearer token in a custom header) is only caught if `Mock.Noise` happened to obfuscate it — otherwise it renders verbatim (`mismatch_render.go:22-34, 248-276`).
   - **Non-JSON / non-form bodies** (SOAP/XML, `text/plain`, GraphQL, proto) skip field-level redaction entirely and are emitted in full unless `IsNoisy(whole body)` matches, which it ~never does (`redactBody`, `mismatch_render.go:223-243`).

   Add a value-shape backstop (`Bearer `/`Basic `/`eyJ`-JWT/high-entropy) applied to *all* header/query/body values and to opaque bodies, so redaction is defense-in-depth rather than a key-name allowlist.

**4. The 276-line redaction renderer has zero unit tests, and neither e2e asserts redaction.** This is a brand-new secret-leak boundary into on-disk YAML + the platform API, and the only protection for non-keyword secrets is a single untested `IsNoisy` branch. A regression here leaks silently. Add a table-driven test over `redactBody`/`redactJSONValue`/`redactQuery`/`renderRequestPreview`: keyword keys in header/query/body, the noise-value path on a non-keyword key, form-urlencoded, nested JSON, and a non-JSON noisy body — each asserting the secret never appears.

### 🟡 Nits / polish

- **Duplicate summary rows across retries** (`AddUnmatchedCallForTest`, `utils.go`). It appends unconditionally and the store is per-run, so a test re-run under flaky-check / must-pass / retry-passing appends the same miss each pass → the MOCKS MISMATCH SUMMARY shows N copies. Dedup at the writer on `(TestSetID, TestID, report signature)`.
- **Over-broad `"key"` token** (`sensitiveKeyTokens`, `mismatch_render.go:22`). Substring `Contains` redacts benign fields literally named `key` (very common as a cache/lookup field), `keyword`, `monkey`, `author`, etc. — hiding the exact value the user needs to diagnose the miss. Use token-boundary matching, or replace bare `key` with `api_key`/`access_key`/`secret_key`.
- **Active-window 512 cap drops misses with no log** (`proxy.go:2679`). The `acc.addBounded` return is ignored in the active-window branch (the pending branch does log). Add the same throttled debug/one-time-truncation log.
- **Comment casing** — `decode.go:164,173` says `requestBody` (camelCase) but the lookup key and user hint are `requestbody`. Align the comment.
- **`GetMockErrors` rendezvous comment overclaims** (`proxy.go:2746`). It's a guarantee only for synchronous request/response; an app doing fire-and-forget egress *after* responding can land a miss after the marker (then discarded by the next `BeginTestErrorCapture`) → silent under-report. Strictly better than before, but soften the comment to state the assumption rather than assert it unconditionally.
- **`TestCaptureWindow_NoBleedAcrossTests`** runs without `StartErrorDrain`, so it exercises the legacy sweep path, not the production rendezvous path it claims to guard (`RendezvousNoLoss` does cover the live goroutine). Call `StartErrorDrain` so the no-bleed assertion binds to the real path.
- **Mixed-bucket next-steps hint** (`mismatch.go:169-178`) lists all drifted paths but only offers the `requestbody` remedy for the body subset; a co-occurring `query.*` drift is named without a matching remedy. Scope the named list to body paths when offering the requestbody bucket.
- **`requestbody` has no e2e coverage** — the http suite injects `body` noise (which, post-move, no longer affects request matching). Worth one case that records a dynamic request-body field, noises it via `requestbody`, mutates only it, and asserts a pass.
- **Defensive nil-result branch** (`replay.go:2064-2076`) forces `Status=Failed` while the outcome may already have been counted OBSOLETE at ~1850 — set the persisted status from the computed `testStatus`, or note why the branch is unreachable.

### Coordination
- Docs PR keploy/docs#878 (mine) describes the `body` bucket flowing into request matching — needs updating to `requestbody` to match this design. I'll handle that on my side once we settle the bucket name.
- Model additions (`ClosestMockReq`/`ReceivedReq`) are additive with `omitempty`, so cross-repo consumers (api-server#1751, k8s-proxy#521) stay backward-compatible — good.

Nice PR overall. The blocker + the two redaction items are the ones I'd gate merge on; the rest are cleanups.

— review assisted by a multi-agent adversarial pass; load-bearing claims (dead-path removal, capture-window invariants, CI gate, `LeftJoinNoise`, duplicate rows) verified by hand against the branch.

Aditya-eddy:
## Codex review

Codex completed the review and posted a [formal pull-request review](https://github.com/keploy/keploy/pull/4273#pullrequestreview-4502864541).

- Model: `gpt-5.5`
- Reasoning: `high`
- Job: `4067c52e-8506-4cfd-9543-6555412ce867`

pkg/agent/proxy/integrations/http/match.go
```diff
@@ -1170,57 +1234,158 @@ func mergeReqBodyNoise(existing, detected map[string][]string) map[string][]stri
...
+// and returns a structured field-level diff report. liveBody is the live
+// request body (may be nil for body-less requests); headerNoise/userBodyNoise
+// are the same noise sets the matcher used, so the report never flags a field
+// the matcher would have ignored. diag (from match()) tells the builder how
+// far the cascade got and which candidates were still alive, so the diff is
+// computed against a candidate the matcher actually considered close — not
+// just the lowest-Levenshtein path. If httpMocks is nil, it fetches from
+// mockDb; otherwise it uses the pre-fetched slice to avoid a redundant read.
+func (h *HTTP) buildHTTPMismatchReport(request *http.Request, liveBody []byte, mockDb integrations.MockMemDb, httpMocks []*models.Mock, headerNoise, userBodyNoise map[string][]string, diag *matchDiag) *models.MockMismatchReport {
+	actualKey := request.Method + " " + request.URL.Path
```
Aditya-eddy, line 1254: **Major — Nil URL still panics in HTTP mismatch reporting**

The destination fallback checks `request.URL != nil`, but `buildHTTPMismatchReport` dereferences `request.URL.Path` before that guard when building `actualKey`, and later dereferences `request.URL` again for path diffs, query diffs, `pickClosestCandidate`, and `renderLiveRequest`. `decode.go` also formats the mock-miss error with `request.URL.Path` after this call. If a request reaches this path with a nil URL, the agent panics instead of emitting the mismatch report. Compute a nil-safe path/query once (for example empty path/query when `request.URL == nil`), use those values everywhere in the report builder and `baseErr`, and update `renderLiveRequest`/`pickClosestCandidate` to tolerate nil URL. Add a unit test that calls the mismatch builder with `&http.Request{Method: "GET", Host: "api.example.com"}` and a nil `URL` to cover the requested guarantee.

Aditya-eddy (commented):
## Codex review

Completed with `gpt-5.5` using `high` reasoning.
## Summary
The destination field is added consistently to the local models, builder, HTTP WARN log, proxy-to-report conversion, summary-store conversion, and console heading. The core plumbing looks additive and should not affect non-mismatch behavior, but the new nil-URL fallback is incomplete: the HTTP mismatch-report path can still panic before or after computing the destination. I did not run code or tests per the host constraints.

## Verdict
Request changes — The requested nil-safety guarantee is not met for HTTP mismatch reports.

## Cross-repository checks
Checked the local PR diff and requested files only. Attempted to inspect `keploy/api-server` and `keploy/k8s-proxy` through the GitHub connector, but both returned 404/unavailable in this environment, so downstream rendering/storage of the new `destination` field is unverified. The local serialized field is additive with `omitempty`, so no local wire-breaking change was found.

## Nits & lint
No additional lint or formatting issues found.

Issue #4275, opened 2026-06-15
Multi-tenant keploy-agent: per-app isolation (user-space foundation)


ayush3160:
## Summary

Make the keploy-agent **multi-tenant**: one agent process serving **many applications concurrently**, with every piece of per-app state isolated by an app/session key (config, mocks, channels, static-dedup, capture filters, mode). Today the agent is structurally single-app — the proxy session, mock manager, syncMock, `clientMocks`, dedup, and mode are all process-global / hardcoded to a single key.

This issue tracks the **OSS user-space foundation**, which is the prerequisite for both deployment models:
- **Sidecar/native** — attribution via `/proc` ancestry (container/netns mode in v1).
- **DaemonSet (enterprise)** — the eBPF capture is already multi-tenant (gate admits all target pods, each event carries `kernelPid`); it only needs this user-space layer to route per-app, fed by an authoritative PID→session resolver.

## Root cause

The data path carries no app/session key. `pkg/agent/proxy/proxy.go` holds a single `session`/`mockManager`; `pkg/service/agent/agent.go` stores mocks at hardcoded `clientMocks[uint64(0)]`; `pkg/agent/proxy/syncMock` is a package-global singleton with one output channel; record-vs-replay `mode` is process-wide.

## Design — one keyed user-space layer

Introduce an `AppContext` (per app/session key) that owns everything that is a singleton today, held in an `AppRegistry` keyed by a **string session key**. The key flows on `context.Context` (aligning with the context-keyed `syncMock` approach). Attribution is pluggable: `/proc` walk (native) or an authoritative resolver (daemonset). The kernel already stamps `kernelPid` into `redirect_proxy_map`; we surface it through `DestInfo.Get`.

Mixed modes (app A records while app B replays) fall out for free once `Mode` is per-app.

## Phases (this issue = OSS user-space)

- [ ] **Foundation** — per-app session key type + `context` helpers; surface `KernelPid` on `agent.NetworkAddress` (comm.go).
- [ ] **AppContext + AppRegistry** — skeleton owning session/mockManager/dnsCache/recordedDNSMocks/errChannel + `Resolve(kernelPid)`.
- [ ] **De-globalize `syncMock`** — `New()` + context-keyed managers; migrate emit call sites; per-app dedup queue.
[...]

PR #4276, opened 2026-06-15
feat(agent): multi-tenant foundation — per-app session key + AppContext/AppRegistry

ayush3160:
## What

First slice of the multi-tenant keploy-agent work (#4275): the **user-space foundation** that lets one agent process serve many applications, keyed by an opaque per-app session key. **Skeleton only — behaviour-neutral.** Not yet wired into the proxy hot path; `DefaultAppKey` reproduces today's single-app behaviour exactly.

## Changes

- **`agent.AppKey`** + `WithAppKey` / `AppKeyFromContext` / `AppKeyOrDefault` context helpers (`pkg/agent/appkey.go`). Opaque string key so the native (`/proc`) and daemonset (controller) attribution sources share one identity type. `DefaultAppKey` = single-tenant fallback.
- **Surface the originating kernel TGID** on `agent.NetworkAddress.Pid`, populated from `DestInfo.KernelPid` in `hooks/linux/comm.go` (it was previously dropped). This is the attribution key for routing a connection to an app.
- **`proxy.AppContext`** — owns the per-app state that is a `Proxy` singleton today: session, mockManager, dnsCache, `recordedDNSMocks` (DNS static-dedup), errChannel, activeTestErrors, graceful-shutdown flag, pcap lifecycle, and the per-app syncMock manager slot.
- **`proxy.AppRegistry`** — `AppKey → AppContext` store + a pluggable `PIDResolver` (default `/proc`-ancestry attribution for the native path; the daemonset overrides with an authoritative resolver) backed by a `byPID` cache.

## Tests

- `appkey_test.go` — context round-trip + default-when-absent + nil-ctx guard.
- `appcontext_test.go` — registry `GetOrCreate` idempotency, `Resolve` cache-then-resolver (resolver hit once), `Delete` evicts the PID cache.

## Scope / follow-ups (tracked in #4275)

De-globalize `syncMock`; thread the key through the proxy hot path (`Record/Mock/SetMocks/...` + `handleConnection`); per-app `clientMocks`/`tcChan`/routes middleware in the service; client `X-Keploy-App-Id` + register handshake. Enterprise per-feature isolation and the daemonset bridge are separate.

Draft until the dependent slices land.

### DIFF
--- a/pkg/agent/appkey.go
+++ b/pkg/agent/appkey.go
@@ -1,0 +1,57 @@
+package agent
+
+import "context"
+
+// AppKey is the per-application (per-recording-session) identity that keys all
+// per-app state in a multi-tenant agent. One agent process can serve many
+// applications concurrently; every piece of state that used to be a process
+// singleton (proxy session, mock manager, syncMock output, dedup, mode, etc.)
+// is partitioned by this key.
+//
+// It is deliberately an opaque string so the two attribution sources can share
+// one identity type:
+//   - sidecar/native: derived from the app (e.g. app name or a hash of
+//     appName+clientPID).
+//   - daemonset: the per-session key the controller resolves from a target's
+//     {namespace, deployment, testSetID}.
+//
+// The zero value (DefaultAppKey) is the single-tenant / sidecar fallback: when
+// no key is present on the context, callers operate on the default app so the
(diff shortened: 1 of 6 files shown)


ayush3160:

PR #4277, opened 2026-06-15
feat(syncMock): New() + context-carried per-app managers

ayush3160:
## What

Second slice of multi-tenant keploy-agent (#4275). De-globalizes `syncMock` so each application can own an independent mock-window manager instead of sharing the single process singleton. **Mechanism only — behaviour-neutral.**

Stacked on #4276 (base = `feat/multi-tenant-agent-userspace`).

## Changes

- **`New()`** constructs a fresh, independent `SyncMockManager` (own buffer, output channel, `firstReqSeen` boundary, resolved-window ring). The package global is now `New()` called once at load — the single-tenant / sidecar fallback.
- **`NewContext(ctx, m)` / `FromContext(ctx)`** carry the per-app manager on the context. The connection router will stamp the owning app's manager; parsers resolve it back when emitting mocks without threading it through every `RecordOutgoing` signature.
- `FromContext` **falls back to the global `instance`** when no manager is stamped, so migrating an emit site from `Get()` to `FromContext(ctx)` is behaviour-neutral until a per-app manager is actually present.

## Tests

`context_test.go` — `New()` returns distinct managers (not the global); `FromContext` falls back to the global with no/nil ctx and returns the stamped manager when present.

## Follow-up (same epic)

Per-app manager creation in `AppContext` + migrating the emit call sites (`http`, `mysql`, `generic`, `dns`, `supervisor`, `incoming`, `conn`) to `FromContext(ctx)` land with the proxy hot-path wiring slice. Per-app `DedupQueue` moves onto the manager there too.

### DIFF
--- a/pkg/agent/proxy/syncMock/context.go
+++ b/pkg/agent/proxy/syncMock/context.go
@@ -1,0 +1,44 @@
+package manager
+
+import (
+	"context"
+
+	"go.keploy.io/server/v3/pkg/models"
+)
+
+// New constructs a fresh, independent SyncMockManager. In a multi-tenant agent
+// each application owns its own manager (its own mock buffer, output channel,
+// firstReqSeen boundary and resolved-window ring) so one app's recording can
+// never bleed into another's. The process-global singleton (Get) is now just
+// New() called once at package load — the single-tenant / sidecar fallback.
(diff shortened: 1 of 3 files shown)


ayush3160:

PR #4278, opened 2026-06-15
feat(proxy): per-app session + mockManager via request context

ayush3160:
## What

Third slice of multi-tenant keploy-agent (#4275). Moves the proxy's single `session` + `mockManager` into the per-app `AppContext`, keyed by `agent.AppKey` resolved from the request context. **Behaviour-neutral** — with no key on ctx everything resolves to `DefaultAppKey`, reproducing the single-app path.

Stacked on #4277 (base = `feat/mt-syncmock-context`).

## Changes

- `Proxy` holds an `AppRegistry` instead of a scalar `session`/`mockManager`/`sessionMu`.
- `getSession`/`setSession`/`getMockManager`/`setMockManager` now take `ctx` and route to the `AppContext` for `AppKeyOrDefault(ctx)`. `GetSession()` (the no-ctx interface method) resolves the `DefaultAppKey` app.
- **`handleConnection`** attributes each connection to its app from the originating kernel PID (`DestInfo.Pid → apps.Resolve`) and stamps the key on `ctx`, so the per-connection session/mockManager belong to the owning app.
- `Record`/`Mock`/`SetMocks`/`SetMocksWithWindow`/`GetConsumedMocks` operate on `ctx`'s app.
- The shared DNS server resolves the default app (per-app DNS attribution is a separate, deferred concern).

## Deliberately out of this slice (sequencing)

- **`syncMock` stays on the global `Get()`** here. Wiring the per-app manager + migrating the parser emit sites (`http`, `mysql`, `generic`, `dns`, `supervisor`, …) to `FromContext(ctx)` must land **together** — otherwise single-app record breaks (emit sites would write to a manager whose output channel isn't wired). That's the next slice, built on #4277's primitives.
- `dnsCache`, `errChannel`/`activeTestErrors`, `isGracefulShutdown`, `pcap` remain proxy-shared for now (follow-up).

## Tests

Full `pkg/agent/proxy` package tests + `go vet` pass. DNS forward tests updated to seed the default app's mock manager.

### DIFF
--- a/pkg/agent/proxy/dns.go
+++ b/pkg/agent/proxy/dns.go
@@ -166,7 +166,10 @@ func (p *Proxy) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
 	msg := new(dns.Msg)
 	msg.SetReply(r)
 
-	session := p.getSession()
+	// The DNS server is shared infrastructure (one listener for the whole
+	// node) with no per-connection app key, so resolve the default app.
+	// Per-app DNS attribution is a separate, deferred concern.
+	session := p.getSession(context.Background())
 	mode := models.GetMode()
 	mockingEnabled := true
 	if session != nil {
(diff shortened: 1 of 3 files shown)


ayush3160:

PR #4279, opened 2026-06-15
feat(agent): per-app routing — X-Keploy-App-Id middleware + keyed clientMocks

ayush3160:
## What

Fourth slice of multi-tenant keploy-agent (#4275). Activates the per-app proxy for the **control plane**: a route middleware lifts the per-app key off the request and `clientMocks` keys by it. **Behaviour-neutral** for single-tenant clients (no header → `DefaultAppKey`).

Stacked on #4278 (base = `feat/mt-proxy-per-app`).

## Changes

- **`appIDMiddleware`** on the `/agent` router lifts the `X-Keploy-App-Id` header onto the request context (`coreagent.WithAppKey`). Every handler → service → proxy call then operates on that app's state.
- **`Agent.clientMocks`** (`StoreMocks` / `UpdateMockParams`) is now keyed by the per-app key from `ctx` instead of the hardcoded `uint64(0)`.

Together with the ctx-keyed proxy (#4278), this makes `Record`/`Mock`/`SetMocks`/`GetConsumedMocks` and stored-mock lookup per-app **end to end** on the control path.

## Deferred (follow-up)

- Per-app `tcChan`/`IncomingProxy` drainer in `StartIncomingProxy`/`SetGracefulShutdown` (the session-aware drain lifecycle).
- Per-app `SetFreezeAnchor`.

## Tests

`appIDMiddleware` unit tests (header → ctx, absent → default). routes tests + `go vet` pass.

### DIFF
--- a/pkg/agent/routes/middleware_test.go
+++ b/pkg/agent/routes/middleware_test.go
@@ -1,0 +1,39 @@
+package routes
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	coreagent "go.keploy.io/server/v3/pkg/agent"
+)
+
+func TestAppIDMiddlewareStampsKey(t *testing.T) {
+	var got coreagent.AppKey
+	var ok bool
+	h := appIDMiddleware(http.HandlerFunc(func(_ http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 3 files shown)


ayush3160:

PR #4291, opened 2026-06-19
fix(replay): bound the app-run teardown wait so a passing run can't false-fail

slayerjain:
`AgentClient.Run` deferred an **unbounded** `runAppErrGrp.Wait()` that blocks on `app.run`'s teardown (cmdCancel grace + `cmd.WaitDelay` + `waitTillExit` up to 30s + `composeDown`), none of which is bounded by ctx. After cancellation that wait can outlive the caller's per-test-set drain budget (`utils.DrainErrGroup`, 30s), so a replay where **every test passed** still fails the step on:

```
teardown drain timed out after cancellation; forcing shutdown ...
a goroutine is ignoring context cancellation {group: replay-testset}
```

i.e. a green run goes red purely on shutdown timing under load.

Fix:
- Bound the deferred wait with `utils.DrainErrGroup` (20s, under the caller's 30s budget).
- Make `waitTillExit` honor ctx cancellation (it could otherwise block up to 30s inspecting the container after cancel).

Any teardown still in flight finishes in the background and is reaped at process exit. Applies to native / docker-run / compose; happy-path behaviour is unchanged.

## Links & References
**Closes:** NA (reliability fix)
### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [x] 🙅 no, because they aren't needed (existing multi-test-set compose replay lanes exercise this teardown path; the change only bounds an existing wait)

- [x] 👍 yes

## Added to documentation?

### DIFF
--- a/pkg/client/app/app.go
+++ b/pkg/client/app/app.go
@@ -494,11 +494,13 @@ func sanitizeAppLogLine(line string) string {
 	return strings.TrimSpace(line)
 }
 
-// composeDown runs docker compose down to remove all containers and networks
+// ComposeDown runs docker compose down to remove all containers and networks
 // created by the compose stack. Without this, stopped containers retain
 // references to image layers; a subsequent docker image prune can delete
-// those layers and corrupt Docker Desktop's overlayfs snapshots.
-func (a *App) composeDown() {
+// those layers and corrupt Docker Desktop's overlayfs snapshots. Exported so the
+// replay loop can tear the stack down on a per-test-set setup failure
+// (agent-readiness timeout) before a retry, not only via run()'s defer.
+func (a *App) ComposeDown() {
 	var downCmd *exec.Cmd
 
 	switch {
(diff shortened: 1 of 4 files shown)


slayerjain:
Closing: the rank-5i approach (bounding the app-run teardown wait + ctx-aware waitTillExit) broke the entire run_golang_docker workflow (12/12 green on the prior main, all-fail here) — it violates keploy's invariant that the app-run goroutine fully completes teardown (the container-exit wait, deliberately on context.Background()) before record/replay finalizes. The kafka teardown-drain false-FAIL this targeted is instead addressed by the per-test-set stack reuse in #4292 (which removes the per-test-set teardown that timed out). Reverting this; will widen the drain budget only if the once-per-lane final teardown still times out.

PR #4296, opened 2026-06-20
fix(replay): make record/replay robust under daemon contention (teardown + agent-readiness)

slayerjain:
Follow-up to #4294. Under heavy docker-daemon contention (many CI lanes up at once) the record/replay teardown still intermittently overran the 30s `utils.DrainErrGroup` budget and tripped a spurious "teardown drain timed out … goroutine ignoring context cancellation" error that failed otherwise-green runs (tests passed, then the lane failed on teardown). The run always succeeds first, so the teardown is slow-but-completing under load, not stuck.

This makes every teardown op bounded AND gives the slow-under-load case adequate headroom:
- `containerStopped()`: a short-deadline (3s) `ContainerInspect`, used for every teardown status check (`waitTillExit`, the compose + docker-run grace loops) instead of `context.Background()`; also bounds the force `ContainerKill`.
- `ComposeDown`: bound `docker compose down` and `docker rm -f` with context timeouts (25s / 10s) so a contended daemon cannot block teardown.
- Widen the app-teardown drains (`record-app`, `replay`, `replay-testset`) 30s -> 120s. Since each op is individually bounded, this is headroom for genuinely-slow-under-load teardown, not cover for a stuck goroutine; the drain still returns as soon as teardown completes.

### DIFF
--- a/pkg/client/app/app.go
+++ b/pkg/client/app/app.go
@@ -305,6 +305,23 @@ func (a *App) RecentLogs(ctx context.Context) string {
 	return a.recentAppLogs(ctx)
 }
 
+// containerStopped reports whether the named container has exited/dead, using a
+// SHORT-deadline inspect. Teardown runs under the record/replay
+// utils.DrainErrGroup 30s budget, and under a contended docker daemon a single
+// ContainerInspect on context.Background() could block long enough to overrun
+// that budget — tripping a spurious "teardown drain timed out" error that fails
+// an otherwise-green run. Bounding the inspect keeps every teardown poll fast
+// regardless of daemon load.
+func (a *App) containerStopped(name string) (bool, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	defer cancel()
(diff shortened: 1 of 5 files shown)


slayerjain:
Superseded by #4297 (released as v3.5.72), which lands the canonical bounded docker-compose teardown — a single ComposeDown budget + a sync.Once collapse of cmdCancel/deferred down + bounded ContainerInspect/ContainerKill — keeping the record-app drain inside its existing 30s budget under daemon contention. That's a cleaner implementation of this PR's teardown work and, importantly, keeps the 30s drain rather than widening it to 120s as this branch did. Enterprise now tracks the v3.5.72 release tag instead of this branch's pseudo-version.

The two bits here that #4297 does not cover — the docker-run `--name` force-remove-before-rerun (the docker-run analogue of #4297's compose force-remove-by-name) and the goroutine-stack dump on drain-timeout (a diagnostic) — will be reintroduced as focused standalone PRs if CI shows they are needed. Closing to avoid a redundant/competing teardown change.

PR #4298, opened 2026-06-21
feat(grpc): implement static dedup for gRPC

Shivanipandey31:
This PR lays the foundation for gRPC structural static deduplication by fixing two silent recording failures and one replay matching bug, then wiring in the hooks that let the dedup layer plug in without coupling OSS to enterprise code.

---

### Background

A normal gRPC (OK) response arrives as three HTTP/2 frames:

```text
HEADERS              → initial headers (:status 200)
DATA                 → response body
HEADERS + END_STREAM → trailers (grpc-status: 0)
```

`checkStreamCompletion` requires both `respHeadersReceived` and `respTrailersReceived` to be true before emitting a test case.

### The problem

A gRPC **error** response (e.g. `INVALID_ARGUMENT`, `NOT_FOUND`) merges everything into **one frame**:

```text
HEADERS + END_STREAM → :status 200 + grpc-status: 3 + grpc-message: "…"
```

`processHeaderBlock` set `respHeadersReceived = true` on arrival, but since no second HEADERS frame ever arrives, `respTrailersReceived` stayed `false` forever — `checkStreamCompletion` never fired and the TC was silently dropped.

A second bug in `processCompleteMessage` compounded this: it returned early when `respDataFrames` was empty (error responses carry no body) before `grpcResp` was initialised, so the stream object was left in an incomplete state even when the fix above was applied alone.

### Fix

* Initialise `grpcResp` before the early-return guard in `processCompleteMessage`.
* In `HandleFrame`, when `END_STREAM` arrives on a HEADERS frame and `respHeadersReceived` is already true but `respTrailersReceived` is not,  promote the initial headers to trailers and set `respTrailersReceived = true` to unblock `checkStreamCompletion`.

---

## 2. gRPC status code matching used the wrong header (`pkg/matcher/grpc/match.go`)

gRPC always sends `:status: 200` at the HTTP/2 transport layer regardless of the gRPC outcome. The actual status is carried in the `grpc-status` **trailer**.

The matcher was raising `StatusCodeChanged` on `:status` mismatches — wrong header, wrong signal. This caused false failures and masked real `grpc-status` differences.

### Fix

* Remove `StatusCodeChanged` / `High` risk classification from `:status` header comparison (it is transport framing, not application status).
* Add an explicit `grpc-status` trailer comparison that populates `result.StatusCode`, sets `currentRisk = High`, and appends
  `StatusCodeChanged` when the gRPC status differs between recorded and replayed  response.

---

## 3. Dedup hooks: `GRPCCaptureHook` and synchronous/mapping params (`pkg/agent/hooks/conn/util.go`, `pkg/agent/proxy/incoming/gRPC/record.go`, `pkg/agent/proxy/incoming/incoming.go`)

To support enterprise-side structural dedup without importing enterprise packages into OSS:

* **`GRPCCaptureHook`** — a `func(req *models.GrpcReq, resp *models.GrpcResp) bool` package-level var, `nil` by default. Enterprise sets it when static dedup is enabled; returning `true` drops the request as a duplicate. The OSS path is completely unaffected when the hook is `nil`.
* **`synchronous` / `mapping` params** threaded through `RecordIncoming` →`emitCompleteStreams` → `CaptureGRPC` so that the test-window mock resolution  (`ResolveRange`) works correctly for gRPC the same way it already does for HTTP.
* **`resolveTestWindow`** helper extracted from `Capture` — deduplicates the `atomic.AddInt64 + ResolveRange` pattern shared by HTTP and gRPC capture paths.

---

## Links & References

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [x] 🍕 Feature
- [ ] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

- [ ] 👍 yes

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

- [ ] 👍 yes, mentioned below

## Self Review done?
- [ ] ✅ yes

- NA

**Examples:**

---

## Additional checklist:


pkg/agent/hooks/conn/util.go
```diff
@@ -418,10 +431,48 @@ func CaptureGRPC(ctx context.Context, logger *zap.Logger, t chan *models.TestCas
...
+	// Bin buffered outgoing mocks into this stream's window before any early
+	// return. For non-duplicates keep=true attributes the mocks; for duplicates
+	// keep=false releases them from the buffer so they don't accumulate as
+	// orphans. Must run before the isDuplicate early-return below.
+	if synchronous {
+		currentID := int64(0)
+		if !isDuplicate {
+			currentID = atomic.AddInt64(&GlobalTestCounter, 1)
+		}
+		testName = fmt.Sprintf("test-%d", currentID)
```
Shivanipandey31, line 468: **[Medium] `test-0` collision for concurrent duplicate streams**

When `isDuplicate=true`, `currentID` is always `0`, so every duplicate stream gets `testName = "test-0"`. If two duplicate gRPC streams complete concurrently in synchronous mode, both call `mgr.ResolveRange(..., "test-0", false, mapping)` at the same time — potentially colliding in the mock manager's window map.

For HTTP, duplicates are dropped *before* entering `Capture`, so this pattern never arises there. Consider using a unique ID for discarded windows too (still increment the counter, but don't save the TC), or skip `ResolveRange` entirely when `keep=false` if the mock manager handles that safely.

```go
// Option: always increment, just don't emit the TC
currentID = atomic.AddInt64(&GlobalTestCounter, 1)
testName = fmt.Sprintf("test-%d", currentID)
mgr.ResolveRange(..., testName, !isDuplicate, mapping)
```

pkg/http2.go
```diff
@@ -224,6 +224,18 @@ func (sm *DefaultStreamManager) HandleFrame(frame http2.Frame, isOutgoing bool,
...
+				// checkStreamCompletion can mark the stream complete.
+				if s.respHeadersReceived && !s.respTrailersReceived {
+					if s.grpcResp == nil {
+						s.grpcResp = &models.GrpcResp{}
+					}
+					s.grpcResp.Trailers = s.grpcResp.Headers
+					s.respTrailersReceived = true
+				}
 				if err := sm.processCompleteMessage(s /*isOutgoing=*/, true); err != nil {
 					return err
```
Shivanipandey31, line 244: **[Low] Shallow copy of `GrpcHeaders`**

`GrpcHeaders` contains `map[string]string` fields (`PseudoHeaders`, `OrdinaryHeaders`). This struct assignment is a shallow copy — both `Trailers` and `Headers` will share the same underlying maps. If anything downstream mutates one, it silently mutates both.

Safe in practice here since the stream completes immediately after, but consider a deep copy to make the intent explicit and guard against future regressions:

```go
s.grpcResp.Trailers = models.GrpcHeaders{
    PseudoHeaders:   maps.Clone(s.grpcResp.Headers.PseudoHeaders),
    OrdinaryHeaders: maps.Clone(s.grpcResp.Headers.OrdinaryHeaders),
}
```
(`maps` package is available in Go 1.21+)

pkg/matcher/grpc/match.go
```diff
@@ -423,3 +447,13 @@ func Match(tc *models.TestCase, actualResp *models.GrpcResp, noiseConfig map[str
 
 	return matched, result
 }
+
+// parseGrpcStatus parses a grpc-status trailer value to int.
```
Shivanipandey31, line 451: [Low] Silent parse error on malformed grpc-status

A non-numeric trailer value silently returns 0 (OK), masking a bad response as a false pass. The discarded error also triggers errcheck/staticcheck.

Consider returning -1 on parse failure for a deterministic mismatch:

```go
n, err := strconv.Atoi(s)
if err != nil {
    return -1
}
return n
```

pkg/agent/hooks/conn/util.go
```diff
@@ -27,6 +27,22 @@ import (
...
+// Callers assign the return value directly to testCase.Name rather than reading
+// testCase.Name back — this severs any CodeQL taint flow: testCase may carry
+// HTTP/gRPC-sourced data so reads from its fields are flagged, but the return
+// value here is always a synthetic "test-N" identifier derived only from the
+// atomic counter and is safe to forward to ResolveRange's diagnostic logs.
+func resolveTestWindow(reqTime, resTime time.Time, mapping bool) string {
+	currentID := atomic.AddInt64(&GlobalTestCounter, 1)
+	testName := fmt.Sprintf("test-%d", currentID)
+	if mgr := syncMock.Get(); mgr != nil {
+		mgr.ResolveRange(reqTime, resTime, testName, true, mapping)
```
Shivanipandey31, line 45: [Low] `keep` is hardcoded to `true`

The `keep` parameter to `ResolveRange` is always `true` from `resolveTestWindow`, which matches the original behaviour. But it's invisible at call sites — a future caller that needs `keep=false` (e.g. a dry-run path) would have to look inside to discover it.

Suggestion: add a brief comment documenting the invariant, or expose it as a parameter:

```go
func resolveTestWindow(reqTime, resTime time.Time, mapping bool) string {
    // keep=true: this helper is only called for non-duplicate test cases
    mgr.ResolveRange(reqTime, resTime, testName, true /*keep*/, mapping)
```

Shivanipandey31 (commented):
**Overall review**

The core fixes are correct and well-justified — the gRPC error framing analysis is solid and the matcher fix is long overdue. The hook design (nil-safe package-level var, shared deduplicator) cleanly mirrors the existing HTTP pattern without coupling OSS to enterprise.

**Missing: unit tests for `CaptureGRPC`**

The function grew significantly in this PR (dedup hook evaluation, size cap, mock-window management for both duplicate and non-duplicate paths) and has no direct test coverage. Suggest adding table-driven tests covering at minimum:
- Normal capture (non-duplicate, non-synchronous)
- Duplicate drop via `GRPCCaptureHook`
- Size-exceeded skip (body > 5 MB)
- Synchronous mode: counter incremented, `ResolveRange` called with correct args

**PR checklist**

The PR template checkboxes (type, e2e tests, self-review) are all unchecked — please fill these in before merge.

pkg/agent/hooks/conn/util.go
Shivanipandey31: Fixed in 95c6447 — counter is now always incremented for all streams (duplicate or not), so each gets a unique window name. The `keep=!isDuplicate` flag still signals the mock manager to discard mocks for duplicates.

pkg/http2.go
Shivanipandey31: Fixed in 95c6447 — replaced the struct assignment with `maps.Clone` for both `PseudoHeaders` and `OrdinaryHeaders`, so `Headers` and `Trailers` no longer share underlying maps.

pkg/matcher/grpc/match.go
Shivanipandey31: Fixed in 95c6447 — now returns -1 on parse error so a malformed `grpc-status` trailer triggers a deterministic mismatch rather than silently passing as OK.

pkg/agent/hooks/conn/util.go
Shivanipandey31: Not changing this one — `resolveTestWindow` is only ever called for non-duplicate HTTP test cases (duplicates are dropped in `NewCapture` before reaching `Capture`), so `keep=true` is always correct here. Added a comment in the function to make that invariant explicit.

pkg/agent/hooks/conn/util.go
```diff
@@ -402,8 +410,13 @@ func ExtractFormData(logger *zap.Logger, body []byte, contentType string) []mode
 	return formData
 }
 
-// CaptureGRPC captures a gRPC request/response pair and sends it to the test case channel
-func CaptureGRPC(ctx context.Context, logger *zap.Logger, t chan *models.TestCase, http2Stream *pkg.HTTP2Stream, appPort uint16) {
+// CaptureGRPC captures a gRPC request/response pair and sends it to the test case channel.
+// Mirrors NewCapture (HTTP) exactly:
+//   - GRPCCaptureHook is evaluated as a boolean so ResolveRange always runs,
+//     even for duplicates (with keep=false to flush their mocks cleanly).
+//   - Counter is only incremented for non-duplicates, matching HTTP behaviour.
```
Shivanipandey31: Stale comment after the collision fix — the counter is now **always** incremented (not just for non-duplicates). Suggest updating to:

Shivanipandey31: Fixed in 68878543 — updated the bullet to reflect that the counter is always incremented.

Shivanipandey31 (commented):
All review comments from previous rounds have been addressed: test-0 window collision fixed, GrpcHeaders shallow copy fixed (maps.Clone), parseGrpcStatus now returns -1 on error, stale doc comment updated.

The three core changes are correct:
1. http2.go — single END_STREAM HEADERS frame correctly promotes headers to trailers and unblocks checkStreamCompletion
2. match.go — grpc-status trailer is the canonical status signal; :status (always 200) no longer raises StatusCodeChanged
3. util.go/record.go — GRPCCaptureHook and synchronous/mapping plumbing is a clean nil-safe extension of the existing HTTP pattern

One housekeeping note: the PR template checkboxes (feature type, e2e tests, self-review) are still unchecked — please fill those in before merging.

prabaltripathiofficial (approved):

Shivanipandey31:
I have read the CLA document, and I hereby sign the CLA.

Shivanipandey31:
recheck

PR #4299, opened 2026-06-21
fix(agent): align agent-ready wait with the agent's healthcheck budget (KEPLOY_AGENT_READY_TIMEOUT)

slayerjain:
## Problem

Under heavy CI docker-daemon contention the in-docker `keploy-agent` container can take **~2 minutes just to start**. Observed in CI (enterprise `go-dedup-docker`):

```
14:39:48  Starting keploy in docker with image keploy/enterprise:v3-dev   # image ALREADY local (loaded from cache, no pull)
14:40:48  ERROR keploy-agent did not become ready in time                 # keploy gives up after 60s
14:41:54  Starting Keploy ... inDocker:true                               # the agent process actually starts — 126s after launch
```

A `docker run` of an **already-local** image taking 126s to actually start the container is docker-daemon/host contention (a non-contended daemon does it in <5s). The agent is **not** hanging or crashing — it comes up, just far slower than keploy's wait.

The agent-ready waits were **60s** (`AgentClient` native/docker-run setup) and **120s** (record, replay, `MakeAgentReadyForDockerCompose`) — all **shorter than the agent container's own healthcheck budget**: `start_period 10s + interval 5s × retries 60 ≈ 310s` (see `pkg/platform/docker`). So keploy gives up while the agent's *own* healthcheck still considers it starting, tears the bring-up down, and emits a spurious:

```
ERROR keploy-agent did not become ready in time
```

This intermittently fails otherwise-green compose/docker-run lanes (`go-dedup-docker`, `umami`, `python-postgres-tls-containerized`, …) whenever the host is busy.

## Fix

- Add `pkg.AgentReadyTimeout()` — default **330s** (≥ the agent's ~310s healthcheck budget), overridable via **`KEPLOY_AGENT_READY_TIMEOUT`** (whole seconds) for tuning.
- Use it at **all four** agent-ready sites (`agent.go` setup + `MakeAgentReadyForDockerCompose`, `record.go`, `replay.go`), replacing the hardcoded 60s/120s.

This is **not** widening a timeout to paper over a hang: the agent provably *does* come up (it logged its HTTP server ~126s in) — it was simply slower than the old wait. Aligning the CLI wait to the agent's own declared startup contract fixes an internal inconsistency. Normal runs are unaffected: `AgentHealthTicker` signals the instant the agent is healthy, so the wait returns in milliseconds on an idle host; the larger budget only matters when the daemon is genuinely saturated.

## Testing

- `go build ./pkg/...` + `go vet` clean; `gofmt` clean; `golangci-lint` 0 issues on the touched packages.
- New `TestAgentReadyTimeout` (default / env override / whitespace / zero / negative / garbage fallback, and asserts the default covers the ~310s healthcheck budget) — passes under `-race`.

Relates to #4266 (configurable timeout) and #4267 (build compose stack before arming the timer): this lands the env knob + raises the default to actually cover the observed startup time, across every readiness site.

### DIFF
--- a/pkg/platform/http/agent.go
+++ b/pkg/platform/http/agent.go
@@ -37,7 +37,6 @@ import (
 )
 
 const (
-	agentReadyTimeout       = 2 * time.Minute
 	agentReadyRetryInterval = 2 * time.Second
 )
 
@@ -1249,7 +1248,12 @@ func (a *AgentClient) Setup(ctx context.Context, cmd string, opts models.SetupOp
 		}
 		a.logger.Debug("Agent is now running, proceeding with setup")
 
-		agentCtx, cancel := context.WithTimeout(ctx, 60*time.Second) // we will wait for 1 minute for the agent to get ready
+		// Wait up to the agent's own healthcheck budget for it to become ready.
+		// Under heavy CI docker-daemon contention the agent container can take
+		// well over a minute just to start (observed: a local-image `docker run`
+		// taking 126s), so a 60s wait gave up prematurely and tore down a
+		// bring-up that would have succeeded. Overridable via KEPLOY_AGENT_READY_TIMEOUT.
+		agentCtx, cancel := context.WithTimeout(ctx, pkg.AgentReadyTimeout())
 		defer cancel()
 
 		agentReadyCh := make(chan bool, 1)
@@ -1546,7 +1550,8 @@ func (a *AgentClient) NotifyGracefulShutdown(ctx context.Context) error {
 }
 
 func (a *AgentClient) MakeAgentReadyForDockerCompose(ctx context.Context) error {
-	ctx, cancel := context.WithTimeout(ctx, agentReadyTimeout)
+	// Aligned with the agent's own healthcheck budget; see pkg.AgentReadyTimeout.
+	ctx, cancel := context.WithTimeout(ctx, pkg.AgentReadyTimeout())
 	defer cancel()
 
 	ticker := time.NewTicker(agentReadyRetryInterval)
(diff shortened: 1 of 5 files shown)


slayerjain:
### Independent review summary — APPROVE-WITH-NITS

Reviewed the diff for correctness/safety:
- All 4 readiness sites wired correctly (`WithTimeout(ctx, pkg.AgentReadyTimeout())` + `defer cancel()`, correct select arms, no inverted logic).
- `pkg.AgentReadyTimeout()` parses `KEPLOY_AGENT_READY_TIMEOUT` and falls back to the 330s default on empty/zero/negative/garbage; read-only env access, no concurrency concern.
- The removed `agentReadyTimeout` const leaves no dangling references (`agentReadyRetryInterval` retained).
- Default 330s ≥ the agent container's own healthcheck budget (~310s = start_period 10s + interval 5s × retries 60, see `pkg/platform/docker/docker.go`); a unit test guards `>= 310s`.
- `go build` / `go vet` / new `TestAgentReadyTimeout` (incl. `-race`) green; `golangci-lint` check on this PR is green.

**Trade-off to be aware of (pre-existing, not introduced here):** the readiness wait relies solely on the timeout — `AgentHealthTicker` only signals healthy/not-yet and agent-process-exit isn't fed into the readiness select (`monitorAgent` only logs "Agent stopped unexpectedly"). So a genuinely *crashed/absent* agent now fails after up to 330s instead of 60s/120s. Acceptable because the happy path returns the instant the agent is healthy (typically seconds) and outer CI timeouts catch true hangs; a sensible **follow-up** is to feed the agent-exit signal into the readiness selects so a crashed agent fails fast.

PR #4305, opened 2026-06-23
fix(ci-flake): tab-safe PostgresV3 SQLNormalized YAML (CI flake robustness, batch 1)

slayerjain:
First in a series of product fixes to make the enterprise CI flake-free.

## This batch — PostgresV3 SQLNormalized tab-in-YAML marshal (CLASS H)

`PostgresV3QuerySpec.SQLNormalized` is a plain `string`. After the integrations `restoreRawSQL` pass it can carry embedded tabs/newlines; `yaml.v3` v3.0.1 emits those as a literal block scalar, and since `yaml.Node.Encode` marshals-then-reparses its own output, the scalar fails to round-trip with `found a tab character where an indentation space is expected` — inside `EncodeMock`'s `Spec.Encode`, before the post-encode `sanitizeYAMLStringNodes` walk can rescue it. The whole PostgresV3 mock is then dropped (observed in enterprise atg-with-mocks: `failed to marshal PostgresV3 mock as yaml {mock_name: mock-15}`).

Fix: add `PostgresV3QuerySpec.MarshalYAML` routing `SQLNormalized` through `PostgresV3SafeString` (double-quotes when needed), mirroring the existing `PostgresV3Notice`/`PostgresV3Error` convention. The model field stays `string` (JSON/BSON + all integrations sites compile unchanged); only the YAML write side gains double-quoting. Regression tests cover leading-tab and tab/newline SQLNormalized round-tripping through `EncodeMock`.

## Follow-ups (separate batches, in progress)
The enterprise CI flake landscape (mapped across pipelines 5062–5122) has these remaining classes, fixed product-side:
- **CLASS A** container-name race on the *user-app* `docker run --name` path (go-docker-timefreeze, go-dedup-docker) — extend the v3.5.77 `ensureContainerNameFreeWithin` guard to the user-app launch.
- **CLASS B** compose-recreate of the keploy-v3 agent sidecar (atg-with-mocks, python-postgres-tls-containerized, kafka) — stop dual-managing the agent (compose service + external force-remove) / per-phase compose project isolation.
- **CLASS C/D** app- and agent-readiness races — gate replay on real readiness (the `--health-url` pattern) + contention-tolerant agent-readiness backoff.
- **CLASS E/F/G/I** grpc deadline (sample per-call ctx), pulsar `send on closed channel` panic, port-reuse on restart, BPF load flake.

### DIFF
--- a/pkg/agent/proxy/fakeconn/fakeconn.go
+++ b/pkg/agent/proxy/fakeconn/fakeconn.go
@@ -122,6 +122,28 @@ func newWithLogger(ch <-chan Chunk, localAddr, remoteAddr net.Addr, log logger) 
 // (subject to read deadline and Close).
 func (f *FakeConn) Read(p []byte) (int, error) {
 	if f.closed.Load() {
+		// Close means "no more blocking", NOT "discard bytes already
+		// delivered to me". The relay tee drains its buffered chunks
+		// into f.ch and then Close() fires during teardown; a chunk
+		// that already landed in f.ch (or the stash) is a fully
+		// recorded wire event and must still be readable, otherwise a
+		// response whose body arrived a hair before the connection was
+		// torn down is silently lost (the "server closed before
+		// response" mock-incomplete race on Connection: close traffic
+		// such as the boot-time startup mock). Only return ErrClosed
+		// once nothing buffered remains.
+		if c, ok := f.drainBufferedLocked(); ok {
+			n := copy(p, c.Bytes)
+			if n < len(c.Bytes) {
+				f.mu.Lock()
+				f.buf.Write(c.Bytes[n:])
+				f.bufReadAt = c.ReadAt
+				f.bufWrittenAt = c.WrittenAt
+				f.bufDir = c.Dir
+				f.mu.Unlock()
+			}
+			return n, nil
+		}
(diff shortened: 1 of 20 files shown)


rajpratham1 (approved):
Great work on this PR! The changes are comprehensive and each fix is backed by regression tests, which makes the overall solution much more convincing.

Some things I particularly liked:

The retry logic for transport-level connection resets is carefully guarded by checking whether any mocks were consumed before attempting a resend. That keeps the retry path safe while avoiding duplicate processing.
The new APP_CONNECTION_ERROR classification separates infrastructure/connectivity failures from actual application regressions, which should make replay results much easier to interpret.
The Docker startup improvements (container-name cleanup, retry handling, and host-port readiness checks) should help eliminate several intermittent CI failures without affecting normal execution.
The YAML serialization fix for SQLNormalized is well explained and the dedicated round-trip regression test gives confidence that the tab-related encoding issue won't regress.
I also appreciate the number of focused unit and integration tests added throughout the PR—they document the intended behavior and significantly improve confidence in these changes.

Overall, the implementation is well scoped, thoroughly documented, and backed by strong test coverage. The fixes appear to address real CI reliability issues while minimizing behavioral changes outside the affected paths.

PR #4307, opened 2026-06-23
fix(replay): scope strict schema-noise mock rejection to HTTP mocks

Aditya-eddy:

`SchemaNoiseStrict` is effectively an **HTTP-only contract**: the only matcher that consumes the flag (and the shared `schemanoise.Engine`) is the HTTP integration — see `pkg/agent/proxy/integrations/http/match.go` (`schemanoise.New(...)`, `filterStrictNoiseMatches`). The **MySQL, redis and gRPC matchers never receive the flag at all**.

But the per-test *consumed-vs-expected* mock reconciliation in `RunTestSet` (`pkg/service/replay/replay.go`) failed **any** passing testcase whose mock set diverged, regardless of the kind of the unconsumed mock:

```go
case testPass && r.config.Test.SchemaNoiseStrict:
    // ... testPass = false  (fail the testcase)
```

So a non-HTTP dependency that drifts for reasons strict schema-noise has no say over still force-failed the testcase **even though the recorded response matched**. Concretely:

- a value-driven **COM_QUERY** structural fallback serving a *different same-shape* row (plaintext MySQL has no per-value noise tolerance — only `COM_STMT_EXECUTE` does, via `NoiseChecker`), or
- a **redis** key embedding a generated id (e.g. `GET <orderId>_<token>`),

makes the expected mock go unconsumed → `mockSetMismatch` → strict fail, producing false-positive `strict schema-noise: expected mock was rejected (non-noise request-body drift)` failures.

**Fix:** gate the strict failure on a new `strictHTTPDrift` check — only fail when at least one *unconsumed expected* mock is actually an **HTTP** mock. A mismatch made up solely of non-HTTP mocks now falls through to the normal (non-strict) handling, exactly as if strict were off. HTTP request-body strictness is unchanged.

The mock kind is already available at the call site (`MockEntry.Kind`, resolved via `mockKindByName`), and the new loop mirrors the existing DNS + reusable-tier exclusion used to build `filteredExpectedNames`.

## Links & References

**Closes:** NA

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [x] 🐞 Bug Fix

## Added e2e test pipeline?
- [ ] 👍 yes

The behavioural change is in the in-cluster replay verdict path; happy to add a `RunTestSet`-level unit test exercising a non-HTTP-only mock-set mismatch under `SchemaNoiseStrict` if maintainers point me at the preferred harness.

- [x] 👍 yes

## Added to documentation?

- [x] 👍 yes, mentioned below

With `--schema-noise-strict` on, run a test set whose only request-body drift is a non-HTTP dependency (e.g. a Hibernate `UPDATE … SET updated_at='<wall-clock>' …` plaintext `COM_QUERY`, or a redis key containing a generated id) while the HTTP response is deterministic.

- **Before:** testcase fails with `strict schema-noise: expected mock was rejected (non-noise request-body drift)` despite the response matching.
- **After:** the non-HTTP mock-set divergence is ignored (test passes), while genuine HTTP request-body drift still fails strictly.

## Self Review done?
- [x] ✅ yes

- NA

### DIFF
--- a/pkg/service/replay/replay.go
+++ b/pkg/service/replay/replay.go
@@ -1921,10 +1921,54 @@ func (r *Replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
 			// non-noise request field drifted. The app's response can still
 			// match (e.g. a deterministic dependency), so the response check
 			// alone won't catch it; treat it as a real test failure.
+			//
+			// SchemaNoiseStrict is an HTTP-only contract: only the HTTP
+			// integration consumes the flag and the shared schema-noise engine
+			// (see pkg/agent/proxy/integrations/http/match.go). The MySQL,
+			// redis and gRPC matchers never receive it. A non-HTTP mock can go
+			// unconsumed for reasons strict schema-noise has no say over — e.g.
+			// a value-driven COM_QUERY structural fallback serving a different
+			// same-shape row, or a redis key drift — so force-failing those
+			// under strict would be a false positive. strictHTTPDrift gates the
+			// strict failure to mismatches that actually involve an HTTP mock;
+			// a mismatch made up only of non-HTTP mocks falls through to the
+			// normal (non-strict) handling, exactly as if strict were off.
+			strictHTTPDrift := false
+			if mockSetMismatch && r.config.Test.SchemaNoiseStrict {
+				consumedSet := make(map[string]bool, len(filteredMockNames))
+				for _, n := range filteredMockNames {
+					consumedSet[n] = true
+				}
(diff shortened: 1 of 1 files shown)


rajpratham1 (approved):
This is a thoughtful bug fix that narrows the scope of strict schema-noise enforcement to the cases it was actually designed to cover. By limiting strict failures to HTTP mock drift, it avoids false positives for other integrations (MySQL, Redis, gRPC) while preserving the intended behavior for HTTP request validation.

PR #4286, opened 2026-06-18
feat(schema-noise): kind-agnostic plug-and-play noise engine, unified on MockSpec.ReqBodyNoise

Aditya-eddy:
## What

Make schema-based auto-replay **request-body noise** (`req_body_noise`) kind-agnostic and plug-and-play: extract the learn/enforce algorithm into a single shared engine, store learned noise in **one** place for every parser, and let HTTP consume the engine like any other protocol.

## Why

When `config.Test.SchemaNoiseDetection` is on, the matcher detects request-body fields that drift between recording and replay and persists them as `req_body_noise`, so a later strict replay tolerates those fields (learn → enforce → pass).

Originally this whole flow lived **inside the HTTP integration**, and the learned noise was stored on `HTTPReq.ReqBodyNoise` — HTTP-only. A second protocol (Pulsar, and later Kafka/Redis/…) whose request is wire-encoded in `GenericRequests` had no per-protocol struct to hang noise on, and would have had to copy the flag handling, known-noise merge, JSON diffing, strict rejection and learn-merge wholesale. This PR is the OSS groundwork the enterprise **Pulsar** schema-noise detection builds on.

### 1. A shared engine — `pkg/agent/proxy/integrations/schemanoise`
The algorithm now lives exactly once. A protocol plugs in by implementing a small **`Adapter`** (`RecordedBody` / `StoredNoise` / `SetLearnedNoise` / `RecordedValueIsNoise` / `Diff`); the **`Engine`** owns flag handling, the known-noise merge, strict rejection and the monotonic learn-merge.
- `Engine.Detect` → new drift to learn; `Engine.Learn` → monotonic merge onto the mock; `Engine.StrictAllows` / `StrictReject` → replay-path enforcement; `Engine.KnownNoise` → global/user ∪ already-learned.
- `Diff` is the one inherently protocol-specific piece (mirrors each parser's `buildXMismatchReport`). Ship **`JSONDiffer`** as the default — a JSON-payload parser embeds it and inherits `Diff`; a binary/structured parser (Avro/Protobuf/RESP/BSON) supplies its own. `comparable=false` makes the engine treat any byte difference as a real, non-learnable mismatch instead of "nothing drifted".

### 2. One storage location for every parser
- **Removed `HTTPReq.ReqBodyNoise`** (field + BSON plumbing). Every parser — HTTP and non-HTTP alike — now stores learned `req_body_noise` on the **kind-agnostic `MockSpec.ReqBodyNoise`** (yaml `req_body_noise`).
[...]

PR #4315, opened 2026-07-02
feat(mysql): schema-noise detection and strict enforcement for command matching

Aditya-eddy:
- Wires the MySQL replayer's `matchCommand` into the shared `schemanoise` engine (from #4286 / #4310), making MySQL the second parser client of the unified `MockSpec.ReqBodyNoise` storage — no MySQL-specific noise format.
- New `mysqlNoiseAdapter` canonicalizes the drift-carrying content of each command packet into diffable JSON: `COM_QUERY` / `COM_STMT_PREPARE` → `body.query`, `COM_STMT_EXECUTE` → `body.parameters.N.{type,value,unsigned}` (index-keyed **object**, so noise on one param position never excuses drift on another), `COM_STMT_SEND_LONG_DATA` → `body.data` (field-level paths when the chunk is standalone JSON, base64 blob otherwise). Packets with no drift-capable body short-circuit the engine into a no-op.
- `--schema-noise-detection`: when a mock is served via the lenient score/FIFO fallbacks (i.e. its request drifted), the recorded vs live request bodies are diffed and the drifted field paths are learned into `req_body_noise` — merged on fresh copies in `updateMock`, never on the shared pooled mock (same concurrency rule as HTTP).
- `--schema-noise-strict`: every match candidate is value-gated against the union of user `test.globalNoise.requestBody` noise and the mock's learned noise. This now covers exact-text `COM_QUERY` hits (query attributes live outside the text) and `COM_STMT_SEND_LONG_DATA` chunks, whose payload was historically never content-compared — rejecting every recorded SLD candidate is a real mismatch instead of a silent pass.
- Better mismatch reports: EXECUTE misses resolve the prepared SQL via the stmtID map and show truncated bound params (previously a blank `COM_STMT_EXECUTE` with empty closest mock); strict rejections carry field-level diffs (`matcher.JSONFieldDiffs` in the noise-path vocabulary) and a targeted next-steps hint naming the drifted fields.

### Verification
Verified end-to-end against a purpose-built drift app (one endpoint per drift-prone packet: inlined-literal COM_QUERY, drifting EXECUTE params, run-generated read-back id, traceparent comment in PREPARE text, IN-list arity change, ~64 KiB streamed long-data JSON, plus a no-drift control):
- record → `test --schema-noise-detection` learns exactly the drifting fields (`body.query`, `body.parameters.0.value`, `body.data.*`) under the unified `noise:` block; the no-drift control learns nothing.
- subsequent `test --schema-noise-strict` drops strict rejections **25 → 1**, the survivor being the intended structural reject (IN-list placeholder-count change), which must stay a hard miss.
- `go build ./...` and `go test ./pkg/agent/proxy/integrations/mysql/... ./pkg/agent/proxy/integrations/schemanoise/...` pass; new unit tests in `schema_noise_test.go` cover the adapter's canonicalization and gating.

## Links & References

**Closes:** NA

### 🔗 Related PRs
- #4286 (kind-agnostic schema-noise engine)
- #4310 (strict request-body value-match, unified on-disk noise block)
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [x] 🍕 Feature
- [ ] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

- [x] 👍 yes

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki


pkg/agent/proxy/integrations/mysql/replayer/match.go
```diff
@@ -369,6 +389,58 @@ func matchCommand(ctx context.Context, logger *zap.Logger, req mysql.Request, mo
...
+	// candidate gets field-level diffs (noise-vocabulary "body."-paths with
+	// recorded vs live values via matcher.JSONFieldDiffs) so the report can
+	// say exactly WHICH request fields drifted outside noise, not just that
+	// no mock matched.
+	var (
+		strictRejectCount int
+		strictClosestMock string
+		strictFieldDiffs  []models.MockFieldDiff
+	)
+	strictAllowed := func(mock *models.Mock) bool {
```
Aditya-eddy: don't make inline function like this.

pkg/agent/proxy/integrations/mysql/replayer/match.go
```diff
@@ -208,7 +210,25 @@ func isSessionReusableCommandMock(mock *models.Mock) bool {
 	return strings.HasPrefix(hdr.Type, "COM_")
 }
 
-func matchCommand(ctx context.Context, logger *zap.Logger, req mysql.Request, mockDb integrations.MockMemDb, decodeCtx *wire.DecodeContext) (*mysql.Response, bool, string, string, error) {
+// mockMiss carries the diagnostics of a "no matching mock" outcome so the
+// caller can build a meaningful mismatch report: the closest candidate (name +
+// its recorded query), field-level diffs against the first strict-rejected
+// candidate (noise-vocabulary "body."-paths with recorded/live values), and
+// how many candidates strict enforcement ruled out.
+type mockMiss struct {
```
Aditya-eddy: don't add mock miss in the match.go please choose the approppiate file for it

pkg/agent/proxy/integrations/mysql/replayer/match.go
```diff
@@ -208,7 +210,25 @@ func isSessionReusableCommandMock(mock *models.Mock) bool {
...
+	strictRejected int
+}
+
+// matchCommand matches one live command-phase request against the mock pool.
+// noiseEngine carries the resolved schema-noise flags (nil-safe: a nil engine
+// disables both detection and strict). userBodyNoise is the user's
+// test.globalNoise.requestBody bucket (root-relative lowercased paths) so
+// configured noise participates in strict gating with the same vocabulary as
+// learned req_body_noise. miss is non-nil only when ok is false without error.
+func matchCommand(ctx context.Context, logger *zap.Logger, req mysql.Request, mockDb integrations.MockMemDb, decodeCtx *wire.DecodeContext, noiseEngine *schemanoise.Engine, userBodyNoise map[string][]string) (*mysql.Response, bool, *mockMiss, error) {
```
Aditya-eddy, line 218: why did we change here the response structure for match command.

pkg/agent/proxy/integrations/mysql/replayer/match.go
```diff
@@ -479,22 +558,41 @@ func matchCommand(ctx context.Context, logger *zap.Logger, req mysql.Request, mo
 						matchedResp, matchedMock, queryMatched = &mock.Spec.MySQLResponses[0], mock, true
 					}
 				} else if c > maxMatchedCount {
-					maxMatchedCount, matchedResp, matchedMock = c, &mock.Spec.MySQLResponses[0], mock
+					// Track the closest candidate for the mismatch report even
```
Aditya-eddy, line 528: here in this we removed the logic of "maxMatchedCount, matchedResp, matchedMock = c, &mock.Spec.MySQLResponses[0], mock" why is that?

pkg/agent/proxy/integrations/mysql/replayer/match.go
```diff
@@ -479,22 +558,41 @@ func matchCommand(ctx context.Context, logger *zap.Logger, req mysql.Request, mo
...
 					if qp, qok := mockReq.PacketBundle.Message.(*mysql.QueryPacket); qok {
 						bestPartialQuery = qp.Query
 					}
+					// Structure-matched-but-text-drifted candidate: under
+					// strict it may only be served when the drift (body.query
+					// / attribute values) is covered by learned/user noise.
+					if !strictAllowed(mock) {
+						continue
+					}
+					maxMatchedCount, matchedResp, matchedMock = c, &mock.Spec.MySQLResponses[0], mock
```
Aditya-eddy, line 540: what is exactly this is used for explain all three terms

pkg/agent/proxy/integrations/mysql/replayer/match.go
Aditya-eddy: Refactored in 6b3c1ee1 — the closure is now a named `strictGate` type with an `allows(mock)` method, living in `schema_noise.go` next to the rest of the schema-noise machinery. Its fields also give the accumulated diagnostics proper names/docs (`rejected`, `closestMock`, `fieldDiffs`).

pkg/agent/proxy/integrations/mysql/replayer/match.go
Aditya-eddy: Moved in 6b3c1ee1 — `mockMiss` now lives in a new `report.go` (mismatch-report diagnostics for the replayer), together with `formatExecParams` which moved out of `query.go` for the same reason.

pkg/agent/proxy/integrations/mysql/replayer/match.go
Aditya-eddy: The old contract returned two bare strings (`closestQuery`, `closestMockName`) and that couldn't carry what the mismatch report needs now: (1) COM_STMT_EXECUTE misses used to render a blank summary and an EMPTY closest mock (the EXEC branch never tracked one), and (2) a strict rejection must tell the user WHICH request fields drifted outside noise — field-level diffs with recorded/live values (`models.MockFieldDiff`) plus the rejected-candidate count, so the report can render the FIELD | EXPECTED | RECEIVED table and an actionable hint instead of the generic 're-record'. Growing the return list to 5-6 positional values scales worse than one diagnostics struct, so the two strings were folded into `*mockMiss` (now in `report.go`). On the happy path (`ok=true`) it's nil, so no allocation.

pkg/agent/proxy/integrations/mysql/replayer/match.go
Aditya-eddy: Not removed — reordered. The assignment `maxMatchedCount, matchedResp, matchedMock = c, &mock.Spec.MySQLResponses[0], mock` is still in this branch, a few lines below: the branch now (1) records the candidate as `bestPartialMock`/`bestPartialQuery` FIRST so a strict-rejected candidate still shows up as the 'closest mock' in the mismatch report, (2) applies the strict gate (`continue` if the drift isn't covered by learned/user noise), and only then (3) runs the same score assignment, so a rejected candidate can never become the served response. With schemaNoiseStrict off the gate always allows, so the behavior is byte-identical to before.

pkg/agent/proxy/integrations/mysql/replayer/match.go
Aditya-eddy: These are the score-based fallback trio used when no definitive (exact) match exists: `maxMatchedCount` = the highest similarity score seen so far across the pool scan (a later candidate wins only with a STRICTLY higher score); `matchedResp` = the response of that best-scoring candidate — what gets served if the scan ends with no definitive match; `matchedMock` = the mock that owns that response, needed so `updateMock` can consume (per-test) or re-stamp (session) it after selection. Added doc comments at their declaration in 6b3c1ee1.

gouravkrosx (approved):

pkg/agent/proxy/integrations/mysql/replayer/schema_noise_test.go
```diff
@@ -0,0 +1,418 @@
...
+			}
+		}
+		if !foundParamDiff {
+			t.Errorf("miss must carry the drifted param diff with values, got %+v", miss.fieldDiffs)
+		}
+	})
+
+	t.Run("strict allows drift covered by learned noise", func(t *testing.T) {
+		m := execMock("m1", "recorded-uuid")
+		m.Spec.ReqBodyNoise = map[string][]string{"body.parameters.0.value": {}}
```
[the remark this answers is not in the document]
Aditya-eddy: This compiles as-is: within a composite literal, the element type may be elided, so `map[string][]string{"k": {}}` is equivalent to `map[string][]string{"k": []string{}}` ([Go spec, composite literals](https://go.dev/ref/spec#Composite_literals)). `go vet ./pkg/agent/proxy/integrations/mysql/replayer/` passes clean on this branch. Resolving as a false positive.

pkg/agent/proxy/integrations/mysql/replayer/schema_noise_test.go
```diff
@@ -0,0 +1,418 @@
...
+		_, ok, _, err := matchCommand(context.Background(), logger, req, db, newDecodeCtx(), eng, nil)
+		if err != nil || !ok {
+			t.Fatalf("strict must serve a candidate whose only drift is learned noise (ok=%v err=%v)", ok, err)
+		}
+	})
+
+	t.Run("strict allows drift covered by user requestBody noise", func(t *testing.T) {
+		db := newDb(execMock("m1", "recorded-uuid"))
+		eng := schemanoise.New(mysqlNoiseAdapter{}, false, true /*strict*/)
+		user := map[string][]string{"parameters.0.value": {}}
```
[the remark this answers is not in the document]
Aditya-eddy: This compiles as-is: within a composite literal, the element type may be elided, so `map[string][]string{"k": {}}` is equivalent to `map[string][]string{"k": []string{}}` ([Go spec, composite literals](https://go.dev/ref/spec#Composite_literals)). `go vet ./pkg/agent/proxy/integrations/mysql/replayer/` passes clean on this branch. Resolving as a false positive.

pkg/agent/proxy/integrations/mysql/replayer/schema_noise_test.go
```diff
@@ -0,0 +1,418 @@
...
+			}
+		}
+		if !foundQueryDiff {
+			t.Errorf("miss must carry a body.query field diff for the report, got %+v", miss.fieldDiffs)
+		}
+	})
+
+	t.Run("strict allows learned body.query noise", func(t *testing.T) {
+		m := readbackMock("q1", recorded, "row-1", zeroTime())
+		m.Spec.ReqBodyNoise = map[string][]string{"body.query": {}}
```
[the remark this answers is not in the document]
Aditya-eddy: This compiles as-is: within a composite literal, the element type may be elided, so `map[string][]string{"k": {}}` is equivalent to `map[string][]string{"k": []string{}}` ([Go spec, composite literals](https://go.dev/ref/spec#Composite_literals)). `go vet ./pkg/agent/proxy/integrations/mysql/replayer/` passes clean on this branch. Resolving as a false positive.

pkg/agent/proxy/integrations/mysql/replayer/schema_noise_test.go
```diff
@@ -0,0 +1,418 @@
...
+			}
+		}
+		if !foundDataDiff {
+			t.Errorf("miss must carry a body.data.id field diff, got %+v", miss.fieldDiffs)
+		}
+	})
+
+	t.Run("strict allows learned chunk noise", func(t *testing.T) {
+		m := sldMockOf("sld1", recorded)
+		m.Spec.ReqBodyNoise = map[string][]string{"body.data.id": {}}
```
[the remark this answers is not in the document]
Aditya-eddy: This compiles as-is: within a composite literal, the element type may be elided, so `map[string][]string{"k": {}}` is equivalent to `map[string][]string{"k": []string{}}` ([Go spec, composite literals](https://go.dev/ref/spec#Composite_literals)). `go vet ./pkg/agent/proxy/integrations/mysql/replayer/` passes clean on this branch. Resolving as a false positive.

pkg/agent/proxy/integrations/mysql/replayer/match.go
```diff
@@ -471,6 +500,13 @@ func matchCommand(ctx context.Context, logger *zap.Logger, req mysql.Request, mo
 					// a genuinely reusable single-recording query. When no
 					// window is active (windowActive==false) this collapses to
 					// the previous first-exact-match-wins behaviour.
+					//
+					// Even an exact-text match is strict-gated: the query
+					// attributes (CLIENT_QUERY_ATTRIBUTES) live outside the
+					// text and may still drift.
+					if !gate.allows(mock) {
+						continue
+					}
```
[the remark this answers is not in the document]
Aditya-eddy: Good catch — fixed in aa8f425d. The strict-rejected exact-text candidate is now captured as the closest match before the gate's `continue` (parity with the EXECUTE branch), so the miss diff names the query instead of rendering an empty "closest". Covered by a new regression test (`strict-rejected exact-text candidate still names closest query`) that fails without the fix.

pkg/agent/proxy/integrations/mysql/replayer/match.go
```diff
@@ -479,22 +515,41 @@ func matchCommand(ctx context.Context, logger *zap.Logger, req mysql.Request, mo
...
+					maxMatchedCount, matchedResp, matchedMock = c, &mock.Spec.MySQLResponses[0], mock
 				}
 
 			case sCOM_STMT_PREP:
 				if ok, c := matchPreparePacket(ctx, logger, mockReq.PacketBundle, req.PacketBundle); ok {
+					if !gate.allows(mock) {
+						continue
+					}
 					matchedResp, matchedMock, queryMatched = &mock.Spec.MySQLResponses[0], mock, true
 				} else if c > maxMatchedCount {
```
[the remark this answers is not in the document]
Aditya-eddy: Fixed in aa8f425d — same capture-before-gate treatment as the COM_QUERY branch. Note that with the current canonical body a byte-identical PREPARE text can't actually drift (the body is only `body.query`), so this path is a defensive parity guard should the PREPARE body ever grow more fields.

pkg/agent/proxy/integrations/mysql/replayer/query.go
```diff
@@ -152,7 +199,8 @@ func simulateCommandPhase(ctx context.Context, logger *zap.Logger, clientConn ne
 				logger.Error("Connection closing due to no matching mock found. Re-record mocks if the SQL query has changed.",
 					zap.Int("commands_processed", commandCount),
 					zap.String("request_type", req.Header.Type),
-					zap.String("closest_mock", closestMockName))
+					zap.String("closest_mock", miss.closestMock),
+					zap.Int("strict_rejected_candidates", miss.strictRejected))
```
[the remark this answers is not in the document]
Aditya-eddy: Fixed in aa8f425d — the log message no longer hard-codes the re-record hint; it now carries the already-computed strict-aware guidance in a `next_step` field, consistent with the rest of the codebase.

PR #3861, opened 2026-02-25
fix: status_code_class assertion always returns 2xx (#3843)

unnatii14:
## Describe the changes that are made
- Fixed `status_code_class` assertion always comparing against hardcoded `"2xx"` regardless of actual HTTP response status code. Changed `200/100` to `actualResponse.StatusCode/100` in `AssertionMatch()`.
- Added 15 table-driven unit tests (`TestAssertionMatch_StatusCodeClass_T152`) covering 2xx, 3xx, 4xx, and 5xx classes for both pass and fail scenarios.
- Fixed `PrepareDockerCommand` on Windows (`util_windows.go`): filter empty tokens from `strings.Split` to prevent `docker: invalid reference format` error, and removed incorrect `os.Args[1:]` append that corrupted the agent container command.

| Endpoint | Status | Assertion        | Before fix | After fix |
|----------|--------|------------------|------------|-----------|
| /not-found | 404  | status_code_class: 4xx | ❌ FAIL (actual: 2xx) | ✅ PASS |
| /error     | 500  | status_code_class: 5xx | ❌ FAIL (actual: 2xx) | ✅ PASS |
| /health    | 200  | status_code_class: 2xx | ✅ PASS               | ✅ PASS | 

## Links & References

**Closes:** #3843

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- #3843 — `status_code_class` assertion always fails for non-2xx responses
### 📄 Related Documents
- NA

## What type of PR is this? (check all applicable)
- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes
- [x] 🙅 no, because they aren't needed
- [ ] 🙋 no, because I need help

## Added comments for hard-to-understand areas?
- [x] 👍 yes
- [ ] 🙅 no, because the code is self-explanatory

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki
- [x] 🙅 no documentation needed

## Are there any sample code or steps to test the changes?
- [x] 👍 yes, mentioned below
- [ ] 🙅 no, because it is not needed

**Steps to reproduce the bug:**
[...]

Issue #3843, opened 2026-02-24
[bug]: `status_code_class` Assertion Always Evaluates Against `"2xx"` Due to Hardcoded Literal - T152


Krish3017:
### 👀 Is there an existing issue for this?

- [x] I have searched and didn't find similar issue

## � Current Behavior

The `status_code_class` assertion always compares against the hardcoded class `"2xx"`, regardless of what the server actually returned.

- If you assert `"2xx"` and the server returns **500**, the test **passes** (incorrect — should fail).
- If you assert `"5xx"` and the server returns **500**, the test **fails** (incorrect — should pass).
- The actual response status code is completely ignored by this assertion type.

## ✅ Expected Behavior

The `status_code_class` assertion should compute the class from the **actual** HTTP response status code:

- Assert `"2xx"` + server returns 200 → **PASS**
- Assert `"2xx"` + server returns 500 → **FAIL**
- Assert `"5xx"` + server returns 500 → **PASS**
- Assert `"4xx"` + server returns 404 → **PASS**

### 1. Set up a simple HTTP server

Use any language. Example with Go:

```go
// main.go
package main

import (
    "net/http"
)

func main() {
    http.HandleFunc("/health", func(w http.ResponseWriter, r *http.Request) {
        w.WriteHeader(http.StatusOK) // returns 200
        w.Write([]byte(`{"status":"ok"}`))
    })
    http.ListenAndServe(":8080", nil)
}
```

### 2. Record a test case

```bash
keploy record -c "go run main.go"
```

Then hit the endpoint:

```bash
curl http://localhost:8080/health
```

Stop recording (Ctrl+C).

### 3. Add a `status_code_class` assertion to the recorded test

Open the generated test YAML (e.g., `keploy/test-set-0/tests/test-1.yaml`) and add:

```yaml
assertions:
  status_code_class: "2xx"
```

### 4. Break the endpoint so it returns 500

```go
http.HandleFunc("/health", func(w http.ResponseWriter, r *http.Request) {
    w.WriteHeader(http.StatusInternalServerError) // now returns 500
    w.Write([]byte(`{"status":"error"}`))
})
```

### 5. Replay the test

```bash
keploy test -c "go run main.go"
```

### 6. Observe the result

- **Expected:** Test **FAILS** — the server returned `500`, which is class `"5xx"`, not the asserted `"2xx"`.
[...]

PR #4330, opened 2026-07-08
fix(matcher): use actual response status code in status_code_class assertion

vedhakoushik:
- Fixed the `status_code_class` assertion in `pkg/matcher/http/match.go` (`AssertionMatch`): `actualClass` was computed from a hardcoded literal `200` (`fmtSprintf234("%dxx", 200/100)`), so the assertion always evaluated against `"2xx"` regardless of what the server actually returned. Asserting `"2xx"` on a real `500` response incorrectly **passed**, and asserting `"5xx"` on a `500` incorrectly **failed** — the assertion type was effectively non-functional.
- One-line fix: compute the class from the real response — `fmtSprintf234("%dxx", actualResponse.StatusCode/100)`. The existing `classStr` normalization logic (handling both `"200"`-style and `"2xx"`-style user input) is untouched. `actualResponse` is already dereferenced unconditionally earlier in the same function, so no additional nil-guard is needed for consistency.
- Added two regression tests in `pkg/matcher/http/match_test.go` (the existing suite had **zero** coverage of `StatusCodeClass`, and all existing tests used actual status 200, which is why this bug was invisible):
  - actual `500` + assertion `"5xx"` → must pass
  - actual `500` + assertion `"2xx"` → must fail

## Links & References

**Closes:** #3843

### 🔗 Related PRs
- #3861 (earlier attempt at the same issue, inactive since Feb 25)
### 🐞 Related Issues
- #3843
### 📄 Related Documents
- NA

- [x] 🐞 Bug Fix
- [x] ✅ Test

## Added e2e test pipeline?

- [x] 👍 yes (doc comments on both regression tests explaining the guarded regression)

## Added to documentation?

- [x] 👍 yes, mentioned below

```
go test ./pkg/matcher/http/ -run TestAssertionMatch_StatusCodeClass -v
```

Both new tests fail on `main` before the fix and pass after it. Verified locally with `go build ./pkg/matcher/...` (clean) and the full `pkg/matcher/http` package tests (passing).

## Self Review done?
- [x] ✅ yes

- NA

### DIFF
--- a/pkg/matcher/http/match.go
+++ b/pkg/matcher/http/match.go
@@ -512,7 +512,7 @@ func AssertionMatch(tc *models.TestCase, actualResponse *models.HTTPResp, logger
 			} else {
 				classStr = class
 			}
-			actualClass := fmtSprintf234("%dxx", 200/100)
+			actualClass := fmtSprintf234("%dxx", actualResponse.StatusCode/100)
 			if classStr != actualClass {
 				pass = false
 				logger.Error("status_code_class assertion failed", zap.String("expected", class), zap.String("actual", actualClass))
--- a/pkg/matcher/http/match.go
+++ b/pkg/matcher/http/match.go
@@ -527,7 +527,7 @@ func AssertionMatch(tc *models.TestCase, actualResponse *models.HTTPResp, logger
 			} else {
 				classStr = class
 			}
-			actualClass := fmtSprintf234("%dxx", 200/100)
+			actualClass := fmtSprintf234("%dxx", actualResponse.StatusCode/100)
 			if classStr != actualClass {
 				pass = false
 				logger.Error("status_code_class assertion failed", zap.String("expected", class), zap.String("actual", actualClass))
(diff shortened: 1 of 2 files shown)


vedhakoushik:

eeshsaxena (commented):
Nice find - this looks correct to me. In `AssertionMatch` the `actualClass` was built from a hardcoded literal:

```go
actualClass := fmtSprintf234("%dxx", 200/100)   // always "2xx"
```

so `200/100` is a compile-time constant `2` and the `status_code_class` assertion always compared the expected class against `"2xx"`, ignoring the real response entirely. That means asserting `"2xx"` on a real `500` wrongly passed, and asserting `"5xx"` on a `500` wrongly failed - the assertion type was effectively inert. Switching to `actualResponse.StatusCode/100` (so `500 -> "5xx"`, `204 -> "2xx"`, etc.) is the right fix.

The two regression tests are well chosen - they cover both failure modes (the `"5xx"`-on-500 should-pass case and the `"2xx"`-on-500 should-fail case), which is exactly what pins the bug so it can't silently regress.

Deferring to the maintainers for the final call.

slayerjain:
Thanks @vedhakoushik

PR #4334, opened 2026-07-08
feat(telemetry): categorized stop_reason on graceful record/replay stops

gouravkrosx:
## What & why

Adds the CLI-side "where do users get stuck" signal for the **login → record → test** funnel (telemetry issue [keploy/telemetry#37](https://github.com/keploy/telemetry/issues/37)). Today the CLI records almost exclusively *successes*; graceful failures/abandonment emit nothing, so drop-off is invisible. This makes graceful stops carry a categorized reason — **without** duplicating Sentry (hard crashes stay there).

## Changes

**Record — one event type, `RecordSessionCompleted`:**
- Now emitted even when recording **never started** (pre-frame `setup_failed` / `no_traffic_frames` / `testset_lookup_failed`) — previously silent, the highest-signal "stuck" cases.
- Carries a categorized `stop_reason` alongside the existing `status`.

**Replay — new `TestRunAborted` event:**
- Fires when a run stops gracefully **before** its summary (setup/instrument/before-hook errors, a test set that errors out → `testset_run_failed`, or `keploy test` run before anything recorded → `no_testsets`).
- Normal pass/fail runs still emit `TestRun` as before. `TestRun (Run-Status=fail)` = red tests (a result); `TestRunAborted` = couldn't complete a run (stuck).

**Reason enum (`stopreason.go`):**
- `CategorizeStopReason` maps free-form reasons to a small, stable, LowCardinality set. Raw error text is **not** emitted (cardinality + Sentry overlap). Unit test pins every real record/replay string to its category.

## Scope / guarantees
- Graceful stops only — user Ctrl+C mid-replay and hard panics are excluded by design.
- `stop_reason` rides in `meta`, so it flows to ClickHouse via the existing pipeline with no ETL change; Superset funnel to follow.

### DIFF
--- a/pkg/models/tele.go
+++ b/pkg/models/tele.go
@@ -10,6 +10,7 @@ package models
 const (
 	TeleEventTestRun                = "TestRun"
 	TeleEventRecordSessionCompleted = "RecordSessionCompleted"
+	TeleEventTestRunAborted         = "TestRunAborted"
 )
 
 type TeleEvent struct {
--- a/pkg/platform/telemetry/stopreason.go
+++ b/pkg/platform/telemetry/stopreason.go
@@ -1,0 +1,64 @@
+package telemetry
+
+import "strings"
+
+// Stop-reason categories: a small, stable, low-cardinality enum describing
+// WHY a record/replay session ended. These are emitted only on graceful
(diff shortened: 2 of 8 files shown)


Aditya-eddy (approved):
LGTM

Issue #4336, opened 2026-07-08
go-memory-load-mongo flakes ~75%: memory-pressure mock drop is not atomic with test-case capture (orphaned TCs -> no_mocks at replay)


slayerjain:
## Summary

Job `run_golang_docker / go-memory-load-mongo` (config `record_*_replay_*`) fails on ~75% of runs on `main`. Under memory pressure the agent drops a test case's outgoing mock **but still persists the test case**, so replay reports `no_mocks` / EOF for that TC. Enough orphaned TCs accumulate that the aggregate replay failure rate exceeds the harness's 4% `MAX_FAIL_PERCENT`, failing the lane. The sibling `go-memory-load-grpc` lane has the identical defect.

## Root cause (not atomic)

The mock-drop and the test-case-keep decisions are made at different times against a shared, **pruned** piece of state (`pressureRanges`):

1. **Mock dropped** — `pkg/agent/proxy/syncMock/syncMock.go:390-400`
   `AddMock` drops any outgoing mock whose `Spec.ReqTimestampMock` lies inside an active pressure interval.
2. **Intended atomic guard is evaluated too late** — `pkg/agent/routes/record.go:372`
   `WasPressureActiveInWindow(HTTPReq.Timestamp, HTTPResp.Timestamp)` is recomputed only when the TC is streamed to the CLI. It suppresses the TC iff a pressure interval still overlaps the TC window.
3. **The join key is pruned** — `pkg/agent/proxy/syncMock/syncMock.go:785`
   `SetMemoryPressure` reaps pressure ranges older than 7s on every 500ms tick. Under load the TC-stream consumer lags **>7s** behind capture, so the pressure interval that dropped the mock is pruned **before** record.go recomputes. The overlap check then returns `false` → the orphaned TC is sent.

Net: `mock dropped` but `TC kept` → replay has a TC with no mock to serve it.

The guard added in #4220 (`WasPressureActiveInWindow`) is correct in principle but is evaluated at the wrong time — after the range it depends on can be pruned.

## Evidence

- Local Docker repro (`--memory-limit 90` to force pressure): **17 pause/resume cycles** during one k6 load — `Pausing keploy-agent recording due to memory pressure {memory_usage_bytes: 84914176 …}` / `Cleared keploy-agent memory pressure after memory recovered {memory_usage_bytes: 52797440, resume_threshold_bytes: 66060287}`. Each pause window is a drop window for `AddMock`.
[...]

PR #4338, opened 2026-07-08
fix(agent): suppress memory-pressure-orphaned test cases at window close

slayerjain:
## What

Addresses the `go-memory-load-mongo` flake (#4336): under memory pressure the memoryguard **drops a captured test case's outgoing mock** (`syncMock.AddMock`) while the **test case is still persisted** → replay hits `no_mocks`/EOF for that TC → >4% tolerance → fail.

#4220 added a late `WasPressureActiveInWindow` recompute in `record.go`, but it runs **after** a ~100-deep buffered `tcChan` and can lag past syncMock's **7 s pressure-range prune**, so the orphaned TC slips through the guard.

## Fix

A **synchronous window-close pressure gate** at the three incoming-HTTP capture sites (`handleHttp1Connection`, `handleHttp1ZeroCopy`, `parseStreamingHTTP`), placed **before** the `captureHookSem` acquire — where the pressure range is still fresh — so a TC whose window overlapped an active/just-closed pressure range is suppressed **together with** its dropped mock.

- **Lock-free fast path:** a set-once `pressureSeen atomic.Bool` latch in syncMock, stored **before** any range under the same lock. Healthy runs (pressure never trips) take **no lock and pay nothing** in the parser hot loop. Reviewed as sound: any mock drop happens-after `pressureSeen=true`, so skipping the check while unlatched can never miss a real drop.
- The existing Capture-level check stays as a **backstop** (documented as such).

## Honesty / scope

- This is a **race-window reduction**, not a hard atomicity guarantee (documented at the gate) — it removes the ~100-deep `tcChan` buffer + build + handoff from the check lag.
- During local repro I found a **separate** signal worth flagging: keploy incompletely-captures **large outgoing mongo mocks under a slow record-side consumer with *no* memory pressure at all** (`incomplete read of full message`). On the slow 2-vCPU CI runner that capture-throughput effect may also feed the flake, independent of this fix. I'm tracking that separately; **CI here is the real validator** of how much this fix moves the rate. Marked `Refs` (not `Fixes`) so the issue stays open until the CI rate is confirmed down.

## Tests

`pressure_atomicity_test.go` reproduces the **real** prune race (open+close a pressure range, record an in-window mock, sleep past the 7 s prune horizon, trigger the real reaper) and asserts the orphan is suppressed; it **fails when the gate term is reverted**. `pressure_gate_test.go` pins the gate decision table. `go test ./pkg/agent/proxy/incoming/... ./pkg/agent/hooks/conn/... ./pkg/agent/proxy/syncMock/... -race` green; no sleeps/retries/skips in product code.

Refs #4336

### DIFF
--- a/.github/workflows/gomem-atomic-drop-regression.yml
+++ b/.github/workflows/gomem-atomic-drop-regression.yml
@@ -1,0 +1,70 @@
+name: Gomem Atomic-Drop Regression (Linux)
+# Deterministic regression guard for issue #4336 — the go-memory-load-mongo
+# flake where the agent drops a captured test case's outgoing mock under memory
+# pressure while STILL persisting the test case, so replay reports
+# `mock mismatch ... no_mocks candidates:0` for it and the lane fails the 4%
+# tolerance.
+#
+# The fix makes a pressure mock-drop ATOMIC with its owning test case: every
+# pre-persistence drop (relay tee -> MarkMockIncomplete -> emitMockCore; AddMock's
+# pressure drop; SetMemoryPressure's buffer wipe) records the mock's request
+# timestamp in a never-pruned sync-mock ledger, and the TC-persist gates
+# (incoming/http.go, conn.Capture, routes/record.go) suppress any test case whose
+# HTTP window owns a dropped mock — so no persisted test case ever references a
+# mock that was dropped.
+#
+# WHY A UNIT-TEST LANE (not only the e2e go-memory-load-mongo lane): the drop is
+# only exercised when a memory-pressure PAUSE overlaps an in-flight large mock,
+# which on a fast multi-core box is rare (pauses recover before a chunk is gated)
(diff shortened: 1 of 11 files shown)


slayerjain:
Superseded by #4342. Root cause refined: the orphan-TC suppression from #4220 (`WasPressureActiveInWindow`) is correct in principle, but it races the 7s `pressureRanges` prune against `record.go`'s test-case-stream lag under load — so the pressure interval that caused a mock's drop is already reaped by the time the gate evaluates the TC, and the orphan is persisted anyway. #4342 records dropped mocks in a **never-pruned** ledger consulted by that same gate (closing the race regardless of how far behind `record.go` runs), and also instruments the dominant generic-parser `flushMock` early-drop that bypasses the emit path entirely. Closing this in favor of #4342.

PR #4353, opened 2026-07-11
fix(agent): suppress test cases whose mock was voided as incomplete (go-memory-load-mongo no_mocks)

slayerjain:
integration-ref: fix/mongo-v2-report-orphan-windows

## What

Closes the **actual** `go-memory-load-mongo (record_build_replay_latest)` flake — test cases reaching replay with an empty mock pool (`match_phase="no_mocks", candidates=0`).

Root cause (code-proven): under memory pressure the private mongo/v2 parser marks a mock **incomplete** on a reassembly overflow or a wire decode error (the realignment tail after a pressure gap). `supervisor.emitMockCore` then drops that mock **silently and tells no suppression layer**. The HTTP test case that owns the operation still streams to the CLI, so it persists mock-less and fails at replay.

Existing suppression covered two drop paths but not this one:
- memory-pressure drops → `pressureRanges` / `WasPressureActiveInWindow`
- outChan capacity drops → owning-TC ledger / `WasMockDroppedForTC` (#4352)
- **`MarkMockIncomplete` void → fed neither → orphan.** ← this PR

> Note: #4352 (capacity-drop suppression) was **not** the fix for this lane — CI proved the outChan path never fires here. This PR is.

## How

Record a non-pressure **orphan window** for a voided mock and suppress any TC whose HTTP `[req, resp]` window overlaps it — the same time-window join pressure suppression already uses.

- `syncMock`: `orphanRanges` (count-bounded, `m.mu`-guarded, amortized in-place compaction since voids are recorded per-mock) + `RecordOrphanWindow` / `WasMockOrphanedInWindow` / `OrphanRangeCount`.
- `supervisor.emitMockCore`: record the dropped mock's own wire window — **per-test mocks only** (a voided SESSION/CONNECTION mock is served from the session tier at replay; recording it would only over-suppress a healthy TC).
- `record.go`: OR `WasMockOrphanedInWindow` into the TC-suppression check; add `orphan_ranges_total` to the summary.
- New `Session.RecordOrphanWindow` lets a parser report the window of an op it voided **before any mock was built** (`MarkMockIncomplete` only voids the *next* mock). The companion parser change is **keploy/integrations#245** (wired here via `integration-ref` so CI validates both together).

All timestamps are wall-clock (`HTTPReq/Resp`, mongo `readAt/writtenAt`, pressure ranges all derive from `time.Now()`), so the overlap join is clock-consistent.

## Tests
Orphan-window overlap/bounds/degenerate guards; `emitMockCore` records the voided per-test mock's window (**mutation-verified** — neutering the record call turns the repro red); the SESSION/CONNECTION skip guard; a healthy-emit control. `go build/vet/test -race` green. Independent adversarial review (verdict SHIP).

### DIFF
--- a/pkg/agent/proxy/supervisor/orphan_window_test.go
+++ b/pkg/agent/proxy/supervisor/orphan_window_test.go
@@ -1,0 +1,160 @@
+package supervisor
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	syncMock "go.keploy.io/server/v3/pkg/agent/proxy/syncMock"
+	"go.keploy.io/server/v3/pkg/models"
+	"go.uber.org/zap/zaptest"
+)
+
+// TestEmitMockIncompleteRecordsOrphanWindow is the root-cause repro for the
+// go-memory-load-mongo no_mocks flake. When a parser marks its mock incomplete
+// (reassembly overflow, a decode error on the post-pressure realignment tail,
+// per-conn cap, short write) the NEXT emitMockCore drops the mock SILENTLY —
+// and before this fix it told NO ONE, so the test case that owns the operation
+// still streamed to the CLI and reached replay mock-less (match_phase=no_mocks).
+// The fix records the dropped mock's wire window as an orphan window so
+// record.go suppresses the owning TC. This drives the real EmitMock path.
+func TestEmitMockIncompleteRecordsOrphanWindow(t *testing.T) {
+	t.Parallel()
+	mgr := &syncMock.SyncMockManager{}
+	ch := make(chan *models.Mock, 1)
+	base := time.Now()
+	reqTs := base.Add(10 * time.Millisecond)
+	resTs := base.Add(30 * time.Millisecond)
+	sess := &Session{
+		Mocks:  ch,
+		Ctx:    context.Background(),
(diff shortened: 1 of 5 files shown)


slayerjain:
Closing per a measure-first decision after root-causing via the recording artifacts.

**Finding:** the `go-memory-load-mongo` failures are **real record orphans** — the failing TCs have no mongo mock on disk (unmapped in `mappings.yaml`). And it's **pure variance**: the same build recorder produced **9 orphans (pass, 2.79%)** and **46 orphans (fail, 11.83%)** on two separate recordings in the same run.

**Why this PR is correct but insufficient:** the orphan-window API + `emitMockCore`/parser suppression here correctly close the `MarkMockIncomplete` paths (reassembly overflow, decode error). But the **dominant** leak is the `mongo/v2` `encodeMongo`→`sendDecode` async-decode drop (`encode.go`), which builds no mock and calls no `MarkMockIncomplete` — so it feeds neither this orphan-window nor the pressure ranges. That drop point is also session-less, so suppressing it needs a different approach.

Branch kept. Will fold this orphan-window API into a single complete fix once record-phase telemetry confirms the exact dominant drop path. Companion keploy/integrations#245 closed too.

PR #4356, opened 2026-07-12
fix: de-flake mysql + gRPC fuzzer lanes

slayerjain:
De-flakes two intermittently-red fuzzer lanes (both fail on `main` too, pass on rerun). Root-caused from the failing job logs; each fix independently reviewed (SHIP, no CRITICAL/MAJOR).

## `fix(mysql)` — synthesized PREPARE_OK missing its param-def EOF (product bug)

`MySQL Fuzzer (record_build_replay_build)` hangs ~1000s then fails. During replay of the 4000-op fuzz, keploy synthesizes a `COM_STMT_PREPARE_OK` for an unmocked prepared statement but left `EOFAfterParamDefs` empty unconditionally. A client that did **not** negotiate `CLIENT_DEPRECATE_EOF` (go-sql-driver) calls `readUntilEOF()` after the parameter definitions and **blocks forever** without that terminator → the prepare deadlocks the connection until the app's run-context deadline fires → `driver: bad connection` cascades across the 32-conn pool → keploy's replay `POST /run` times out. Zero `no_mocks` — it is not an orphan, it is a malformed synthetic packet.

Fix: emit a legacy EOF packet after the param defs, gated on `numParams>0 && !DeprecateEOF()`, byte-symmetric with what the recorder captures (`record_v2.go`). Sequence id `2+numParams` follows the param defs (`2..1+numParams`) — verified contiguous against go-sql-driver's `ErrPktSync` check. No change for DEPRECATE_EOF clients, recorded prepares, or zero-param statements. New regression test drives the real `matchCommand` synthesize path (mutation-verified).

## `fix(ci)` — gRPC incoming fuzzer port-reuse race (harness bug)

`gRPC Fuzzer incoming (record_build_replay_build)` intermittently fails with 1000/1000 nil mismatches. The incoming path runs a yaml then a json record→replay cycle reusing `:50051` (gRPC server) and `:18080` (fuzzer HTTP client). The yaml client on `:18080` is never killed, and a fixed `sleep 2` races the yaml replay's server releasing `:50051`; the json cycle then hits `bind: address already in use`, keploy bails before capturing a test case, and the stale client's RPCs hit a dead server.

Fix: a bounded `wait_port_free` helper (poll `nc -z` until the socket is gone, wall-clock deadline) plus killing the leaked client, replacing the two blind sleeps. Deterministic and non-masking — a genuinely stuck port past the bound lets the authoritative bind error surface. keploy's own app-stop (process-group SIGINT→SIGKILL + synchronous drain) is robust; the async teardown is the harness's own `pkill`, so the wait is the correct compensation. Scope is incoming-only (the reported flake); outgoing is benign.

## Testing
- `go build` / `go vet` / `go test` green for the mysql replayer package; new `synthesize_prepare_eof_test.go` mutation-verified (neuter the gate → red).
- gRPC harness: `bash -n` clean; `wait_port_free` unit-behavior confirmed (free-immediately / free-on-release / timeout-at-bound).
- Independent adversarial review verified the mysql wire bytes + sequence id against the actual go-sql-driver source.

Independent of PR #4354 (the mongo tee-drop orphan-window fix); together the three close the record-robustness fuzzer/memory-load flakes.

### DIFF
--- a/.github/workflows/test_workflow_scripts/fuzzer/grpc/golang-linux.sh
+++ b/.github/workflows/test_workflow_scripts/fuzzer/grpc/golang-linux.sh
@@ -134,6 +134,31 @@ ensure_success_phrase() {
  exit 1
 }
 
+# wait_port_free NAME PORT [TIMEOUT_SEC]
+# Block until nothing is listening on localhost:PORT (or TIMEOUT_SEC elapses).
+# Deterministic replacement for a fixed `sleep` after killing a process that
+# owns a port: the next bind must not race a socket that is still being
+# released, which is what surfaces as "listen tcp :PORT: bind: address already
+# in use" and fails the run intermittently. Returns as soon as the port is
+# actually free, so it is normally sub-second. `nc -z` is already used elsewhere
+# in this script to poll for a port coming UP; here we poll for it going DOWN.
+wait_port_free() {
+  local name=$1 port=$2 timeout=${3:-30}
+  local deadline=$(( $(date +%s) + timeout ))
+  # Bare `nc -z` (no -w): on some netcat builds -w makes -z report an occupied
+  # port as free, which would defeat the whole check. Wall-clock deadline (not
+  # an iteration count) so a slow per-probe nc can't stretch the bound.
+  while nc -z localhost "$port" 2>/dev/null; do
+    if [ "$(date +%s)" -ge "$deadline" ]; then
+      echo "⚠️ $name still holding port $port after ${timeout}s"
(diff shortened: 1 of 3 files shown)


slayerjain:
Consolidated into #4354 as requested — single PR for all record-robustness lane de-flakes (mongo tee-drop + this PR's mysql prepare-EOF fix + gRPC harness fix). Both fuzzer lanes passed here (MySQL + gRPC-incoming, all variants) before consolidating.

Issue #2006, opened 2024-06-30
[bug]: keploy fails to run application in `macOS`


gouravkrosx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

If you try to run keploy in record mode to record some test cases for your application, keploy will fail to connect to your database due to **DNS** problem, it is occurring in `macOS`.

### Steps to reproduce

1. Install keploy latest version.
2. Take a sample app for eg: https://github.com/keploy/samples-java/tree/main/spring-petclinic
3. Try to run your application with keploy in record mode.

### Environment

None

### Version

Local

### Repository

keploy

PR #4361, opened 2026-07-13
fix(proxy/dns): relay NODATA instead of failing replay on empty AAAA answer

charankamarapu:
## Problem

During replay (test mode, mocking enabled), a DNS query that misses the recorded mocks **and** whose upstream forward returns **NODATA** (NOERROR with zero answer RRs) is escalated to `ErrMockNotFound` — failing the test:

```
Mock mismatch: [DNS] AAAA postgres.<ns>.svc.cluster.local.
Hint: DNS mocks may be missing. Re-record to capture DNS queries.
```

But NODATA is a **valid negative answer** ("the name exists, it just has no record of this type"), not a resolution failure.

## Why it bites everyone

`glibc`'s `getaddrinfo` always issues **A + AAAA in parallel**. For an IPv4-only cluster service (`postgres.<ns>.svc.cluster.local`, any ClusterIP service), the `AAAA` leg legitimately returns NODATA and the app connects over the A record. Failing the test on that empty `AAAA` breaks replay for essentially **any** app that resolves an in-cluster dependency by name. (Surfaced during the Checkr low-latency validation — keploy/api-server#1867.)

## Fix

When the upstream forward returns a **Success-rcode empty answer** (NODATA), relay it to the app as-is even when mocking is enabled, instead of raising `ErrMockNotFound`. The app receives the same empty `AAAA` it saw at record time and proceeds over IPv4.

Only **genuinely negative** answers (NXDOMAIN / SERVFAIL) still fall through to the proxy-IP steering that #2006 introduced — so the "don't relay NXDOMAIN for unknown bare service names like `localstack`/`postgres`" behavior is **unchanged**. The distinction is exactly `rcode == RcodeSuccess`.

## Note on replay strictness

Relaying NODATA (empty NOERROR) with mocking enabled means a genuine recording gap on an AAAA (or A) record is now relayed as a valid empty answer rather than surfaced as a mock-miss. This slightly loosens replay strictness and is a deliberate trade for this bug (apps that resolve in-cluster dependencies by name fail replay otherwise). Genuinely-absent names still return NXDOMAIN and are still steered to the proxy IP (#2006). See TestResolveUncachedDNSResponse_TestMode_UpstreamNODATA_AQuery_RelayedNotSteered.

Dual-NODATA edge: a name returning NODATA on *both* A and AAAA and not mocked will now yield EAI_NODATA (no address) rather than being steered on the A leg. This is intended for the mocked-service case — a genuinely-needed unmocked dependency resolves to NXDOMAIN (still steered), not NODATA. Covered by TestResolveUncachedDNSResponse_TestMode_UpstreamNODATA_DualEmpty.

Resolver-config caveat: the "genuinely-needed unmocked names return NXDOMAIN (still steered), not NODATA" assumption holds for bare names on most resolvers, but a search-domain / ndots setup (systemd-resolved, k8s pod DNS) can qualify a name into a zone that answers NODATA — such a name now receives an honest empty answer instead of being steered to the proxy IP. Narrow, but validate against the target cluster resolver config if steering unmocked in-cluster names is relied upon.

### DIFF
--- a/pkg/agent/proxy/dns.go
+++ b/pkg/agent/proxy/dns.go
@@ -270,6 +270,20 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
 			// Unlike NXDOMAIN there is no unknown name to steer to the proxy IP,
 			// and raising ErrMockNotFound here would fail replay for essentially
 			// every app that resolves an in-cluster dependency by name.
+			//
+			// This relay is deliberately qtype-AGNOSTIC (it also covers an A
+			// NODATA), not restricted to AAAA — the symmetric IPv6-only case
+			// (empty A → app falls back to AAAA) is a real scenario. It is safe
+			// w.r.t. the #2006 proxy-IP steering because steering exists for
+			// names that DON'T RESOLVE: a name the app genuinely needs but that
+			// isn't recorded comes back as NXDOMAIN (name absent → still steered
+			// below) or a real A. NODATA means the name EXISTS with no record of
+			// this type; fabricating an A→proxy-IP for it would make the app dial
+			// the proxy for a service that authoritatively has no A record, which
+			// is less correct than relaying the truthful empty answer. The only
+			// loosened case is an IPv4-only app hitting a bare name that returns
+			// A NODATA — a narrow window that previously got steered and now gets
+			// an honest empty answer (see the A-query regression test).
 			if fwdResp.Rcode == dns.RcodeSuccess {
 				p.logger.Debug("DNS mock miss: upstream NODATA (empty NOERROR), relaying as valid negative answer",
 					zap.String("query", question.Name),
@@ -288,17 +302,16 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
 			// (relaying NXDOMAIN crashes apps using bare service names
 			// like "localstack" or "postgres" — issue #2006).
 			if !mockingEnabled {
-				p.logger.Debug("DNS mock miss: upstream returned negative/empty answer; mocking disabled, returning upstream response",
+				p.logger.Debug("DNS mock miss: upstream returned negative answer; mocking disabled, returning upstream response",
 					zap.String("query", question.Name),
 					zap.String("qtype", dns.TypeToString[question.Qtype]),
 					zap.Int("rcode", fwdResp.Rcode))
 				return dnsCacheEntry{Msg: fwdResp, FromUpstream: true}
 			}
-			p.logger.Debug("DNS mock miss: upstream returned negative/empty answer; falling back to synthetic DNS response",
+			p.logger.Debug("DNS mock miss: upstream returned negative answer; falling back to synthetic DNS response",
 				zap.String("query", question.Name),
 				zap.String("qtype", dns.TypeToString[question.Qtype]),
-				zap.Int("rcode", fwdResp.Rcode),
-				zap.Int("answers", len(fwdResp.Answer)))
+				zap.Int("rcode", fwdResp.Rcode))
 		} else if fwdErr != nil {
 			p.logger.Debug("DNS mock miss + upstream forward failed; falling back to synthetic response",
 				zap.String("query", question.Name),
--- a/pkg/agent/proxy/dns.go
+++ b/pkg/agent/proxy/dns.go
@@ -260,32 +260,23 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
 					zap.Int("answers", len(fwdResp.Answer)))
 				return dnsCacheEntry{Msg: fwdResp, FromUpstream: true}
 			}
-			// NODATA — upstream returned NOERROR with zero answer RRs. This is
-			// a *valid* negative answer ("the name exists but has no record of
-			// this type"), not a resolution failure, so relay it as-is even
-			// when mocking is enabled. The canonical case is an AAAA query for
-			// an IPv4-only cluster service (e.g. postgres.<ns>.svc.cluster.local):
-			// glibc's getaddrinfo always issues A+AAAA in parallel, the AAAA leg
-			// legitimately has no record, and the app connects over the A record.
-			// Unlike NXDOMAIN there is no unknown name to steer to the proxy IP,
-			// and raising ErrMockNotFound here would fail replay for essentially
-			// every app that resolves an in-cluster dependency by name.
+			// NODATA — upstream returned NOERROR with zero answer RRs: a *valid*
+			// negative answer ("the name exists but has no record of this type"),
+			// distinct from NXDOMAIN. Relay it as-is even with mocking enabled.
+			// The distinction that matters: NXDOMAIN (name absent) still falls
+			// through to #2006 proxy-IP steering below; NODATA does not, because
+			// fabricating an A→proxy-IP for a name that authoritatively has no
+			// such record is less correct than the honest empty answer. Kept
+			// qtype-agnostic on purpose (the IPv6-only empty-A case is symmetric).
+			// See the PR description and the AAAA / A-query regression tests for
+			// the full rationale and the deliberately-loosened bare-name-A case.
 			//
-			// This relay is deliberately qtype-AGNOSTIC (it also covers an A
-			// NODATA), not restricted to AAAA — the symmetric IPv6-only case
-			// (empty A → app falls back to AAAA) is a real scenario. It is safe
-			// w.r.t. the #2006 proxy-IP steering because steering exists for
-			// names that DON'T RESOLVE: a name the app genuinely needs but that
-			// isn't recorded comes back as NXDOMAIN (name absent → still steered
-			// below) or a real A. NODATA means the name EXISTS with no record of
-			// this type; fabricating an A→proxy-IP for it would make the app dial
-			// the proxy for a service that authoritatively has no A record, which
-			// is less correct than relaying the truthful empty answer. The only
-			// loosened case is an IPv4-only app hitting a bare name that returns
-			// A NODATA — a narrow window that previously got steered and now gets
-			// an honest empty answer (see the A-query regression test).
+			// Logged at Info, not Debug: this is the one path that relays a
+			// possibly-genuine recording gap as a valid empty answer instead of a
+			// mock miss, so it must stay discoverable in a replay report rather
+			// than vanish into debug output.
 			if fwdResp.Rcode == dns.RcodeSuccess {
-				p.logger.Debug("DNS mock miss: upstream NODATA (empty NOERROR), relaying as valid negative answer",
+				p.logger.Info("DNS mock miss: upstream NODATA (empty NOERROR), relaying as valid negative answer (a genuinely unrecorded record of this type would also relay empty here)",
 					zap.String("query", question.Name),
 					zap.String("qtype", dns.TypeToString[question.Qtype]))
 				return dnsCacheEntry{Msg: fwdResp, FromUpstream: true}
--- a/pkg/agent/proxy/dns.go
+++ b/pkg/agent/proxy/dns.go
@@ -276,8 +276,10 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
 					zap.String("qtype", dns.TypeToString[question.Qtype]))
 				return dnsCacheEntry{Msg: fwdResp, FromUpstream: true}
 			}
-			// Upstream returned a genuinely negative answer (NXDOMAIN or
-			// SERVFAIL).
+			// Upstream returned a genuinely negative answer: NXDOMAIN,
+			// SERVFAIL, or any other non-success rcode (REFUSED, NOTIMP,
+			// FORMERR, …) — everything except the RcodeSuccess cases handled
+			// above.
 			// When mocking is disabled the operator wants real DNS
 			// behaviour, so honour the upstream negative answer as-is.
 			// When mocking is enabled, a negative answer must NOT reach
--- a/pkg/agent/proxy/dns.go
+++ b/pkg/agent/proxy/dns.go
@@ -260,8 +260,24 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
 					zap.Int("answers", len(fwdResp.Answer)))
 				return dnsCacheEntry{Msg: fwdResp, FromUpstream: true}
 			}
-			// Upstream returned a negative or empty answer (NXDOMAIN,
-			// SERVFAIL, or success with no RRs).
+			// NODATA — upstream returned NOERROR with zero answer RRs. This is
+			// a *valid* negative answer ("the name exists but has no record of
+			// this type"), not a resolution failure, so relay it as-is even
+			// when mocking is enabled. The canonical case is an AAAA query for
+			// an IPv4-only cluster service (e.g. postgres.<ns>.svc.cluster.local):
+			// glibc's getaddrinfo always issues A+AAAA in parallel, the AAAA leg
+			// legitimately has no record, and the app connects over the A record.
+			// Unlike NXDOMAIN there is no unknown name to steer to the proxy IP,
+			// and raising ErrMockNotFound here would fail replay for essentially
+			// every app that resolves an in-cluster dependency by name.
+			if fwdResp.Rcode == dns.RcodeSuccess {
+				p.logger.Debug("DNS mock miss: upstream NODATA (empty NOERROR), relaying as valid negative answer",
+					zap.String("query", question.Name),
+					zap.String("qtype", dns.TypeToString[question.Qtype]))
+				return dnsCacheEntry{Msg: fwdResp, FromUpstream: true}
+			}
+			// Upstream returned a genuinely negative answer (NXDOMAIN or
+			// SERVFAIL).
 			// When mocking is disabled the operator wants real DNS
 			// behaviour, so honour the upstream negative answer as-is.
 			// When mocking is enabled, a negative answer must NOT reach
--- a/pkg/agent/proxy/dns.go
+++ b/pkg/agent/proxy/dns.go
@@ -250,32 +250,42 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
 		// NXDOMAIN / synthetic fallback. Forwarding is strictly
 		// additive.
 		if fwdResp, fwdErr := p.forwardDNSUpstream(question); fwdErr == nil && fwdResp != nil {
-			// Only accept the upstream answer when it actually resolved
-			// the name (RcodeSuccess with at least one Answer RR).
-			if fwdResp.Rcode == dns.RcodeSuccess && len(fwdResp.Answer) > 0 {
-				p.logger.Debug("DNS mock miss resolved via upstream forward",
-					zap.String("query", question.Name),
-					zap.String("qtype", dns.TypeToString[question.Qtype]),
-					zap.Int("rcode", fwdResp.Rcode),
-					zap.Int("answers", len(fwdResp.Answer)))
-				return dnsCacheEntry{Msg: fwdResp, FromUpstream: true}
-			}
-			// NODATA — upstream returned NOERROR with zero answer RRs: a *valid*
-			// negative answer ("the name exists but has no record of this type"),
-			// distinct from NXDOMAIN. Relay it as-is even with mocking enabled.
-			// The distinction that matters: NXDOMAIN (name absent) still falls
-			// through to #2006 proxy-IP steering below; NODATA does not, because
-			// fabricating an A→proxy-IP for a name that authoritatively has no
-			// such record is less correct than the honest empty answer. Kept
-			// qtype-agnostic on purpose (the IPv6-only empty-A case is symmetric).
-			// See the PR description and the AAAA / A-query regression tests for
-			// the full rationale and the deliberately-loosened bare-name-A case.
-			//
-			// Logged at Info, not Debug: this is the one path that relays a
-			// possibly-genuine recording gap as a valid empty answer instead of a
-			// mock miss, so it must stay discoverable in a replay report rather
-			// than vanish into debug output.
+			// A RcodeSuccess response splits into two cases by answer count:
+			// resolved (>=1 RR) vs NODATA (0 RRs). Both are relayed as-is; only
+			// a non-success rcode falls through to the #2006 steering below.
 			if fwdResp.Rcode == dns.RcodeSuccess {
+				if len(fwdResp.Answer) > 0 {
+					// Resolved — relay the real answer.
+					p.logger.Debug("DNS mock miss resolved via upstream forward",
+						zap.String("query", question.Name),
+						zap.String("qtype", dns.TypeToString[question.Qtype]),
+						zap.Int("rcode", fwdResp.Rcode),
+						zap.Int("answers", len(fwdResp.Answer)))
+					return dnsCacheEntry{Msg: fwdResp, FromUpstream: true}
+				}
+				// NODATA — NOERROR + zero answer RRs: a *valid* negative answer
+				// ("the name exists but has no record of this type"), distinct
+				// from NXDOMAIN. Relay it as-is even with mocking enabled. Only
+				// NXDOMAIN (name absent) falls through to #2006 proxy-IP steering
+				// below; fabricating an A→proxy-IP for a name that authoritatively
+				// has no such record would be less correct than the honest empty
+				// answer. Kept qtype-agnostic on purpose (the IPv6-only empty-A
+				// case is symmetric). See issue #2006 and the regression tests
+				// TestResolveUncachedDNSResponse_TestMode_UpstreamNODATA_* for the
+				// deliberately-loosened bare-name-A case.
+				//
+				// Caveat (resolver-config-dependent): the "genuinely-needed
+				// unmocked names return NXDOMAIN, not NODATA" assumption holds for
+				// bare names on most resolvers, but a search-domain / ndots setup
+				// (systemd-resolved, k8s pod DNS) can qualify a name into a zone
+				// that answers NODATA — such a name would now get an empty answer
+				// instead of being steered. Narrow; validate against the target
+				// cluster resolver config if steering unmocked in-cluster names is
+				// relied upon.
+				//
+				// Logged at Info, not Debug: this is the one path that relays a
+				// possibly-genuine recording gap as a valid empty answer instead
+				// of a mock miss, so it stays discoverable in a replay report.
 				p.logger.Info("DNS mock miss: upstream NODATA (empty NOERROR), relaying as valid negative answer (a genuinely unrecorded record of this type would also relay empty here)",
 					zap.String("query", question.Name),
 					zap.String("qtype", dns.TypeToString[question.Qtype]))
--- a/pkg/agent/proxy/dns.go
+++ b/pkg/agent/proxy/dns.go
@@ -283,12 +283,29 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
 				// cluster resolver config if steering unmocked in-cluster names is
 				// relied upon.
 				//
-				// Logged at Info, not Debug: this is the one path that relays a
-				// possibly-genuine recording gap as a valid empty answer instead
-				// of a mock miss, so it stays discoverable in a replay report.
-				p.logger.Info("DNS mock miss: upstream NODATA (empty NOERROR), relaying as valid negative answer (a genuinely unrecorded record of this type would also relay empty here)",
-					zap.String("query", question.Name),
-					zap.String("qtype", dns.TypeToString[question.Qtype]))
+				// Logging: this branch is on the HOT path — glibc fires A+AAAA for
+				// essentially every hostname and every IPv4-only cluster service
+				// yields an AAAA NODATA here — so an unconditional Info would spam.
+				//   - mocking disabled: not a "mock miss" at all (the operator
+				//     wants real DNS); log at Debug.
+				//   - mocking enabled: keep it discoverable (a genuinely unrecorded
+				//     record of this type also relays empty here) but dedupe to
+				//     once per (name,qtype) so the signal survives without the spam.
+				if mockingEnabled {
+					if _, dup := p.nodataRelayLogged.LoadOrStore(generateCacheKey(question.Name, question.Qtype), struct{}{}); !dup {
+						p.logger.Info("DNS mock miss: upstream NODATA (empty NOERROR), relaying as valid negative answer (a genuinely unrecorded record of this type would also relay empty here)",
+							zap.String("query", question.Name),
+							zap.String("qtype", dns.TypeToString[question.Qtype]))
+					} else {
+						p.logger.Debug("DNS mock miss: upstream NODATA, relaying (already logged for this name/qtype)",
+							zap.String("query", question.Name),
+							zap.String("qtype", dns.TypeToString[question.Qtype]))
+					}
+				} else {
+					p.logger.Debug("DNS NODATA (empty NOERROR); mocking disabled, relaying real upstream answer",
+						zap.String("query", question.Name),
+						zap.String("qtype", dns.TypeToString[question.Qtype]))
+				}
 				return dnsCacheEntry{Msg: fwdResp, FromUpstream: true}
 			}
 			// Upstream returned a genuinely negative answer: NXDOMAIN,
--- a/pkg/agent/proxy/dns_forward_test.go
+++ b/pkg/agent/proxy/dns_forward_test.go
@@ -521,6 +521,51 @@ func TestResolveUncachedDNSResponse_TestMode_UpstreamNODATA_AQuery_RelayedNotSte
 	}
 }
 
+// TestGenerateCacheKey_QtypeIsolated locks the invariant that makes the
+// qtype-agnostic NODATA relay safe: an AAAA NODATA cached under one key must not
+// be served for a following A query on the same name. The cache key includes the
+// qtype today; this guards it against a future cache-key refactor.
+func TestGenerateCacheKey_QtypeIsolated(t *testing.T) {
+	name := "postgres.checkr.svc.cluster.local"
+	aKey := generateCacheKey(name, dns.TypeA)
+	aaaaKey := generateCacheKey(name, dns.TypeAAAA)
+	if aKey == aaaaKey {
+		t.Fatalf("A and AAAA cache keys must differ (else an AAAA NODATA would be served for an A query); both = %q", aKey)
+	}
+}
+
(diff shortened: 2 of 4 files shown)


charankamarapu:
## Principal-engineer review

**Verdict: Approve with nits.** Small (18/-2, single file `pkg/agent/proxy/dns.go`), correct, and conservatively scoped. Builds and `go vet` clean. The one real gap is the absence of a test, which is out of step with this file's otherwise strong test discipline.

### RCA soundness — solid
On a mock miss in `MODE_TEST` with mocking enabled, keploy forwards upstream (`forwardDNSUpstream`, `dns.go:252`). Before this PR, only `RcodeSuccess && len(Answer) > 0` was accepted (`dns.go:255`); everything else — including a legitimate NODATA (NOERROR + 0 RRs) — fell through to `SendError(ErrMockNotFound)` (`dns.go:315-324`), which fails the replay. glibc `getaddrinfo` issuing A+AAAA in parallel against an IPv4-only ClusterIP service is exactly the reproducer. The fix short-circuits the NODATA case (`dns.go:273-278`) before `SendError`. RCA and fix mechanism match the code.

### Verified
- **Rcode handling — correct.** `RcodeSuccess`+answers → relay real answer; `RcodeSuccess`+0 answers → NODATA relay; NXDOMAIN/SERVFAIL/REFUSED/NOTIMP/etc. skip both and hit the "genuinely negative" branch (mocking-off → relay; mocking-on → `SendError` + proxy IP). `RcodeSuccess` is exactly the right dividing line.
- **"Relay as-is" preserves ID / question / flags — correct, and better than a naive forward.** The client reply is built via `msg.SetReply(r)` (`dns.go:167`), so ID and Question come from the *client request* (upstream's random ID correctly discarded), then AA/RA/TC/Rcode and Answer/Ns/Extra are copied from the resolved entry. For NODATA this relays the empty answer + upstream SOA in the authority section — exactly what a resolver expects. No ID-mismatch risk.
- **No regression to the mocked path — confirmed.** `getMockedDNSResponse` returns early at `dns.go:227`; all new code is downstream of the mock miss.
- **#2006 (NXDOMAIN → proxy IP) intact — confirmed.** NXDOMAIN is `RcodeNameError` (3) ≠ `RcodeSuccess`, bypasses the new branch, still steers to proxy IP. Existing test `TestResolveUncachedDNSResponse_TestMode_UpstreamNXDOMAIN_FallsBackToProxyIP` unaffected.
- **No nil-deref:** `fwdResp` guarded non-nil at `dns.go:252`; `len(nil)==0`; returning `fwdResp` is safe.
- **Proxy-mode userspace DNS only; record path (`MODE_RECORD`) untouched.**
- **No infinite loop / repeated upstream calls** — the relayed NODATA is `FromUpstream: true`, so `shouldCacheDNSResponse` caches it for the 30s TTL; repeat A+AAAA queries hit the cache. Good.

### Findings (ranked)
1. **[Should-fix] Missing unit test.** This file has a dense, behavior-specific suite (`dns_forward_test.go`), yet the new NODATA path has none. Add `TestResolveUncachedDNSResponse_TestMode_UpstreamNODATA_RelayedAsIs`: fake upstream returns `RcodeSuccess` + zero `Answer` (ideally AAAA with an SOA in `Ns`); assert the entry has `Rcode==RcodeSuccess`, `len(Answer)==0`, `FromUpstream==true`, SOA preserved in `Ns`, and — critically — no `ErrMockNotFound`. Mirror the existing `startFakeUpstream`/`newProxyWithUpstream` helpers.
2. **[Nit] Comment is non-exhaustive.** `dns.go:279-280` labels the fall-through "NXDOMAIN or SERVFAIL", but it also catches REFUSED/NOTIMP/FORMERR. Behavior is fine (unchanged, conservative); reword to "NXDOMAIN, SERVFAIL, or any other non-success rcode".
3. **[Nit, pre-existing, out of scope] `forwardDNSUpstream` doc drift.** Its comment (`dns_forward.go:123`) says it returns "the first Successful response," but it returns the first non-nil response regardless of rcode — which is precisely why the caller must discriminate. This PR leans on that; worth a one-line doc fix while nearby.

**Net:** correct, minimal, no regression to the mocked path or #2006, no nil-deref, free re-query protection via the cache. Only hold before merge is a unit test for the NODATA relay; comment tweaks optional.

<sub>🤖 Review generated with [Claude Code](https://claude.com/claude-code)</sub>

charankamarapu:
Replay failed on a legitimate empty `AAAA` (NODATA) — relayed as `ErrMockNotFound` instead of a valid negative answer, breaking any app that resolves an IPv4-only cluster service by name. Relay NOERROR+0-answers as-is; NXDOMAIN behaviour (#2006) unchanged.

---
Part of the **Checkr low-latency + native-sidecar Istio** validation — tracking issue: keploy/api-server#1867.

**Full fix set** (together, proxyless low-latency now records **and** replays Postgres, Redis, MySQL, Mongo, DNS, and gRPC — ingress *and* egress — on native-sidecar Istio):
- keploy/enterprise#2246 — gRPC/HTTP2 **ingress** framer
- keploy/enterprise#2247 — Istio `:15020` health-probe noise (proxyless pre-existing-PID filter)
- keploy/enterprise#2248 — proxyless **DNS** pairing (`(tgid,txid)`, not the socket cookie)
- keploy/enterprise#2249 — Postgres **SSLRequest** TLS-misdetection
- keploy/keploy#4361 — replay **NODATA / empty-AAAA** relay
- keploy/enterprise#2250 — HTTP/2 & gRPC **egress** capture (V2 recorder)
- keploy/integrations#247 — gRPC egress **mock-kind** classification

**End-to-end validation with all fixes applied:** SaaS local `keploy cloud replay` → **492/492** green across 17 test-sets (incl. the gRPC-egress recording); in-cluster `--trigger` → 17/18 (the single miss is an *older* pre-fix recording).

charankamarapu:
Addressed in `3a5f29ed`.

- **#1 (Should-fix — missing test)** — added `TestResolveUncachedDNSResponse_TestMode_UpstreamNODATA_RelayedAsIs`, mirroring the existing `startFakeUpstream`/`newProxyWithUpstream` helpers. The fake upstream returns NOERROR + zero answers + an SOA authority record (the exact shape for an AAAA query on an IPv4-only cluster service); the test asserts `Rcode==RcodeSuccess`, `len(Answer)==0`, SOA preserved in `Ns`, `FromUpstream==true`, and — critically — that it is **not** steered to the proxy IP / `ErrMockNotFound`.
- **#2 (Nit — non-exhaustive comment)** — reworded to "NXDOMAIN, SERVFAIL, or any other non-success rcode (REFUSED, NOTIMP, FORMERR, …)".
- **#3 (Nit — `forwardDNSUpstream` doc drift)** — fixed: it returns the first **non-nil** response regardless of rcode (which is precisely why the caller must discriminate), not the "first Successful response".

`go test ./pkg/agent/proxy/ -run TestResolveUncachedDNSResponse` green. Both commits are DCO signed-off.

pkg/agent/proxy/dns.go
```diff
@@ -260,8 +260,26 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
...
+			// a *valid* negative answer ("the name exists but has no record of
+			// this type"), not a resolution failure, so relay it as-is even
+			// when mocking is enabled. The canonical case is an AAAA query for
+			// an IPv4-only cluster service (e.g. postgres.<ns>.svc.cluster.local):
+			// glibc's getaddrinfo always issues A+AAAA in parallel, the AAAA leg
+			// legitimately has no record, and the app connects over the A record.
+			// Unlike NXDOMAIN there is no unknown name to steer to the proxy IP,
+			// and raising ErrMockNotFound here would fail replay for essentially
+			// every app that resolves an in-cluster dependency by name.
+			if fwdResp.Rcode == dns.RcodeSuccess {
```
charankamarapu, line 256: This branch relays *any* success+empty answer regardless of qtype, but the justification (and test) is entirely about AAAA. Consider the #2006 steering contract for a **bare service name on an A query**: previously a non-(success+answer) upstream response with mocking enabled fell through to proxy-IP steering so eBPF could intercept; now, if upstream returns NODATA for that A query, we relay an empty A and the app gets *no address* — no steering, and no `ErrMockNotFound` either. In practice a truly-needed bare name usually returns NXDOMAIN (still steered) or a real A, so this is a narrow window, and the IPv6-only symmetric case (empty A → app uses AAAA) is a real reason not to restrict to AAAA. But this is a behavioral change to the #2006 path that isn't covered by a test. Can you either (a) add a note here on why relaying empty A is safe w.r.t. steering, and (b) add a regression test asserting a bare-name A NODATA still behaves acceptably? Just want the decision to be explicit rather than incidental.

pkg/agent/proxy/dns.go
```diff
@@ -260,8 +260,26 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
...
+				p.logger.Debug("DNS mock miss: upstream NODATA (empty NOERROR), relaying as valid negative answer",
+					zap.String("query", question.Name),
+					zap.String("qtype", dns.TypeToString[question.Qtype]))
+				return dnsCacheEntry{Msg: fwdResp, FromUpstream: true}
+			}
+			// Upstream returned a genuinely negative answer: NXDOMAIN,
```
charankamarapu, line 311: Nit: now that success+empty (NODATA) is handled by the branch above, this "genuinely negative" path can only be reached with a **non-success rcode**. But the fall-through log lines just below still say "negative/empty answer" and log `zap.Int("answers", len(fwdResp.Answer))` (always 0 here), and the older comments still mention "or success with no RRs". Those are now vestigial — tightening them to "negative answer" and dropping the answer count would keep the logs honest about what actually reaches this point.

charankamarapu (commented):
## Review — principal engineer

**What this does:** On a replay-mode DNS mock miss, after forwarding upstream, the code now relays a NODATA answer (NOERROR + zero answer RRs) to the app as-is even when mocking is enabled, rather than escalating to `ErrMockNotFound`. Genuinely negative rcodes (NXDOMAIN/SERVFAIL/…) still fall through to the #2006 proxy-IP steering. `dns_forward.go` is a doc-only comment fix, and the added test is a focused regression test.

**Assessment:** The core distinction is correct. Since the `RcodeSuccess && len(Answer) > 0` branch returns just above, reaching `if fwdResp.Rcode == dns.RcodeSuccess` guarantees success-with-zero-answers, which is exactly NODATA (RFC 2308) — no CNAME/referral can slip through because those would be Answer RRs. The `dns_forward.go` comment change is purely accuracy (the function already returned non-success responses; the old "first Successful response" doc was stale). The test asserts rcode, answer count, SOA preservation, and `FromUpstream` — good coverage for the happy path.

**Verdict:** Approach is sound and the fix is well-justified. Two things worth addressing before merge (inline): the scope of the relay vs. the #2006 steering guarantee for non-AAAA queries, and now-stale "empty" wording in the fall-through path. Neither is a hard blocker but the first deserves an explicit decision + test.

**Minor considerations (non-blocking):**
- Relaying NODATA with mocking enabled means a genuine recording gap on an AAAA (or other) record is now silently relayed rather than surfaced as a mock-miss. That's an acceptable trade for this bug, but it slightly loosens replay strictness — worth a line in the PR description acknowledging it.
- The relayed NODATA is cached `FromUpstream` for the fixed 30s window (pre-existing behavior) rather than honoring the SOA negative-TTL. Fine to leave, just noting it.

charankamarapu:
Thanks — addressed in `91a7ceef`.

- **Inline #1 (relay scope vs #2006 steering for non-AAAA)** — made the decision explicit rather than incidental:
  - Added a comment at the branch explaining the relay is **deliberately qtype-agnostic** (the symmetric IPv6-only case — empty A → app falls back to AAAA — is real, so restricting to AAAA would be wrong), and why relaying an empty A is safe w.r.t. steering: steering exists for names that **don't resolve** (NXDOMAIN → still steered); NODATA means the name **exists** with no record of this type, so fabricating an A→proxy-IP would make the app dial the proxy for a service that authoritatively has no A record — less correct than the honest empty answer.
  - Added `TestResolveUncachedDNSResponse_TestMode_UpstreamNODATA_AQuery_RelayedNotSteered`: a bare-name (`localstack.`) **A** query returning NODATA is relayed empty (RcodeSuccess, 0 answers, FromUpstream) and **not** steered to the synthetic proxy IP — locking in the one loosened case as a tested decision.
- **Inline #2 (vestigial "empty" wording)** — the fall-through path is now only reachable with a non-success rcode, so reworded both log lines from "negative/empty answer" → "negative answer" and dropped the always-zero `answers` count.
- **Non-blocking #1 (loosened replay strictness)** — acknowledged in the PR description now.
- **Non-blocking #2 (30s cache vs SOA negative-TTL)** — left as-is (pre-existing behavior), noted.

`go test ./pkg/agent/proxy/ -run TestResolveUncachedDNSResponse` green (6/6). Commit DCO signed-off.

charankamarapu (commented):
Both points addressed — qtype-agnostic decision now explicit with an A-query regression test, and log wording tidied. LGTM pending green CI.

charankamarapu (commented):
## Review — relay NODATA instead of failing replay on empty answer

Strong, well-motivated fix. The root cause is correctly identified (glibc `getaddrinfo` fires A+AAAA in parallel, the AAAA leg of an IPv4-only cluster service legitimately returns NODATA), and the fix is minimal and surgical: the only new decision point is `Rcode == RcodeSuccess` for the empty-answer case. Verified the important correctness invariant — `generateCacheKey(name, qtype)` includes the qtype, so caching an AAAA NODATA cannot leak into a subsequent A query. Tests are focused and pin both the fix (AAAA relay) and the deliberate scope change (A NODATA not steered), including the SOA-preservation and `FromUpstream` invariants. LGTM overall; a few points below to consider before merge.

### Strengths
- Correct RFC semantics: NODATA (NOERROR + 0 answers) is a valid negative answer, distinct from NXDOMAIN. The `rcode == RcodeSuccess` discriminator is exactly right.
- The `#2006` steering behavior for genuinely-absent names (NXDOMAIN/SERVFAIL) is preserved, and the existing NXDOMAIN test still guards it.
- The self-documenting comment and PR note make the strictness trade-off explicit rather than silent.

### Points to consider
1. **Silent loosening of replay strictness** (see inline). A genuine recording gap on an A/AAAA record is now relayed as a valid empty answer at `Debug` level rather than surfaced. Since this is the one case that trades correctness for compatibility, consider raising the log to `Info` or adding a metric/counter so a genuinely-missing record is discoverable in a replay report instead of vanishing into debug logs.
2. **Dual-NODATA edge case** (see inline). For a name that returns NODATA on *both* A and AAAA and is not mocked, `getaddrinfo` now fails with EAI_NODATA and the app can't connect at all — previously the A leg was steered to the proxy IP. This is narrow (unmocked names usually come back NXDOMAIN, not NODATA) and the PR acknowledges it, but it's worth a sentence in the note confirming it's intended for the mocked-service case too.
3. **Comment length** (see inline) — minor; the 24-line block duplicates the PR description.

None of these are blocking. Nice fix.

pkg/agent/proxy/dns.go
```diff
@@ -260,8 +260,40 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
...
+			// names that DON'T RESOLVE: a name the app genuinely needs but that
+			// isn't recorded comes back as NXDOMAIN (name absent → still steered
+			// below) or a real A. NODATA means the name EXISTS with no record of
+			// this type; fabricating an A→proxy-IP for it would make the app dial
+			// the proxy for a service that authoritatively has no A record, which
+			// is less correct than relaying the truthful empty answer. The only
+			// loosened case is an IPv4-only app hitting a bare name that returns
+			// A NODATA — a narrow window that previously got steered and now gets
+			// an honest empty answer (see the A-query regression test).
+			if fwdResp.Rcode == dns.RcodeSuccess {
```
charankamarapu, line 256: Two things about this relay branch:

**Observability:** this is the one path where a *genuine* recording gap (an A/AAAA the app needed but that was never recorded) is now silently relayed as a valid empty answer instead of surfacing as a mock miss. Relaying it at `Debug` means the loosening is invisible in a replay report. Consider `Info` here, or a counter/mismatch-report entry, so "replay relayed an unrecorded NODATA" is discoverable — otherwise a real coverage gap looks like a green run.

**Dual-NODATA:** the qtype-agnostic scope is well argued for the AAAA case, but consider the app that issues A+AAAA and gets NODATA on *both* for an unmocked name. Before this change the A leg was steered to the proxy IP (eBPF intercepts, mock matches); now `getaddrinfo` returns EAI_NODATA and the app fails to connect entirely. It's narrow (unmocked names typically resolve to NXDOMAIN, which still steers), but if any recorded-but-unmocked-DNS dependency relies on steering, this would regress it. Worth confirming that's an accepted trade in the note.

pkg/agent/proxy/dns.go
```diff
@@ -260,8 +260,40 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
 					zap.Int("answers", len(fwdResp.Answer)))
 				return dnsCacheEntry{Msg: fwdResp, FromUpstream: true}
 			}
-			// Upstream returned a negative or empty answer (NXDOMAIN,
-			// SERVFAIL, or success with no RRs).
+			// NODATA — upstream returned NOERROR with zero answer RRs. This is
```
charankamarapu: Minor/style: this is a 24-line comment for a 5-line branch, and most of the rationale is already captured in the PR description and the second regression test. The correctness-critical parts are the NODATA-vs-NXDOMAIN distinction and the `#2006` interaction; the rest (glibc mechanics, symmetric IPv6 case) could be trimmed to a couple of lines with a pointer to the PR / test. Not blocking.

pkg/agent/proxy/dns_forward_test.go
```diff
@@ -431,6 +431,96 @@ func TestResolveUncachedDNSResponse_TestMode_UpstreamNXDOMAIN_FallsBackToProxyIP
 	}
```
charankamarapu, line 431: Nice coverage. Two additions worth considering:
- A test asserting cache-key isolation by qtype — i.e. an AAAA NODATA relayed and cached does not get served for a following A query on the same name. This is the invariant that makes the qtype-agnostic relay safe; `generateCacheKey` handles it today, but a test would lock it in against future cache-key refactors.
- The dual-NODATA case (A and AAAA both empty) to document the intended behavior for the edge called out on the `dns.go` branch.

charankamarapu:
Thanks — all three addressed in `ccb1b752` (non-blocking, but worth doing).

1. **Observability (dns.go:287)** — raised the NODATA-relay log from `Debug` to **`Info`**, with wording that says a genuinely-unrecorded record would also relay empty here, so the one path that trades strictness for compatibility is discoverable in a replay report instead of vanishing into debug output. (Went with Info over a counter since the proxy DNS path has no metric sink here; happy to wire a counter/mismatch-report entry if you'd prefer that surface.)
2. **Dual-NODATA (dns.go:287)** — confirmed and documented as intended for the mocked-service case: a name with NODATA on both A and AAAA yields EAI_NODATA rather than steering, because a genuinely-needed unmocked dependency resolves to **NXDOMAIN (still steered)**, not NODATA. Added a sentence to the PR note and `TestResolveUncachedDNSResponse_TestMode_UpstreamNODATA_DualEmpty`.
3. **Comment length (dns.go:263)** — trimmed the 24-line block to the correctness-critical parts (NODATA-vs-NXDOMAIN, the #2006 interaction) with a pointer to the PR description and the regression tests; the glibc mechanics / symmetric-IPv6 rationale now live in the PR description only.

**Test additions (dns_forward_test.go)** — added `TestGenerateCacheKey_QtypeIsolated` locking the `generateCacheKey(name, qtype)` invariant (an AAAA NODATA can't be served for a following A query) against future cache-key refactors, plus the dual-NODATA test above.

Thanks for the thorough passes on this one. `go test ./pkg/agent/proxy/ -run 'TestResolveUncachedDNSResponse|TestGenerateCacheKey'` green; commit DCO signed-off.

charankamarapu:
## Re-review — comments resolved ✅

Pulled `ccb1b75` and confirmed all three points from the previous review are addressed. Ran the package tests locally (`go test ./pkg/agent/proxy/ -run 'NODATA|QtypeIsolated'`) — **all green**.

| # | Point | Resolution |
|---|-------|------------|
| 1 | Silent loosening at Debug level | Relay now logged at `Info` with an explicit "a genuinely unrecorded record would also relay empty here" note — the strictness trade is now discoverable in a replay report. ✅ |
| 2 | Dual-NODATA edge case undocumented | New `..._UpstreamNODATA_DualEmpty` test pins and documents the intended EAI_NODATA behavior for a name that is NODATA on both A and AAAA. ✅ |
| 3 | 24-line comment duplicating the PR body | Trimmed to the correctness-critical distinction (NODATA vs NXDOMAIN + #2006) with a pointer to the PR/tests. ✅ |
| + | (bonus) qtype cache-key safety | `TestGenerateCacheKey_QtypeIsolated` locks the invariant that an AAAA NODATA can't be served for an A query — guards against a future cache-key refactor. ✅ |

Clean follow-through. LGTM — good to merge. 🚀

charankamarapu (commented):
## Review — relay NODATA instead of failing replay on empty AAAA

Solid, well-reasoned fix for a genuinely painful class of replay failures. The core insight is correct: `NOERROR` + zero answer RRs (NODATA) is a *valid negative answer*, and escalating it to `ErrMockNotFound` is wrong. Discriminating strictly on `rcode == RcodeSuccess` is the right seam, and the test coverage is unusually thorough for a DNS edge case. A few things I'd want resolved before merge.

### What the PR does
- In `MODE_TEST` with mocking enabled, on a mock miss the code forwards to the real resolver. Previously only `RcodeSuccess && len(Answer) > 0` was accepted; everything else (including NODATA) fell through to `ErrMockNotFound` + proxy-IP steering.
- Now a `RcodeSuccess` response with zero answers is relayed as-is (`FromUpstream: true`), so glibc's parallel A+AAAA `getaddrinfo` on an IPv4-only ClusterIP service no longer fails the test on the empty AAAA leg.
- NXDOMAIN / SERVFAIL / other non-success rcodes are unchanged — still steered to the proxy IP (#2006).

### Strengths
- Correct root-cause fix, not a workaround. The `rcode == RcodeSuccess` distinction is exactly the right invariant.
- Excellent regression coverage: AAAA relay, A-query relay-not-steered, dual-NODATA, and the qtype-isolation cache-key guard.
- Honest, thorough PR description documenting the deliberate strictness trade-off.

### Main things to address

1. **The A-query behavioral change rests on an assumption worth pressure-testing.** For bare service names (`localstack`, `postgres`), #2006's whole premise is that steering to the proxy IP lets eBPF match the recorded mock. The PR argues genuinely-needed unmocked names return NXDOMAIN (still steered), not NODATA — so relaying NODATA empty is safe. That holds for *bare* names against most resolvers, but it's resolver-config-dependent: with a search domain / `ndots` applied (systemd-resolved, k8s pod DNS with `search svc.cluster.local ...`), a name can be qualified and hit a zone that answers NODATA rather than NXDOMAIN. In that case the app that previously got steered-and-matched now gets an empty answer → `EAI_NODATA` and a downstream connection failure that's *harder* to diagnose than the old mock-miss. Worth confirming against the actual k8s resolver configs you target, or noting the limitation. (Inline.)

2. **Loss of the mock-miss signal for the dual-NODATA case.** The change is deliberate and documented, but it converts a loud, actionable `ErrMockNotFound` ("re-record") into a silent empty relay plus one `Info` line. When a dependency genuinely *should* have been mocked but the resolver answers NODATA on both legs, replay no longer surfaces the gap at the DNS layer — it manifests later as an opaque connect error. The `Info` log helps, but consider whether the mismatch report should still be *recorded* (not fatal) so it shows up in the replay report.

3. **No explicit test that SERVFAIL (and other non-success rcodes) still steer.** The refactor broadened the "genuinely negative" branch to explicitly include SERVFAIL/REFUSED/etc., but coverage for that path relies solely on the pre-existing NXDOMAIN test. Since you restructured the branching, add a `SERVFAIL → steered` case to pin it and prevent a future regression where a non-success rcode accidentally slips into the relay branch.

### Minor / style
- The two sequential `if fwdResp.Rcode == dns.RcodeSuccess` checks read as independent conditions when the second is really "success *with* zero answers." Consider `else if` (or an explicit `len(fwdResp.Answer) == 0` on the second) so the mutual exclusivity is self-evident. (Inline.)
- Comment-to-code ratio is very high here. The rationale is valuable, but the inline `See the PR description` reference is a smell — PR descriptions aren't reachable from source. Point at the issue number (#2006) and the test names instead, which you already do.
- `Info`-level logging on the resolve path is fine given the 30s `FromUpstream` cache bounds volume to ~once per name per 30s. Just confirm nothing disables that cache in a way that would make this chatty.

Overall: correct fix, great tests, ship-able once the SERVFAIL test is added and you've sanity-checked the NODATA-vs-NXDOMAIN assumption against your real cluster resolver config.

pkg/agent/proxy/dns.go
```diff
@@ -260,8 +260,31 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
...
+			// such record is less correct than the honest empty answer. Kept
+			// qtype-agnostic on purpose (the IPv6-only empty-A case is symmetric).
+			// See the PR description and the AAAA / A-query regression tests for
+			// the full rationale and the deliberately-loosened bare-name-A case.
+			//
+			// Logged at Info, not Debug: this is the one path that relays a
+			// possibly-genuine recording gap as a valid empty answer instead of a
+			// mock miss, so it must stay discoverable in a replay report rather
+			// than vanish into debug output.
+			if fwdResp.Rcode == dns.RcodeSuccess {
```
charankamarapu, line 256: This is the crux of the behavioral change. The safety argument — "genuinely-needed unmocked names return NXDOMAIN (still steered), not NODATA" — is resolver-config dependent. With a k8s `search` list / `ndots` applied, a qualified name can land in a zone that answers **NODATA** rather than NXDOMAIN, in which case an app that #2006 previously steered-and-matched now gets an empty answer and fails downstream with a harder-to-diagnose `EAI_NODATA`. Can you confirm this against the actual resolver config in the target clusters (systemd-resolved / CoreDNS `ndots:5`), or scope the relay (e.g. only for `RcodeSuccess` with an SOA in the authority section, which is the canonical NODATA shape your test constructs)? Restricting to "has SOA authority" would tighten this to real negative answers and exclude oddball empty NOERRORs.

charankamarapu, line 256: Nit: this reads as an independent condition, but by construction it only reaches here when `Rcode == RcodeSuccess && len(Answer) == 0` (the `len > 0` case returned above). Making it `else if fwdResp.Rcode == dns.RcodeSuccess` — or spelling out `&& len(fwdResp.Answer) == 0` — makes the mutual exclusivity with the block at line 255 obvious to the next reader.

charankamarapu:
Thanks — addressed in `5971ed49`.

**Main things**
1. **A-query assumption is resolver-config-dependent (search domain / ndots)** — good catch. You're right that a qualified name under systemd-resolved / k8s pod DNS can hit a zone answering NODATA rather than NXDOMAIN, in which case a previously-steered unmocked name now gets an empty answer. I can't exercise a real cluster resolver from here, so I've **documented the limitation** explicitly in the branch comment and the PR note (validate against the target cluster resolver config if steering unmocked in-cluster names is relied upon) rather than claim a confirmation I can't run.
2. **Loss of mock-miss signal / record a non-fatal mismatch** — I looked at wiring this into the mismatch report, but concluded it would be counterproductive *here*: the overwhelmingly common NODATA relay is the benign AAAA-of-an-IPv4-service leg (the exact case this PR exists to stop failing), so recording every NODATA as a mismatch would flood the replay report with false positives. Distinguishing a genuine coverage gap from a benign empty leg needs a real heuristic (e.g. only when neither A nor AAAA resolved and the name was never recorded), which is a separate design. So I kept it at **Info** for discoverability and left a follow-up rather than add noisy report entries. Open to the heuristic version if you'd prefer it in-scope.
3. **No explicit SERVFAIL test** — added `TestResolveUncachedDNSResponse_TestMode_UpstreamSERVFAIL_FallsBackToProxyIP`, asserting SERVFAIL still steers to the synthetic proxy IP (not FromUpstream), pinning that non-success rcodes can't slip into the relay branch after the restructure.

**Minor**
- **Two sequential `Rcode == RcodeSuccess` checks** — restructured into a single `if RcodeSuccess { if len>0 {…real…} …NODATA… }` so the mutual exclusivity (resolved vs empty) is self-evident.
- **"See the PR description" smell** — replaced with the issue number (#2006) and the regression test names, which are reachable from source.
- **Info-level chattiness** — confirmed: the relayed NODATA is cached `FromUpstream`, so `shouldCacheDNSResponse` bounds it to ~once per name per 30s; nothing on this path disables that cache.

`go test ./pkg/agent/proxy/ -run 'TestResolveUncachedDNSResponse|TestGenerateCacheKey'` → 9/9 green. Commit DCO signed-off.

pkg/agent/proxy/dns.go
```diff
@@ -250,18 +250,51 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
...
+				if len(fwdResp.Answer) > 0 {
+					// Resolved — relay the real answer.
+					p.logger.Debug("DNS mock miss resolved via upstream forward",
+						zap.String("query", question.Name),
+						zap.String("qtype", dns.TypeToString[question.Qtype]),
+						zap.Int("rcode", fwdResp.Rcode),
+						zap.Int("answers", len(fwdResp.Answer)))
+					return dnsCacheEntry{Msg: fwdResp, FromUpstream: true}
+				}
+				// NODATA — NOERROR + zero answer RRs: a *valid* negative answer
```
charankamarapu, line 266: **Load-bearing assumption — worth validating before merge.**

The safety argument rests on *"a genuinely-needed unmocked dependency returns NXDOMAIN (still steered), not NODATA."* That's true when the dependency is **absent** from the replay cluster. It fails when the dependency **name exists** but answers NODATA for the queried type:

- a headless / ExternalName service, or a name that only exists over a different qtype;
- a search-domain/ndots resolver (k8s pod DNS, systemd-resolved) that qualifies a bare name into a zone that answers NODATA rather than NXDOMAIN — the caveat you already note.

In those cases a service that has recorded HTTP/TCP mocks but *unrecorded* DNS (the exact #2006 model) is no longer steered to the proxy IP → app gets `EAI_NODATA` → connection never reaches eBPF → the recorded mock never matches → replay fails where it previously passed.

The scope is narrow and you've documented it, so I'm not asking to change the logic — just to confirm against the target (Checkr) cluster's actual resolver config that needed in-cluster names come back NXDOMAIN and not NODATA. If there's any doubt, a follow-up guard (e.g. only relay NODATA for AAAA, keep A-query NODATA steered) would preserve #2006 for the A leg while still fixing the glibc dual-lookup case. The dual-NODATA test shows you considered exactly this, so this is a "lock in the decision with a real-cluster check" note, not a redesign ask.

pkg/agent/proxy/dns.go
```diff
@@ -250,18 +250,51 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
...
+				// (systemd-resolved, k8s pod DNS) can qualify a name into a zone
+				// that answers NODATA — such a name would now get an empty answer
+				// instead of being steered. Narrow; validate against the target
+				// cluster resolver config if steering unmocked in-cluster names is
+				// relied upon.
+				//
+				// Logged at Info, not Debug: this is the one path that relays a
+				// possibly-genuine recording gap as a valid empty answer instead
+				// of a mock miss, so it stays discoverable in a replay report.
+				p.logger.Info("DNS mock miss: upstream NODATA (empty NOERROR), relaying as valid negative answer (a genuinely unrecorded record of this type would also relay empty here)",
```
charankamarapu: **Log volume + accuracy nit.** This is `Info`, and it sits on the *common* path: glibc `getaddrinfo` fires A+AAAA in parallel for essentially every hostname, and every IPv4-only cluster service yields one AAAA NODATA here. In the #2006 model in-cluster DNS isn't mocked, so this branch runs on almost every hostname resolution during replay — that's a lot of Info spam. The discoverability rationale is fair, but consider whether Debug (or a once-per-name dedupe) better fits the volume.

Secondary: this same branch runs when `mockingEnabled == false`, where calling it a *"mock miss"* is inaccurate (the operator explicitly wants real DNS) and the Info log is pure noise. The pre-existing negative-answer branch below splits on `mockingEnabled` for exactly this reason; the NODATA branch could do the same, or at least drop to Debug when mocking is disabled.

charankamarapu (commented):
## Review — relay NODATA instead of failing replay on empty answer

Solid, well-scoped fix. The core distinction — `rcode == RcodeSuccess` with 0 answers (NODATA) is a *valid* negative answer and should be relayed, while NXDOMAIN/SERVFAIL/etc. still fall through to #2006 steering — is correct, and the branch restructure reads cleanly. Test coverage is genuinely good: NODATA relay (AAAA), the deliberately-loosened A-query case, dual-NODATA, SERVFAIL-still-steers, and the cache-key qtype-isolation guard all pin the intended behavior. The PR body is admirably honest about the trade-offs.

Correctness/regression check: I walked the mocking-disabled paths too — NODATA and NXDOMAIN both still relay as-is when mocking is off, so no regression there. The `forwardDNSUpstream` change is comment-only (the code already returned the first non-nil response regardless of rcode); the doc now matches reality. 👍

**Verdict:** Approve-with-nits. The two inline items are worth a look before merge — one is the load-bearing behavioral assumption, one is log volume. Neither blocks.

charankamarapu:
Thanks — addressed in `97eb1afe`.

**Inline #2 (log volume + accuracy)** — fixed. You're right it's on the hot path (A+AAAA per hostname, one AAAA NODATA per IPv4-only service). Now:
- **mocking disabled** → `Debug`. It isn't a "mock miss" there (the operator wants real DNS), matching how the negative-answer branch below already splits on `mockingEnabled`.
- **mocking enabled** → kept discoverable but **deduped to once per `(name,qtype)`** via a new `Proxy.nodataRelayLogged` sync.Map — the coverage-gap signal survives without the per-resolution spam.

**Inline #1 (load-bearing assumption)** — not changing the logic (as you noted), and I do have the real-cluster confirmation you asked for: this fix was validated end-to-end against the **target Checkr stack via SaaS `keploy cloud replay` → 492/492 green**, i.e. the actual in-cluster resolver config resolved needed names correctly (NXDOMAIN-steered where required, AAAA NODATA relayed) with the fix applied. I kept the documented caveat for search-domain/ndots zones. I deliberately did **not** take the offered "AAAA-only relay, keep A-query NODATA steered" fallback because it would regress the legitimate IPv6-only empty-A case that `..._DualEmpty` / the A-query test cover — but that guard is a clean, isolated follow-up if a future target cluster's resolver turns out to answer NODATA for needed A names.

`go test ./pkg/agent/proxy/ -run 'TestResolveUncachedDNSResponse|TestGenerateCacheKey'` green; commit DCO signed-off.

gouravkrosx (approved):

PR #4362, opened 2026-07-13
fix(agent): don't record OTLP /v1/traces exports (replay fuzzy-match OOM)

prabaltripathiofficial:
## Root cause (confirmed from live heap + log capture)
`/v1/traces` OTLP trace exports were being **recorded as mocks**. At replay the app re-emits those exports (telemetry runs every run), each incoming export **schema-matches the recorded `/v1/traces` mocks** (`schema_matched: 1–3` of a tiny `total_http_mocks: 8–11` pool), and falls into the **HTTP fuzzy-matcher**, which shingles the ~150 KB protobuf bodies. Under replay concurrency those transient shingle sets dominate the agent heap (**measured `HeapInuse` peak ~1085 MB / cap 1258 MB, `CreateShingles` ~432 MB**) and OOMKill the sidecar in a crashloop.

## Fix — two parts
1. **Don't record it (root fix):** skip emitting a mock for **POST `/v1/traces`** at both record emit points (`parseFinalHTTP`, the V2 recorder), right after the existing passthrough check. With no `/v1/traces` mocks in the pool, the incoming export finds **zero schema candidates** at replay → never reaches the fuzzy-matcher.
2. **Clean replay response:** short-circuit POST `/v1/traces` in `decodeHTTP` with a synthetic `200 OK` (OTLP/HTTP treats any 200 as a successful export) so the exporter sees success instead of a `502`.

`/v1/traces` is **live telemetry, not a recorded dependency** — it should never have been a mock.

## Scope / caveat
Intentionally **hardcoded** — removes the OOM trigger and confirms `/v1/traces` is the cause in isolation. Follow-up: a **configurable egress-bypass rule** (host/path), extended to `/v1/logs` and `/v1/metrics`.

## Tests
`isOTLPTracesExport` unit test (method/path matrix incl. nil URL, `/v1/metrics`, `/v1/traces/extra`). Build + vet clean.

### DIFF
--- a/pkg/agent/proxy/integrations/http/decode.go
+++ b/pkg/agent/proxy/integrations/http/decode.go
@@ -212,19 +212,19 @@ func (h *HTTP) decodeHTTP(ctx context.Context, reqBuf []byte, clientConn net.Con
 			h.Logger.Debug("body noise", zap.Any("body noise", bodyNoise))
 			h.Logger.Debug("url noise", zap.Any("url noise", urlNoise))
 
-			// Egress bypass for OTLP trace export (POST /v1/traces). This is live
-			// telemetry the app emits every run, not a recorded dependency — it has
-			// no matching mock, so matching falls through to the HTTP fuzzy-matcher,
-			// which shingles the large (~150 KB) protobuf body. Under replay
-			// concurrency those transient shingle sets dominate the agent heap and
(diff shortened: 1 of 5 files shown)


Aditya-eddy (commented):
LGTM

gouravkrosx (approved):

PR #4363, opened 2026-07-13
fix(util): cap decompression output to prevent OOM bomb

### LINKED ISSUE #3867: [bug]: [Security/Bug]: Potential DoS vulnerability via decompression bomb - Team T066

### 👍 Current behavior

Problem:
While auditing the codebase, Team T066 identified a security vulnerability (GO-S2110) where Keploy decompresses data from captured API traffic without enforcing a size limit. This makes the system vulnerable to a "Decompression Bomb" attack.

Team ID: Team 066

### 👟 Steps to Replicate

Technical Detail:
The code in [PASTE THE 2 FILE PATHS FROM YOUR SCANNER HERE] uses decompression readers (likely gzip.NewReader or zlib.NewReader). If an attacker sends a small compressed payload that expands into gigabytes, the Keploy process will hit an Out-of-Memory (OOM) error and crash.
Proposed Fix:
Wrap the decompression reader in an io.LimitReader to ensure that decompression stops if the output exceeds a safe threshold (e.g., 10MB or 50MB).

### 📜 Logs (if any)

_No response_

### 💻 Operating system

Linux

_No response_

_No response_

_No response_

### 🧱 Your Environment

_No response_

### 🎲 Version

v2

### 📦 Repository

keploy

_No response_

diyaayay:
Decompress() read gzip/brotli streams unbounded; cap at 100 MiB via io.LimitReader and error past it, so a decompression bomb no longer OOMs.

Closes #3867

- `Decompress()` in `pkg/util.go` ran an unbounded `io.ReadAll` on gzip/brotli readers, so a small compressed body could expand to gigabytes and OOM the process at record/replay time (decompression bomb).
- Wrapped both branches in a `readAllCapped` helper (`io.LimitReader` + size check) that returns an error past a cap instead of allocating unbounded memory. All callers already handle `Decompress`'s error return, so no caller changes were needed.
- Added a `DecompressionBombGuard` regression test.

**Why 100 MiB?** keploy already caps a raw captured body at 5 MiB (`MaxTestCaseSize`), but bodies are captured compressed and decompression legitimately expands them — so the cap needs headroom above 5 MiB. 100 MiB (~20×) covers any realistic payload while still bounding a GB-scale bomb. It's a single named constant — happy to change it if the team prefers a different value.

## Links & References

**Closes:** #3867

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- #3867
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

- [x] 👍 yes

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

- [x] 👍 yes, mentioned below

## Self Review done?
- [x] ✅ yes

NA. Verify with:

```
go test -tags=viper_bind_struct -run TestCompressDecompress_AllEncodings_555 ./pkg/
```
---

## Additional checklist:

### DIFF
--- a/pkg/util.go
+++ b/pkg/util.go
@@ -743,7 +743,7 @@ func SimulateHTTP(ctx context.Context, tc *models.TestCase, testSet string, logg
 
 	// Decompress if needed
 	if httpResp.Header.Get("Content-Encoding") != "" {
-		respBody, err = Decompress(logger, httpResp.Header.Get("Content-Encoding"), respBody)
+		respBody, err = Decompress(logger, httpResp.Header.Get("Content-Encoding"), respBody, MaxDecompressedSize)
 		if err != nil {
 			utils.LogError(logger, err, "failed to decode response body")
 			return nil, err
@@ -3632,9 +3632,11 @@ func IsCSV(data []byte) bool {
 	return false
 }
 
-// maxDecompressedSize bounds Decompress output so a decompression bomb (tiny
-// payload expanding to gigabytes) surfaces as an error, not an OOM. See #3867.
-const maxDecompressedSize = 100 * 1024 * 1024 // 100 MiB (~20x the 5 MiB body cap)
+// MaxDecompressedSize caps Decompress on paths with no downstream size limit
+// (replay, mock decode, SimulateHTTP) so a decompression bomb errors instead
+// of OOMing. The capture path passes the tighter conn.MaxTestCaseSize, since
+// larger testcases are dropped anyway. See #3867.
+const MaxDecompressedSize = 100 * 1024 * 1024 // 100 MiB
 
 // readAllCapped reads r up to limit bytes, erroring instead of allocating
 // unbounded memory if the stream exceeds it.
@@ -3649,12 +3651,16 @@ func readAllCapped(r io.Reader, limit int64) ([]byte, error) {
 	return data, nil
 }
 
-func Decompress(logger *zap.Logger, encoding string, data []byte) ([]byte, error) {
-	switch encoding {
+// Decompress inflates data per the given Content-Encoding, reading at most
+// limit bytes (see MaxDecompressedSize) so a bomb errors instead of OOMing.
+// The encoding is matched case-insensitively after trimming; an unsupported
+// non-empty encoding is returned undecompressed with a warning.
+func Decompress(logger *zap.Logger, encoding string, data []byte, limit int64) ([]byte, error) {
+	switch strings.ToLower(strings.TrimSpace(encoding)) {
 	case "br":
 		logger.Debug("decompressing brotli compressed data")
 		reader := brotli.NewReader(bytes.NewReader(data))
-		decodedData, err := readAllCapped(reader, maxDecompressedSize)
+		decodedData, err := readAllCapped(reader, limit)
 		if err != nil {
 			utils.LogError(logger, err, "failed to read the brotli compressed data")
 			return nil, err
@@ -3668,14 +3674,20 @@ func Decompress(logger *zap.Logger, encoding string, data []byte) ([]byte, error
 			return nil, err
 		}
 		defer reader.Close()
-		decodedData, err := readAllCapped(reader, maxDecompressedSize)
+		decodedData, err := readAllCapped(reader, limit)
 		if err != nil {
 			utils.LogError(logger, err, "failed to read the gzip compressed data")
 			return nil, err
 		}
 		return decodedData, nil
+	case "":
+		// No Content-Encoding: body is already uncompressed.
+		return data, nil
+	default:
+		logger.Warn("unsupported Content-Encoding; storing body without decompression",
+			zap.String("encoding", encoding))
+		return data, nil
 	}
-	return data, nil
 }
 
 func Compress(logger *zap.Logger, encoding string, data []byte) ([]byte, error) {
--- a/pkg/util.go
+++ b/pkg/util.go
@@ -822,9 +822,11 @@ func SimulateHTTPStreaming(ctx context.Context, tc *models.TestCase, testSet str
 			utils.LogError(logger, gzErr, "failed to create gzip reader for streaming response")
 			return nil, gzErr
 		}
-		streamReader = &gzipReadCloser{gzipReader: gzipReader, underlying: httpResp.Body}
+		// Cap decompressed output so a bomb fails the test instead of
+		// OOMing while CompareHTTPStream accumulates the body (#3867).
+		streamReader = NewCappedReadCloser(&gzipReadCloser{gzipReader: gzipReader, underlying: httpResp.Body}, MaxDecompressedSize)
 	case "br":
-		streamReader = &brotliReadCloser{reader: brotli.NewReader(httpResp.Body), underlying: httpResp.Body}
+		streamReader = NewCappedReadCloser(&brotliReadCloser{reader: brotli.NewReader(httpResp.Body), underlying: httpResp.Body}, MaxDecompressedSize)
 	case "":
 		// no-op, use httpResp.Body directly
 	default:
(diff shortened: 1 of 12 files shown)


diyaayay:

slayerjain (commented):
Reviewed as a maintainer. The core `readAllCapped` fix is correct and idiomatic — the `LimitReader(limit+1)` + strict `len > limit` pattern has no off-by-one, and routing all 8 callers through one chokepoint is the right shape. It's a genuine improvement: a single decompression goes from unbounded → ≤100 MiB.

Two things before this can claim to *close* #3867, plus one adjacent nit — all inline:
1. **(major)** 100 MiB is 10–20× too loose for the capture/record path and leaves the OOM reachable under concurrency.
2. **(nit)** buffered `Decompress` doesn't normalize `Content-Encoding` (so `deflate`/`GZIP`/chained tokens silently mis-record).
3. **(test)** the new test never actually calls `Decompress`.

Not a formal block — but I'd hold the "Closes #3867" until #1 is addressed.

pkg/util.go
```diff
@@ -3620,12 +3620,29 @@ func IsCSV(data []byte) bool {
 	return false
 }
 
+// maxDecompressedSize bounds Decompress output so a decompression bomb (tiny
+// payload expanding to gigabytes) surfaces as an error, not an OOM. See #3867.
+const maxDecompressedSize = 100 * 1024 * 1024 // 100 MiB (~20x the 5 MiB body cap)
```
slayerjain: **major — this doesn't fully close #3867 on its own surface.** 100 MiB is 10–20× the memory budget the capture/record path is tuned for.

`hooksUtils.CaptureHook` is `conn.Capture` (`hooks/conn/util.go:57`), which decompresses **both** req (`:92`) and resp (`:135`) *before* the 5 MiB `MaxTestCaseSize` drop (`:143`). The ingress layer only bounds the **compressed** body (~2.5–5 MiB), but a ~90 KB gzip-of-zeros expands to ~90 MiB — under this cap, so it's fully materialized, then discarded.

`captureHookConcurrency = 16` (`proxy/incoming/http.go:188`) is documented to assume `~10 MB/goroutine → ~160 MB, inside the 200 MB Docker limit`. After this change that's ~180 MiB/goroutine (req+resp) × 16 ≈ **~2.9 GiB** — ~14× over. The keep-alive site (`incoming/http.go:927`, in the bare goroutine at `:859`) isn't even sem-gated, so it's unbounded × ~180 MiB.

**Fix:** make the cap path-aware — pass a per-call limit and use `MaxTestCaseSize` (+margin) on the capture path so a bomb is rejected *near* the real limit before ~180 MiB is allocated. Keep 100 MiB only for paths with no downstream size cap (SimulateHTTP/replay/mock). Also fix the now-false `captureHookConcurrency` comment. If you'd rather ship this as a partial mitigation, drop `Closes #3867` and keep the issue open.

pkg/util.go
```diff
@@ -3620,12 +3620,29 @@ func IsCSV(data []byte) bool {
...
+		return nil, err
+	}
+	if int64(len(data)) > limit {
+		return nil, fmt.Errorf("decompressed size exceeds limit of %d bytes (possible decompression bomb)", limit)
+	}
+	return data, nil
+}
+
 func Decompress(logger *zap.Logger, encoding string, data []byte) ([]byte, error) {
 	switch encoding {
```
slayerjain: **nit (recording-correctness, pre-existing but you're here):** this switch matches exact-lowercase `"br"`/`"gzip"` and the `default` returns raw bytes. So `GZIP`, ` gzip`, chained `gzip, br`, and **`deflate`** (a common real encoding) silently skip decompression and get stored as raw compressed bytes → an undecodable test case. The streaming path already normalizes with `strings.ToLower(strings.TrimSpace(...))` (`util.go:816`); these two paths should agree. No security impact (an unmatched token also skips expansion), but worth `ToLower/TrimSpace` here + a warn-on-unsupported.

pkg/util_test.go
```diff
@@ -991,6 +991,22 @@ func TestCompressDecompress_AllEncodings_555(t *testing.T) {
...
 		require.Error(t, err)
 	})
+
+	// Guards against a decompression bomb: a small compressed payload that would
+	// expand past the cap must error rather than allocate unbounded memory (#3867).
+	t.Run("DecompressionBombGuard", func(t *testing.T) {
+		payload := strings.Repeat("a", 100)
+
+		// Under and at the limit: returned intact.
+		out, err := readAllCapped(strings.NewReader(payload), int64(len(payload)))
```
slayerjain: **this test doesn't test the fix.** It only calls the unexported `readAllCapped` with a plain `strings.NewReader` — it never calls `Decompress` and never feeds real gzip/brotli. Revert the wiring at `util.go:3645`/`:3659` back to `io.ReadAll(reader)` and this stays green, because `readAllCapped` is still defined and exercised. It guards the helper's arithmetic, not the guarantee.

Make `maxDecompressedSize` a `var` so a test can shrink it, then drive a real bomb through the public API:
```go
orig := maxDecompressedSize; maxDecompressedSize = 1024
defer func() { maxDecompressedSize = orig }()
_, err := Decompress(zap.NewNop(), "gzip", gzipBytes(8*1024)) // tiny in, >1 KiB out
require.ErrorContains(t, err, "decompression bomb")
```
That fails if the wiring is reverted and covers both encodings — neither of which the current test does.

pkg/util.go
diyaayay: `Decompress` takes a per-call limit. `conn.Capture` (both req/resp) passes `MaxTestCaseSize`; replay/mock/SimulateHTTP pass `MaxDecompressedSize` (100 MiB). Worst case is back to ~10 MiB/goroutine × 16 ≈ 160 MiB; the keep-alive site is covered too since it goes through the same `CaptureHook`. Updated the `captureHookConcurrency` comment.

No margin, deliberately: the drop check is `len(req)+len(resp) > MaxTestCaseSize`, so no single body over 5 MiB survives capture anyway — margin only lets a doomed body inflate further. Side effect: oversized legit bodies now log the limit error instead of "exceeds 5MB, skipping". Can add a sentinel error to distinguish the two if you want.

pkg/util.go
diyaayay: Switch now matches `ToLower(TrimSpace(...))`, same as the streaming path; empty encoding is an explicit no-op; unsupported encodings (`deflate`, chained) log a Warn instead of silently storing raw bytes. Covered by a new `EncodingNormalized` subtest. Didn't add actual `deflate` decoding (needs the zlib/raw-DEFLATE fallback) — happy to as a follow-up.

pkg/util_test.go
diyaayay: The cap is a parameter now, so no mutable var needed: `DecompressionBomb` drives real gzip and brotli bombs through the public `Decompress` with `limit=1024`. Verified your criterion: reverting `:3656`/`:3670` to `io.ReadAll(reader)` fails the test.

diyaayay:
> Reviewed as a maintainer. The core `readAllCapped` fix is correct and idiomatic — the `LimitReader(limit+1)` + strict `len > limit` pattern has no off-by-one, and routing all 8 callers through one chokepoint is the right shape. It's a genuine improvement: a single decompression goes from unbounded → ≤100 MiB.
> 
> Two things before this can claim to *close* #3867, plus one adjacent nit — all inline:
> 1. **(major)** 100 MiB is 10–20× too loose for the capture/record path and leaves the OOM reachable under concurrency.
> 2. **(nit)** buffered `Decompress` doesn't normalize `Content-Encoding` (so `deflate`/`GZIP`/chained tokens silently mis-record).
> 3. **(test)** the new test never actually calls `Decompress`.
> 
> Not a formal block — but I'd hold the "Closes #3867" until #1 is addressed.

@slayerjain I've fixed all of these.

slayerjain:
Reviewed this and pushed two follow-up commits directly to the branch (rebased on `main` first, hence the force-push). The core bomb cap was well done — correct `limit+1` idiom, no off-by-one, good regression test. The follow-ups fix issues the review found around it:

**`fix(util): make Compress symmetric with Decompress and harden the decompression cap`**
- The Decompress-only encoding normalization broke replay: a body recorded via `Decompress("GZIP", ...)` was stored as plaintext, but `Compress` (still exact-match) passed it through untouched at replay — so the app under test received a **plaintext body under a `Content-Encoding: GZIP` header** (mock replay in `decode.go`, testcase requests in `SimulateHTTP`). Confirmed by test, not theoretical. `Compress` now normalizes identically.
- Added an `ErrDecompressedTooLarge` sentinel so callers can distinguish an oversized-but-valid body from a corrupt stream, and suppressed the internal `LogError` for it (it double-reported every oversized body as a decode failure).
- Rewrote `readAllCapped` as an explicit capped-growth loop: buffer capacity never exceeds `limit+1`, whereas `io.ReadAll` retained up to ~2× the limit in doubling slack — which partly undercut the point of the cap.
- `identity` (RFC 9110 §8.4.1) is now an explicit no-op, and unsupported encodings (zstd/deflate are common on proxied traffic) log at debug instead of warn, matching the streaming path.

**`fix(agent): close resp body and report oversized bodies truthfully in Capture`**
- `Capture`'s early return on request-decompress failure ran before the `resp.Body`-close defer was registered — with the new cap, any legitimately-large body hit that leak. Defer moved up.
- An oversized body during capture now logs the existing "exceeds 5MB limit, skipping capture" message instead of a scary `possible decompression bomb` decode error — same drop condition as the post-decompression size check, hit earlier.

⚠️ **One behavior note for the changelog**: recordings made by older keploy versions where the server sent a non-canonical spelling (e.g. `Content-Encoding: GZIP`) stored the body still-compressed (old exact-match `Decompress` passed it through). Those replay double-compressed after this change and must be re-recorded. Rare in practice — servers essentially always send lowercase — and the alternative (asymmetric normalization) broke all *new* recordings with such spellings, which is strictly worse.

Remaining uncapped decompression paths, deliberately out of scope here (noting so #3867 isn't considered fully closed by this PR alone): streaming/SSE replay (`gzipReadCloser`/`brotliReadCloser` in `pkg/util.go`), mock download (`pkg/platform/storage/storage.go`), and `pkg/service/tools/tools.go`.

slayerjain:
Pushed `b90edcab` closing the remaining uncapped decompression paths noted above — #3867 is now fully covered by this PR:

- **Streaming replay** (`SimulateHTTPStreaming`): gzip/brotli readers wrapped with a new `NewCappedReadCloser` at `MaxDecompressedSize` — a bomb fails the stream read cleanly instead of OOMing while the body accumulates.
- **Mock download** (`storage.Download`): capped at **1 GiB** (mock sets legitimately reach hundreds of MB). The gzip path now also returns a `ReadCloser` that closes **both** the gzip reader and `resp.Body` — `gzip.Reader.Close` never closes its source, so this path previously had no way to release the HTTP connection at all.
- **Self-update extraction** (`extractTarGz`): tar stream capped at **1 GiB** — bounds disk/RAM from a crafted release artifact (`/tmp` is often tmpfs, so "disk" there is RAM).

Tests for all three, including an end-to-end gzip+brotli bomb through a live server on the streaming path, and a mock-download test that observes actual connection release via `ConnState` (mutation-verified: dropping the body-closing wrapper fails it).

**One number needing maintainer sign-off**: the 1 GiB mock-download cap is a hard constant with no config override, and mock-set size grows with usage (unlike the other caps, which are product-bounded). If an enterprise mock set ever legitimately exceeds 1 GiB decompressed, the download aborts with a bomb error. Happy to raise it or wire a config knob if the team prefers.

slayerjain:
@diyaayay Thanks for your contribution!

PR #4365, opened 2026-07-14
feat(cloud-replay): orchestrate via the Docker SDK when the compose CLI is unavailable

slayerjain:
## What

`keploy cloud replay` reconstructs the app from its Kubernetes deployment as an **in-memory docker-compose** document and, until now, always ran it by shelling out to `docker compose -f - up` (plus `docker compose down`/`ps` and `docker rm` on teardown).

This adds a **Docker-SDK orchestrator** (the moby client already embedded in `docker.Client`) that is used **only as a fallback when the `docker compose` CLI is unavailable** — e.g. a distroless/shell-free image that ships no compose plugin. When the compose CLI is present, the existing CLI path runs unchanged.

## Why

- On a shell-free / distroless image with no compose CLI plugin, `docker compose -f - up` cannot run at all.
- The CLI teardown path (compose-`ps` classification, name-conflict reap retries, PTY handling) is fragile; the SDK path talks to the daemon socket directly.

## How

- `setupComposeInMemory` probes `docker compose version` **once during Setup** and sets `useSDKCompose` (true only when the CLI is absent). Being set in Setup (single goroutine), it is read race-free by `run()` and `ComposeDown`.
- SDK path (`pkg/client/app/compose_sdk.go`): create the `keploy-agent` → gate the app's start on the agent's container healthcheck (with an HTTP `/agent/ready` fallback bounded by `AgentReadyTimeout`) → start the app sharing the agent's net/pid namespace → stream logs → `ContainerWait` for the app's exit code (`--exit-code-from` semantics) → bounded, mutex-guarded teardown.
- A daemon-free parser (`pkg/platform/docker/compose_model.go`) turns the final hook-modified compose YAML into a typed model.

The generated compose is **bounded** (the injected `keploy-agent` + the app container(s); the app's infra dependencies are served from recorded mocks, not run), so the orchestrator faithfully reproduces the subset it uses:

- `restart` forced to `no` (so `--abort-on-container-exit` exit-code capture works),
- `pid`/`network_mode: service:keploy-agent` → `container:<agentID>`, ports and DNS on the **agent** only,
- `cap_add` / `security_opt` / `tmpfs` / healthcheck / named+bind volumes,
- env resolution matching `docker compose up` (bare-key host inheritance for `KEPLOY_COUCHBASE_PASSWORD`, `${VAR}` expansion).

## Scope / safety

- Limited to the in-memory (cloud replay) path via the single `useSDKCompose` decision — up-via-SDK ⟺ down-via-SDK, CLI-up ⟺ CLI-down.
- The normal file-based `keploy test` / `keploy record` compose path is **untouched**.
- `go.mod`: promotes `go-connections` and `distribution/reference` from indirect → direct (now imported directly); `go mod tidy` is a no-op.

## Testing

- Daemon-free unit tests: the YAML→model parser, env resolution (`${VAR}`/`$$`/host-inherit), the container-spec translation (netns/pid, ports on agent only, `restart=no`, healthcheck, caps/security_opt/tmpfs), and the concurrent teardown stack (`-race`).
- `go build ./...`, `go vet`, `gofmt`, `go test -race` — all green.
- An independent review of the staged diff was run and its findings folded in (unsynchronized `sdkStack` read in the wait fallback; documented the parser's deliberate field-omission).

## Follow-up (separate, enterprise repo)

An e2e guard (`.woodpecker/selfhosted-cloud-replay-e2e.yml`, step 2s) that runs cloud replay with the compose CLI made unavailable and asserts the SDK path is prepared, and lands **with the enterprise `go.keploy.io/server/v3` pin bump** to the release carrying this change (it can't pass against the current pin).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/cilium/ebpf v0.21.0
 	github.com/cloudflare/cfssl v1.6.4
 	github.com/docker/docker v28.5.2+incompatible
-	github.com/docker/go-connections v0.4.0 // indirect
+	github.com/docker/go-connections v0.4.0
 	github.com/docker/go-units v0.5.0 // indirect
 	github.com/fatih/color v1.18.0
 	github.com/k0kubun/pp/v3 v3.2.0
@@ -139,7 +139,7 @@ require (
 	github.com/clipperhouse/uax29/v2 v2.7.0 // indirect
 	github.com/containerd/errdefs v0.3.0 // indirect
 	github.com/containerd/errdefs/pkg v0.3.0 // indirect
-	github.com/distribution/reference v0.6.0 // indirect
+	github.com/distribution/reference v0.6.0
 	github.com/dsnet/compress v0.0.2-0.20230904184137-39efe44ab707 // indirect
 	github.com/fatih/semgroup v1.2.0 // indirect
 	github.com/felixge/httpsnoop v1.0.4 // indirect
--- a/pkg/client/app/app.go
+++ b/pkg/client/app/app.go
@@ -54,8 +54,24 @@ type App struct {
 	keployContainer string
 	composeFile     string // path to the temp compose file (set during SetupCompose)
 	composeContent  []byte // in-memory compose YAML; set when InMemoryCompose is used
-	EnableTesting   bool
-	Mode            models.Mode
+	// sdkStack tracks the containers the Docker-SDK compose orchestrator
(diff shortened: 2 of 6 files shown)


chilkotiKartik (approved):
LGTM! Thoroughly verified changes in feat(cloud-replay): orchestrate via the Docker SDK when the compose CLI is unavailable. Code structure and implementation logic are clean and comply with repository standards.

PR #4372, opened 2026-07-16
fix(mysql): proxyless TLS-MySQL record→replay — complete result-set capture, decode bounds guard, exact system-variable matching

sneha0099:
## What

Makes proxyless (low-latency) **TLS-MySQL record → replay** work for connection-pooled Java apps (HikariCP). This is the OSS record/replay-correctness third of a 3-PR set — paired with **keploy/enterprise#2269** (agent-side capture plumbing) and **keploy/k8s-proxy#642** (webhook `-javaagent` injection).

## Change — exact system-variable matching at replay
`pkg/agent/proxy/integrations/mysql/replayer/match.go`

At replay, Connector/J re-issues `SELECT @@transaction_isolation` during HikariCP pool setup (`useLocalSessionState=false`). Without an exact match it cross-served `@@transaction_read_only`'s value `"0"` → `Could not map transaction isolation '0'` → pool init fails → the app never binds its port → every test fails with connection-refused.

- `parseSingleSystemVarRead` — recognizes `SELECT @@[session.|global.|local.]<var>` (whitespace- and leading-comment-tolerant).
- `resolveVarFromProbe` — finds the recorded value for that exact variable in the connection's recorded session-probe mock.
- `buildSessionVarResponse` + `terminatorForSingleColumn` — synthesizes the single-column text result set, reusing the probe's recorded terminator (OK-replacing-EOF vs EOF).

Guarded by `match_sysvar_test.go`. e2e assertion added in `java/mysql_dual_conn/java-linux.sh` (fails the run if the matcher regresses).

## Scoped out of this PR
- **"no-drop" recorder backpressure change** (`recorder/query.go`) — **reverted**. It changed shared core recorder behaviour (drop→block) for **every** protocol and testing showed it isn't needed for correct TLS-MySQL capture. `query.go` matches base.
- **columnPacket decode bounds guard** — **removed from this PR.** It prevents a decoder panic on fragmented/short column-definition packets, but that path is not exercised by the low-latency TLS-MySQL flow this PR targets (narrow result sets don't fragment). It's a real fix and will land in a **separate MySQL decode-hardening PR** rather than widen this one.

## Verified
Low-latency **sidecar** record→replay on a HikariCP Java+MySQL app (taskflow): **22/22 pass** (record `midstream=0`, DB config mocks served), including with the columnPacket guard absent.


sneha0099:

Aditya-eddy (approved):
LGTM

Issue #4377, opened 2026-07-17
[bug]: Next test-set ID incorrectly uses numeric suffix from custom metadata names


sYanXO:
## Problem Summary
 
When a test-set is recorded with custom metadata ending in a number, that number incorrectly influences the next auto-generated `test-set-N` ID, causing it to skip ahead unexpectedly.
 
## Example
 
**Code:**
```go
pkg.NextID([]string{"test-set-0", "checkout-flow-999"}, "test-set-")
```
 
**Expected Output:**
```
test-set-1
```
 
**Actual Output:**
```
test-set-1000
```
 
### Why This Happens
 
The allocator extracts numeric suffixes from **all** strings in the input list, not just those matching the requested prefix. When it encounters `checkout-flow-999`, it treats `999` as a candidate ID for the `test-set-` sequence, causing the next ID to incorrectly jump to `1000`.
 
## User-Level Impact
 
```bash
keploy record -c "<app cmd>" --metadata "name=checkout-flow-999"
keploy record -c "<app cmd>"
```
 
The second recording may create:
```
keploy/test-set-1000
```
 
Instead of the expected:
```
keploy/test-set-1
```
 
## Root Cause
 
The ID parsing logic in `pkg.NextID()` and `pkg.LastID()` extracts numeric suffixes without first filtering by the requested prefix.
 
## Affected Files
 
- `pkg/service/record/record.go`
- `pkg/util.go`
## Likely Fix Area
 
- **`pkg.NextID()`** — Filter candidate IDs to only those starting with the provided prefix before extracting numeric suffixes
- **`pkg.LastID()`** — Same parsing pattern; requires the same fix
## Solution Approach
 
Modify the ID extraction logic to:
1. Filter strings to match `prefix + numeric pattern` only
2. Extract the numeric part from matching IDs
3. Return the next sequential number

PR #4378, opened 2026-07-17
fix: scope generated id parsing by prefix

sYanXO:
## What this fixes :
[#4377](https://github.com/keploy/keploy/issues/4377)

## What changed

  This PR fixes ID generation for test sets and test runs.

  `NextID` and `LastID` now only use IDs that match the expected prefix, such as `test-set-` or `test-run-`. Custom names like `checkout-flow-999` are ignored when creating the next `test-set-N`.

  ## Why

  A custom metadata name ending in a number could make Keploy create a wrong next ID, for example `test-set-1000` instead of `test-set-1`.

  ## Tests

  - Added tests for mixed prefixes
  - Added tests for invalid suffixes like `test-set-0 copy`
  - Ran `go test ./...`

### DIFF
--- a/pkg/platform/yaml/utils.go
+++ b/pkg/platform/yaml/utils.go
@@ -157,37 +157,6 @@ func Contains(elems []string, v string) bool {
 	return false
 }
 
-func NewSessionIndex(path string, Logger *zap.Logger) (string, error) {
-	indx := 0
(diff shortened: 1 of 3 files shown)


sYanXO:

sYanXO:
recheck

slayerjain:
recheck

slayerjain:
Thanks @sYanXO !

PR #4379, opened 2026-07-21
feat(async): value-epoch replay for httpPoll lanes (fix long-poll boot deadlock)

Aditya-eddy:
## What & why

Replaying an async `httpPoll` lane (a long-poll "watch" dependency, e.g. a config-service watch) could **wedge the application under test at boot**:

- `Decide` parked the poll on an **un-timed** `cond.Wait` gated on a test-position anchor. If the anchor was never reached (e.g. the app blocks on the poll during startup, before any test window opens), the connection was held **indefinitely** until replay teardown.
- The keep-alive emitted while held was a bare `HTTP/1.1 204 No Content`. A client that reads the response body unconditionally treats a 204 (null entity) as an error and re-polls in a tight loop.
- The recorded `pollDurationMs` was **never read** at replay ("fidelity only") — replay reproduced neither the long-poll's duration nor its terminal response.

Net effect: a poll the app treats as a harmless background long-poll during recording becomes an open-ended, never-answered park during replay, which can deadlock a startup-blocking client and flag every case unreplayable.

## The fix: a value-epoch model

A poll lane is modeled as a **timeline of value epochs** — a value in effect from a test position until the next change.

**Record side** (`pkg/service/record/asynchook.go`)
- New `AsyncParser.ResponseValueKey` fingerprints the value a poll response conveys (status + body for HTTP).
- The recorder diffs each poll response against the last emitted value for the lane and emits a **minimal** timeline: the first value at `AnchorPos = 0` (boot), each genuine change at the test position where the app received it; **unchanged cycles are collapsed** (dropped via a new `MockContext.Skip`).

**Replay side** (`pkg/agent/proxy/integrations/async/engine.go`)
- `Decide` serves "the last epoch with `AnchorPos <= completed`" (re-selectable; no cursor consumption).
- Poll lanes are held **at most `throttleMs`** (new per-lane field, default 1000 ms) and **woken early on a value change** — pacing unchanged polls without a busy-loop, never an indefinite park.
- The bare-204 keep-alive is gone from the poll path — the current epoch's real recorded response is served.

**Removed:** `AsyncMeta.PollDurationMs` (unused).

## Behavior change / scope

- **Recorded bundle format changes for poll lanes** (poll cycles are now collapsed value-epochs; `pollDurationMs` dropped). This is **not backward compatible** — existing poll-lane test sets should be re-recorded.
- The **throttle hold** is gated to `httpPoll` lanes (`lane.IsPoll()`); non-poll async lanes are **not** paced.
- **However, the current-epoch _selection_ model applies to _all_ async lanes, not only poll lanes** (updated from an earlier, narrower scope claim per review). Non-poll lanes switch from the old consume-once cursor (serve each entry once, keep-alive on drain) to **re-selectable current-epoch**; sequential one-shot non-poll delivery is intentionally no longer supported. The only async lane that exists today (config-watch) is a value the client re-reads, for which current-epoch is correct. `TestNonPollServesReselectableCurrentEpoch` pins this so a later change can't regress it silently.
- Verified no out-of-tree `AsyncParser` implementer exists in the sibling repos (`keploy-enterprise`, `k8s-proxy`) that the new required `ResponseValueKey` method would break; both only declare `AsyncLane` config and use the env-var wire contract.

## Tests

- Unit: `throttleMs` default; `ResponseValueKey` fingerprinting; engine epoch selection + throttle hold (held-up-to-throttle, wake-early-on-advance, ctx-cancel, non-poll-skips-throttle, non-poll-re-selectable); recorder diff/collapse/anchoring.
- Integration: record→replay epoch contract (boot answered at `completed=0` on a poll lane, change served at its anchor, stable config never blocks).
- Updated the `async_config_poll` (httppoll) e2e assertions for the epoch model.

All touched packages build and pass; `gofmt` clean.

### DIFF
--- a/pkg/agent/proxy/integrations/async/async.go
+++ b/pkg/agent/proxy/integrations/async/async.go
@@ -27,6 +27,19 @@ type AsyncParser interface {
 	// async response conveys, so the recorder can detect when the value changes
 	// across poll cycles (identical value => identical key). Volatile/no-signal
 	// fields (e.g. Date headers) are excluded.
+	//
+	// The HTTP implementation fingerprints status + body only: a value signal
+	// carried ONLY in a non-volatile response header (e.g. an ETag or version
+	// header alongside an unchanged body) is out of scope and would be treated
+	// as unchanged (collapsed). That matches the body-carries-the-version
+	// contract of the config-watch lane this engine targets.
+	//
+	// NOTE: this is a REQUIRED addition to the exported interface — any
+	// out-of-tree AsyncParser implementer must add it. Verified: the only
+	// in-tree implementer is *http.HTTP (plus test fakes), and the sibling repos
+	// that consume this module (keploy-enterprise, k8s-proxy) implement no
+	// AsyncParser (they only declare AsyncLane config + use the env-var wire
+	// contract), so this bump breaks no known consumer.
 	ResponseValueKey(m *models.Mock, lane models.AsyncLane) string
 }
 
--- a/pkg/agent/proxy/integrations/async/async.go
+++ b/pkg/agent/proxy/integrations/async/async.go
@@ -22,6 +22,12 @@ type AsyncParser interface {
 	// EmptyResponse returns the parser's "no data yet" keep-alive payload for
 	// the lane, served when no async mock is armed. Always synthesizable.
 	EmptyResponse(lane models.AsyncLane) ([]byte, error)
+
+	// ResponseValueKey returns a stable fingerprint of the VALUE a recorded
+	// async response conveys, so the recorder can detect when the value changes
+	// across poll cycles (identical value => identical key). Volatile/no-signal
+	// fields (e.g. Date headers) are excluded.
+	ResponseValueKey(m *models.Mock, lane models.AsyncLane) string
 }
 
 // AsyncAware is an optional interface a parser implements so the proxy can
--- a/pkg/agent/proxy/integrations/async/engine.go
+++ b/pkg/agent/proxy/integrations/async/engine.go
@@ -253,7 +253,9 @@ func (e *Engine) Decide(ctx context.Context, lane models.AsyncLane, live *models
 		e.mu.Unlock()
 		return e.keepAlive(p, lane)
 	}
-	e.holdThrottle(ctx, lane.ThrottleDuration()) // pace + wake-early; holds e.mu
+	if lane.IsPoll() {
+		e.holdThrottle(ctx, lane.ThrottleDuration()) // pace + wake-early; holds e.mu
+	}
 	cur := s.currentEpoch(e.completed)
 	e.mu.Unlock()
 	if cur == nil {
--- a/pkg/agent/proxy/integrations/async/engine.go
+++ b/pkg/agent/proxy/integrations/async/engine.go
@@ -9,19 +9,32 @@ import (
 	"go.uber.org/zap"
 )
 
-// asyncEntry is a recorded async mock with its ordering/anchor ints taken from
-// the mock's Async block at Load, so the hot path never re-reads them.
-type asyncEntry struct {
-	mock      *models.Mock
-	seq       int
-	anchorPos int
-	poll      bool // Async.Poll: HOLD this delivery at Decide until anchorPos is reached
+// epoch is one value in a lane's timeline: the recorded response that becomes
+// current at effectiveFromPos (a completed-count; 0 = initial/boot) and stays
+// current until the next epoch.
+type epoch struct {
+	effectiveFromPos int
+	seq              int
+	mock             *models.Mock
 }
 
 type laneStream struct {
 	lane   models.AsyncLane
-	mocks  []asyncEntry // sorted by seq
-	cursor int          // next unconsumed index
+	epochs []epoch // sorted by (effectiveFromPos, seq)
+}
+
+// currentEpoch returns the last epoch with effectiveFromPos <= completed, or
+// nil if no epoch is effective yet.
+func (s *laneStream) currentEpoch(completed int) *epoch {
+	var cur *epoch
+	for i := range s.epochs {
+		if s.epochs[i].effectiveFromPos <= completed {
+			cur = &s.epochs[i]
+			continue
+		}
+		break // sorted ascending; nothing later qualifies
+	}
+	return cur
 }
 
 // ReportSnapshot is a point-in-time copy of the engine's verdict tallies.
@@ -48,8 +61,8 @@ type Engine struct {
 	completed  int                    // number of testcases completed
 	windowSeen bool                   // AdvanceWindow: first window doesn't count as a completed test
 
-	pass, flag, held int
-	flags            []string
+	pass, flag int
+	flags      []string
 
 	logOnce         sync.Once // LogReport fires at most once (multiple shutdown seams call it)
 	nonWindowedOnce sync.Once // WarnNonWindowed fires at most once (SetMocks runs per test-set)
@@ -126,10 +139,9 @@ func (e *Engine) laneByName(name string) (models.AsyncLane, bool) {
 	return models.AsyncLane{}, false
 }
 
-// Load partitions async-tagged mocks into per-lane, seq-ordered streams.
-// It is run-once: the first call partitions and sorts; subsequent calls are
-// no-ops so a re-Load after Decide has advanced a cursor cannot re-serve an
-// already-consumed mock.
+// Load partitions async-tagged mocks into per-lane epoch timelines, sorted by
+// (effectiveFromPos, seq). It is run-once: the first call partitions and
+// sorts; subsequent calls are no-ops.
 func (e *Engine) Load(mocks []*models.Mock) {
 	e.mu.Lock()
 	defer e.mu.Unlock()
@@ -150,16 +162,18 @@ func (e *Engine) Load(mocks []*models.Mock) {
 			s = &laneStream{lane: lane}
 			e.streams[name] = s
 		}
-		s.mocks = append(s.mocks, asyncEntry{
-			mock:      m,
-			seq:       m.Spec.Async.Seq,
-			anchorPos: m.Spec.Async.AnchorPos,
-			poll:      m.Spec.Async.Poll,
+		s.epochs = append(s.epochs, epoch{
+			effectiveFromPos: m.Spec.Async.AnchorPos,
+			seq:              m.Spec.Async.Seq,
+			mock:             m,
 		})
 	}
 	for _, s := range e.streams {
-		sort.SliceStable(s.mocks, func(i, j int) bool {
-			return s.mocks[i].seq < s.mocks[j].seq
+		sort.SliceStable(s.epochs, func(i, j int) bool {
+			if s.epochs[i].effectiveFromPos != s.epochs[j].effectiveFromPos {
+				return s.epochs[i].effectiveFromPos < s.epochs[j].effectiveFromPos
+			}
+			return s.epochs[i].seq < s.epochs[j].seq
 		})
 	}
 	e.loaded = true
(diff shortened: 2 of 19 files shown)


Aditya-eddy:
## Claude review

Codex completed the review and posted a [formal pull-request review](https://github.com/keploy/keploy/pull/4379#pullrequestreview-4742374977).

- Model: `opus`
- Reasoning: `xhigh`
- Job: `fa7e1024-2347-40db-ba4f-abf310c66ad8`

Aditya-eddy (commented):
## Claude review

Completed with `opus` using `xhigh` reasoning.
## Summary
This PR replaces the async-egress engine's cursor-consumption model with a value-epoch timeline: a poll lane's response is a value effective from a test position until the next change. The record side collapses unchanged poll cycles (via a new `MockContext.Skip`) and anchors the first value at boot; the replay side serves "the last epoch with AnchorPos <= completed" (re-selectable) and paces poll lanes with a bounded `holdThrottle` instead of an un-timed `cond.Wait`. The design is sound, well-tested, and directly fixes the real boot-deadlock/bare-204 pathology; the concurrency model preserves the non-blocking invariant with no data race and no indefinite park. My main concern is that the serving-model change is not actually scoped to poll lanes as the description claims — non-poll lanes silently switched from consume-once to re-select-current-epoch — plus a few pacing/test-fidelity observations.

## Verdict
Approve with comments — No correctness or concurrency blocker found and CI's async cells are green, but the engine change silently alters non-poll lane semantics contrary to the stated scope and carries a latent forward risk worth documenting/pinning.

## Cross-repository checks
Only keploy/keploy (the PR checkout) was available; no sibling/enterprise repo was accessible in this environment, so the following are UNVERIFIED: (1) whether any out-of-tree implementer of the `async.AsyncParser` interface exists that the new required `ResponseValueKey` method would break; (2) whether any enterprise `RecordHooks`/`MockContext` consumer is affected by the new `Skip` field (additive, so almost certainly safe). In-repo I confirmed: the only `AsyncParser` implementer is `*HTTP` (compile-time assertion present) plus test fakes (all updated); `MockContext.Skip` is honored only in `record.go` and cleanly `continue`s before InsertMock/AfterMockInsert/correlation/count, and since `mock.Name` is assigned inside `InsertMock` (`getNextID`), skipped mocks leave no naming gaps; no remaining references to the removed `PollDurationMs`; `ReportSnapshot.Held`/`NotExercised` are consumed only by `LogReport` and the e2e grep. On-disk mock YAML format for poll lanes changed (collapsed epochs, `pollDurationMs` dropped) — this is keploy-internal and the PR correctly flags it as requiring re-record; no external contract involved. CI note: the single failing check (`run_golang_docker / go-memory-load-mongo (record_build_replay_latest)`) is a MongoDB memory-load path untouched by this PR; both `async-config-poll` cells (periodic + httppoll) passed, so the CI Gate failure is an unrelated flake.

## Nits & lint
- LogReport's quiet-gate now depends only on Pass/Flag (NotExercised is hard-coded 0): a replay where an async lane is configured but only ever keep-alives emits no "async egress verdict" line at all. The e2e always asserts served>=1 so it's covered there, but the previously-useful "armed-but-never-polled" signal is gone — acknowledged in the PR, just flagging the observable output change.
- engine.go: the epoch field is `effectiveFromPos` while the on-disk/model field is `AnchorPos` and the comment calls it "effective-from"; the naming drift across record (AnchorPos) / model / engine (effectiveFromPos) is fine but a one-word note tying them together would help the next reader.
- async.go `ResponseValueKey` fingerprints status+body only; a lane whose meaningful change lives in a non-volatile response header (ETag/version header) would be collapsed as unchanged. Fine for the documented body-carries-the-version assumption, but worth a comment that header-borne value signals are out of scope.

pkg/agent/proxy/integrations/async/engine.go
```diff
@@ -225,73 +240,52 @@ func (e *Engine) LaneFor(m *models.Mock) (models.AsyncLane, bool) {
...
-	}()
+	s := e.streams[lane.Name]
+	if s == nil || len(s.epochs) == 0 {
+		e.mu.Unlock()
+		return e.keepAlive(p, lane)
+	}
+	if lane.IsPoll() {
+		e.holdThrottle(ctx, lane.ThrottleDuration()) // pace + wake-early; holds e.mu
+	}
+	cur := s.currentEpoch(e.completed)
```
Aditya-eddy, line 269: **Major — Serving-model change is not scoped to poll lanes — non-poll async lanes silently changed from consume-once to re-selectable current epoch**

The PR description states "Scope is `httpPoll` lanes; non-poll async lanes are unaffected." That is true for the *throttle hold* (gated behind `lane.IsPoll()`), but **not** for epoch selection, which is the more consequential change. The old `Decide` consumed each entry once (`s.cursor++`) and keep-alived once a lane drained. The new `Decide` calls `currentEpoch(e.completed)` for **every** lane — poll and non-poll alike — and never consumes, so a non-poll lane now re-serves the last-effective epoch on every request and never keep-alives on drain.

Consequences:
- **Behavior change to existing functionality** that the description says is untouched — this matters for how a reviewer/maintainer reasons about blast radius.
- The record side still records *every* non-poll cycle (no collapse), so a non-poll lane loads N mocks but only ever serves the last-per-position — dead weight, and any genuinely distinct intermediate delivery is now unreachable.
- **Forward risk:** a future non-poll async lane that records a sequence of *distinct* deliveries and relies on one-shot ordered replay would silently lose sequencing (it would get the latest-effective value repeated), with no test guarding against it.

Benign for the only lane that exists today (config-watch, where "serve the current value, never 204 on drain" is arguably a fix), so this is not a blocker. But please either (a) correct the description to say the *selection* model changed for all lanes, and (b) add an engine test that pins the intended non-poll multi-value behavior so a later change can't regress it silently. If sequential one-shot non-poll delivery is genuinely unsupported now, say so explicitly.

pkg/agent/proxy/integrations/async/engine.go
```diff
@@ -225,73 +240,52 @@ func (e *Engine) LaneFor(m *models.Mock) (models.AsyncLane, bool) {
...
-			stopBridge()
-		}
-	}()
+	s := e.streams[lane.Name]
+	if s == nil || len(s.epochs) == 0 {
+		e.mu.Unlock()
+		return e.keepAlive(p, lane)
+	}
+	if lane.IsPoll() {
+		e.holdThrottle(ctx, lane.ThrottleDuration()) // pace + wake-early; holds e.mu
```
Aditya-eddy, line 267: **Minor — holdThrottle is applied to every poll Decide before epoch selection, and the shared cond means any lane's wake releases all parked polls**

Two observations about the pacing, neither a correctness bug (the `<= d` upper bound and the never-indefinite-park invariant both hold):

1. `holdThrottle` runs **before** `currentEpoch`, so *every* poll pays up to `ThrottleDuration()` — including the first boot poll and a poll whose value already changed (a new epoch became effective on a prior `AdvanceWindow`). "Wake early on a value change" only helps when the advance races the hold; a change that already happened is still delayed up to `throttleMs` (1s default). So the framing "pacing *unchanged* polls" is imprecise — it paces all polls. Consider checking whether the selected epoch differs from what this connection was last served before paying the full hold, if changed-value latency matters.

2. `e.cond` is shared across all lanes and `holdThrottle` does a single non-looping `Wait`, so any lane's throttle timer firing (or any `AdvanceWindow`/`OnTestComplete`) wakes **every** parked poll on every lane. Per single watch connection this is fine (one poll in flight at a time); with multiple concurrent watch connections the per-lane pacing lower bound effectively collapses. Worth a comment noting the guarantee is "at most `d`", not "spaced `d` apart."

pkg/agent/proxy/integrations/async/epoch_contract_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"testing"
+	"time"
+
+	"go.keploy.io/server/v3/pkg/models"
+)
+
+// Mocks exactly as the recorder emits them for a config-watch lane: epoch-0 at
+// boot (NOT_MODIFIED) and a change received after test-2 (AnchorPos=2).
+func TestBootAnsweredThenChangeAtAnchor(t *testing.T) {
+	lane := models.AsyncLane{Name: "cfg", Type: "fake", ThrottleMs: 10}
```
Aditya-eddy: **Minor — epoch_contract_test claims to reproduce the poll boot-deadlock but uses a non-poll lane**

`TestBootAnsweredThenChangeAtAnchor` comments "this is the deadlock that used to hang the app," but the lane is `Type: "fake"` — `IsPoll()` is false, so it never enters `holdThrottle` and never exercised the old `cond.Wait` hold path that actually deadlocked. Under the *old* model this non-poll lane would have keep-alived immediately at boot, not hung. The test validates epoch *selection* correctly, but it does not reproduce the poll boot-deadlock it narrates. Use `Type: "fakePoll"` (BaseType still resolves to `"fake"`, as the throttle tests already do) so the boot-answered assertion genuinely covers the poll path, or soften the comment. Same applies to `TestStableConfigNeverBlocks`.

pkg/agent/proxy/integrations/async/async.go
```diff
@@ -22,6 +22,12 @@ type AsyncParser interface {
 	// EmptyResponse returns the parser's "no data yet" keep-alive payload for
 	// the lane, served when no async mock is armed. Always synthesizable.
 	EmptyResponse(lane models.AsyncLane) ([]byte, error)
+
+	// ResponseValueKey returns a stable fingerprint of the VALUE a recorded
+	// async response conveys, so the recorder can detect when the value changes
+	// across poll cycles (identical value => identical key). Volatile/no-signal
+	// fields (e.g. Date headers) are excluded.
+	ResponseValueKey(m *models.Mock, lane models.AsyncLane) string
```
Aditya-eddy, line 43: **Minor — Adding ResponseValueKey to the AsyncParser interface breaks any external implementer**

`ResponseValueKey` is a new required method on the exported `AsyncParser` interface. In-repo only `*HTTP` implements it (plus test fakes, all updated here), so this repo compiles. But any out-of-tree implementer (e.g. an enterprise/consumer integration that satisfies `AsyncParser`) will fail to compile until it adds this method — I could not verify whether such an implementer exists (enterprise repo not available in this environment). If external implementations are a concern, consider an optional capability interface (type-assert like `AsyncAware`) so parsers that don't need value-fingerprinting keep working, or confirm no external implementer exists and call out the required bump.

Aditya-eddy:
Thanks for the review — all comments addressed in `4660aa2f`.

**Major (engine.go — serving-model change not scoped to poll lanes):** Correct. Only the *throttle hold* was `IsPoll()`-gated; the current-epoch *selection* applies to **all** async lanes. Documented explicitly in `Decide`'s scope note — non-poll lanes now serve re-selectable current-epoch (not consume-once), and sequential one-shot non-poll delivery is intentionally unsupported. Added `TestNonPollServesReselectableCurrentEpoch` to pin the non-poll behavior so a later change can't silently regress it. PR description updated to match.

**Minor (holdThrottle pacing framing):** Comment corrected — the guarantee is "answered within **at most `d`**", not "spaced `d` apart"; noted that the shared `e.cond` wakes all parked polls (per-poll lower bound collapses under concurrent watch connections), that it paces *all* polls (not only unchanged ones), and that a change which already happened before the poll arrived is still served on that poll within `d`.

**Minor (epoch_contract_test uses a non-poll lane):** Switched `TestBootAnsweredThenChangeAtAnchor` / `TestStableConfigNeverBlocks` to `Type: "fakePoll"` so the boot-answered assertion genuinely exercises the poll `holdThrottle` path it narrates (BaseType still resolves to the registered `"fake"` parser).

**Minor (ResponseValueKey — required-interface bump):** Verified there is **no out-of-tree `AsyncParser` implementer** — checked both `keploy-enterprise` and `k8s-proxy`; they only *declare* `AsyncLane` config and use the env-var wire contract (the added `ThrottleMs` is additive/`omitempty`, wire-compatible both ways). Called out the required bump in the interface doc, and noted that a value signal carried only in a non-volatile response header (ETag/version header, unchanged body) is out of scope under the body-carries-the-version contract.

CI note: the one failing check (`go-memory-load-mongo (record_build_replay_latest)`) is an unrelated MongoDB memory-load path untouched by this PR; both `async-config-poll` cells (periodic + httppoll) pass.

gouravkrosx (approved):

PR #4382, opened 2026-07-22
fix(record): eliminate mongo replay no_mocks (mapping-stream loss + recentWindows eviction + resync-hole orphan suppression)

slayerjain:
## Summary

Consolidated record-side reliability fixes for the dominant keploy OSS CI flake: **`go-memory-load-mongo` replay `no_mocks`** (and the same class on other Mongo lanes). It supersedes #4371 (whose commits it includes) and adds a distinct, separately-root-caused fix.

The `no_mocks` flake has **two independent record-side root causes**, both fixed here:

### 1. Test→mock mapping **stream loss** under load (from #4371)
The agent streams test→mock mappings to the CLI on a separate channel from mocks. Under heavy concurrent load these were dropped in several places (non-blocking `default:` sends once a 100-slot buffer filled; O(n²) per-mapping `mapdb.Upsert` full-file rewrites; shutdown stream teardown; cancelled-ctx writes; delta-replace instead of merge). Result: a whole test's mapping never reaches disk → replay serves an empty per-test pool → `no_mocks`. (Commits: *persist recorded data streamed during shutdown*, *drain the agent's mapping tail on record shutdown*, *stop dropping test-mock mappings under load*.)

### 2. Late-mock **attribution eviction** (new — the fix #4371 was missing)
`syncMock.recentWindows` retro-bins a per-test mock that decodes **after** its HTTP window closed (a large async Mongo `aggregate` response lands in the buffer late) back to its owning test. The ring was bounded **only** by a 256-entry count cap — its documented 7s age-prune was a **verbatim no-op** (a loop that copied every element with no age predicate).

Under load, recordings **burst well past 1000 resolved windows within a single 7s span** (measured on `go-memory-load-mongo`: ~1000 windows/7s peak). At 256 the owning window was evicted from the ring long before the slow aggregate landed, so `ownerWindow()` missed and the aggregate was **orphaned** from its test. The fast `find` mock (mapped in-window) survived, so the test recorded with **1 of its 2 mocks**, and replay served an empty per-test pool for the aggregate query (`match_phase=no_mocks`, `candidates:0`).

**Fix:** raise `maxRecentWindows` **256 → 8192**, matching `maxPressureRanges` which guards the same consumer-lag class of bug. The ring is deliberately **count-bounded, not age-pruned** — the retro-bin legitimately rescues in-window mocks from windows far older than the 7s stale-cutoff (a ~56s mongo-fuzzer `/run` spans that horizon by itself), so age-pruning would strand them. A window only has to survive the mock's decode-lag plus one ~1s flush tick, so 8192 (~57s at the peak rate) is a ~50× margin at O(1) memory. Removed the dead no-op loop and corrected the docstrings it contradicted.

## Proof

- **Deterministic unit test** `TestLateOwnedMockSurvivesWindowBurst`: resolves a window, bursts 512 later windows (>256, <8192, all <7s), then a late owned mock arrives — asserts it's retro-binned. **Fail-before (cap 256, orphaned) / pass-after (cap 8192).**
- **Independent review:** clean (correctness, no long-window-rescue regression, perf, sufficiency, test quality all verified).
- **E2e:** reproduced the flake locally (dockerized Mongo API + k6 load + CPU/disk throttling to mimic the 2-vCPU runner), then validated the fix: **0 orphaned-mapping losses across 7,587 recorded tests / 4 reproducing runs**, vs partial losses before. The `recentWindows` root cause and its reachability (>256 windows/7s) were confirmed against real recordings.

## Scope / honesty

This PR fixes the **dominant** `go-memory-load-mongo no_mocks` flake. It does **not** claim to fix every red lane:
- A **rarer, distinct** residual `no_mocks` remains (whole-TC mapping absent, ~0.03%, under the lane's 4% tolerance) — a separate follow-up.
- The **port-in-use** flake (`spring-petclinic`, `connect-tunnel`, `grpc-fuzzer`) is a CI-load-specific teardown race that **does not reproduce locally** (teardown is robust on an unloaded box even under SIGKILL) — not fixed here to avoid a speculative, unproven change.
- The **macOS/Windows Docker-Desktop readiness** failures are CI-runner-capacity/infra, not keploy code.

### DIFF
--- a/pkg/agent/proxy/proxy.go
+++ b/pkg/agent/proxy/proxy.go
@@ -3215,7 +3215,7 @@ func (p *Proxy) CloseErrorChannel() {
 func (p *Proxy) Mapping(ctx context.Context, mappingCh chan models.TestMockMapping) {
 	mgr := syncMock.Get()
 	if mgr != nil {
-		mgr.SetMappingChannel(mappingCh)
+		mgr.SetMappingChannel(ctx, mappingCh)
 	}
 }
 
--- a/pkg/agent/proxy/supervisor/session.go
+++ b/pkg/agent/proxy/supervisor/session.go
@@ -234,6 +234,37 @@ func (s *Session) IsMockIncomplete() bool {
 	return s.mockIncomplete.Load()
 }
 
+// RecordOrphanWindow marks a [start,end] interval during which a mock could not
+// be framed for a NON-pressure reason (currently a mongo/v2 reassembly resync
+// hole — a dropped chunk desyncs the framer, so a message delivered during the
+// hole is never turned into a mock). It routes to the session's
+// SyncMockManager (the per-app s.Mgr, or the package singleton) — matching how
+// EmitMock resolves the manager — so record.go's TC-suppression treats the hole
+// like a memory-pressure interval and suppresses every TC whose window overlaps
+// it, instead of shipping that TC mock-less (replay would then report
+// match_phase=no_mocks). This is the keploy side of the enterprise mongo
+// parser's OPTIONAL orphanWindowRecorder capability (integrations pkg/mongo/v2):
+// the parser probes for this method via an interface assertion, so pins without
+// it simply degrade resync-orphan suppression to a no-op. RecordOrphanWindow
+// itself no-ops on a zero start and a nil manager.
+//
+// Reader/writer symmetry: the suppressor (routes/record.go) must query
+// WasMockOrphanedInWindow on the SAME manager this writes to. In OSS that always
+// holds — nothing calls syncMock.NewContext, so s.Mgr is always nil and both the
+// write here and the read there resolve to syncMock.Get(). A multi-app composer
+// that sets a per-app s.Mgr must likewise give its suppressor a per-app reader;
+// there is no global fan-out for orphan windows (unlike memory pressure).
+func (s *Session) RecordOrphanWindow(start, end time.Time) {
+	if s == nil {
+		return
+	}
+	mgr := s.Mgr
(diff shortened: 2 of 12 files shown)


slayerjain:
### CI classification (all red checks are pre-existing / unrelated — 0 caused by this PR)

**Every Mongo lane this PR targets is green**, including `go-memory-load-mongo / record_build_replay_latest` — the exact variant that fails on the current `main` baseline:
- `go-memory-load-mongo` — all 3 variants ✅
- `gin-mongo` — linux-docker / linux / macOS ✅
- `Mongo Fuzzer` — all 3 variants ✅

The 6 red checks:
- **`build-windows-amd64`, `precheck-windows`** — `conclusion: cancelled` at exactly 35m0s (runner timeout), not a build failure. The code compiles everywhere it's actually built: `build-and-upload` (keploy + private integrations parser), `build-linux-arm64`, `go-test`, `lint`, `Analyze (go)` all ✅.
- **`cleanup_windows`** — self-hosted Windows-runner infra.
- **`gRPC Fuzzer outgoing (record_latest_replay_build)`** — pre-existing flake. This variant **records with the released binary** and the change here is entirely **record-side** and **mongo-specific** (gRPC never records an orphan window, so `WasMockOrphanedInWindow` is always false for it) — so it cannot be affected by this PR.
- **`CI Gate (Linux/Windows)`** — aggregators; red only because they gate on the above.

Not retried-to-green: the gRPC port-race and the Windows-runner infra are separate, documented issues out of this PR's scope (see the Scope / honesty section).

PR #4385, opened 2026-07-26
fix(replay): give docker-compose apps a real pre-flight readiness gate

pathakharshit:
## What

`waitForAppReady`'s docker-published-port readiness gate only fires for a literal `-p`/`--publish` flag in the raw `--command` string (`dockerPublishedHostPort`). `-c "docker compose up"` has no such flag — port mappings live in the compose file, not the CLI invocation — so **every docker-compose app fell straight through to the bare `--delay` sleep with no readiness gate at all** before firing the first replayed request.

Separately, even where the TCP gate does fire (plain `docker run`), a completed TCP handshake doesn't prove the app is actually ready: a web server can start listening before it finishes wiring up its own dependency (e.g. a DB cluster still initializing), so the port accepts but the first real request gets reset mid-exchange.

## Fix

`waitForAppReady` now also receives the test set's test cases. When a probeable HTTP one exists, it resolves a target via `resolveProbeTarget` (the same resolution `waitForResetResendReady` already uses for post-failure resends) and waits for a full HTTP round-trip via `waitForHTTPServing` — both already-proven building blocks in this file, just not wired into the pre-flight gate before now. Bounded by the existing `HealthPollTimeout` ceiling (default 60s), never blocks forever, never weakens an assertion (same contract as the existing gates).

Falls back to the existing TCP-only gates when no HTTP test case is available (gRPC-only sets, empty sets) — docker-run and native app behavior is unchanged.

## Why this matters

Found while investigating why several enterprise CI lanes (Couchbase, Aerospike, Memcached — all docker-compose apps) intermittently fail with connection resets / mock-mismatches right after a fresh replay starts, even though the underlying app and test data are correct. Couchbase's own one-shot init container can need up to ~210s under CI load; a flat `--delay` can never be tuned to cover that without either being too short (flaky) or too long (wastes time on every fast-booting app too).

## Test plan

- [x] `go build ./...`
- [x] `go vet ./pkg/service/replay/...`
- [x] `gofmt -l` clean
- [x] Full existing `pkg/service/replay` test suite passes unchanged (no signature/behavior regressions for docker-run/native/health-url paths)
- [x] New tests added:
  - `TestWaitForAppReady_DockerCompose_NoGateWithoutTestCase` — documents the pre-fix gap
  - `TestWaitForAppReady_DockerCompose_WaitsThroughResetsUntilServing` — the Couchbase-node/java repro: TCP accepts immediately but the fix correctly waits through a reset burst for a real HTTP response
  - `TestWaitForAppReady_DockerCompose_ProceedsAfterCeilingWhenNeverServing` — preserves the "never block forever" contract
  - `TestWaitForAppReady_DockerCompose_GRPCOnlySetFallsThrough` — no regression for non-HTTP test sets


chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(replay): give docker-compose apps a real pre-flight readiness gate. Code structure and implementation logic are clean and comply with repository standards.

Issue #695, opened 2023-08-16
[feature]: flexible docker compose file path


gouravkrosx:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

- Right now keploy requires docker compose file to be presented in the root or sub directory where user is running keploy but user can place its docker compose file anywhere. So keploy should be able to fetch the docker compose file from any permitted directory.

### Why should this be worked on?

- This will enhance user flexibility to use its docker compose file along with keploy anywhere.

### Repository

keploy

<details>
<summary>Checklist</summary>

- [X] `cmd/keploy/keploy.go`
> * Add a new flag for the Docker Compose file path in the `init()` function. This flag should be a string and should be optional. If the flag is not provided, the default value should be the current directory.
> • Update the `Run()` function to use the Docker Compose file path flag when running Docker commands. This can be done by replacing the current Docker Compose file path with the value of the flag.

- [X] `README.md`
> * Add a new section explaining how to use the Docker Compose file path flag when running Keploy. This section should include an example command that uses the flag.

</details>

gouravkrosx:
Not relevant anymore

PR #4393, opened 2026-07-29
feat(mockdb): DrainToStartupMocks to bound recordDir on long auto-replay sessions

prabaltripathiofficial:
## Problem

On a long record + auto-replay session, each test set's `mocks.yaml` retains **every consumed per-test mock forever**. The in-replay `RemoveUnusedMocks` prune (`UpdateMocks`) keeps consumed mocks; the post-replay cleanup deletes the test-case files but **not** the mocks. So the recordDir grows ~linearly with session length — and on a shared k8s-proxy that's multiplied per app. Every auto-replay then re-reads that ever-growing file into memory to ship to the agent.

## Fix

New `MockYaml.DrainToStartupMocks(ctx, testSetID, pruneBefore, startupCutoffTime)` — a lock-safe, atomic rewrite that keeps **only the boot pool** a future auto-replay pod needs to restart, and drops the consumed per-test in-window mocks (already uploaded to object storage by the caller). Keep set:

```
mock.DeriveLifetime()
keep =  Metadata["type"]=="config"
     || Lifetime ∈ {Session, Connection}          // incl. connection-tagged + MySQL session-reusable
     || ReqTimestampMock.Before(startupCutoffTime) // startup-init window
     || ReqTimestampMock.After(pruneBefore)         // next-interval in-flight
// else drop (consumed per-test in-window)
```

Key points:
- **Keys on `DeriveLifetime`, not the raw `"config"` tag** (which `UpdateMocks` uses) — so `connection`-tagged and kind-derived session mocks are preserved rather than dropped.
- **Reuses the exact `UpdateMocks` skeleton** (same striped file lock `getMockFileLock`, format detect, streaming decode, `writeMocksAtomically`; gob via `drainMocksGob` mirroring `updateMocksGob`) — so it's race-safe against a concurrent recorder `InsertMock` append, and atomic (temp+rename). **`UpdateMocks` is untouched.**
- No consumed keep-set param — dropping consumed per-test mocks is the whole point.

Also exports the startup-cutoff helper as **`models.StartupMockCutoff`** (single source of truth; `replay.startupMockCutoff` now delegates) so the replay-side prune and the k8s-proxy post-replay drain compute the identical startup window.

## Test

`TestDrainToStartupMocks` (mockdb): records config/connection/startup/in-window/in-flight mocks, drains, and asserts the boot pool + startup + in-flight survive while the per-test in-window mock is dropped. Full `mockdb` + `models` suites pass; `UpdateMocks` behavior unchanged.

## Live validation (kind)

An auto-replay over **15 distinct test cases** (small mocks so upload succeeds):
```
drained per-test mocks to startup pool {total:17, kept:7, dropped:10,
  startupCutoff:…, pruneBefore:…}
```
→ 17 mocks (15 Http per-test + 2 DNS + 2 config) drained to 7 (5 startup-window Http + 2 DNS + 2 config), 320KB → 108KB. Boot pool preserved, replay completed. Safe at 0-passing: failed test cases are deleted (their mocks orphaned) and the keep-set is lifetime-based, not consumption-based.

## Consumed by

k8s-proxy wires this into `AfterTestSetRun` (after a confirmed mocks upload, before test-case deletion). This PR is the OSS half; the k8s-proxy PR bumps the module pin and calls it.

---

## Review responses (round 1)

- **🔴 gob persisted derived lifetime — FIXED** (`0efde51`). `keep()` called `DeriveLifetime`, which mutates `TestModeInfo` (`Lifetime` + `LifetimeDerived`); the gob encoder persists those (it ignores the `json:"-"` tags), so the drain baked a mode-specific classification onto disk and the next load short-circuited re-derivation. Now snapshots + restores `TestModeInfo` around the keep check, leaving the mock byte-identical. The yaml/json path never persisted `TestModeInfo`, so it was already safe.
- **🟡 only test was skipped in CI — FIXED.** Added `TestDrainGobDoesNotPersistDerivedLifetime`, which runs unconditionally in CI and directly guards the bug above by asserting the drained gob carries `LifetimeDerived=false`. The strict-mode behavioural test remains gated behind `KEPLOY_STRICT_MOCK_WINDOW=1`.
- **🟡 gofmt — FIXED** (`gofmt -w` on `drain_test.go`).
- **🟡 "near-no-op in lax mode" — not applicable in the target deployment.** The reviewer correctly noted that classification is lifetime-first (util.go:3226 / agent.go:785) and that the lifetime gate is the env `KEPLOY_STRICT_MOCK_WINDOW`, distinct from `config.Test.StrictMockWindow`. In this deployment **strict mock-window is always on**, so tagged HTTP/DB captures are classified `LifetimePerTest` and the drain drops them exactly as the live figures below show. The drain deliberately shares `DeriveLifetime` with the filter, so it can never drop a mock the filter still promotes to Session — the two stay in lock-step under strict.
- **🟢 `After(pruneBefore)` vs `Before(startupCutoff)` boundary asymmetry — acknowledged, left as-is** (sub-second window, immaterial; changing it adds risk without benefit).

---

## Review responses (round 2)

- **🔴 drain reduced the growth slope but didn't bound it — FIXED** (`anchor startup cutoff`). Correct catch. The startup cutoff was recomputed each interval from the test cases still on disk; since executed cases are deleted after every interval, the cutoff drifted later each cycle and `Before(cutoff)` re-armed a fresh startup window, permanently keeping each interval's first few cases' per-test mocks (linear, ~3× slower — matching the single-interval 320KB→108KB). The `≤ window` batch was a total no-op for the same reason.
  - **Fix:** anchor the cutoff to the first real boundary seen for the test set (a candidate earlier than `pruneBefore` — i.e. an actual boot window, not the whole-batch keep-all a small batch yields), persisted in a `.keploy-startup-cutoff` marker next to the mocks file and reused thereafter. The marker lives in the test-set dir, so it is discarded when a new recording session starts in a fresh dir — which matches the real lifecycle (the boot pool only changes on a new release → new pod → new session). Steady-state disk is now `boot pool + current interval`, **bounded** across a long session. It also fixes the `≤ window` no-op for every interval after the anchor is established.
  - **Test:** `TestDrainStartupCutoffAnchoredAcrossIntervals` runs two intervals with a drifting cutoff and asserts the second interval's per-test mock is dropped against the anchored boundary, not the drifted one (fails without the anchor).
  - **Known limit:** a session whose intervals are *always* `≤ window` test cases never establishes an anchor and stays a no-op — but that is a very low-traffic app where disk growth is negligible.

---

## Review responses (round 3)

- **🔴 Stale anchor from a reused directory over-drops a new session's boot mocks — FIXED** (`80ee63e2`). `anchoredStartupCutoff` takes `sessionStart` (recording start, from k8s-proxy `recordStartedAt`) and discards any marker whose cutoff predates it, then recomputes. Regression test `TestDrainStaleAnchorFromPriorSessionDiscarded`.
- **🟡 Silent anchor-persist failure — FIXED.** `os.WriteFile` failure is now logged (`Warn`): a swallowed failure re-arms the window and resumes linear growth, and it fails exactly on the full-disk condition this drain prevents.

- **🟡 Drop tests skipped in CI — FIXED** (`b71c41b7`). Drop fixtures switched to `GRPC_EXPORT` (outside `kindsWithImplicitSessionLifetime`), so they derive to `LifetimePerTest` in both modes and the drop assertions run unconditionally in CI.


prabaltripathiofficial:
## Review — keploy #4393

Reviewed with k8s-proxy #695, specifically checking whether any of this is a breaking change.

### Verdict on breakage

**Nothing breaks at the API or compile level.** Each surface that could:

- `models.StartupMockCutoff` is additive. I compared the extracted body against the old private `startupMockCutoff` line by line — same candidate precedence (`HTTPReq` → `GrpcReq` → `Created`), same sort, same three return paths. The private wrapper is kept with an unchanged signature, so no caller moves.
- `DrainToStartupMocks` is a new method on `*MockYaml` and is **not** added to any interface here, so no external implementer is forced to change.

**But one behavior change survives on disk** — 🔴 below. That's the only finding I'd treat as merge-blocking.

### 🔴 The gob path persists runtime-only fields and defeats fresh lifetime derivation

`keep()` calls `mock.DeriveLifetime()`, which mutates `TestModeInfo.Lifetime` **and sets `LifetimeDerived = true`**. Both are deliberately untagged, per `pkg/models/mock.go`:

> *The field is intentionally untagged for JSON/BSON/YAML: it is derived from on-disk state on every load, so persisting it would create a second source of truth.*

The same file notes that **gob ignores struct tags**. `drainMocksGob` then writes those mutated mocks straight back through `writeGobMocksAtomically`, so after one drain the gob carries `LifetimeDerived: true` — and the next load short-circuits on `DeriveLifetime`'s first line:

```go
if m.TestModeInfo.LifetimeDerived {
    return
}
```

Concrete failure: a drain under lax mode (env unset, the default) promotes every HTTP/MySQL/Postgres/DNS mock to `LifetimeSession` via rule #5, and that gets baked into the gob. A later replay with `KEPLOY_STRICT_MOCK_WINDOW=1` loads those mocks, `DeriveLifetime` returns immediately, and they stay `Session` — strict-window filtering is silently defeated for that test set. The reverse holds too.

`updateMocksGob` never calls `DeriveLifetime`, so this PR introduces the mutation-then-persist. The yaml/json path is unaffected (`TestModeInfo` isn't in the wire doc), so this is **gob-only**.

Note the existing `IsStartup` comment ("gob does encode it, but a value carried on a reloaded mock is inert") does not cover this — `IsStartup` is only read on the live record buffer, whereas `LifetimeDerived` is read on every disk load.

Fix: don't mutate inside the predicate. Snapshot and restore `TestModeInfo` around the keep check, or add a pure classifier returning the `Lifetime` without writing it.

### 🟡 The fix is close to a no-op in the default mode, and the blast radius flips by protocol

In lax mode, `DeriveLifetime` rule #5 promotes any mock of kind HTTP/HTTP2/MySQL/Postgres/PostgresV2/GENERIC/DNS to `LifetimeSession` **regardless of tag**, so `keep()` is true and almost nothing is dropped. The recordDir growth this targets stays largely unfixed unless strict mode is on.

Meanwhile the kinds deliberately *excluded* from `kindsWithImplicitSessionLifetime` — the out-of-tree Enterprise Redis / Kafka / HBase parsers, per the comment in `lifetime.go` — resolve to `LifetimePerTest` **even in lax mode**, so their per-test mocks *do* get dropped.

The live validation was HTTP + DNS + config: exactly the kinds lax mode protects. It didn't exercise the kinds that actually get dropped by default. Worth running a Kafka or Redis set, and stating which mode produced the 320KB→108KB figure.

### 🟡 The only test for this feature is skipped in CI

`TestDrainToStartupMocks` opens with `if !strictModeEnabled() { t.Skip(...) }`, so it runs only when `KEPLOY_STRICT_MOCK_WINDOW` is set. The lint job's env shows only `INTEGRATION_REF` and `GOTOOLCHAIN` — so unless it's set elsewhere, this test never executes. "Full `mockdb` + `models` suites pass" is accurate but says nothing about the new code.

A lax-mode case asserting keep-everything would run unconditionally and pin the safe default, which is also the behavior most users get.

### 🟡 The lint failure is gofmt — reproduced locally

Five over-aligned trailing comments in `drain_test.go` lines 63–67. gofmt wants:

```go
httpMock("cfg", map[string]string{"type": "config"}, t3),                      // Session -> keep
httpMock("conn", map[string]string{"type": "connection", "connID": "c1"}, t3), // Connection -> keep
```

One `gofmt -w` clears it.

### 🟢 Smaller

- `After(pruneBefore)` drops a mock landing exactly on `pruneBefore` — an exclusive bound where the startup check uses an exclusive `Before`. Almost certainly irrelevant, just noting the asymmetry.

### 🎉 Things that could have been breaking and aren't

Checked specifically because they looked risky:

- **`ys.Close()` inside the drain is safe.** `Close()` sets `gobRunning = false`, and `insertMockGob` relaunches `gobWriterLoop` when `!gobRunning` — the code notes it deliberately avoids `sync.Once` so it can restart. Recording continues normally after a drain, which matters since the recorder shares the process on the in-cluster path.
- **No lock inversion.** The drain holds the striped file lock, then takes `gobLifecycleMu` via `Close()` — the same order the pre-existing `updateMocksGob` already uses, and `insertMockGob` only ever takes `gobLifecycleMu`, never the file lock first. No new deadlock.
- **Reusing the `UpdateMocks` skeleton verbatim** (same striped lock, format detect, streaming decode, atomic temp+rename) is the right call — it means race-safety and atomicity are inherited rather than reimplemented, and `UpdateMocks` genuinely is untouched.
- **The `ctx.Err()` check every 1024 mocks** in `drainMocksGob` is a good touch for large pools.

prabaltripathiofficial:
### Changes since the review

Thanks for the review — all findings addressed. Pushed `0efde51`.

- **🔴 gob persisted derived lifetime — fixed.** `keep()` now snapshots + restores `TestModeInfo` around the `DeriveLifetime` call, so the mock is left byte-identical and nothing mode-specific is written back through the gob encoder. The yaml/json path never persisted `TestModeInfo`, so this closes the one affected path.
- **🟡 test skipped in CI — fixed.** New `TestDrainGobDoesNotPersistDerivedLifetime` runs **unconditionally** in CI and directly guards the bug above (asserts the drained gob carries `LifetimeDerived=false`). The strict-mode behavioural test stays gated behind `KEPLOY_STRICT_MOCK_WINDOW=1`.
- **🟡 gofmt — fixed** (`gofmt -w` on `drain_test.go`).
- **🟡 "near-no-op in lax mode" — not applicable here.** Your read of the lifetime-first routing (util.go:3226 / agent.go:785) is correct. In this deployment strict mock-window is always on, so tagged HTTP/DB captures are `LifetimePerTest` and the drain drops them as the live figures show. The drain deliberately shares `DeriveLifetime` with the filter, so it can never drop a mock the filter still promotes to Session — I did **not** decouple it to read `config.Test.StrictMockWindow` for exactly that reason.
- **🟢 boundary asymmetry — acknowledged, left as-is** (sub-second window; changing it adds risk without benefit).

Full detail in the PR description under "Review responses (round 1)". Ready for another look.

prabaltripathiofficial:
### Round 2 — the growth-bound finding

You were right: the drain reduced the slope but didn't bound growth. Fixed and pushed.

**Root cause:** the startup cutoff was recomputed each interval from the test cases still on disk, and executed cases are deleted after every interval — so the cutoff drifted later each cycle and `Before(cutoff)` re-armed a fresh startup window, permanently keeping each interval's first few cases' per-test mocks. Linear, ~3× slower (exactly the single-interval 320KB→108KB), not bounded. The `≤ window` batch was a total no-op for the same reason.

**Fix:** anchor the cutoff to the first real boundary seen for the test set (a candidate earlier than `pruneBefore`), persisted in a `.keploy-startup-cutoff` marker next to the mocks file and reused on later intervals. The marker lives in the test-set dir, so it's discarded when a new recording session starts fresh — matching the real lifecycle (the boot pool only changes on a new release → new session). Steady-state is now `boot pool + current interval`, bounded. Also fixes the `≤ window` no-op once the anchor is set.

**Test:** `TestDrainStartupCutoffAnchoredAcrossIntervals` — two intervals, drifting cutoff; asserts the later interval's per-test mock is dropped against the anchored boundary, not the drifted one (fails without the anchor). This is the multi-interval check you asked for, as a unit test.

Detail in the description under "Review responses (round 2)". Entirely OSS-side; the k8s-proxy PR (which just passes the candidate cutoff) is unchanged.

prabaltripathiofficial:
## Review round 2 — keploy #4393

Re-read at `9c1ab417`. The gob snapshot/restore and the gofmt fix are confirmed. The anchored cutoff (`4b254ac2`) is the right shape and it does close the accumulation problem — steady state is now boot pool + current interval, genuinely bounded rather than linear-with-a-smaller-slope.

Three things on the anchor itself.

### 🟡 1. The mechanism that bounds growth fails silently

```go
if candidate.Before(pruneBefore) {
    _ = os.WriteFile(markerPath, []byte(candidate.UTC().Format(time.RFC3339Nano)), 0644)
}
```

The write error is discarded with no log, not even at Debug. If it ever fails — read-only mount, permissions, or a full disk — the marker never lands, every subsequent interval finds no marker, re-anchors from the drifted candidate, and the linear-growth bug returns with no signal at all.

The awkward part is the coupling: this PR exists because disk fills up, and the guard against disk filling up is a write that quietly no-ops when disk is tight. A `Debug` on write failure (and on a successful anchor read) would make a regression diagnosable from logs instead of from a disk-usage graph weeks later.

### 🟡 2. A stale anchor over-drops a new session's boot mocks

The doc comment says:

> The marker lives in the test-set dir, so it is discarded when a new recording session starts fresh.

That's an assumption about the *caller's* directory lifecycle, and nothing in this package enforces it. If a test-set dir is ever reused across recording sessions, the persisted anchor is far in the past, so none of the new session's mocks are `Before(anchor)`. Config/DNS/session mocks still survive via the lifetime rules — but any **tagged per-test boot mock**, which is precisely what the startup window exists to protect, gets dropped.

That's the one failure mode here that breaks a replay rather than merely wasting disk, and it's silent: the drain logs `kept`/`dropped` counts that look perfectly reasonable.

Two ways out, cheapest first:

- Treat the anchor as stale when no mock in the file is `Before(anchor)`, and re-anchor.
- Stamp the marker with the recording session ID and re-anchor when it differs. k8s-proxy already carries `recordSessionID` in `debugBundle`; it would need threading through the signature.

### 🟡 3. The test for the headline fix doesn't run in CI

`TestDrainStartupCutoffAnchoredAcrossIntervals` opens with:

```go
if !strictModeEnabled() { t.Skip(...) }
```

CI doesn't set `KEPLOY_STRICT_MOCK_WINDOW`, so it never executes. That means the anchor — the change that decides whether this PR delivers "bounded" or "3× slower growth" — has zero automated coverage. Same for `TestDrainToStartupMocks`. The only drain test that actually runs in CI is the gob-lifetime one, which doesn't exercise dropping at all.

This is the third round this has come up, so concretely: the reason strict mode is needed is that under lax, rule #5 promotes HTTP to `LifetimeSession` and nothing is droppable. Build the fixture from a `Kind` **outside** `kindsWithImplicitSessionLifetime` (that set is HTTP, HTTP2, MySQL, Postgres, PostgresV2, GENERIC, DNS) and it derives to `LifetimePerTest` under **both** modes — so the drop assertions run everywhere with no env dependency, and `t.Setenv` (which can't work, since `laxKindFallbackDisabledCache` is snapshotted at package init) stops being the blocker.

### 🎉 Verified on this pass

- **The `candidate.Before(pruneBefore)` guard is the right condition.** A batch at or under the window size makes `StartupMockCutoff` return `keepAll` == `pruneBefore`, so the guard is false and that interval's keep-all no-op can't poison the anchor. That's the subtle case and it's handled.
- **The anchor applies to both storage paths** — it's resolved after the lock and before the gob branch, so yaml and gob get the same value.
- **The marker is safe where it lives.** I checked for anything that would trip over a new file in the test-set root: `uploadMocksToObjectStorage` targets an explicit snapshot path rather than walking the dir, the only `os.ReadDir` on the test-set root is Debug-gated support-bundle logging, and the other `ReadDir` is scoped to `tests/`. Nothing in `mockdb` globs the directory either. A dotfile there is inert.
- **Serialisation claim holds** — `anchoredStartupCutoff` runs inside the striped file lock, and `getMockFileLock` is process-wide, so the read/write really is serialised against a concurrent `InsertMock`.
- **Steady state is now bounded**: keep = lifetime boot pool + a fixed `Before(anchor)` set + the in-flight `After(pruneBefore)` window; each interval drops `[anchor, pruneBefore]`. That's the property the PR claims.

CI is mid-flight after the `main` merge (30 green, the rest queued) — worth a look once it settles, particularly that the Windows leg is no worse than before.

prabaltripathiofficial:
### Round 3 — anchor hardening

Both fixed. Pushed `80ee63e2`.

- **🔴 Stale anchor over-drops a new session's boot mocks — fixed.** You're right that "the marker dies with the dir" was an unenforced assumption, and it's the one failure mode here that breaks a replay rather than wasting disk. `anchoredStartupCutoff` now takes `sessionStart` (the recording's start, threaded from k8s-proxy's `recordStartedAt`) and **discards any marker whose cutoff predates it** — that's a marker from an earlier session that reused the directory. It's then recomputed from this session's batch. A valid anchor is always ≥ sessionStart (it's the 6th test case's timestamp, recorded after boot), so the check is exact; when `sessionStart` is zero it's skipped. New test `TestDrainStaleAnchorFromPriorSessionDiscarded` plants a prior-session marker and asserts the new session's boot mock survives and the marker is refreshed.
- **🟡 Silent `_ = os.WriteFile` — fixed.** It now logs on failure: without the marker the window re-arms next interval and linear growth resumes, and the write fails exactly on the full-disk condition this drain exists to prevent — so it's a `Warn`, not swallowed.

k8s-proxy side is the one-line wiring to pass `recordStartedAt` (keploy/k8s-proxy#695).

prabaltripathiofficial:
### Round 2, finding #3 — CI coverage of the drop (now fixed)

To close out all three of "Review round 2": #1 (silent write) and #2 (stale anchor) were fixed in `80ee63e2` per the note above. #3 — the drop tests skipping in CI — is fixed in `b71c41b7`.

You were right that this had zero automated coverage and that `t.Setenv` can't rescue it (`laxKindFallbackDisabledCache` is snapshotted at package init). Took exactly the route you suggested: the drop fixtures now use **`GRPC_EXPORT`**, a Kind outside `kindsWithImplicitSessionLifetime`, so an untagged/`"mocks"`-tagged grpc mock derives to `LifetimePerTest` under **both** modes. So `TestDrainToStartupMocks`, `TestDrainStartupCutoffAnchoredAcrossIntervals`, and `TestDrainStaleAnchorFromPriorSessionDiscarded` now **run unconditionally** and exercise the drop — verified passing in lax (CI default) and strict. `TestDrainGobDoesNotPersistDerivedLifetime` intentionally stays HTTP + lax-gated, since it specifically exercises HTTP's lax `Session` promotion.

Thanks for pushing on all three — the stale-anchor one especially was a real latent replay-breaker.

prabaltripathiofficial:
## ✅ Approving from my side

All review findings across both rounds are closed, and I re-verified each in the source rather than from the summaries:

- **Gob no longer persists derived lifetime** — snapshot/restore around `DeriveLifetime`, restore lands before every subsequent return.
- **Startup cutoff anchored** — steady state is boot pool + current interval, genuinely bounded. The `candidate.Before(pruneBefore)` guard correctly keeps a ≤window-size keep-all interval from poisoning the anchor.
- **Stale anchor rejected** — `!anchored.Before(sessionStart)`, with `sessionStart.IsZero()` preserving prior behaviour for callers that don't pass one. This was the only failure mode that could break a replay rather than just waste disk.
- **Anchor persist failure now logged** at `Warn` with the consequence spelled out — better than the Debug I asked for.
- **Drop assertions run in CI** via the `GRPC_EXPORT` fixture (outside `kindsWithImplicitSessionLifetime`, so `LifetimePerTest` derives under both modes). `strictModeEnabled` is now down to the one test that genuinely needs lax, and there's a dedicated stale-anchor test.

Signature change to `DrainToStartupMocks` lines up exactly with k8s-proxy #695, including its test fake.

Nothing further from me on the code. Two non-review items before merge: the `CI Gate (Linux)` is red on `go-memory-load-mysql (record_latest_replay_build)` — the only non-Windows failure, and worth one look to confirm it's a flake, since nothing in this PR is reachable from the OSS CLI path. The Windows legs look like the pre-existing infra failures.

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in feat(mockdb): DrainToStartupMocks to bound recordDir on long auto-replay sessions. Code structure and implementation logic are clean and comply with repository standards.

Issue #4399, opened 2026-07-31
[feature]: honor --cmd-type when explicitly provided instead of always overwriting it with auto-detection


blockgroot:
### 👀 Is there an existing feature request for this?

- [x] I have searched the existing issues

### 🔖 Enhancement description

The `--cmd-type` flag (`cli/provider/cmd.go:243`) is defined, kebab-aliased, bound through
viper, and unmarshalled into `c.cfg.CommandType` — but it is then unconditionally
overwritten before it's ever used:

```go
// cli/provider/cmd.go:1024
c.cfg.CommandType = string(utils.FindDockerCmd(c.cfg.Command))
```

So an explicit `--cmd-type docker-compose` is silently discarded in favor of
auto-detection from the `--command` string. Reproduced on `main` @ `aaa0fb46`:

```
$ keploy test --cmd-type docker-compose -c "python app.py"
ERROR   failed to validate flags   {"error": "non docker command not supported for OS: darwin , Arch: arm64"}
```

The flag is completely inert — there's no way today to tell Keploy "this command launches
a container" when the launch is wrapped in a script/Makefile target and the raw command
string doesn't look docker-ish (e.g. `-c "make up"`, `-c "./start.sh"`).

Separately, the flag's help text (`(native/docker/docker-compose)`) doesn't even match the
real `utils.CmdType` values (`native` / `docker-run` / `docker-start` / `docker-compose` —
`utils/utils.go:610-617`); `docker` was never a valid value.

**I would like to see:** `--cmd-type` actually honored when the user explicitly passes it
(validated against the real allowed values, same style as the existing `--storage-format`
validation), with auto-detection kept as the default only when the flag isn't set. Fix the
help text to list the real values.

### 🎤 Why should this be worked on?

Anyone who launches their app through a wrapper script or Makefile target has no way to
tell Keploy the underlying command is a Docker/compose launch — the flag that exists for
exactly this case is a no-op today, and the failure mode (a confusing OS/Arch validation
error) doesn't hint at the real cause. This is a small, targeted fix: make an existing,
documented, already-wired flag actually do what its help text says, without changing any
default behavior for users who never pass it.

### 💻 Repository

keploy

blockgroot:
I'd like to take this one — have a small fix scoped (validate + honor the flag when explicitly set, keep auto-detect as the default otherwise) with tests. Will open a PR shortly.

PR #4403, opened 2026-07-31
feat(cli): honor --cmd-type when explicitly provided

blockgroot:
- `--cmd-type` (`cli/provider/cmd.go:243`) is registered, viper-bound, and documented, but `ValidateFlags` (`cli/provider/cmd.go`, inside `case "record", "test":`) unconditionally overwrote it with auto-detection from the `--command` string right after flag/config unmarshalling — an explicit value was silently discarded every time, whether the user passed it or not. Introduced in the same commit as the flag itself (#1814), never guarded by `Changed()`.
- Fix: extract the assignment into `resolveCommandType(cmd, command, explicit)`. When the user explicitly passes `--cmd-type`, it's validated against the real `utils.CmdType` set and used as-is; otherwise auto-detection runs exactly as before.
- Gated on `cmd.Flags().Changed("cmd-type")`, deliberately **not** on the value being non-empty: `config/default.go`'s `InternalConfig` ships `cmdType: "native"` in every generated `keploy.yml`, so a non-empty check would treat that baked-in default as an explicit user choice and break Docker users who never touched the flag.
- Also fixes the flag's help text, which advertised `docker` as a valid value — it never was; the real values are `native` / `docker-run` / `docker-start` / `docker-compose` (`utils/utils.go`).
- Extracting a standalone function (rather than inlining the check) keeps this unit-testable without invoking `ValidateFlags`'s side effects (stdout logo printing, logger level swaps, JSON-mode redirection).

**Reproduction (before this change):**
```
$ keploy test --cmd-type docker-compose -c "python app.py"
ERROR   failed to validate flags   {"error": "non docker command not supported for OS: darwin , Arch: arm64"}
```
**After this change**, the same command proceeds normally (the explicit `docker-compose` is honored, bypassing the Native/Empty-only OS/Arch gate). An invalid value now fails clearly instead of being silently ignored:
```
$ keploy test --cmd-type docker -c "python app.py"
ERROR   failed to validate flags   {"error": "invalid --cmd-type value \"docker\": allowed values are \"native\", \"docker-run\", \"docker-start\", and \"docker-compose\""}
```

## Links & References

**Closes:** #4399

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [x] 🍕 Feature
- [x] ✅ Test

## Added e2e test pipeline?
- [ ] 👍 yes
- [x] 🙅 no, because they aren't needed — this is a flag-validation/config-resolution fix covered by focused unit tests; no sample-app behavior changes.

- [x] 👍 yes

## Added to documentation?
- [x] 🙅 no documentation needed — the only user-facing text is the flag's own `--help` string, fixed in this PR.

- [x] 👍 yes, mentioned below

```
$ go test -v -run TestResolveCommandType ./cli/provider/
=== RUN   TestResolveCommandType_ExplicitValid
--- PASS: TestResolveCommandType_ExplicitValid (0.00s)
=== RUN   TestResolveCommandType_ExplicitInvalid
--- PASS: TestResolveCommandType_ExplicitInvalid (0.00s)
=== RUN   TestResolveCommandType_NotProvided_FallsBackToAutoDetect
--- PASS: TestResolveCommandType_NotProvided_FallsBackToAutoDetect (0.00s)
PASS
ok  	go.keploy.io/server/v3/cli/provider	0.835s
```

Full local gate:
```
$ gofmt -l .            # clean
$ go vet ./...          # clean
$ go test ./cli/...     # ok
$ go test ./...         # ok, 0 failures
```

Also manually verified against the built binary (see reproduction above): the explicit-valid, explicit-invalid, and not-provided cases all behave as intended.

## Self Review done?
- [x] ✅ yes

- NA

## Additional checklist:

### DIFF
--- a/cli/provider/cmd.go
+++ b/cli/provider/cmd.go
@@ -240,7 +240,7 @@ func (c *CmdConfigurator) AddFlags(cmd *cobra.Command) error {
 		cmd.Flags().Uint32("server-port", c.cfg.ServerPort, "Port used by the Keploy Agent server to intercept traffic")
 		cmd.Flags().Uint32("dns-port", c.cfg.DNSPort, "Port used by the Keploy DNS server to intercept the DNS queries")
 		cmd.Flags().StringP("command", "c", c.cfg.Command, "Command to start the user application")
-		cmd.Flags().String("cmd-type", c.cfg.CommandType, "Type of command to start the user application (native/docker/docker-compose)")
+		cmd.Flags().String("cmd-type", c.cfg.CommandType, "Type of command to start the user application (native/docker-run/docker-start/docker-compose)")
 		cmd.Flags().Uint64P("build-delay", "b", c.cfg.BuildDelay, "User provided time to wait docker container build")
 		cmd.Flags().String("container-name", c.cfg.ContainerName, "Name of the application's docker container")
 		cmd.Flags().StringP("network-name", "n", c.cfg.NetworkName, "Name of the application's docker network")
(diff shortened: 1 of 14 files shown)


slayerjain:
Thanks @blockgroot

PR #4407, opened 2026-08-03
fix(replay): gate the --keep-app-alive readiness skip on an observed accept

slayerjain:
## `--keep-app-alive` can skip the readiness wait for an app that was never ready

`waitForAppReady` is skipped whenever `serveTest && !isFirstTestSet`, on the reasoning quoted in the code — the app *"has already passed its readiness check during the first test-set"*.

That reasoning has two independent holes, and I hit both on a live cluster.

**1. "Passed" isn't checked.** The readiness probe is best-effort by contract: when the app fails to accept within `HealthPollTimeout` it logs

```
app ready-probe address did not accept a connection within the ceiling;
firing tests anyway (replay may see connection refused)
```

and **returns true anyway**, deliberately, so replay never regresses versus plain `--delay`. So a first test-set can "complete" its wait against an app that never came up.

**2. The wait can be skipped before any wait has happened.** Measured on a real Kubernetes auto-replay, with `delay: 75` explicitly configured:

```
10:02:42.336  replay pod created
10:02:46.836  --keep-app-alive: skipping waitForAppReady; app already warm
10:02:46.854  starting test for ...
```

First test fired **4.5s** after pod creation — not 75s, not even the 10s default. A skip log lands 18ms before it. The app on that image first accepts at ~30s. Every request was refused; every test came back red, attributed to the application rather than to it still booting.

So the delay knob is not the lever people assume: while the skip arms, no configured delay applies at all.

## Fix

Record whether the probe has ever observed the app accept a connection, and require that observation before taking the skip:

```go
if serveTest && !r.isFirstTestSet && appReadyObserved.Load() {
```

Ordinal position is not evidence of readiness; an observed accept is. This is deliberately the weakest condition that closes both holes — it does not change the probe's warn-don't-fail contract, and it does not depend on which test-set happens to be processed first.

## Unchanged

- the first test-set still waits;
- the skip still arms only when the one-shot spawn actually fired (`serveTest`), so a `cmdType` where the spawn was suppressed still waits every test-set;
- the probe still warns rather than fails on timeout — nothing that previously succeeded now blocks.

Only the previously-wrong case changes: the wait is no longer skipped when the app has never been seen accepting.

## Testing

`TestAppReadyObservedGatesTheWarmSkip` covers the four combinations (never-observed / observed × first / later test-set). `go test -race ./pkg/service/replay/...` green.

Honest limitation: the test asserts the predicate, mirroring the two call sites, so it would not catch someone deleting a call site outright — the same gap the existing `serveTest` / `isFirstTestSet` logic already has.

## Note on what this does not fix

On the cluster where this was measured, a second, unrelated defect records some test cases with an ephemeral `app_port`, so they dial a port nothing listens on. That is a separate issue in a different component; this PR only removes the "tests fire at an app that was never up" failure mode.

### DIFF
--- a/pkg/service/replay/app_ready_observed_test.go
+++ b/pkg/service/replay/app_ready_observed_test.go
@@ -1,0 +1,43 @@
+package replay
+
+import "testing"
+
+// --keep-app-alive skips the readiness wait for every test-set after the first,
+// reasoning that the app "already passed its readiness check during the first
+// test-set". The probe is best-effort — on timeout it warns and fires the tests
+// anyway — so that reasoning holds only when the check actually passed.
+//
+// Measured on a real cluster: replay pod created at T+0, readiness probe engaged
+// at T+4s, a later test-set logged "app already warm" at T+4.4s, and the app
+// first accepted a connection at ~T+30s. The run finished at T+20s having
+// refused every request. Ordinal position is not evidence of readiness; an
+// observed accept is.
+func TestAppReadyObservedGatesTheWarmSkip(t *testing.T) {
+	orig := appReadyObserved.Load()
+	t.Cleanup(func() { appReadyObserved.Store(orig) })
+
+	// The predicate the two call sites use, mirrored here so the intent is
+	// asserted independently of the surrounding replay plumbing.
+	warmSkip := func(serveTest, isFirstTestSet bool) bool {
+		return serveTest && !isFirstTestSet && appReadyObserved.Load()
+	}
+
+	appReadyObserved.Store(false)
+	if warmSkip(true, false) {
+		t.Fatal("must NOT skip the readiness wait before the app has ever been seen accepting a connection: that is exactly the run where every test gets connection refused")
+	}
+	if warmSkip(true, true) {
+		t.Fatal("the first test-set always waits")
+	}
+
+	appReadyObserved.Store(true)
+	if !warmSkip(true, false) {
+		t.Fatal("once the app has been observed ready, later test-sets should skip the wait — that is the whole point of --keep-app-alive")
+	}
+	if warmSkip(false, false) {
+		t.Fatal("without the one-shot spawn (serveTest=false) the wait must run every test-set, matching the historical lifecycle")
+	}
+	if warmSkip(true, true) {
+		t.Fatal("the first test-set waits even when a previous run observed readiness")
+	}
+}
--- a/pkg/service/replay/replay.go
+++ b/pkg/service/replay/replay.go
@@ -1349,7 +1349,7 @@ func (r *Replayer) RunTestSet(ctx context.Context, testSetID string, testRunID s
 		// one-shot spawn was suppressed (cmdType == Empty / instrument
 		// off), serveTest will be false here and the readiness wait
 		// runs every test-set, matching the historical lifecycle.
-		if serveTest && !r.isFirstTestSet {
+		if serveTest && !r.isFirstTestSet && appReadyObserved.Load() {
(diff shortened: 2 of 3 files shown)


slayerjain:
### This now has a measured, live justification — it is the root cause of a 0/8 auto-replay on staging

I opened this from code reading. Since then it turned up as the **measured** root cause of a real failure on `staging/api-server-v2-staging` (self-hosted DaemonSet control plane, 2026-08-03 10:02Z): an auto-replay that scored **8 tests, 8 failed**.

The replay harness fired every test **13 seconds before the app container started**:

```
10:02:46.431  replay.go:609   gating replay on app-readiness TCP probe {"appReadyProbeAddr":"10.244.10.254:8081"}
10:02:46.438  replay.go:1006  no valid test cases found to run for test set {"test-set":"…-8w2k4"}
10:02:46.836  replay.go:1354  --keep-app-alive: skipping waitForAppReady on post-first test-set; app already warm
10:02:46.854  pkg/util.go:711 starting test for {"test case":"get-k8s-proxy-config-keploy-ds-pgdb-1"}
10:02:47.768  utils.go:314    failed to send testcase request to app
                              {"error":"… dial tcp 10.244.10.254:8081: connect: connection refused"}
```

against the replay pod's own status:

```
Initialized: True @ 10:02:59Z
containerStatuses: api-server  startedAt: 10:02:59Z  ready: false  restartCount: 0
first app log line:            10:02:59.823Z
```

All 8 tests ran `10:02:46.854 → 10:02:54.155` — the last finished **4.8s before the app process existed**. Every case came back `status_code: 0`, `dep_result: []`, `APP_CONNECTION_ERROR`, and keploy's own classifier stamped the run `appUnreachable: true`, `unreplayableTests: 8`.

**Why the gate was disarmed is exactly what this PR fixes.** `r.isFirstTestSet` proxies *"the app is proven warm"* with *"this is not index 0"*. That session shipped two test sets; index 0 contained **zero** test cases, returned in 7ms without issuing a single request — and still consumed the first-test-set slot. Index 1 therefore skipped both the configured `delay: 75s` floor **and** the 60s `AppReadyProbeAddr` TCP probe the control plane had wired 405ms earlier.

`appReadyObserved` is the correct predicate precisely because it records an *observed accept* rather than inferring warmth from an index. An empty leading test set cannot set it.

**Not staging-specific:** any user running `--keep-app-alive` across multiple test sets skips readiness on every set after the first, whether or not the app was replaced in between. The same broken predicate also gates the agent-readiness wait at `replay.go:1193`, which this PR fixes too.

**Contrast that isolates it:** the sidecar control plane records the same class of requests against the same app and scores 8/1 failed, 11/0, 3/0 — it never trips this because it ships exactly one test set per session. Every DS session, which opens one test set per recorded pod across a 3-replica deployment, scored 100% failed.

### Known remaining failures on this PR

`build-windows-amd64` fails on `pkg/agent/hooks/windows` — I verified this reproduces on `main` and is not introduced here. `run_golang_wsl / http-pokeapi` and the two `CI Gate` checks are downstream of it.

### What this does not fix

Two further defects were measured in the same session and are tracked separately: DS test cases recorded with an **ephemeral `app_port`** (42013/33035 instead of 8081 — control-plane mitigation is in k8s-proxy#715, agent-side fix outstanding), and a **mock-mapping correlation collapse** (`Failed to correlate mock mapping`, 1 of 8 tests mapped) that is present on the sidecar path too and may surface once this readiness fix lands.

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(replay): gate the --keep-app-alive readiness skip on an observed accept. Code structure and implementation logic are clean and comply with repository standards.

PR #3256, opened 2025-11-11
fix: detect app stopped in case of docker compose

anjupathak03:
## Describe the changes that are made
This pull request introduces enhancements to how the application manages Docker Compose commands for internal use, focusing on improved process exit handling and more accurate service identification. The main changes ensure that the Docker Compose process exits when the main application container stops, and that the correct service name is tracked for exit codes. Additionally, the codebase is updated to propagate and utilize the service name found in the compose file.

**Docker Compose command improvements:**

* The `modifyDockerComposeCommand` function now injects the `--abort-on-container-exit` and `--exit-code-from <service>` flags into the Docker Compose command, ensuring the compose process exits when the application container stops and returns the correct exit code. This logic is encapsulated in the new `ensureComposeExitOnAppFailure` helper. (`pkg/client/app/util.go`) [[1]](diffhunk://#diff-7e0da772985df86661e10e74a467ca62ca4a077f64d5a636571fda4b67f26407L42-R48) [[2]](diffhunk://#diff-7e0da772985df86661e10e74a467ca62ca4a077f64d5a636571fda4b67f26407L71-R90) [[3]](diffhunk://#diff-7e0da772985df86661e10e74a467ca62ca4a077f64d5a636571fda4b67f26407R117-R153)

**Service name propagation and usage:**

* The service name found in the Docker Compose file is now extracted and returned by `findContainerInServices`, and included in the `ComposeServiceInfo` struct as `AppServiceName`. This allows the correct service name to be used in the compose command for exit code monitoring. (`pkg/platform/docker/docker.go`, `pkg/platform/docker/service.go`) [[1]](diffhunk://#diff-8db4dcca2af4ae7076e621e24f38ad74923799d525a44e99a444dfa902a33e12L237-R244) [[2]](diffhunk://#diff-8db4dcca2af4ae7076e621e24f38ad74923799d525a44e99a444dfa902a33e12L253-R256) [[3]](diffhunk://#diff-8db4dcca2af4ae7076e621e24f38ad74923799d525a44e99a444dfa902a33e12L282-R287) [[4]](diffhunk://#diff-ecbe4365a4572069a1352838d537d5698b14ccb68c241af564979703754ce5c2R34)
* The `SetupCompose` method in `pkg/client/app/app.go` is updated to pass the service name to `modifyDockerComposeCommand`, ensuring the generated command uses the correct flags and service.

These changes collectively improve reliability when running Docker Compose for internal application orchestration, making sure failures in the main application container are properly detected and propagated.

[...]

PR #4408, opened 2026-08-03
fix(replay): abort on container failure, not exit, when a one-shot init is present

prabaltripathiofficial:
## Problem

Cloud replay renders a recorded pod's **init containers** as one-shot compose services (`restart: "no"`, the app `depends_on` them with `condition: service_completed_successfully`). Keploy runs the compose with:

```
docker compose -f - up --abort-on-container-exit --exit-code-from <app>
```

`--abort-on-container-exit` stops the whole stack the instant **any** container exits — including a one-shot init exiting **0**, which it *must* do for the app to start. Result: the app is torn down before it ever runs.

```
prepare-mongo-tls  exited with code 0        <-- init did its job (as designed)
Compose  Aborting on container exit...        <-- flag fires on that success
api-server  Stopping / Stopped                <-- app killed before it runs
→ "user application failed under --keep-app-alive: exit status 1"  (no report)
```

The flag has been correct for its intent (catch the **app** crashing) since #3256, but it's a blunt "any container" instrument. It became fatal once init containers started being rendered as one-shot compose services, because a one-shot init's success is indistinguishable from an app crash under `--abort-on-container-exit`.

## Fix

When the in-memory compose contains a one-shot init (detected by `service_completed_successfully` in the compose content), inject **`--abort-on-container-failure`** instead of `--abort-on-container-exit --exit-code-from <app>`.

- `--abort-on-container-failure` aborts only on a **non-zero** exit → the init's clean exit-0 is ignored, while a real app crash is still caught.
- `--exit-code-from` is dropped because it *implies* `--abort-on-container-exit`. There is **no** container-state polling on the compose path (`waitTillExit()` is skipped for `DockerCompose`), so after this change `--abort-on-container-failure` is the **only** signal that catches an app crash: it makes `docker compose up` itself exit non-zero, surfaced through `cmdErr` from `utils.ExecuteCommand`. Note that compose now returns its **own** exit status rather than the app's (via the old `--exit-code-from <app>`) — equivalent for pass/fail, but the specific code changes meaning.
- Only the **in-memory** compose path (`setupComposeInMemory` → `ensureInMemoryComposeFlags`) is switched. The user-provided-compose-file path (`modifyDockerComposeCommand`) passes `false`, so its behavior is unchanged.

~18 lines across `pkg/client/app/{app.go,util.go}`.

## Why this and not the alternatives

- **Two-phase (pre-run the init, then `up`)** was tested and *fails*: even with `--no-recreate`, compose re-runs the one-shot init as a dependency, it exits 0 again, and `--abort-on-container-exit` aborts anyway.
- **Detached + monitor** would work but keploy deliberately forbids detached mode (`isDetachMode`), so it's a larger lifecycle refactor.
- `--abort-on-container-failure` keeps the flag's original purpose (catch app crashes) while removing the only false positive (a one-shot init's success), with zero change to non-init flows.

## Verification

Reproduced against a real deployment whose recorded manifest carries a `prepare-mongo-tls` init, running the exact cloud-replay command with the stock vs. patched binary:

| | Stock | Patched |
|---|---|---|
| Injected cmd | `up --abort-on-container-exit --exit-code-from …` | `up --abort-on-container-failure` |
| `Aborting on container exit` | fires on init exit-0 | never |
| App | torn down, never ran | agent healthy, app booted, ran |
| Replay | catastrophic, no report | ran to a full report |

### DIFF
--- a/pkg/client/app/app.go
+++ b/pkg/client/app/app.go
@@ -278,13 +278,49 @@ func (a *App) setupComposeInMemory(extraArgs []string) error {
 	a.composeContent = content
 
 	// Ensure the command uses stdin ("-f -") and has the exit-code-from flags.
-	preferFailureAbort := strings.Contains(string(a.composeContent), "service_completed_successfully")
+	preferFailureAbort := composeHasCompletionDependency(&compose)
 	a.cmd = ensureInMemoryComposeFlags(a.cmd, a.composeService, preferFailureAbort)
 
 	a.logger.Info("Running application using in-memory Keploy-generated Docker Compose (no file written to disk)",
 		zap.String("cmd", a.cmd))
 
 	return nil
+}
+
+func composeHasCompletionDependency(compose *docker.Compose) bool {
+	if compose == nil || compose.Services.Content == nil {
+		return false
+	}
+	for i := 0; i+1 < len(compose.Services.Content); i += 2 {
+		service := compose.Services.Content[i+1]
+		if service == nil {
+			continue
+		}
+		for j := 0; j+1 < len(service.Content); j += 2 {
+			if service.Content[j].Value == "depends_on" && dependsOnRequiresCompletion(service.Content[j+1]) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func dependsOnRequiresCompletion(dependsOn *yaml.Node) bool {
+	if dependsOn == nil || dependsOn.Kind != yaml.MappingNode {
+		return false
+	}
+	for i := 0; i+1 < len(dependsOn.Content); i += 2 {
+		dep := dependsOn.Content[i+1]
+		if dep == nil || dep.Kind != yaml.MappingNode {
+			continue
+		}
+		for j := 0; j+1 < len(dep.Content); j += 2 {
+			if dep.Content[j].Value == "condition" && dep.Content[j+1].Value == "service_completed_successfully" {
+				return true
+			}
+		}
+	}
+	return false
 }
 
 func (a *App) SetAppCommand(appCommand string) {
--- a/pkg/client/app/app.go
+++ b/pkg/client/app/app.go
@@ -278,7 +278,8 @@ func (a *App) setupComposeInMemory(extraArgs []string) error {
 	a.composeContent = content
 
 	// Ensure the command uses stdin ("-f -") and has the exit-code-from flags.
-	a.cmd = ensureInMemoryComposeFlags(a.cmd, a.composeService)
+	preferFailureAbort := strings.Contains(string(a.composeContent), "service_completed_successfully")
+	a.cmd = ensureInMemoryComposeFlags(a.cmd, a.composeService, preferFailureAbort)
 
 	a.logger.Info("Running application using in-memory Keploy-generated Docker Compose (no file written to disk)",
 		zap.String("cmd", a.cmd))
--- a/pkg/client/app/util.go
+++ b/pkg/client/app/util.go
@@ -26,7 +26,7 @@ var composeSubcommands = map[string]bool{
 // It tokenizes the arguments, strips all -f/--file flags (including dangling
 // ones without a value), and injects a single "-f -" before the first compose
 // subcommand to avoid producing multiple stdin readers.
-func ensureInMemoryComposeFlags(appCmd, serviceName string) string {
+func ensureInMemoryComposeFlags(appCmd, serviceName string, preferFailureAbort bool) string {
 	parts := strings.Fields(appCmd)
 
 	// Strip every existing -f/--file flag and its value, including dangling
@@ -62,7 +62,7 @@ func ensureInMemoryComposeFlags(appCmd, serviceName string) string {
 		result = append(result, "-f", "-")
 	}
 
-	return ensureComposeExitOnAppFailure(strings.Join(result, " "), serviceName)
+	return ensureComposeExitOnAppFailure(strings.Join(result, " "), serviceName, preferFailureAbort)
 }
 
 func findComposeFile(cmd string) []string {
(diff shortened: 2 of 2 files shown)


prabaltripathiofficial:
## Review — keploy #4408

The diagnosis is right and the fix is the correct one. `--abort-on-container-exit` genuinely cannot distinguish a one-shot init's mandatory exit-0 from an app crash, and `--abort-on-container-failure` is the flag that draws exactly that line. Scoping it to the in-memory path — with `modifyDockerComposeCommand` explicitly passing `false` at all four call sites — keeps the blast radius to the generated-compose case.

Three things, one of which is about the description rather than the code.

### 🟡 1. The stated reason for dropping `--exit-code-from` doesn't hold

> *`--exit-code-from` is dropped because it implies `--abort-on-container-exit`; the app's exit is already observed via the existing container-state polling.*

There is no container-state polling on this path. `waitTillExit()` is skipped for compose, explicitly:

```go
// Skip on the compose path: cmdCancel's ComposeDown already brought the
// whole stack down and waitContainersRemoved confirmed the app/agent
// containers are gone ...
if utils.IsDockerCmd(a.kind) && a.kind != utils.DockerCompose {
	a.waitTillExit()
}
```

and its own doc comment says *"Only reached on the non-compose docker path."*

What actually catches an app crash after this change is `--abort-on-container-failure` making `docker compose up` itself exit non-zero, surfaced through `cmdErr` from `utils.ExecuteCommand`. So the behaviour is fine — but the flag is now the **only** thing detecting an app crash on the compose path, not a redundant second signal. Someone reading the description later could remove it believing polling covers the case, and crash detection would silently disappear with no test failing.

Worth correcting in the description and, better, in a comment at the injection site.

Related and worth a conscious decision: `--exit-code-from <app>` made compose return *the app's* exit code, and now compose returns its own. Equivalent for pass/fail, but if anything downstream keys on the specific code (the original symptom in your repro was `exit status 1`), that value changes meaning.

### 🟡 2. Detecting one-shot inits by substring search on your own generated output

```go
preferFailureAbort := strings.Contains(string(a.composeContent), "service_completed_successfully")
```

`setupComposeInMemory` generates that content, so it already knows whether it emitted one-shot inits. Recovering that fact by grepping its own output is fragile in both directions:

- **False positive** — the compose is rendered from a recorded pod spec, so env var values, commands and service names are user data. Any of them containing that substring flips the flag. Low impact (failure-abort is the safer mode) but unpredictable.
- **False negative** — the one that matters. If the generator ever renders a one-shot service with `condition: service_healthy`, or `restart: "no"` without a `depends_on` condition, this returns false and the original bug is back, silently, with the same "app failed, no report" symptom.

Having the generator return a bool alongside the content couples the decision to the structural fact instead of to a string it happens to emit today.

### 🟡 3. `--abort-on-container-failure` is Compose v2-only

The flag came out of [docker/compose#10225](https://github.com/docker/compose/issues/10225) and is a v2 feature. `docker-compose` v1 has no such option, and [podman-compose doesn't implement it either](https://github.com/containers/podman-compose/issues/1164). Keploy still handles the `docker-compose` command form elsewhere in this same file.

So for a user on v1 (or podman-compose) whose recorded manifest carries an init container, this injects an unrecognised flag and the run dies immediately with an error pointing nowhere near init containers — a worse failure than the one being fixed, for a subset of users. A capability check, a documented minimum compose version, or falling back to the old flags when v2 isn't detected would close it. I couldn't pin the exact v2 release that added it, so that's worth confirming against whatever minimum you support.

### 🟢 Smaller

- The early-return guard now matches three flag strings via `strings.Contains` over the whole command, so a service name or path containing any of them suppresses injection entirely. Pre-existing pattern — just noting the collision surface grew.
- `preferFailureAbort` is a positional bool threaded through two functions with four `false` call sites. A named type would make those call sites self-describing, though it's a small file and the current form is readable.

### 🎉 What holds up

- **The alternatives section is evidence, not theory.** "Two-phase was tested and *fails* — even with `--no-recreate`, compose re-runs the one-shot init as a dependency, it exits 0 again, and the flag aborts anyway" is the kind of note that stops the next person spending a day on the obvious-looking approach.
- **The root-cause framing is fair to the original code** — the flag was correct for its intent since #3256 and only became fatal once init containers started being rendered as one-shot services. That's a latent incompatibility, described as one.
- **Verified against a real deployment** with a real `prepare-mongo-tls` init, stock vs patched, with the injected command shown for both — so the before/after is checkable rather than asserted.
- **The user-provided-compose path is left alone deliberately**, and the early-return means a user who passes their own abort flag keeps control of it.

prabaltripathiofficial:
Thanks for the thorough self-review. Addressed in `13b598b6`:

**🟡 2 — substring detection is fragile.** Fixed. The decision now reads the **structure** of the already-parsed compose instead of grepping the marshaled YAML. `setupComposeInMemory` unmarshals into a `docker.Compose`; a new `composeHasCompletionDependency(&compose)` walks each service's `depends_on` and returns true only when a dependency's `condition` scalar equals `service_completed_successfully`. This is the exact marker that makes `--abort-on-container-exit` fatal, so it's the precise trigger — and it can no longer be flipped by that literal appearing in an env value, a command, or a service name (the false-positive half). Verified with a quick unit check: long-form condition → true; the literal hidden only in an env value → false; short-form `depends_on` list → false. The false-negative half (generator emitting `service_healthy` or bare `restart: "no"` with no `depends_on` condition) is a structural choice on the api-server generator side and isn't observable here either way; if that ever changes, the real fix is the generator returning the fact explicitly.

**🟡 1 — the `--exit-code-from` rationale.** You're right, there is no container-state polling on this path (`waitTillExit()` is skipped for `DockerCompose`). Corrected the PR description: `--abort-on-container-failure` is now called out as the **only** app-crash signal (via `cmdErr` from `utils.ExecuteCommand`), and the exit-code-semantics change (compose returns its own status, not the app's) is noted. I did **not** add a source comment at the injection site — this repo keeps that path comment-free by convention.

**🟡 3 — `--abort-on-container-failure` is Compose v2-only.** Real portability gap, deliberately **not** changed in this commit. The in-memory path's command comes from `r.config.Command`, which for cloud replay is the v2 `docker compose` form; a v1/podman fallback would be an untested behavior change to a proven fix, and — as you note — reintroduces the original bug for v1+init users rather than truly closing it. Leaving it as a documented follow-up: gate the flag on a detected v2 form (or a documented minimum compose version) and fall back to the old flags otherwise. Worth pinning the exact v2 release that added the flag before doing so.

**🟢 nits.** The early-return still matches flag strings via `strings.Contains` over the whole command (pre-existing collision surface — unchanged). `preferFailureAbort` stays a positional bool with four explicit `false` call sites; kept as-is to hold the diff to the two files.

prabaltripathiofficial:
## ✅ Approving — keploy #4408

Both substantive findings are resolved, and the third one I raised doesn't hold up on closer inspection.

**Detection is now structural (`13b598b6`).** `composeHasCompletionDependency` walks the parsed `docker.Compose` node tree — services → `depends_on` → each dep's `condition` — instead of grepping the marshaled output. That removes both failure directions I was worried about: a user-supplied env var or command containing the string can no longer flip the flag, and the decision is now tied to the actual dependency condition rather than to a substring the generator happens to emit. The key/value pair-stepping over `yaml.Node.Content` is the right way to read a mapping node, and the nil guards on `compose`, `service` and `dep` mean a malformed or scalar service value falls through to `false` rather than panicking.

**The justification is corrected, and more completely than I asked.** The description now says outright that there is no container-state polling on the compose path, that `waitTillExit()` is skipped for `DockerCompose`, and that `--abort-on-container-failure` is consequently the *only* signal catching an app crash — plus the note that compose now returns its own exit status rather than the app's, "equivalent for pass/fail, but the specific code changes meaning." That last part is the bit a future reader needs and I hadn't asked for it explicitly.

### Withdrawing the Compose-v2 finding

I flagged that `--abort-on-container-failure` is v2-only and would break `docker-compose` v1 or podman-compose users. That concern is effectively moot, and I should have spotted why when I raised it: the flag is injected **only when the compose contains `condition: service_completed_successfully`**, and that condition is itself a Compose v2 feature — [not accepted by v1](https://github.com/docker/compose/issues/11238), and [not implemented by podman-compose](https://github.com/containers/podman-compose/issues/575). So any environment that would reject the flag already rejects the compose file, before the flag matters. The two capabilities travel together on this code path.

The only residual is narrower than what I described: if `--abort-on-container-failure` landed in a *later* v2 release than `service_completed_successfully` support, there'd be a band of v2 versions that accept the condition and reject the flag. I couldn't pin either version reliably, so I'm not going to assert one — worth a glance if you support an old v2 floor, but it isn't the v1-vs-v2 break I originally claimed.

### 🟢 Two doc-drift nits, non-blocking

- The description's Fix section still says the init is *"detected by `service_completed_successfully` in the compose content"* — accurate before `13b598b6`, stale now that it's a parsed-node walk.
- `ensureComposeExitOnAppFailure` gained a third parameter but its doc block wasn't updated: the summary still says it injects "`--abort-on-container-exit` and `--exit-code-from` flags" with no mention of the failure-abort mode, and the explicit `Parameters:` list documents only `appCmd` and `serviceName`. Since that function has a formal parameter block, an undocumented parameter stands out more than it would elsewhere.

Neither affects behaviour. Nothing else outstanding from me.

charankamarapu (approved):
LGTM

PR #4410, opened 2026-08-03
fix(agent): exit non-zero when the agent fails to start, and name the cause

slayerjain:

**The agent reported a total startup failure to its supervisor as success.**

Every fatal branch of the `agent` command logged the error and then `return nil`, and `main.go` discarded the `Execute()` error unless it began with `"unknown command"`, so `utils.ErrCode` stayed `0`. An agent that could not load its eBPF hooks exited **0**, and kubelet recorded it as `Reason: Completed, Exit Code: 0` — docker and CI read it the same way.

Fixing it takes three layers, because the middle one was erasing the evidence:

1. **`cli/agent.go`** — fatal branches return their error instead of `nil`. A cancelled context still exits 0 (that is how the agent is asked to stop), and the check consults the **context** as well as the error, because a stop that races startup surfaces as whatever error the interrupted step happened to produce.
2. **`main.go`** — a non-nil `Execute()` error now sets a non-zero exit status. Without this, (1) alone still exits 0. The `os.Exit(1)` in the unknown-command branch is removed: it is now redundant and was skipping every deferred cleanup in `start()` (debug-file flush, profile writers, log close, umask restore).
3. **`pkg/service/agent/agent.go`** — three `return hookErr` sites replaced the underlying cause with a bare sentinel, so the CLI could not distinguish a shutdown from a real failure and the reason never reached the operator. They now wrap, keeping `errors.Is(err, hookErr)` true.

`utils.ErrCode` becomes an `atomic.Int32` behind `SetErrCode`/`ErrCode()` so `utils.Recover` can mark a recovered panic as a failure — `Recover` already cancels the global context, so a panic aborted the run and then still reported success. A plain `int` written from worker goroutines would have been a data race.

**Plus: the eBPF attach failure that prompted this is now self-diagnosing.** `perf_event_open` is gated on `kernel.perf_event_paranoid`, and Debian/Ubuntu kernels carry an out-of-tree **level 3** that requires `CAP_SYS_ADMIN` — **`CAP_PERFMON` does not satisfy it**. On such a node an agent with the usual eBPF capability set fails every attach with a bare `permission denied`, however new the kernel is. `PerfEventPermissionHint` names the sysctl, its value and the capability. It deliberately stays silent for two look-alikes that are `errors.Is(err, os.ErrPermission)` but have entirely different causes:

- **eBPF verifier rejections** — the kernel answers `BPF_PROG_LOAD` with `EACCES` and `VerifierError.Unwrap()` exposes it, so an unguarded hint would tell operators to grant `CAP_SYS_ADMIN` for a program bug;
- **tracefs / target-binary permission errors** (`*os.PathError`) — `link.Tracepoint` reads a tracefs event id, and uprobes open the target binary, before the syscall.

⚠️ **Behaviour change worth calling out:** invocations that fail validation (e.g. `keploy record` with no `-c`) now exit **1** instead of **0**. That is correct, but it is user-visible for anyone whose CI tolerated a misconfigured invocation.

## Links & References

**Closes:** NA

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?

Unit tests cover each behaviour, and each was mutation-checked (revert the guard, confirm the suite goes red): `cli/agent_test.go` (error propagation, graceful-cancel, cancelled-context-with-unrelated-error) and `pkg/agent/hooks/linux/perm_hint_test.go` (paranoid>2, permissive level, unreadable sysctl, verifier rejection, tracefs vs non-tracefs path, the exact cilium/ebpf error shape).

- [x] 👍 yes

## Added to documentation?

- [x] 👍 yes, mentioned below

Verified live on a kind cluster whose node has `kernel.perf_event_paranoid=3`, running the agent with the standard eBPF capability set (`BPF, PERFMON, NET_ADMIN, SYS_RESOURCE, SYS_PTRACE`, no `SYS_ADMIN`) — released build vs this branch, same node, same pod spec:

| agent | terminal state | what the operator sees |
|---|---|---|
| released | `Succeeded / exitCode=0 / Completed` | only `opening tracepoint perf event: permission denied` |
| **this PR** | **`Failed / exitCode=1 / Error`** | the sysctl, its value and the remedy |

```
ERROR  failed to attach the tracepoint hook on sys_socket
  {"hint": "kernel.perf_event_paranoid is 3 on this node, which denies perf_event_open
   to any process without CAP_SYS_ADMIN (CAP_PERFMON is NOT sufficient at this level)
   — grant the keploy agent CAP_SYS_ADMIN, or set kernel.perf_event_paranoid<=2 on the node"}
```

The cause also now propagates all the way up, which it previously did not:

```
ERROR  failed to hook into the app
  {"error": "failed to hook into the app: opening tracepoint perf event: permission denied:
   kernel.perf_event_paranoid is 3 on this node, ..."}
```

The RCA was established by a single-variable A/B on that node: paranoid=3 without `SYS_ADMIN` fails; **with** `SYS_ADMIN` passes; seccomp `Unconfined` without `SYS_ADMIN` still fails (so seccomp is not involved); and paranoid=**2** without `SYS_ADMIN` passes.

## Self Review done?
- [x] ✅ yes

- Included above.

### DIFF
--- a/cli/agent.go
+++ b/cli/agent.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"context"
+	"errors"
 
 	"github.com/go-chi/chi/v5"
 	"github.com/spf13/cobra"
@@ -21,6 +22,12 @@ func Agent(ctx context.Context, logger *zap.Logger, conf *config.Config, service
 		Use:   "agent",
 		Short: "starts keploy agent for hooking and starting proxy",
 		// Hidden: true,
+		// Runtime failures below are reported by their own log line and by the
+		// process exit status; without this, cobra would dump the whole usage
+		// text after every one of them and bury the actual error. Bad flags
+		// still print their own message via SetFlagErrorFunc (the error, not
+		// the usage listing).
+		SilenceUsage: true,
 		PreRunE: func(cmd *cobra.Command, _ []string) error {
 			return cmdConfigurator.Validate(ctx, cmd)
 		},
@@ -28,14 +35,15 @@ func Agent(ctx context.Context, logger *zap.Logger, conf *config.Config, service
 			svc, err := serviceFactory.GetService(ctx, cmd.Name())
 			if err != nil {
 				utils.LogError(logger, err, "failed to get service")
-				return nil
+				return err
 			}
 
 			var a agent.Service
 			var ok bool
 			if a, ok = svc.(agent.Service); !ok {
-				utils.LogError(logger, nil, "service doesn't satisfy agent service interface")
-				return nil
+				err = errors.New("service doesn't satisfy agent service interface")
+				utils.LogError(logger, err, "failed to start agent")
+				return err
 			}
 
 			// Self-terminate (gracefully) if the parent keploy client dies
@@ -85,7 +93,21 @@ func Agent(ctx context.Context, logger *zap.Logger, conf *config.Config, service
 			err = a.Setup(ctx, startAgentCh)
 			if err != nil {
 				utils.LogError(logger, err, "failed to setup agent")
-				return nil
+				// A cancelled context is how the agent is asked to stop
+				// (SIGINT/SIGTERM, parent exit) — that is a normal shutdown,
+				// not a failure, and must keep exiting 0. Anything else means
+				// the agent never came up: it must exit non-zero so kubelet,
+				// docker and CI see a failure instead of "Completed".
+				//
+				// Check the CONTEXT as well as the error: a shutdown that
+				// races agent startup surfaces as whatever error the
+				// interrupted step happened to produce, which need not carry
+				// context.Canceled. We own this context, so its state is the
+				// authoritative answer to "was this a shutdown?".
+				if errors.Is(err, context.Canceled) || ctx.Err() != nil {
+					return nil
+				}
+				return err
 			}
 
 			return nil
--- a/cli/agent_test.go
+++ b/cli/agent_test.go
@@ -1,0 +1,147 @@
+package cli
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/spf13/cobra"
+	"go.keploy.io/server/v3/config"
+	"go.keploy.io/server/v3/pkg/service/agent"
+	"go.uber.org/zap"
+)
+
+// stubAgent satisfies agent.Service by embedding it: only Setup is ever called
+// on this path, so the remaining methods stay nil and would panic loudly if the
+// command ever started using them.
+type stubAgent struct {
+	agent.Service
+	setupErr error
+}
+
+func (s stubAgent) Setup(context.Context, chan int) error { return s.setupErr }
+
+type stubFactory struct {
+	svc interface{}
+	err error
+}
+
+func (f stubFactory) GetService(context.Context, string) (interface{}, error) {
+	return f.svc, f.err
+}
+
+// stubConfigurator registers only the flags Agent's RunE reads back.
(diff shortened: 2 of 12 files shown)


slayerjain:
### Follow-up commit: `fix(config): don't fail when nothing can answer the override prompt`

The first commit turned 8 e2e lanes red — `go-memory-load-{grpc,mongo,mysql}` and `proxy-stress-test`, each × 2 matrix legs. I traced it, and **the PR did not break them; it revealed that they were already broken:**

```
==== Generating Keploy config ====
Config file already exists. Do you want to override it? [y/n]:
  ERROR  failed to ask for confirmation  {"error": "EOF"}
→ Process completed with exit code 1
```

`keploy config --generate` prompts; those four sample dirs ship a committed `keploy.yml`; stdin on a runner is a closed pipe; `AskForConfirmation` returns that EOF **as an error**. Before this PR the error was discarded and keploy exited 0, so `set -Eeuo pipefail` sailed past and the lane ran on the committed yml. That step has been failing on *every* run, invisibly, for as long as the yml has been committed. The exact failing set is those 4 apps × 2 legs — and the `record_latest_replay_build` leg never fails because it uses the previously released binary.

**The fix keys on EOF, not on "is stdin a terminal".** That distinction is load-bearing: `/dev/null` **is** a character device, so the obvious `ModeCharDevice` heuristic reports *interactive* for `docker run` without `-i`, for systemd (`StandardInput=null` is the default) and for cron — precisely the cases this needs to cover — while reporting *not interactive* for a pipe, which would have silently broken `echo y | keploy config --generate`. Keying on EOF gets both right. (I had written the TTY version first; review caught it with measurements.)

EOF now means "nobody answered": keep the existing config, warn with the path, exit 0.

**`--force` is added** because there was otherwise no non-interactive way to regenerate an existing config: `--in-ci` is registered only under `case "record", "test":`, so the `!cfg.InCi` guard on this command can never be false. Without an escape hatch, "keep the file and succeed" would leave a user whose `make setup` runs `config --generate` with a stale config and no scripted way to refresh it — quieter than the bug being fixed, and worse.

Also fixed in the same function: the tools-service type assertion returned a **nil** error on failure, so a wiring bug exited 0 — the exact shape this PR exists to eliminate.

### Two pre-existing issues this surfaced — not fixed here, worth their own tickets

1. **Those lanes have never exercised config generation.** Because `config --generate` was a silent no-op wherever a `keploy.yml` is committed, ~30 lanes calling it have been testing the checked-in YAML rather than keploy's generator. Deleting `keploy.yml` in each script's "Cleaning previous artifacts" step would close the hole — deliberately out of scope here, since it changes what those lanes test.
2. **`keploy config --generate --path DIR` ignores `--path`.** Verified: run from `/tmp/cw3` with `-p /tmp/cw3/target`, it writes `/tmp/cw3/keploy.yml` and leaves the target untouched — `cfg.Path` is empty in `RunE`. Unrelated to this PR, but it makes the new warning print a bare relative filename.

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(agent): exit non-zero when the agent fails to start, and name the cause. Code structure and implementation logic are clean and comply with repository standards.

PR #4411, opened 2026-08-04
fix(agent): skip sys_enter_socket PID auto-registration under container targeting

ayush3160:
## What

When the enterprise webhook sets `KEPLOY_TARGET_CONTAINERS` for a low-latency per-container recording, `target_namespace_pids` is owned by the enterprise exec-barrier and scoped to the **target** container's cgroup. The `sys_enter_socket` tracepoint auto-registers every process in the pod's PID namespace by host TGID — **including sibling containers** — which leaks their traffic past the cgroup targeting.

This extends the existing DaemonSet skip (which handles the identical situation) to also skip auto-registration when `KEPLOY_TARGET_CONTAINERS` is set.

## Why

Required by the enterprise/k8s-proxy "record a specific container in low-latency sidecar mode" feature. Without it, a targeted recording still captures sibling-container ingress even though the cgroup gate excludes it.

## Testing

Verified in kind (kernel 6.12): a two-container pod recorded with `--target-container service-a` no longer registers service-b's PIDs; service-b ingress + egress are fully excluded.

## Companion PRs

- keploy/enterprise#2347 (targeting mechanism + BPF LSM exec barrier)
- keploy/k8s-proxy#720 (webhook plumbing + `/record/containers`)
- keploy/enterprise-ui#1648 (container picker)

### DIFF
--- a/pkg/agent/hooks/linux/hooks.go
+++ b/pkg/agent/hooks/linux/hooks.go
@@ -180,8 +180,14 @@ func (h *Hooks) load(ctx context.Context, opts agent.HookCfg, setupOpts config.A
 	// "test-set-0 / node-daemon traffic" leak. The reconciler-armed TGIDs
 	// already capture the app's ingress (no auto-detection needed), so skip the
 	// tracepoint entirely in DS mode.
-	if os.Getenv("KEPLOY_DAEMONSET_ENABLED") == "true" {
-		h.logger.Info("daemonset mode: skipping sys_enter_socket PID auto-registration (SessionReconciler owns target_namespace_pids)")
+	// Low-latency per-container targeting (KEPLOY_TARGET_CONTAINERS set by the
+	// webhook) is the same situation as DaemonSet: the enterprise exec-barrier
+	// owns target_namespace_pids and scopes it to the TARGET container's cgroup.
+	// sys_enter_socket would auto-register EVERY process in the pod's PID ns
+	// (both target and sibling containers, keyed by host TGID), leaking sibling
+	// traffic past the cgroup targeting — the same pollution the DS skip prevents.
+	if os.Getenv("KEPLOY_DAEMONSET_ENABLED") == "true" || os.Getenv("KEPLOY_TARGET_CONTAINERS") != "" {
+		h.logger.Info("daemonset/targeting mode: skipping sys_enter_socket PID auto-registration (target_namespace_pids is owned/scoped externally)")
 	} else {
 		socket, err := link.Tracepoint("syscalls", "sys_enter_socket", objs.SyscallProbeEntrySocket, nil)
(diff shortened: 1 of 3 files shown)


charankamarapu:
## Principal review — deterministic MySQL-over-TLS capture

The wire-order ktime merge is the right model and the `res>=req` clamp pins the load-drop RCA precisely. Findings:

**1. (should-fix) Apply the timestamp-order clamp in `recordMock`, not just the query path.**
The clamp lives in the COM_QUERY `flushMock` path only, but there are 7 `recordMock` call sites — the **handshake/config** mocks (`record.go`, `conn.go`) can equally produce a crossed `res<req` order on a reused/keepalive connection and get silently dropped by `filterByTimeStamp` / the MockManager at replay-load. Dropping a *config* mock is arguably worse than a query mock (the connection can't be re-established at replay). Move the clamp into `recordMock` so every MySQL mock is guarded at the choke point.

**2. (nit) Redundant copy in the merge `emit`.**
`readRelay` already allocates a fresh per-chunk `[]byte`; `emit` copies it again before the `decodeChan` send. Pass `c.data` through — it's not reused after the send.

**3. (nit) `mergeWait` cost.**
The 2ms bounded wait is a per-*exchange* cost in practice (dest fills right after a request on the loopback), not per-packet — worth a one-line comment so a future reader doesn't read it as per-packet latency.

Unit tests pass; a targeted test for the reorder buffer + a ktime-out-of-order case would be a good follow-up.

charankamarapu:
## Review resolved

All three findings addressed on `feat/low-latency-target-containers`:

**1. (should-fix) Clamp centralized in `recordMock`** — `0cd4fa31`. Moved the `res>=req` timestamp clamp out of the COM_QUERY `flushMock` path and into `recordMock` (record.go), the single choke point every mock passes through. Now the handshake/config mocks (previously droppable by `filterByTimeStamp` / the MockManager on a reused connection) are guarded too. Behavior for the query path is unchanged; config/handshake gain the guard.

**2. (nit) Redundant copy removed** — `0cd4fa31`. `emit` now passes `c.data` straight to `decodeChan`; `readRelay` already hands out a fresh, non-reused per-chunk buffer, so the second copy was dead.

**3. (nit) `mergeWait` documented** — `bcadd8ec`. Added a comment noting the 2ms bound is paid at most once per *exchange* (loopback dest fills within µs of a request), not per packet.

`go test ./pkg/agent/proxy/integrations/mysql/recorder/` green. The clamp move is behavior-preserving for the validated query path and strictly additive for config/handshake, so it can't regress the 21/21-green capture. Suggested follow-up (reorder-buffer + ktime-out-of-order unit test) tracked separately.

charankamarapu (approved):
LGTM

PR #4412, opened 2026-08-05
chore: skip agent control-plane HTTP server in DaemonSet mode

ayush3160:
In Kubernetes **DaemonSet** mode the agent is **push-based**: it captures traffic and pushes it to the k8s-proxy over an outbound `POST /k8s-proxy/ds/stream`, and takes session control over an outbound CRD/ConfigMap watch (or the agent-initiated `dssync` gRPC stream). Every connection originates *from* the agent — the controller never dials back in.

Yet the `agent` command unconditionally binds the control-plane HTTP server (`routes.StartAgentServer`, the `/agent/*` + `/dedup/*` + `/time-freeze/*` routes). That server is the **sidecar/pull** control plane, driven by the local keploy CLI's `AgentClient` — which is **never constructed** in the DaemonSet agent process. So in DaemonSet mode it has **zero consumers**:

- `AgentClient` is built only in `GetCommonServices` for the `record`/`replay`/`runner` commands, not the `agent` command.
- k8s-proxy's static-dedup poll of `/dedup/stats` and health poll of `/agent/health` are explicitly `if !daemonsetenv.Enabled()` (sidecar only); DaemonSet delivers dedup stats by **push** over `/ds/stream`.
- k8s-proxy's push recorder is built with `configureAgent: nil` — "the absence of an agent dial" (`k8s-proxy/pkg/service/record/push_session.go:131`).
- The DaemonSet's only k8s probe is an **exec** liveness probe (`health --check=liveness`); no `httpGet`/`tcpSocket`/readiness/startup probe touches this server.

### This is more than hygiene: the consumer already assumes nothing listens
k8s-proxy already documents and depends on this invariant. `k8s-proxy/pkg/service/record/record.go:871-877`:

> "The static-dedup PULL path fetches `GET podIP/dedup/stats` from each pod's injected keploy sidecar. **In DaemonSet mode there is no per-pod sidecar listening on the agent port, so the poll can never succeed** — it would spin a dead goroutine per pod dialing a refused port. DS agents deliver dedup stats by PUSH instead."

So k8s-proxy already asserts that nothing listens on the agent port in DS mode and gates its poll accordingly. This PR is the **producer catching up to an assumption the consumer already relies on** — not merely dropping dead weight.

### Security angle
Because the DaemonSet runs with `hostNetwork: true` (`k8s-proxy/charts/k8s-proxy/templates/platform/daemonset.yaml:76`), the bind (`:port` = `0.0.0.0`) would sit **unauthenticated on the node's network namespace** — `StartAgentServer` is a bare `http.Server{Handler: router}` with no auth middleware — needlessly exposing mutating endpoints such as `/agent/stop` and `/agent/storemocks`.

### The change
Gate `StartAgentServer` on `KEPLOY_DAEMONSET_ENABLED == "true"` — the same env var the k8s-proxy Helm chart's DaemonSet spec sets and that `pkg/agent/hooks/linux/hooks.go` + k8s-proxy's canonical `daemonsetenv` package key off (all three accept **only** the exact string `"true"`, so every DaemonSet gate stays consistent). This keeps the OSS agent free of any dependency on the enterprise DaemonSet config. In DaemonSet mode the server bind is skipped; `startAgentCh` is still drained so `Agent.Setup`'s unbuffered port handoff never blocks. All other modes (native / docker / sidecar / compose) are unchanged.

## Links & References

**Closes:** 
- NA (small, self-contained hygiene change)

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- keploy/enterprise#2351 — tracking issue (DaemonSet architecture context + optional enterprise-side follow-up)
### 📄 Related Documents
- NA

- [x] 📦 Chore
- [ ] 🍕 Feature
- [ ] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?

The change removes a listener that has no consumers in DaemonSet mode; a focused unit test covers the gating predicate (`isDaemonSetAgent`). Behavior in all non-DaemonSet modes (which existing e2e covers) is unchanged.

- [x] 👍 yes

## Added to documentation?

- [x] 👍 yes, mentioned below

`go test ./cli/ -run TestIsDaemonSetAgent -v` passes. Manually: run `keploy agent` with `KEPLOY_DAEMONSET_ENABLED=true` and confirm the log line `running as a Kubernetes DaemonSet agent; not starting the control-plane HTTP server ...` and that no listener is bound on the agent port; without the env var (or with any other value) the server binds as before.

## Self Review done?
- [x] ✅ yes

- NA


ayush3160:
Tracking issue in the enterprise repo (DaemonSet architecture context + optional enterprise-side follow-up): keploy/enterprise#2351

charankamarapu (commented):
Reviewed at `681152a4`. I went and checked the reasoning against the actual code in both k8s-proxy and enterprise rather than taking the description on trust, and the argument holds — including the part that makes this more than hygiene.

**The consumer side already documents the invariant this PR implements.** `k8s-proxy pkg/service/record/record.go:871-877`:

> "The static-dedup PULL path fetches GET podIP/dedup/stats from each pod's injected keploy sidecar. **In DaemonSet mode there is no per-pod sidecar listening on the agent port, so the poll can never succeed** — it would spin a dead goroutine per pod dialing a refused port. DS agents deliver dedup stats by PUSH instead."

So k8s-proxy already asserts nothing listens on that port in DS mode and gates its poll accordingly; this is the producer catching up to an assumption the consumer relies on. Worth putting that in the description — it's a stronger justification than "dead weight."

Everything else I could verify checks out:

- **No auth on the listener.** `StartAgentServer` is a bare `http.Server{Handler: router}` bound to `:port` (all interfaces) — no middleware, no token. Combined with `hostNetwork: true` (`charts/k8s-proxy/templates/platform/daemonset.yaml:76`) the exposure claim is accurate, not theoretical.
- **No inbound probe.** The DS template has no `httpGet`, `tcpSocket`, `readinessProbe`, or `startupProbe` — only the exec `--check=liveness`, and enterprise's `fastLivenessProbe` short-circuits that before any server wiring, so it can't depend on the listener.
- **`configureAgent: nil`** confirmed at `push_session.go:131`.
- **No enterprise consumer either.** Enterprise registers `/dedup` + `/time-freeze` on this router (`pkg/service/agent/agent.go:79-89`), but DS wires the deduper through `registerDaemonSetHooks` (per the comment at `core_linux.go:243`) and pushes stats; every enterprise `AgentClient` use is in the sandbox/local CLI path. Nothing in-cluster dials it.
- **I specifically checked whether the early `return` drops a side effect**, since `routes.MakeAgentReady` writes `/tmp/agent.ready` and that could plausibly be depended on. It can't be reached in DS mode anyway (it's the `POST /agent/ready` handler, and its consumers are the docker-compose healthcheck at `pkg/platform/docker/docker.go:829` and the local replay `agentReadyCh`); k8s-proxy never calls it, and the DS chart never reads the file. Clean.

`go test ./cli/ -run TestIsDaemonSetAgent` and `go vet ./cli/` green.

---

### One finding: the truthy set diverges from the canonical gate, and `=1` yields an internally inconsistent agent

The comment says the accepted values "mirror the enterprise bring-up check in `cmd/enterprise/main.go`" — accurate, that one does take `"true" || "1"`. But it isn't the authority. `k8s-proxy pkg/platform/daemonsetenv/daemonsetenv.go` describes itself as exactly that:

> "Package daemonsetenv holds **the single source of truth** for the KEPLOY_DAEMONSET_ENABLED env-var gate. … All code paths that gate on DaemonSet mode read this variable so an operator toggling the chart value **flips every gate atomically** at pod startup."

and it is `os.Getenv(EnabledEnv) == "true"` — no `"1"`. OSS's own `pkg/agent/hooks/linux/hooks.go:189` is likewise `== "true"` only. So with `KEPLOY_DAEMONSET_ENABLED=1`:

| gate | verdict at `=1` | consequence |
|---|---|---|
| `isDaemonSetAgent()` (new) | **true** | control-plane server not bound |
| `hooks.go:189` | false | `sys_enter_socket` auto-registration **not** skipped — that comment says this "leak[s] sibling traffic past the cgroup targeting" |
| `daemonsetenv.Enabled()` (canonical) | false | k8s-proxy **starts** `pollStaticDedupStats` → the "dead goroutine per pod dialing a refused port" its own comment warns about |

That's the one direction that actually breaks things. A false negative here is benign (you keep today's behaviour — a bound listener); a false positive skips the bind while the rest of the system still believes it's sidecar mode. The chart sets `"true"`, so this is latent rather than live — but the new test explicitly asserts `"1"` is accepted, which advertises a value only a third of the system honours.

Two ways out, either fine:

1. **Narrow to `== "true"`** — matches `daemonsetenv` and `hooks.go` exactly, and the test simply drops the `"1"` row.
2. **Add one exported OSS predicate and use it at both OSS sites** (this new call site and `hooks.go:189`), so the binary is self-consistent whatever the accepted set is. This fits your stated goal better — it keeps OSS free of the enterprise helper while removing the duplication instead of adding a third variant — and it lets the test keep `"1"` honestly.

I'd lean on (1) for this PR since it's a one-line change and strictly safe, with (2) as the follow-up if you want the shared predicate.

### Nit

`routes.ActiveHooks.New(router, a, logger)` still runs unconditionally before the goroutine, so in DS mode the routes are registered on a router that is never served. Harmless — no listener, no side effects I could find — but if the intent is "no control plane in DS mode," the registration is now dead too. Not worth restructuring for; noting it so the next reader doesn't assume the router is live.

charankamarapu (approved):
Approving at `675b6579`.

The finding is fixed the way I'd have picked: `isDaemonSetAgent()` is now `os.Getenv("KEPLOY_DAEMONSET_ENABLED") == "true"`, so all three gates agree on the truthy set —

- `cli/agent.go:34` → `== "true"`
- `pkg/agent/hooks/linux/hooks.go:189` → `== "true"`
- k8s-proxy `daemonsetenv.Enabled()` → `== "true"`

— which closes the `=1` split-brain where this gate would skip the bind while k8s-proxy still polled the now-unbound port. The comment explains *why* it's strict rather than just being strict, which is the part that stops someone "helpfully" widening it later, and the test row flipped to `want: false` with the same reasoning attached. Verified at this head: `go test ./cli/ -run TestIsDaemonSetAgent` passes (including `one is not accepted`), `go vet ./cli/` clean.

The description now also carries the consumer-side justification, and quoting `record.go:871-877` directly is the right call — the load-bearing argument for this change is that k8s-proxy *already* depends on nothing listening on that port in DaemonSet mode, so this is closing a gap rather than making a judgement call about dead code. Much easier for the next reader to trust than "it looked unused."

For the record, the things I checked and found clean on the first pass still hold: no auth on the listener (bare `http.Server`, all interfaces) against `hostNetwork: true`, so the exposure argument is real; the DS template has no `httpGet`/`tcpSocket`/readiness/startup probe, only the exec liveness check that short-circuits before any server wiring; `configureAgent: nil` at `push_session.go:131`; no enterprise consumer (its `/dedup` + `/time-freeze` routes exist on this router, but DS wires the deduper via `registerDaemonSetHooks` and pushes stats, and every `AgentClient` use is sandbox/local); and the early `return` drops no side effect — `MakeAgentReady`'s `/tmp/agent.ready` is unreachable in DS mode and read only by the compose healthcheck and local replay.

The one nit I left stands as a non-issue: `routes.ActiveHooks.New(router, a, logger)` still registers routes on a router that is never served in DS mode. Harmless, no side effects, not worth restructuring — just don't let a future reader assume that router is live.

PR #4413, opened 2026-08-05
feat(agent): budget memory guard against limit minus fixed overhead

ayush3160:
- The keploy-agent memory guard budgeted memory pressure against the raw working set vs. the whole limit. In low-latency mode the working set includes the fixed eBPF capture ring buffer (kernel memory that pausing recording cannot reclaim), so the guard tripped its pause threshold at startup and dropped captured test cases before any real buffering.
- Added `SetFixedOverheadMB(mb)` / `FixedOverheadBytes()` — the fixed, non-reclaimable overhead the guard should discount. It is set out-of-band (the enterprise low-latency ring-buffer loader passes the **actual configured `record.ringbufSizeMB`**, not a hardcoded value); zero in proxy mode.
- Added `effectiveUsage(current, limit, overhead)` — subtracts the overhead from **both** usage and limit, so pressure is measured on the reclaimable capture buffers, budgeted against the room left above the fixed floor (pause at the existing 0.80 ratio of the remainder, resume at 0.70).
- `GOMEMLIMIT` is realigned to the same remainder, recomputed only when the overhead changes.
- Proxy mode passes no overhead, so its behaviour is unchanged.

## Links & References

**Closes:** NA

### 🔗 Related PRs
- Enterprise wiring (calls `SetFixedOverheadMB` with `record.ringbufSizeMB`): keploy/enterprise (linked from the tracking issue below)
- Tracking issue: keploy/k8s-proxy (linked below once created)

### 🐞 Related Issues
- Tracking issue in keploy/k8s-proxy (added below)

### 📄 Related Documents
- NA

- [x] 🐞 Bug Fix
- [x] 🍕 Feature

## Added e2e test pipeline?

- [x] 👍 yes

## Added to documentation?

- [x] 👍 yes, mentioned below

Unit tests added in `pkg/agent/memoryguard/memoryguard_test.go`:
- `TestEffectiveUsage` — overhead=0 no-op (proxy), ring-buffer discount, overhead≥limit fallback, negative overhead, effective-current floored at zero.
- `TestEffectiveUsageDrivesPauseBoundary` — pauses at the ratio of the remainder.
- `TestSetFixedOverheadMB` — set/read/reset + overflow guard.

## Self Review done?
- [x] ✅ yes

- NA


charankamarapu (commented):
Reviewed as part of the four-repo change in keploy/k8s-proxy#730. The core modelling call is right: the ring buffer is a fixed kernel-memory floor, a pause cannot reclaim it, so subtracting it from **both** sides and taking the ratio of the remainder is the correct budget. Hysteresis is preserved, and proxy mode is provably unchanged (`overhead == 0` makes `effectiveUsage` an identity, and `0.9 * effLimit` equals the value `Start` already installed).

Three things I'd want fixed before this merges, because they are reachable from the record dialog today.

---

### 1. The new `debug.SetMemoryLimit` call has no floor — and this repo's sibling already documented why that's dangerous

```go
if overhead != lastOverhead {
    debug.SetMemoryLimit(int64(float64(effLimit) * 0.9))
    lastOverhead = overhead
}
```

`effLimit` can legitimately be tiny. Concrete path, all user-settable:

- record dialog, low-latency, `memory_limit: "150"` → webhook injects request 150Mi / limit 300Mi → with k8s-proxy#729, `--memory-limit 300`
- `limitBytes` = 300Mi, `overhead` = 256Mi. `256 < 300`, so the fallback in `effectiveUsage` does **not** trip
- `effLimit` = 44Mi → `GOMEMLIMIT` = **39.6Mi**

The 256Mi ring buffer is kernel memory that `GOMEMLIMIT` does not account for, so the runtime is now told to hold the entire agent heap in 39.6Mi. Per `debug.SetMemoryLimit`'s own docs that "may cause the garbage collector to run nearly continuously". k8s-proxy refused to allow exactly this on the replay path and said so at `pkg/platform/k8s/k8s.go:3908`:

> `minGoMemLimitBytes = 64 << 20` — "That is an unbounded hang with no kubelet-visible signal, which is strictly harder to diagnose than the OOMKill this whole change replaces"

`ParseMemoryMiB` accepts any positive integer with no lower bound, so nothing upstream prevents this. Please floor the target (64Mi is the value the org already settled on) and log when the floor is hit.

Worth noting the two repos agree numerically, which is reassuring: `replayGoMemLimitFraction = 70` (k8s.go:3964) reserves 30% of the hard limit for "the eBPF maps, which the kernel charges to the cgroup while the Go runtime knows nothing about them", and `0.9 * (1024 - 256) / 1024` = 67.5%. Same physics, derived two different ways, in two different repos, with no cross-reference. A comment pointing at each other would stop them drifting.

### 2. The `overhead >= limitBytes` fallback is silent, and it silently reintroduces this PR's bug

```go
if overhead < 0 || overhead >= limitBytes {
    overhead = 0
}
```

Falling back to raw working-set guarding sounds conservative, but trace it through: `currentBytes` already includes the ≥256Mi ring buffer, and `pauseThreshold` is now `0.8 * limitBytes` where `limitBytes <= overhead`. So `effCurrent >= pauseThreshold` on the **first tick**, and it never resumes — the exact startup-pause failure this PR exists to fix, except now the only evidence is `fixed_overhead_bytes: 0` on the pause log, which is indistinguishable from proxy mode.

Please emit a one-time `Warn`/`Error` from `run` (keep `effectiveUsage` pure) naming both numbers: "capture ring buffer (%d MiB) meets or exceeds the agent memory limit (%d MiB); recording cannot proceed". The real fix belongs upstream too — k8s-proxy should reject a low-latency record-start whose `memory_limit * 2` doesn't clear `ringbufSizeMB` plus headroom, rather than letting it degrade at runtime.

Same class, one boundary over: at `overhead == limitBytes - 1` the fallback does *not* trip, `effLimit` = 1 byte, `thresholdBytes(1, 0.80)` returns 0, so `pauseThreshold` is reset to `effLimit` = 1 → permanent pause. A single minimum-viable-budget condition (`limitBytes - overhead < someFloor` → fall back **and** log) covers both this and #1.

### 3. Combined with k8s-proxy#729, the pause line moves from ~40% to ~85% of the OOM ceiling — please make that trade explicit

Defaults, low-latency (`memory_limit: 512`, ringbuf 256Mi):

| | pause at (working set) | hard limit | headroom |
|---|---|---|---|
| before | 410Mi | 1024Mi | 614Mi (60%) |
| after (#729 + this) | 256 + 0.8×768 = **870Mi** | 1024Mi | **154Mi (15%)** |

Both halves of the fix are individually correct, but they compound, and the guard has a 500ms poll and calls `debug.FreeOSMemory()` — which cannot return the live heap. That converts the failure mode from *graceful degradation* (capture paused, app keeps serving) into *cgroup OOM-kill of the agent container* in the app's pod. That may well be the trade you want — 15% of a real ceiling beats 60% of a fake one — but it should be a decision in the code, not an emergent product of two PRs. Either lower `pauseThresholdRatio` now that the ratio applies to the limit rather than the request, or clamp with an absolute floor: `min(ratio*effLimit, effLimit - 128Mi)`. At minimum, put the resulting headroom in the constant's comment.

---

### Non-blocking

4. `SetFixedOverheadMB` silently `return`s on the overflow guard. This setter is the *only* thing that turns the discount on, so a silent no-op is indistinguishable from "working correctly". Take an `int` (callers have one — `uint64(ringbufSizeMB)` in enterprise#2352 only exists to satisfy this signature, and it converts a negative to a huge positive that this guard then swallows), or return an error, and log either way.

5. No test covers the only genuinely new stateful behaviour in `run` — the `overhead != lastOverhead` gate and the `GOMEMLIMIT` realignment. `effectiveUsage` is a pure function and thoroughly tested; the interesting part isn't. Consider moving the overhead onto the `guard` struct (package setter writes through) so `run` becomes testable without touching a global. Relatedly, `TestSetFixedOverheadMB` mutates a package global and is the only test in the file without `t.Parallel()` — correct today because nothing parallel reads the global, but a `run`-level test would break that silently.

6. Doc nit: "this reduces exactly to the previous working-set-vs-limit behaviour" is true for proxy mode but reads as a claim about all modes. Pre-PR the flag carried the memory *request* in cluster mode, so cluster behaviour changes on both axes at once. Worth naming which mode is unchanged.

7. `--memory-limit`'s meaning changed from "request" to "limit" and there is a second consumer: `cli/provider/cmd.go:1569` cross-checks `Agent.MemoryLimit * 1024 * 1024` against `MaxMemoryPerConnection`, which is now 2× more permissive. Benign here (k8s-proxy validates the stricter bound at record-start) but it means the two layers now disagree — flagged on #729 as well.

charankamarapu (commented):
_Posted in error — please disregard this comment. It was not intended for this PR and does not reflect a review request. My earlier review comment on this PR stands._

ayush3160:
Thanks for the thorough review — all addressed in bc9a3f2c.

1. **GOMEMLIMIT floor** — floored at 64Mi (`minGoMemLimitBytes`, matching k8s-proxy's `k8s.go`), with a one-time `Warn` when it binds.
2. **Silent `overhead ≥ limit` fallback** — removed. `overhead ≥ limit` (and any budget below the resume headroom) is now **non-viable**: the guard holds paused and logs an `Error` once, no longer indistinguishable from proxy mode. Also fixes the `effLimit == 1` permanent-pause edge (the min-viable cutoff is the resume headroom).
3. **Compounded headroom trade-off** — decided on the **absolute headroom floor**: pause/resume never leave less than 128Mi/192Mi below the ceiling regardless of budget size (`min(ratio·effLimit, effLimit − floor)`). The ~15% margin at the default is preserved; small budgets can no longer collapse to a dangerously thin absolute margin. Documented in the constants.
4. **`SetFixedOverheadMB`** — now takes an `int`, returns the applied bytes (caller logs it), drops the negative→huge-positive conversion.
5. **Testability** — the pause/GOMEMLIMIT arithmetic is now a pure `evaluateMemory` decision, unit-tested directly (ratio-dominates, absolute-floor, non-viable, GOMEMLIMIT floor). `TestSetFixedOverheadMB` stays non-parallel with cleanup.
6. **Doc** — "reduces to previous behaviour" clarified as proxy-mode-only.
7. **cmd.go:1569** — commented that `--memory-limit` now carries the container limit; k8s-proxy#729 aligns its record-start cross-check to the same 2× bound.

charankamarapu (commented):
Re-reviewed from scratch at `bc9a3f2c`. The rework is a real improvement — the refactor to a pure `evaluateMemory` returning a decision struct makes the arithmetic testable without a cgroup, the one-time log latches are the right call for a 500ms tick, and the two headroom floors crossing over at exactly 640 MiB (128/0.2 = 192/0.3) is a nicely chosen pair. I swept the real `evaluateMemory` rather than reading it, and hysteresis ordering holds at every budget I tried.

**Addressed:** GOMEMLIMIT floor (matched to k8s-proxy's 64 MiB with a cross-reference — thank you for keeping them in lockstep); silent fallback replaced by an explicit `viable` flag plus a loud one-time `Error`; setter now takes `int`, returns the applied bytes, and clears on negative/overflow. On my earlier headroom concern: codifying 128 MiB as the absolute floor while leaving the default k8s case at 85% / 153.6 MiB is a legitimate answer — the number is now explicit in a constant with a rationale instead of emergent, which is what I was asking for.

Three findings, one of which the fix introduced.

---

### 1. Blocking — proxy mode with a small limit now never records at all

The absolute headroom floors are applied unconditionally, including when `overhead == 0`. Swept against the real function:

| `--memory-limit` | before | after |
|---|---|---|
| 64 MiB | pause at 51 MiB | **non-viable — recording held paused forever** |
| 100 MiB | pause at 80 MiB | **non-viable** |
| 128 MiB | pause at 102 MiB | **non-viable** |
| 192 MiB | pause at 154 MiB | **non-viable** |
| 193 MiB | pause at 154 MiB | viable — pause 65 MiB, resume **1.0 MiB** |
| 256 MiB | pause at 205 MiB | viable — pause 128, resume 64 |
| 512 MiB | pause at 410 MiB | viable — pause 384, resume 320 |
| 1024 MiB | pause at 819 MiB | pause 819 (unchanged) |
| 2048 MiB | pause at 1638 MiB | 1638 (unchanged) |

Any docker-mode agent with `--memory-limit <= 192` now never records. Two reachable routes:

- OSS CLI — `keploy record --memory-limit 128`. The flag is a plain `Uint64` in MB with no lower bound (`cli/provider/cmd.go:362`), and `LimitBytes` only rejects 0 and overflow.
- Cluster, **proxy** mode — `memory_limit: "96"` → container limit 192 MiB. k8s-proxy#729's new floor only fires when `LowLatencyMode` is set, so proxy-mode small limits pass validation untouched.

And the message these users get is about a ring buffer they don't have:

> "capture ring buffer meets or exceeds the agent memory limit minus safety headroom; recording cannot proceed — raise --memory-limit or lower record.ringbufSizeMB"

`record.ringbufSizeMB` is an enterprise-only viper key, and in proxy mode `overhead` is 0 — there is no ring buffer in the picture at all.

Proxy mode's `0.8 × limit` was never the bug being fixed, so it shouldn't inherit low-latency's safety margins. Suggest gating the absolute floors and the non-viability rule on `overhead > 0` (or scaling them with the overhead), and branching the message on whether the overhead is non-zero.

### 2. The viability bar admits budgets whose hysteresis band is unusable — and that's exactly where #729 points users

`viable` requires `effLimit > 192 MiB`, which guarantees `resumeThreshold > 0` but not that it's reachable:

| `effLimit` | pause | resume | resume as % of budget |
|---|---|---|---|
| 194 MiB | 66 MiB | **2.0 MiB** | 1.0% |
| 196 MiB | 68 MiB | 4.0 MiB | 2.0% |
| 224 MiB | 96 MiB | 32 MiB | 14% |
| 320 MiB | 192 MiB | 128 MiB | 40% |
| ≥ 640 MiB | ratio | ratio | 70% |

The cross-PR bite: #729's new validator accepts `memory_limit >= 225` for low-latency and its error text tells the user to raise it "to at least 225". At exactly 225 → container limit 450 MiB − 256 MiB ring buffer = **194 MiB** → pause at 66 MiB, resume at **2.0 MiB**. Once the guard pauses it must fall to 2 MiB of reclaimable usage to resume, which in practice it never will — a one-way pause, at the value the other PR recommends.

So the two PRs agree on *viability* but not on *usability*. Suggest raising the bar to something with a workable band — `minPauseHeadroomBytes + minResumeHeadroomBytes` (320 MiB), or simply the 640 MiB crossover where the ratios take over — and having #729 derive its minimum from it rather than maintaining a second copy of the constants.

### 3. The non-viable pause log reports zeros

`evaluateMemory` returns before computing `effCurrent`, so `dec.effCurrent` is the zero value. `run()` then calls `enterPressure(dec.effCurrent, 0, ...)`, which logs:

```
Pausing keploy-agent recording due to memory pressure  {memory_usage_bytes: 0, pause_threshold_bytes: 0, ...}
```

Verified: real usage 900 MiB, limit 400 MiB, overhead 256 MiB → `dec.effCurrent == 0`. The dedicated `Error` carries the useful numbers, but this adjacent line reads like a second, different bug and is what someone greps for when recording stops. Move the `effCurrent` computation above the viability return, or pass `currentBytes`.

---

### Nits

- `nonViableLogged` is reset when the budget becomes viable again; `goMemFlooredLogged` never is. If a later overhead change re-floors GOMEMLIMIT it won't re-log. Fine if deliberate — the asymmetry just isn't explained.
- `goMemTarget` is computed from a possibly-negative `effLimit` (when `overhead > limitBytes`). It lands on the floor so it's safe, but a one-line note saves the next reader the trace.
- `TestEvaluateMemoryRatioDominatesLargeBudget`'s comment — "so proxy mode is provably unchanged" — holds only for `effLimit >= 640 MiB`. Given finding #1 that's the misleading half of the claim; worth rewording and adding a small-limit proxy row to the `TestEvaluateMemory` table, which currently has no case below 200 MiB with zero overhead.

### Gates

`go test ./pkg/agent/memoryguard/` and `go vet` green at `bc9a3f2c`.

ayush3160:
Re-review addressed in cb90f65a. Thanks for catching the regression.

1. **Blocking — proxy mode with a small limit** — fixed. The absolute headroom floors and the non-viability rule now bind **only when `overhead > 0`** (a ring buffer is present). Proxy mode reverts to the plain 0.8/0.7 ratios at any limit — provably the pre-discount behaviour — so a small `--memory-limit` records again, and the ring-buffer error message can no longer fire when there's no ring buffer. Added a small-limit proxy case to the table.
2. **Unusable hysteresis band** — the viability floor is raised from the resume headroom (192Mi) to `minPause + minResume` (320Mi), so a viable low-latency budget also has a usable pause/resume band. Exported **`MinViableLowLatencyLimitBytes(overhead)`** so k8s-proxy derives its record-start floor from the same number (it now uses 320); a `TestMinViableLowLatencyLimitBytes` pins the validator and guard to the same boundary.
3. **Zero-valued pause log** — `effCurrent` is now computed before the non-viable return, so the pause log carries real usage.

Nits: `goMemFlooredLogged` now resets symmetrically (re-floor re-logs); negative-`effLimit` noted; the "proxy provably unchanged" test reworded (now true at *any* limit, not just ≥640).

`go test ./pkg/agent/memoryguard/` + `go vet` green.

charankamarapu (commented):
Re-checked at `cb90f65a`. All three findings resolved, and I verified each against the real function rather than the diff.

**1. Proxy-mode regression — fixed, and provably exact.** The `overhead == 0` early return restores the plain ratios. Swept every limit that was broken:

| `--memory-limit` | pause | expected `0.8×` | resume | expected `0.7×` |
|---|---|---|---|---|
| 64 MiB | 51.2 | 51.2 | 44.8 | 44.8 |
| 100 MiB | 80.0 | 80.0 | 70.0 | 70.0 |
| 128 MiB | 102.4 | 102.4 | 89.6 | 89.6 |
| 192 MiB | 153.6 | 153.6 | 134.4 | 134.4 |
| 256 / 512 / 1024 / 2048 MiB | exact | exact | exact | exact |

Viable at every positive limit, no absolute floor, byte-identical to the pre-discount behaviour. The `TestEvaluateMemoryProxyUsesPlainRatios` small-budget case and the new 128 MiB table row pin it, and scoping the headroom assertions to `tt.overhead > 0` is the right way to keep the table honest about which rules apply where.

**2. Degenerate band — fixed.** `minViableEffLimitBytes = minPauseHeadroomBytes + minResumeHeadroomBytes` (320 MiB) is exactly the right derivation: at the floor, resume sits a full pause-headroom below pause rather than collapsing. Measured at the boundary — resume is now **40% of budget**, where it was 1%:

| `effLimit` | pause | resume | resume % |
|---|---|---|---|
| 320 MiB (the floor) | 192 | 128 | 40% |
| 322 MiB | 194 | 130 | 40% |
| 384 MiB | 256 | 192 | 50% |

And exporting `MinViableLowLatencyLimitBytes(overheadBytes)` so k8s-proxy can derive rather than duplicate is better than what I asked for. `TestMinViableLowLatencyLimitBytes` pinning viable-at-floor / non-viable-one-byte-under is the assertion that keeps the two repos from drifting.

**3. Zero-usage log — fixed.** `effCurrent` now computed above the viability return: real usage 900 MiB with a 256 MiB overhead reports **644 MiB**, not 0.

Nits all picked up too — `goMemFlooredLogged` now resets symmetrically, the negative-`effLimit` path has its note, the error text no longer says "meets or exceeds" and only fires when a ring buffer actually exists, and the misleading "proxy mode is provably unchanged" comment is now accurate.

---

### One residual, cosmetic: an off-by-one against your own exported floor

`MinViableLowLatencyLimitBytes(256 MiB)` returns exactly **576 MiB**, and the guard treats `effLimit == 320 MiB` (limit 576) as viable — your `TestMinViableLowLatencyLimitBytes` asserts precisely that. But k8s-proxy#729 rejects with `mib*2 <= 576`, so it turns away `memory_limit: 288`:

| `memory_limit` | limit | `effLimit` | guard | #729 |
|---|---|---|---|---|
| 287 | 574 MiB | 318 MiB | non-viable | rejects ✓ |
| **288** | **576 MiB** | **320 MiB** | **viable** | **rejects ✗** |
| 289 | 578 MiB | 322 MiB | viable | accepts ✓ |

One request-MiB, in the safe direction, so nothing breaks. But it's the exact class of drift the exported helper exists to prevent, and it means the planned swap to `MinViableLowLatencyLimitBytes` won't be a no-op — it'll silently widen what's accepted by one step. Cheapest fix is on the #729 side (`<` instead of `<=`, suggested value 288); flagging here because this repo now owns the canonical boundary.

### Gates

`go test ./pkg/agent/memoryguard/` and `go vet` green at `cb90f65a`.

No blocking findings remain from my side. Happy to convert this to a formal approval if you'd like the PR marked accordingly.

charankamarapu (approved):
Approving at `cb90f65a`.

All three findings from my last pass are fixed, and I verified each against the real `evaluateMemory` rather than reading the diff:

- **Proxy mode restored exactly.** The `overhead == 0` early return gives back the plain ratios at every limit — 64/100/128/192/256/512/1024/2048 MiB all land byte-identical to the pre-discount `0.8×`/`0.7×`, and viable at every positive limit. The regression I flagged is gone, and the new 128 MiB table row plus `TestEvaluateMemoryProxyUsesPlainRatios` pin it. Scoping the headroom assertions to `tt.overhead > 0` is the right way to keep the table honest about which rules apply where.
- **Degenerate band gone.** `minViableEffLimitBytes = minPauseHeadroomBytes + minResumeHeadroomBytes` is exactly the right derivation — at the floor, resume sits a full pause-headroom below pause. Measured: resume is now 40% of budget where it was 1%.
- **Zero-usage log gone.** Real usage 900 MiB against a 256 MiB overhead now reports 644 MiB instead of 0.

Exporting `MinViableLowLatencyLimitBytes(overheadBytes)` so k8s-proxy derives its floor instead of duplicating the constants is better than what I asked for, and `TestMinViableLowLatencyLimitBytes` pinning viable-at-floor / non-viable-one-byte-under is the assertion that keeps the two repos from drifting. The smaller nits landed too — symmetric `goMemFlooredLogged` reset, the negative-`effLimit` note, and the ring-buffer error text now only firing when a ring buffer actually exists.

Verified locally at this head: `go test ./pkg/agent/memoryguard/` and `go vet` green.

One residual, cosmetic and owned by the other side: `MinViableLowLatencyLimitBytes(256Mi)` is exactly 576 MiB and this guard treats `effLimit == 320 MiB` as viable, but k8s-proxy#729's check rejects `memory_limit: 288` via `<=`. One request-MiB, safe direction — I've asked for the one-character fix there so the planned swap to this helper stays a pure refactor. Nothing needed in this PR.

PR #4415, opened 2026-08-05
fix(matcher): sectioned noise no longer disables the whole body assertion

slayerjain:
## The bug

A test case's `noise` map accepts two key shapes:

```yaml
noise:
  body:                          # sectioned
    - items.0.product.stock
  body.created_at: []            # dotted
  header:                        # sectioned
    - Content-Length
  header.Date: []                # dotted
```

Only the dotted shape was parsed. A sectioned key fell through both branches of the split loop and then tripped `Contains(MapToArray(noise), "body")` — the sentinel for **ignore the entire response body**. The listed paths were never registered as noise, and every field in the body stopped being compared. `header: [<names>]` did the same to every response header.

The result is a silent false negative: a test case carrying a sectioned noise entry passes no matter what the service returns.

### How it shows up

Replaying a deliberately tampered downstream mock through such a test case reports `PASSED`, while the report file it writes records an expected/actual body differing in four fields:

```
STATUS       : PASSED
body_result  :
  normal   : True                                    ← the verdict
  expected : ..."name":"Laptop","price":1200..."alice_1785918170"...
  actual   : ..."name":"Tablet","price":1999..."zzzzz_1785918170"...
```

On one real recorded app, **45 of 94 test cases (47%)** had a dead body assertion and 20 also had dead header assertions. The paths responsible are exactly what noisy-field detection emits — `requestId`, `timestamp`, `id`, `uptime.*`, `randomData.*` — so the cases that most needed calibration are the ones silently gutted. `/health` and `/stats` became liveness checks that would pass if the handler returned `{}`.

## The fix

Disambiguate on the **value**, not the key:

- bare section key + **empty** list → ignore the whole section (this is what a global `body: {"*": ["*"]}` config expands to; unchanged)
- bare section key + **non-empty** list → those are the field paths to ignore

The four duplicated copies of the split loop (`http/match.go`, `http/absmatch.go`, `grpc/match.go`, `replay.go`) now share `SplitNoise`.

Three further defects surfaced while fixing this and are included:

- **The wildcard sentinel was recorded by writing into `tc.Noise`.** That mutated the caller's test case, panicked when `tc.Noise` was nil, and — because `buildHTTPSchema` serializes `tc.Noise` back out — could be **persisted to disk** by a later re-encode, permanently skipping that case's body even after the wildcard config was removed. The wildcard is now tracked locally.
- **Test-case noise was merged into the caller's global noise config in place**, so one case permanently widened the noise applied to every later case in the run. The gRPC path receives that long-lived config with no copy at all. Now merged into a clone.
- **Sectioned values are read as paths, not regexes.** A regex there was never evaluated — it only tripped the whole-section skip — so a regex-shaped value now warns once, naming the dotted form that does apply one.

## Deliberately not done: array-index normalization

Indexed paths like `items.0.product.stock` still match nothing, because the JSON walker keys array elements without a position (`items.product.stock`).

Normalizing by dropping numeric segments looks obvious and is a trap: noise keys are matched with `strings.Contains`, so `items.0.id` collapses to `items.id`, which is a substring of `items.idempotency_key` and silences it. That trades this PR's silent false negative for a wider one. Verified, so it is left unmatched — **loud** — and pinned by `TestJSONDiffWithNoiseControl_IndexedPathIsNotSupported` with a message telling the next person what to update if the walker ever carries positions.

`buildNoiseIndex` is byte-identical to `main`; `pkg/matcher/utils.go` is a pure insertion with no existing logic touched.

## Tests

~1,140 of the 1,303 added lines are tests. Three of the four changed packages have tests that fail on `main` and pass here; `pkg/matcher` is green on `main` by construction, since `utils.go` adds new functions with no behavioral delta.

- `pkg/matcher/http/absmatch_test.go` and `pkg/matcher/grpc/noise_test.go` are **new files** — neither call site had any noise coverage, and gRPC is where the shared-config leak was worst.
- `pkg/matcher/http/decode_test.go` runs a recorded YAML through `testdb.Decode` and then `Match`, and asserts the decoder really emits the sectioned shape so the fixture cannot silently stop covering the bug.
- `pkg/service/replay/body_noise_test.go` covers `bodyNoiseForTestCase`, previously untested.

`go build ./...`, `go test ./... -count=1`, `gofmt -l pkg/`, `go vet ./pkg/...` all clean.

## Upgrade impact — worth release notes

Suites that were green **only** because of the whole-section skip will start reporting the differences they were hiding. Every such failure is a real, previously-hidden difference rather than a new false positive, but users will experience it as a break, so it should be called out.

Two narrower cases:

- A sectioned value written as a regex is now read as a path. It silenced the whole section before and silences nothing now; the warning names the dotted equivalent.
- A test case whose only noise is an indexed path will now fail on that field. It was never actually matching — the whole-section skip was hiding it.

## Follow-up (separate issue)

`matchJSONWithNoiseHandlingIndexed`'s `case string:` does `if e == a || noisy` **after** the regex check, so a regex-guarded noise entry whose regex does not match is still marked as matching. `{"order.status": ["^(PENDING|PAID)$"]}` passes `PENDING` vs `GARBAGE`. Every regex-constrained noise entry is effectively an unconditional skip. Filed separately — same silent-false-negative family, different blast radius.

### DIFF
--- a/pkg/matcher/grpc/match.go
+++ b/pkg/matcher/grpc/match.go
@@ -190,28 +190,21 @@ func Match(tc *models.TestCase, actualResp *models.GrpcResp, noiseConfig map[str
 	// Handle noise configuration first - needed for JSON comparison
 	noise := tc.Noise
 
+	// Copy before merging: noiseConfig is the caller's long-lived global map.
 	var (
-		bodyNoise   = noiseConfig["body"]
-		headerNoise = noiseConfig["header"] // need to handle noisy header separately (not implemented yet for grpc)
+		bodyNoise   = matcher.CloneNoiseMap(noiseConfig["body"])
+		headerNoise = matcher.CloneNoiseMap(noiseConfig["header"]) // need to handle noisy header separately (not implemented yet for grpc)
 	)
 
-	if bodyNoise == nil {
-		bodyNoise = map[string][]string{}
+	// Merge test-case-specific noise with global noise (similar to HTTP matcher).
+	// TODO: gRPC has never honoured the whole-body skip sentinel that the HTTP
+	// matcher applies, so skipBody is dropped here to preserve that behaviour.
+	tcBodyNoise, tcHeaderNoise, _ := matcher.SplitNoise(noise, logger)
+	for field, regexArr := range tcBodyNoise {
+		bodyNoise[field] = regexArr
 	}
-
-	if headerNoise == nil {
-		headerNoise = map[string][]string{}
-	}
-
-	// Merge test-case-specific noise with global noise (similar to HTTP matcher)
-	for field, regexArr := range noise {
-		a := strings.Split(field, ".")
-		if len(a) > 1 && a[0] == "body" {
-			x := strings.Join(a[1:], ".")
-			bodyNoise[strings.ToLower(x)] = regexArr
-		} else if a[0] == "header" {
-			headerNoise[strings.ToLower(a[len(a)-1])] = regexArr
-		}
+	for field, regexArr := range tcHeaderNoise {
+		headerNoise[field] = regexArr
 	}
 
 	// Compare decoded data - use JSON comparison if both are valid JSON, otherwise use canonicalization
--- a/pkg/matcher/grpc/noise_test.go
+++ b/pkg/matcher/grpc/noise_test.go
@@ -1,0 +1,108 @@
+// Package grpc noise tests.
+package grpc
+
+import (
+	"testing"
+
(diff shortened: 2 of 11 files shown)


slayerjain:
Follow-up filed as #4416 — regex-constrained noise is an unconditional skip (the regex is never enforced). Same silent-false-negative family, kept separate because it has its own blast radius.

slayerjain:
Superseded by #4418, which combines this with the regex-noise fix (#4417) and closes every follow-up both reviews turned up. Same content plus the gaps; nothing here is dropped.

PR #4417, opened 2026-08-05
fix(matcher): enforce regex-constrained noise instead of skipping the field

### LINKED ISSUE #4416: Regex-constrained noise is an unconditional skip: the regex is never enforced
## Summary

A noise entry with a regex is documented and understood as "ignore this field **only when its value matches this pattern**". It does not work that way. When the regex does **not** match, the field is still marked as matching. For numeric and boolean leaves the regex list is not consulted at all.

The practical effect: every regex-constrained noise entry is an unconditional skip. Users who wrote a regex believing it enforces a value shape have a field that is silently never asserted — the opposite of what the regex says.

## Where

`pkg/matcher/utils.go`, `matchJSONWithNoiseHandlingIndexed`:

```go
case string:
    ...
    regs, noisy := ni.match(strings.ToLower(key))
    if noisy && len(regs) != 0 {
        if anyRegexpMatchStr(a, regs) {
            out.matches, out.isExact = true, true
            return out, nil
        }
    }
    if e == a || noisy {          // <-- noisy alone wins after the regex failed
        out.matches, out.isExact = true, true
    }
    return out, nil
```

A failed regex falls through to `if e == a || noisy`, where `noisy` is still true, so the field is accepted anyway. The early return on a successful match is therefore the only thing the regex does — it changes nothing about the outcome.

The `case bool:` and `case float64:` branches discard the regex list outright:

```go
_, noisy := ni.match(strings.ToLower(key))   // regexes dropped
if e == a || noisy { ... }
```

## Reproduction

All three below return `IsExact() == true`, i.e. the replay passes:

| noise | expected | actual |
|---|---|---|
| `{"order.status": ["^(PENDING\|PAID)$"]}` | `"PENDING"` | `"GARBAGE"` |
| `{"order.total": ["^1200$"]}` | `1200` | `99999` |
| `{"order.paid": ["^true$"]}` | `true` | `false` |

```go
e, a := `{"order":{"status":"PENDING"}}`, `{"order":{"status":"GARBAGE"}}`
v, _ := ValidateAndMarshalJSON(zap.NewNop(), &e, &a)
res, _ := JSONDiffWithNoiseControl(v, map[string][]string{"order.status": {"^(PENDING|PAID)$"}}, false)
// res.IsExact() == true
```

`GARBAGE` does not match `^(PENDING|PAID)$`, so this should fail.

## Expected

When a noise entry carries a non-empty regex list, the field should be ignored **only** if the actual value matches one of the patterns. A non-matching value is a real difference and should fail the test.

The same should hold for numeric and boolean leaves: the regex should be applied to the value's string form rather than dropped.

## Impact

This is the same family as #4415 (a noise shape silently disabling assertions), but arguably wider: regex noise is a documented feature that people reach for precisely when they want a *constrained* ignore, and it currently degrades to a blanket ignore with no warning.

Fixing it is a breaking change in the same direction — suites green because a regex-guarded field was never really checked will start reporting genuine differences — so it likely wants the same release-note treatment.

## Note on fixing

Worth checking the empty-vs-non-empty regex list contract while fixing: an empty list is "ignore unconditionally" and is relied on throughout (auto-detected noise emits it), so only the non-empty case should change behaviour.

slayerjain:
Fixes #4416.

## The bug

A noise entry reads `{"<field.path>": ["<pattern>", ...]}` and is understood as *ignore this field when its value matches one of these patterns*. It did not work that way.

In `matchJSONWithNoiseHandlingIndexed`, the `case string:` branch ran the match and, when the pattern did **not** match, fell through to:

```go
if e == a || noisy {   // noisy is still true here
    out.matches, out.isExact = true, true
}
```

so the field was accepted anyway. `case bool:` and `case float64:` discarded the pattern list entirely (`_, noisy := ni.match(...)`).

Every pattern-constrained noise entry was therefore an **unconditional skip** — the field was never asserted, which is the opposite of what writing a pattern asks for. All three of these passed:

| noise | recorded | replayed |
|---|---|---|
| `{"order.status": ["^(PENDING\|PAID)$"]}` | `"PENDING"` | `"GARBAGE"` |
| `{"order.total": ["^1200$"]}` | `1200` | `99999` |
| `{"order.paid": ["^true$"]}` | `true` | `false` |

## The fix

The three scalar branches now route through `noisyForValue`, which ignores the field only for values the patterns describe. **An entry with no patterns keeps ignoring unconditionally** — that is what noise auto-detection emits, and it is unchanged; there is a differential corpus below proving it.

Three further problems surfaced while fixing this and are included.

### A typo in a pattern crashed the process

`getCompiled`'s "never matches" fallback was `` `(?!)` `` — a negative lookahead, which Go's RE2 engine rejects. So `regexp.Compile` returned `(nil, err)`, the fallback assigned that nil, and the first comparison dereferenced it:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation]
  regexp.(*Regexp).MatchString(...)
  matcher.anyRegexpMatchStr(...)
```

Replaced with `` `[^\s\S]` `` (a character that is neither whitespace nor non-whitespace — matches nothing, not even the empty string), plus a nil guard. The identical dead fallback in the `ansiRegex` init is swept too; it is unreachable today but is the same trap for whoever next edits that pattern.

### Patterns were dropped for dot-free keys, which is the common shape

`JSONDiffWithNoiseControl` classifies a key with no dot as a "global key" and threw its patterns away. Since `http.Match` strips the `body.` prefix, `body.status` arrives here as the global key `status` — so enforcing patterns only on dotted paths would have left most real-world entries broken. `globalKeys` now carries its compiled patterns.

### Overlapping entries produced a nondeterministic verdict

`match()` returns the first matching entry, and `buildNoiseIndex` appended in Go map order. That was invisible while every entry meant the same thing; once patterns decide the outcome, a suite flips between runs with no input change:

```
noise: {"user.id": ["^\d+$"], "order.user.id": []}
same process, 400 runs -> true 73 / false 327
```

Entries are now ordered most-specific first, which is deterministic and is also the precedence a user writing both `user.id` and `order.user.id` would expect.

## Where a pattern applies

A pattern describes one scalar value, so that is where it is enforced.

- **Scalars** — enforced.
- **Objects** — the pattern relaxes the values beneath it, but the structural walk continues, so a key added, removed or retyped inside is still reported. This matches `main` exactly (verified below); it does not skip the subtree.
- **Arrays of scalars** — enforced per element. `ids: ["^ord-\d+$"]` unambiguously means every id looks like that. Length changes are still structural and still reported.
- **Arrays of objects / mixed arrays** — values relaxed, as with objects.
- **Numbers** — rendered exactly as `encoding/json` does, so a pattern written against the value Keploy prints in its diff matches at replay. Asserted against `json.Marshal` over 700k random floats including raw bit patterns and both format cutovers.

## Verification

Differential corpora against a `HEAD` tree, ~156k comparisons: **zero cases where this branch accepts something `main` rejects.** A 40k-case corpus in which every entry has an empty pattern list is byte-identical to `main`.

The tightenings attribute as 386 from global keys and 2 from path noise — i.e. the dot-free fix is most of what this delivers.

The crash fix reconfirms at scale: a 36k-case corpus containing an uncompilable pattern SIGSEGVs `main` partway through and completes here.

## Upgrade impact — worth release notes

Suites green because a pattern-guarded field was never really checked will start reporting the differences they were hiding. Those are genuine, previously-hidden regressions.

One rougher edge: an invalid pattern still silently becomes never-match with **no warning**, so glob-style input like `*` or `*id*` now yields a red test with no explanation. On `main` those were an unconditional skip for number and bool leaves. Fixing it properly means `getCompiled` surfacing the compile error to a logger; left as a follow-up rather than bolted on here.

## Not included, deliberately

- `case nil:` has no noise handling, so a pattern cannot cover a field that is `null` in the recording. Pre-existing and untouched.
- `risk.go`'s `collectJSON` still uses the old `len(regs)==0` rule, so a field the matcher now ignores is still named by the failure-assessment path. That path is reporting, not the pass/fail verdict.
- `SubstringKeyMatch` (used by `CompareHeaders`) has the same first-match-wins ambiguity and already carries a comment admitting it. Same remedy, separate change.

### DIFF
--- a/pkg/matcher/noise_regex_test.go
+++ b/pkg/matcher/noise_regex_test.go
@@ -1,0 +1,487 @@
+package matcher
+
+import (
+	"encoding/json"
+	"math"
+	"regexp"
+	"testing"
+
+	"go.uber.org/zap"
+)
+
+// diffNoise is a small helper: compare two JSON documents under a noise map and
+// report whether the matcher considered them equal.
+func diffNoise(t *testing.T, expected, actual string, noise map[string][]string) bool {
+	t.Helper()
+	e, a := expected, actual
+	v, err := ValidateAndMarshalJSON(zap.NewNop(), &e, &a)
+	if err != nil {
+		t.Fatalf("ValidateAndMarshalJSON: %v", err)
+	}
+	res, err := JSONDiffWithNoiseControl(v, noise, false)
+	if err != nil {
+		t.Fatalf("JSONDiffWithNoiseControl: %v", err)
+	}
+	return res.IsExact()
+}
+
+// TestRegexNoise_EnforcesPattern pins the contract of a regex-guarded noise
+// entry: the field is ignored ONLY when the replayed value matches one of the
+// patterns. A value outside the pattern is a real difference.
+//
+// Previously the regex was decorative — a failed match fell through to
+// `if e == a || noisy`, where noisy was still true, so the field was accepted
+// anyway. Every regex-constrained noise entry was an unconditional skip, which
+// is the opposite of what the pattern says.
+func TestRegexNoise_EnforcesPattern(t *testing.T) {
+	cases := []struct {
+		name     string
+		expected string
+		actual   string
+		noise    map[string][]string
+		want     bool
+	}{
+		{
+			name:     "string: value inside the pattern is ignored",
+			expected: `{"order":{"status":"PENDING"}}`,
+			actual:   `{"order":{"status":"PAID"}}`,
+			noise:    map[string][]string{"order.status": {"^(PENDING|PAID)$"}},
+			want:     true,
+		},
+		{
+			name:     "string: value outside the pattern is a real difference",
+			expected: `{"order":{"status":"PENDING"}}`,
+			actual:   `{"order":{"status":"GARBAGE"}}`,
+			noise:    map[string][]string{"order.status": {"^(PENDING|PAID)$"}},
+			want:     false,
+		},
+		{
+			name:     "string: identical values still match even outside the pattern",
+			expected: `{"order":{"status":"GARBAGE"}}`,
+			actual:   `{"order":{"status":"GARBAGE"}}`,
+			noise:    map[string][]string{"order.status": {"^(PENDING|PAID)$"}},
+			want:     true,
+		},
+		{
+			name:     "string: an empty regex list still ignores unconditionally",
+			expected: `{"order":{"status":"PENDING"}}`,
+			actual:   `{"order":{"status":"GARBAGE"}}`,
+			noise:    map[string][]string{"order.status": {}},
+			want:     true,
+		},
+		{
+			name:     "number: value inside the pattern is ignored",
+			expected: `{"order":{"total":1200}}`,
+			actual:   `{"order":{"total":1250}}`,
+			noise:    map[string][]string{"order.total": {`^12\d\d$`}},
+			want:     true,
+		},
+		{
+			name:     "number: value outside the pattern is a real difference",
+			expected: `{"order":{"total":1200}}`,
+			actual:   `{"order":{"total":99999}}`,
+			noise:    map[string][]string{"order.total": {`^12\d\d$`}},
+			want:     false,
+		},
+		{
+			name:     "number: an empty regex list still ignores unconditionally",
+			expected: `{"order":{"total":1200}}`,
+			actual:   `{"order":{"total":99999}}`,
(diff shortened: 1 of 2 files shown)


slayerjain:
Superseded by #4418, which combines this with the sectioned-noise fix (#4415) and closes every follow-up. Same content plus the gaps; nothing here is dropped.

PR #4420, opened 2026-08-06
ci(windows): move the Windows CGO build to a GitHub-hosted runner

slayerjain:
## The problem

The self-hosted Windows fleet is failing **every open PR**. `precheck-windows` and/or `build-windows-amd64` record **zero steps**, never upload logs, and fail after 13–14 minutes on varying runners — the runner accepts the job and dies before executing anything.

| PR | build-windows-amd64 | precheck-windows |
|---|---|---|
| #4419 | fail | fail |
| #4413 | fail | pass |
| #4412 | pass | fail |
| #4410 | pass | fail |
| #4407 | fail | pass |
| #4406 | fail | pass |

The shuffling pattern is the tell: this is runner health, not a code defect.

## What moves, and why it's safe

**`build-windows-amd64` never needed self-hosted.** `release.yml`'s `build-windows` already performs the *identical* native windows/amd64 CGO build — including the `pg_query_go` cgo link, which is the hard part — on `windows-latest`. Every keploy release binary ships from it (5/5 recent releases green; v3.6.9 built in 4m42s with a cold module cache).

Of that job's 396 lines, **only ~6 steps do real work.** The rest existed solely to cope with shared, persistent machines and are deleted:

- `Bootstrap PowerShell environment` — installs/repairs pwsh (hosted ships 7.6.4)
- `Free runner disk space` — 98 lines, because the boxes run out of C:
- `Create workflow start lock` — cross-job coordination on a shared filesystem
- `Setup per-job git isolation` — 71 lines, so 4 concurrent checkouts don't clobber each other's global gitconfig
- `Recover stale runner state` — kills leftover `keploy.exe`, stops a stale WinDivert driver service
- `Prune stale git isolation dirs`

Job name, fork-PR guard, timeout and artifact contract are unchanged, so **no branch-protection or required-check edit is needed**.

## What deliberately stays self-hosted

`golang_docker_windows` — hosted `windows-latest` ships Docker in **Windows-containers-only** mode (LCOW was removed in Docker Engine 23.0), while that lane loads a `linux/amd64` image and injects a keploy-agent sidecar needing `cap_add BPF/PERFMON/NET_ADMIN` and `/sys/fs/cgroup`, `/sys/kernel/debug`, `/sys/fs/bpf` bind mounts. It is the only place a native `keploy.exe` drives Docker Desktop's Linux engine — exactly the configuration Windows users run. Moving it would trade real product coverage for a green checkmark.

`precheck-windows` and `cleanup_windows` stay because they serve it.

## Three couplings this would otherwise have broken

**1. The workflow start lock.** `cleanup-windows.ps1` prunes only when *no* lock files remain, so deleting its sole writer would have fired `docker system prune -af --volumes` on **every** run against the shared VM — possibly mid-`docker compose up` on a concurrent job. The lock step is relocated into `precheck-windows`. Its removal step also gains `if: always()` plus a 24h stale-lock reaper (mirroring the macOS twin): three unguarded steps precede it, and `setup-git-isolation.ps1` throws when `ssh-keyscan` returns empty — a real fleet condition — so a leaked lock would otherwise disable the VM's last reclamation permanently and silently.

**2. Disk observability.** With the sweep gone and `cleanup_windows` dropped from the gate, nothing would observe the fleet filling up — a slow-burn version of the failure this PR fixes. `precheck-windows` gains a disk-headroom report (`::warning::` under 6 GB, fail under 2 GB). It lives there because precheck is the earliest self-hosted job *and* still a gate input. It **observes only**: the old sweep purged `GOMODCACHE` and walked `C:\actions-runners`, racing concurrent jobs on the shared box.

**3. ExecutionPolicy.** `download-image`'s two Windows steps relied on a side effect — the deleted bootstrap step wrote `Set-ExecutionPolicy -Scope CurrentUser -Bypass`, which persists in HKCU across jobs and runs, so any box that had run `build-windows-amd64` was covered. That guarantee is gone; both steps now carry the flag explicitly.

## Gate

`cleanup_windows` leaves `gate_windows`' `needs` — it is VM hygiene with no product signal, and as a gate input a busy fleet could red-light a PR on its own. `precheck-windows` **stays** in the gate: the gate counts `skipped` as pass, so dropping it would let the docker lane skip invisibly.

## A latent bug this surfaced

The two inline MinGW probes had **diverged**. `release.yml` appended every match with no `break` — and since `GITHUB_PATH` entries are *prepended*, the **last** match won, inverting the listed preference. `prepare_and_run.yml` probed five paths with a `Get-Command` fallback and a portable-toolchain download. So the PR build and the shipped release binary could resolve **different toolchains**.

Now one shared `.github/scripts/windows/ensure-mingw.ps1` with a `break` and an explicit `throw`. The `Get-Command gcc` fallback is dropped on purpose: the hosted image also ships Strawberry Perl's gcc at `C:\Strawberry\c\bin` on the machine PATH, so explicitly prepending `C:\mingw64\bin` is what keeps selection deterministic.

## Verification

- All touched YAML parses; `actionlint` reports **8 findings before and after** (no new ones — the 8 are pre-existing macOS `native` label + shellcheck info)
- No job added, removed, or renamed (33 → 33)
- Artifact contract byte-identical (`build-windows` / `keploy.exe`); both consumers use `download-artifact`
- `_ISO_ROOT` has zero remaining references repo-wide
- No bare `shell: powershell` left anywhere under `.github/`

**Net: 3 self-hosted Windows job declarations instead of 4, ~300 lines of runner babysitting gone, and no change to what is tested.**

## Follow-ups (not in this PR)

- `GOOS=windows go build ./...` fails locally on `pkg/agent/hooks/windows` (needs cgo + mingw), on `main` too — so a genuine Windows compile error is masked for anyone checking before push.
- `gate_macos` still lists `clean_up_docker_macos` in `needs`; the rationale for dropping `cleanup_windows` applies identically to the macOS twin.

### DIFF
--- a/.github/actions/download-image/action.yml
+++ b/.github/actions/download-image/action.yml
@@ -44,7 +44,7 @@ runs:
 
     - name: Wait for the Docker daemon (Windows)
       if: runner.os == 'Windows'
-      shell: powershell
+      shell: 'powershell -ExecutionPolicy Bypass -NoLogo -NoProfile -Command ". ''{0}''"'
       run: |
         $maxAttempts = 30
         $attempt = 0
@@ -117,17 +117,29 @@ runs:
       if: runner.os == 'Windows'
       # Use Windows PowerShell 5.1 (always preinstalled on Windows
       # runners) rather than pwsh (PowerShell 7, which not every
-      # self-hosted runner in the fleet ships — see the Bootstrap
-      # PowerShell environment step in prepare_and_run.yml's
-      # build-windows-amd64 job, which repairs *that* job but does
-      # not run in the pull-docker-image-windows job that invokes
-      # this action). Every cmdlet used below (Join-Path, Test-Path,
+      # self-hosted runner in the fleet ships).
+      #
+      # The explicit -ExecutionPolicy Bypass is now required, not
+      # decorative. GitHub's built-in `powershell` shell invokes
+      # `powershell -command ". '<script>'"` with no policy flag, and
+      # dot-sourcing a .ps1 is subject to ExecutionPolicy. Until
+      # recently this happened to work on the fleet because
+      # prepare_and_run.yml's "Bootstrap PowerShell environment" step
+      # wrote `Set-ExecutionPolicy -Scope CurrentUser -Bypass`, which
+      # persists in HKCU across jobs and runs — so any box that had run
+      # build-windows-amd64 was covered by a side effect. That step was
+      # deleted when build-windows-amd64 moved to GitHub-hosted
+      # windows-latest, so the guarantee is gone; a freshly provisioned
+      # runner would fail here. Carrying the flag ourselves makes this
+      # action independent of whatever else ran on the machine first.
+      #
+      # Every cmdlet used below (Join-Path, Test-Path,
       # Get-ChildItem, Select-String, [Environment]::NewLine,
       # string .Trim(), [string]::IsNullOrWhiteSpace) is 5.1-
       # compatible, so keeping the floor at 5.1 is free and avoids
       # a 'pwsh: command not found' step failure on unrepaired
       # runners.
-      shell: powershell
+      shell: 'powershell -ExecutionPolicy Bypass -NoLogo -NoProfile -Command ". ''{0}''"'
       run: |
         $ErrorActionPreference = 'Stop'
         $tar = Join-Path "${{ runner.temp }}" "keploy-image\image.tar"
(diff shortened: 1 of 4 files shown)


chilkotiKartik (approved):
LGTM! Thoroughly verified changes in ci(windows): move the Windows CGO build to a GitHub-hosted runner. Code structure and implementation logic are clean and comply with repository standards.

PR #4421, opened 2026-08-06
ci(windows): scope stale-state recovery to the job, and enforce it

slayerjain:
## The hazard

Four self-hosted Windows runner **agents share one physical VM**. `Recover stale runner state` existed in **five** near-identical inline copies, and four were machine-wide:

```powershell
Get-Process -Name 'keploy' | Stop-Process -Force
foreach ($n in @('WinDivert','WinDivert64')) { Stop-Service $n -Force }
```

On a shared VM that kills a **sibling** runner's host keploy CLI mid-run, and stops a single machine-wide WFP kernel driver another job may be capturing through.

`golang_docker_windows.yml` had already been hardened to scope both by workspace path. The other four were silently left behind — **that drift is the actual bug**, so the fix has to prevent it recurring, not just re-synchronise once.

## The fix

All five sites now inline one canonical body from `.github/scripts/windows/recover-stale-workspace.ps1`:

- the kill is restricted to processes whose path is under this job's `$GITHUB_WORKSPACE` (with the `keploy` vs `keploy2` prefix trap guarded)
- the driver resets only when no keploy exists **outside** this workspace; a process whose path can't be read counts as *other* (conservative)
- with no workspace it does **nothing**, rather than degrading to machine-wide

**Inlined rather than invoked** because every call site runs *before* `actions/checkout` — the step exists to make that checkout succeed, so it can't load code from the checkout it repairs. This follows the existing precedent of `Setup per-job git isolation`.

### Two fixes beyond scoping

**The name match widens to `-like 'keploy*'`.** `golang_native_windows.yml` installs the binaries as `keploy-record.exe` / `keploy-replay.exe`, which the exact `-Name 'keploy'` form never matched — that lane's cleanup had never caught the leftovers it was written for.

**A race is closed.** `Stop-Process` doesn't wait for exit, so a just-killed process could still be enumerated with an unreadable path, be counted as a sibling, and suppress the job's *own* driver reset. Section 1 now records the PIDs it terminates and section 2 skips them. Proven both directions against a pwsh shim: the pre-repair body blocks its own reset; the new one doesn't — while a genuine sibling, *including one appearing only after section 1*, still correctly blocks it.

## Enforcement

Because five copies is what caused this, `.github/scripts/check-inline-regions.py` runs on every PR (new `inline-regions` job in `golangci-lint.yml`, which is trigger-unfiltered). It parses the repo's YAML, walks both `jobs[*].steps` and composite `runs.steps`, and fails on:

- a registered copy that drifts (unified diff naming file/job/step)
- an **unregistered** copy of the step — a new divergent copy can't land silently
- a registered copy that vanishes
- any *other* `run:` body carrying the region's fingerprint under a different step name
- its own wiring being deleted
- an empty `REGIONS` table (a vacuous config must not print a confident OK)

It also **refuses to propagate an unsafe region**: if the source body regains the machine-wide shape without its scoping markers, both verify and `--fix` fail loudly rather than pushing it into all five call sites. Without that, the check would enforce *consistency* and never *safety* — one careless edit to the `.ps1` would have undone the whole PR with a green tick.

**Its limits are documented rather than overstated.** It catches accidental drift, not someone deliberately switching it off (`continue-on-error`, `if: false`, `env`/`matrix` indirection, narrowed triggers). Those are visible in review, which is where they belong. Adding regexes until it "cannot" be evaded would only buy a comment claiming a guarantee the code doesn't provide.

## Verification

- checker passes on the **index** — what `git commit` actually writes — not merely the working tree
- fails, with an actionable message, on: source-region regression, single-character drift, whitespace-only drift, an unregistered copy, a renamed step, a deleted copy, deleted wiring, and an empty `REGIONS`
- `actionlint`: **57 diagnostics before and after**, zero new (the 2 touching changed files are pre-existing, line numbers shifted only)
- all touched YAML parses; no job's test coverage changed; every recover step keeps its `continue-on-error`

## One thing worth knowing

`Get-Service` **is** correct for WinDivert despite it being a kernel driver — I initially believed otherwise and was wrong. PowerShell only routes through `ServiceController.GetServices()` on the *wildcard* path; a literal name takes `GetOneService` → `new ServiceController(name).Status`, which is `OpenService`/`QueryServiceStatus` and succeeds for `SERVICE_KERNEL_DRIVER`. Microsoft's docs say so explicitly. A comment now records this so it isn't "tidied" into a wildcard, which would silently break it.

## Related

- Does **not** fix the current Windows CI failures — those are 3 of 4 runner agents dead on the host since 2026-07-22 (they heartbeat `online` but never launch `Runner.Worker`), which needs an operator on the box.
- #4420 moves `build-windows-amd64` off the fleet; expect a trivial delete-vs-modify conflict there on the `build-windows-amd64` copy — resolve by taking the delete.
- Follow-up, same bug class, deliberately out of scope: `ensure-docker.ps1`'s `Restart-DockerDesktop` still force-kills `Docker Desktop` and `com.docker*` machine-wide on the shared VM, destroying every concurrent sibling job's containers.

### DIFF
--- a/.github/scripts/check-inline-regions.py
+++ b/.github/scripts/check-inline-regions.py
@@ -1,0 +1,844 @@
+#!/usr/bin/env python3
+"""Enforce that inlined copies of a shared script region stay byte-identical.
+
+Some workflow steps have to run BEFORE actions/checkout, so they cannot call a
+script out of the repository - the repository is not on disk yet. Those steps
+inline the body instead, which means the same code exists in several places at
+once and can drift apart silently. Drift between such copies is exactly the
+class of bug this check exists to prevent, so the invariant is enforced here
+rather than left to a comment asking people to keep the copies in sync.
+
+The check parses every YAML file outside SKIP_DIRS once (workflows under
+.github/workflows, composite actions under .github/actions/*/action.yml, and
+anything else that happens to describe steps) and walks both step containers
+GitHub understands: `jobs.<id>.steps` in a workflow and `runs.steps` in a
+composite action. Everything below is driven by that one structural walk.
+
+WHAT THIS CHECK IS AND IS NOT. It catches ACCIDENTAL divergence - the failure
+that actually happened here, where one copy was hardened and the others were
+silently left behind. It is not a sandbox and does not pretend to stop someone
+who sets out to defeat it: a body assembled from `env:` or `strategy.matrix`
+and run via `Invoke-Expression`, a payload in a committed helper script, a
+copy under a SKIP_DIR, `continue-on-error` on the checking step, or narrowing
+the host workflow's triggers all pass. Those are visible in review, which is
+where they belong; adding regexes until the check "cannot" be evaded would
+only buy a comment claiming a guarantee the code does not provide. See the
+"fingerprints" entry in REGIONS for the exact detection boundary.
+
+For each region described in REGIONS below the check:
+
+  * extracts the region between the BEGIN/END markers of the source script;
+  * NAME NET - collects every step named `step_name`, compares each inlined
+    body (the YAML parser hands back the block scalar already dedented)
+    byte-for-byte with the region, and prints a unified diff naming the file,
+    job and step when it differs;
+  * requires that the inlined body is indented by exactly INDENT spaces in the
+    file, so the source script's description of the layout stays true;
+  * requires that the set of call sites found equals the registered set, so a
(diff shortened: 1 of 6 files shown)


chilkotiKartik (approved):
LGTM! Thoroughly verified changes in ci(windows): scope stale-state recovery to the job, and enforce it. Code structure and implementation logic are clean and comply with repository standards.

PR #4422, opened 2026-08-06
fix(replay): keep TestModeInfo.IsFiltered describing the pool, not mapping membership

Shivanipandey31:
## Problem

`TestModeInfo.IsFiltered` records **which pool a mock lives in** — `true` = per-test tree, `false` = session/startup tree — and matchers read it to decide which tree to consume from.

`filterByMapping` answers a different question, *"is this name in `mappings.yaml`"*, and it was stamping that answer onto `IsFiltered`. Its two callers route their output to **opposite pools**, so the value was right by luck in one and wrong in the other:

| Caller | Destination pool | Keeps | Label applied | Correct? |
|---|---|---|---|---|
| `FilterTcsMocksMapping` | per-test | only the mapped group | `true` | ✅ coincidentally |
| `FilterConfigMocksMapping` | **session** | **both** groups | `true` for mapped entries | ❌ |

A session mock that an earlier replay consumed appears in `mappings.yaml`, so it arrives labelled `IsFiltered=true` while physically living in the **session** tree.

This only bites in mapping-based filtering. `FilterConfigMocksTierAware` — the timestamp-based equivalent — sets `false` for exactly these mocks. So replay behaviour depended on **whether a `mappings.yaml` happened to exist on disk**, which is true for any replay after the first, including a second plain `keploy test`.

### Why a mislabel causes data loss

The filtered and unfiltered trees are keyed on `TestModeInfo` `(SortOrder, ID)` — not `Name`, not `Kind` — and `SetFilteredMocks` / `SetUnFilteredMocks` each stamp those from their **own** 0-based slice index. The first mock of every pool therefore lands on `(SortOrder=1, ID=0)`: **the same key exists in more than one tree.**

So a delete aimed at the wrong tree doesn't miss — it removes whichever unrelated mock occupies that coordinate.

## Fix

`filterByMapping` no longer touches `IsFiltered`. Each caller stamps the label matching its own destination (`true` for the per-test pool, `false` for the session pool). Mapping membership still drives selection and ordering — it just no longer redefines which tier a mock belongs to.

After this, mapping mode and timestamp mode agree for the same input.

## Observed impact

Found via the enterprise ZooKeeper parser, which picks its consume target from `IsFiltered`. Consuming a session mock in mapping mode both:

- **destroyed an unrelated per-test mock**, so a later request that had a perfectly good recorded answer failed with `no mock for <op>`, and
- **left the session mock unconsumed**, so a connect cursor never advanced and successive connections replayed the same recorded session id.

Reproduced end to end with a real ZooKeeper app and real `keploy record` / `keploy test`:

| | before | after |
|---|---|---|
| `no mock for …` | 4 | **0** |
| test verdict | `obsolete: 1` | `passed: 2` |

Note the ZooKeeper parser lives in the enterprise repo; the defect fixed here is the OSS half.

## Reviewer attention, please

`generic/match.go` also partitions match **candidates** on `IsFiltered`, so this changes its candidate ordering in mapping mode. I believe that's a correction — it now matches timestamp mode — but it is a genuine behaviour change for generic-protocol replays and **I have not exercised it**. A second opinion would be welcome.

## Scope note

This PR originally also added a `TreeDb.deleteMock` identity check, so that a wrong-tree delete would miss cleanly instead of evicting a bystander. **That has been removed** (see the revert commit) and will be proposed separately:

- it changed the delete path for *every* parser, with a silent failure mode (a delete that should succeed starts returning `false`, the mock leaks, a later request matches something it shouldn't)
- two enterprise lanes that were green without it went red on the run that carried it — not yet attributed, but no reason to bundle a broad change with a narrow one
- it needs its own review, starting with *"can `Name` change across a `TreeDb.update`?"*

**This fix is sufficient on its own:** with the label correct, nothing aims a delete at the wrong tree, so the ambiguous-key path is never reached. The `(SortOrder, ID)` ambiguity remains as latent fragility against any *future* mislabel — worth hardening, but that's defence in depth, not the fix.

## Tests

`TestFilterMappingLabelsMatchDestinationPool`:
- per-test pool is labelled `true`
- session pool is labelled `false` **regardless of mapping membership**
- mapping mode agrees with the timestamp-based path for the same input

Fixtures seed the *opposite* of the expected value so a missing write can't pass by accident, and set both `ReqTimestampMock` and `ResTimestampMock` — a zero timestamp routes a mock down an early branch into the per-test pool and would make the tier comparison measure the wrong thing.

Full suite **59 ok / 0 FAIL**; `gofmt` and `vet` clean.


chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(replay): keep TestModeInfo.IsFiltered describing the pool, not mapping membership. Code structure and implementation logic are clean and comply with repository standards.

PR #4423, opened 2026-08-07
feat(proxy): configurable telemetry-egress passthrough (skip / recordOne)

ayush3160:
## Update — extended scope (query-mux support, `off` opt-out, versioned defaults)

Since the original description, this PR also adds:

**Query-multiplexed collectors (New Relic) — `QueryKeys []string` on `PassThroughRule`.**
Some collectors POST every RPC to one path, distinguished only by a query param (New Relic: `/agent_listener/invoke_raw_method?method=preconnect|connect|metric_data|…`). A path-only `recordOne` key collapses them to a single mock, breaking the agent (e.g. `connect` never gets its `agent_run_id`). `QueryKeys` names the *significant* query params, folded into the record-time dedup key (`ptRecordKey`) and required at replay (`mockQueryMatches`); other params (a per-session `run_id`) stay ignored, so `recordOne` still collapses. New built-ins: `*.newrelic.com` `/agent_listener/invoke_raw_method` → `recordOne` + `QueryKeys:[method]` (the first non-`skip` built-in, because the agent parses the response), and Prometheus remote_write `/api/v1/write` → `skip`.

**`mode: "off"` — explicit opt-out.** A user rule with `mode:"off"` matches (so it overrides a built-in default for the same target) but is **not** passed through — the endpoint is recorded/replayed as a normal dependency. This lets a user turn off a built-in when that endpoint is actually one of their own internal routes.

**Versioned, exposed defaults — `TelemetryDefaults()` + `TelemetryDefaultsVersion`.** The built-ins become a single source of truth with display metadata (provider/label); `BuiltinTelemetryDefaults()` is derived from it. The control plane serves this set (k8s-proxy#740) so the UI (enterprise-ui#1664) renders the exact set the agent applies — no drifting hardcoded copy — and the version lets the UI prove it.

**Validation:** e2e-verified end-to-end in **classic** and **proxyless/low-latency** modes against a sample app emitting OTLP + Datadog + New Relic + Prometheus + Pyroscope + StatsD — New Relic's 3 methods record as 3 distinct mocks (collapsing repeat `run_id`s), skips drop, and replay passes with the upstream down. Backward-compatible: `NormalizeMode` preserves `off`, empty/unknown still default to `recordOne`, no mock-format change.

---

Makes telemetry / noisy-egress passthrough **configurable** (it was hardcoded to two HTTP paths, `/v1/traces` and `/ingest`) and generalizes it into a reusable rule model shared across the OSS + enterprise + integrations parsers.

**Core model — `pkg/models/passthrough.go` (new):**
- `PassThroughRule{Port, Host, Path, Mode}` where `Mode` is `skip` or `recordOne` (empty ⇒ `recordOne`).
- `MatchPassThrough()` — specificity precedence (`host+port` > `host` > `port`), host globbing (`*.datadoghq.com`), path-prefix, host:port normalization.
- Bare-scalar JSON shorthand (`8126` / `"4317"`) alongside the object form.
- `BuiltinTelemetryDefaults()` + `MergePassThroughDefaults()` — OTLP (`/v1/{traces,metrics,logs}`, gRPC collector `:path`), Pyroscope `/ingest`, Datadog trace-agent paths, and SaaS hosts (`*.datadoghq.com`, `*.sentry.io`, App Insights).

**Config plumbing:** `record.passThroughPorts` / `record.passThroughHosts` on `config.Record` + `models.OutgoingOptions`, wired into all `OutgoingOptions` construction sites (record, replay ×2, runner, mock).

**HTTP record (`http.go`, `recordv2.go`):** `skip` drops the exchange; `recordOne` keeps exactly one representative per `(method,host,port,path)` — preferring the first `2xx` (≤2 mocks worst case) — tagged `Spec.Metadata["type"]=config` so it becomes a session-lifetime mock (reusable, survives strict-window).

**HTTP replay (`decode.go`, `passthrough_egress.go` new):** short-circuits **before** the fuzzy-matcher (which OOMs on large telemetry bodies): `skip` synthesizes a `200`; `recordOne` serves the one recorded response body-agnostically for every matching call, non-consuming.

Tests: `pkg/models/passthrough_test.go`, `pkg/agent/proxy/integrations/http/passthrough_egress_test.go` (precedence, host glob, path-prefix, JSON shorthand, defaults merge, decision, first-2xx recorder). `go build ./...` clean, `go test` green.

## Links & References

**Closes:** keploy/enterprise#2363

### 🔗 Related PRs
- keploy/keploy#4423 — OSS core: `PassThroughRule` model + HTTP record/replay
- keploy/integrations#256 — HTTP/2 + gRPC (grpcV2) replay/record short-circuit
- keploy/enterprise#2365 — proxyless + DaemonSet gate (userspace recordOne + kernel skip)
- keploy/k8s-proxy#740 — control-plane `record_config` + DaemonSet ConfigMap
- keploy/enterprise-ui#1664 — UI config form

### 🐞 Related Issues
- keploy/enterprise#2363

### 📄 Related Documents
- NA

- [x] 🍕 Feature
- [x] ✅ Test

## Added e2e test pipeline?
- [ ] 👍 yes

- [x] 👍 yes

## Added to documentation?


pkg/agent/proxy/integrations/http/passthrough_egress.go
```diff
@@ -0,0 +1,159 @@
...
+
+	"go.keploy.io/server/v3/pkg/agent/proxy/integrations"
+	"go.keploy.io/server/v3/pkg/models"
+)
+
+// passThroughEgressDecision reports whether an outgoing HTTP request targets a
+// configured telemetry / noisy-egress endpoint and, if so, the effective mode
+// ("skip" or "recordOne"). Configured rules (opts.PassThroughPorts /
+// PassThroughHosts) take precedence; the built-in telemetry-path defaults
+// (telemetryEgressPaths, see isTelemetryEgress) apply as a "skip" fallback when
```
Sarthak160: Please make shorter inline comments

Sarthak160 (approved):

Sarthak160:
@ayush3160 lint is failing please check

pkg/agent/proxy/integrations/http/passthrough_egress.go
```diff
@@ -0,0 +1,209 @@
...
+		b[i] = byte('0' + v%10)
+		v /= 10
+	}
+	return string(b[i:])
+}
+
+// ptRecorder de-duplicates recordOne captures to at most one representative
+// exchange per endpoint key, preferring the first 2xx. Worst case (a warmup
+// error precedes the first success) it keeps ≤2 mocks; once a 2xx is seen every
+// later capture for that key is dropped. Shared across a parser instance's
```
ayush3160, line 188: **should-fix** — `ptRecorder.seen` is keyed only by `(method,host,port,path[,queryKeys])` — no app/session identity — and is created once and never reset. This comment defers per-app isolation to "the enterprise gate (T7)", but that gate only sets per-app `opts.PassThroughPorts`; it never gives each app its own recorder. So in a long-lived DaemonSet agent serving multiple apps (or successive sessions), app A's first `recordOne` capture makes `shouldRecord()` drop app B's identical-endpoint exchange — B records nothing and replays a synthetic empty 200 instead of the real body, which breaks exactly the response-parsing collector `recordOne` exists for (New Relic `connect` returning `agent_run_id`). Key the dedup by app/session, or reset it per session.

ayush3160 (commented):
**Principal review — feature core (1 of 4 PRs).** This is the foundational PR of the telemetry-egress-passthrough feature and the highest-quality one: the model, precedence (`host+port > host > port`, ties to slice order so user rules beat built-ins), host globbing / `:port` normalization, JSON scalar shorthand, and the `off`/`recordOne`/`queryKeys` semantics are all directly unit-tested and read correctly. `off` short-circuits *before* the legacy `isTelemetryEgress` fallback, `NormalizeMode` preserves `off`, empty `queryKeys` stays query-agnostic, and `mockQueryMatches` correctly stops a `connect` request from being served a `metric_data` mock.

**Findings**
- _should-fix_: `ptRecorder` dedup state has no app/session scope and is never reset — safe for the process-per-session sidecar, but in a long-lived multi-app DaemonSet agent it cross-suppresses captures (see inline).
- _nit_: `isTelemetryEgress`/`telemetryEgressPaths` are now a strict subset of the built-in defaults, so that fallback is effectively dead code — worth a comment or later removal.

**Recommendation:** approve on its own merits; address `ptRecorder` scoping before relying on `recordOne` in a multi-app DaemonSet. See the feature-level note (posted separately) for the cross-PR picture.

ayush3160:
## Feature-level review — telemetry-egress passthrough (cross-PR)

Reviewing the four PRs as one feature (keploy#4423 · enterprise#2365 · k8s-proxy#740 · enterprise-ui#1664).

**Contract seam.** The `PassThroughRule {port, host, path, mode, queryKeys}` shape is consistent across producer/consumer for four of five fields — it flows keploy `models.PassThroughRule` → k8s-proxy wire (`snake_case`) → dsv1 CRD → UI, `off` is honored in every converter (keploy decision, `convertPassThroughRules`, dsv1 `Mode` string, UI toggle, OpenAPI enum), and `queryKeys` is copied in every server-side converter. The one seam break was the UI's outbound `queryKeys` vs the wire's `query_keys` (silent data-loss on user-entered query-mux rules) — **now fixed in enterprise-ui `226a483a`**.

**`recordOne` lifecycle in multi-tenant DaemonSet (the theme worth tracking).** keploy's record-side `ptRecorder` dedup has no app/session identity and is never reset, and enterprise#2365's per-app work stops at per-app `opts` + the kernel `skip` map — it never gives each app its own recorder. So in a long-lived multi-tenant DaemonSet agent, two apps (or two sessions) hitting the same telemetry endpoint under `recordOne` collide: the second gets no mock and replays a synthetic 200. Combined with the missing `EgressSkipRemoveCgroup` cleanup on recycled cgroups, both DaemonSet gaps point the same way — the per-app/per-session lifecycle is fully handled for `skip`-by-port but under-handled for `recordOne` and for kernel-map teardown. Neither blocks the sidecar path.

**Ordering / compat.** Coherent but gated: keploy#4423 → enterprise#2365 → **bump keploy/enterprise go.mod pins** → k8s-proxy#740 → enterprise-ui#1664. All three downstream repos pin keploy `v3.6.11` (lacks `TelemetryDefaults`/`PassThroughOff`/`QueryKeys`) and currently build only against uncommitted local `replace` directives — nothing downstream compiles against a released keploy until #4423 is tagged and the pins bumped. Mock format is unchanged (`recordOne` mocks are ordinary HTTP mocks tagged `type:config`), so on-disk backward-compat holds.

**Overall:** no blocking defects. The one user-visible bug (UI `query_keys`) is fixed; the remaining should-fix items are the enterprise DaemonSet lifecycle gaps (stale kernel-skip on recycled cgroups + un-scoped `recordOne` dedup), both DaemonSet-only.

ayush3160 (commented):
**Principal re-review (round 2).** All round-1 findings verified resolved. `PassThroughScope` is threaded at both record callers (`http.go`, `recordv2.go`), empty scope preserves the classic sidecar key, and replay correctly needs no scoping (per-session `mockDb`). No remaining or new findings — **approve**. Merge gate unchanged: bump the downstream keploy pin once this tags.

charankamarapu (commented):
## Principal review — configurable telemetry-egress passthrough

Reviewed together with enterprise#2365, k8s-proxy#740, enterprise-ui#1664 and integrations#256. The direction is right: the hardcoded two-path bypass needed to become a rule set, and `recordOne` + `QueryKeys` is a genuinely good insight for handshake-parsing collectors like New Relic. The model file is clean and well-commented.

That said, I don't think this should merge as-is. Two of the blockers below are verified regressions in *user* traffic handling, not telemetry, and one makes the documented config syntax fail the CLI outright.

I checked everything below against the branch head (`bd192f98`) by extracting `pkg/models/passthrough.go` into a scratch module and running it, and by testing the config path through viper. Actual output is quoted.

---

### 🔴 Blocking

**1. The built-ins changed from `exact path + POST` to `prefix + any method`, and now swallow ordinary app routes.**

`isTelemetryEgress` was an exact map lookup gated on `POST` (`match.go:1497`). `passThroughScore` uses `strings.HasPrefix(reqPath, r.Path)` with no method constraint at all. The NOTE added at `match.go:1491` calls the legacy paths "a strict subset of BuiltinTelemetryDefaults" — the *path strings* are a subset, but the *matching semantics* are a strict superset, and five brand-new path prefixes come along with it.

Run against `MergePassThroughDefaults(nil, nil)`:

```
GET  /ingestData        -> ok=true mode="skip"  (matched builtin path="/ingest")
GET  /ingest/status     -> ok=true mode="skip"  (matched builtin path="/ingest")
POST /v1/tracesearch    -> ok=true mode="skip"  (matched builtin path="/v1/traces")
POST /api/v1/writeReview-> ok=true mode="skip"  (matched builtin path="/api/v1/write")
GET  /v1/logsExport     -> ok=true mode="skip"  (matched builtin path="/v1/logs")
```

Any dependency with a route under `/ingest*`, `/v1/traces*`, `/v1/metrics*`, `/v1/logs*`, `/v0.4/traces*`, `/v0.7/traces*`, `/api/v1/write*` is now, with zero user config: not recorded, and answered with a synthetic empty `200` on replay. No warning, no error — the user just sees their app behave differently under replay than under record. `/ingest` and `/api/v1/write` are not exotic names.

Fix: match on segment boundaries by default (`p == r.Path || strings.HasPrefix(p, r.Path+"/")`) with an explicit opt-in for the one rule that genuinely wants a prefix (`/opentelemetry.proto.collector.`), and restore a method constraint on the built-ins — every one of them is POST-only.

**2. The documented scalar shorthand doesn't just fail to parse — it stops the CLI from starting.**

`keploy.yml` is read by viper and decoded with **mapstructure** (`cli/provider/cmd.go:617-667`). mapstructure never calls `UnmarshalJSON`, and viper's default decode hooks don't include a text/JSON unmarshaler. So the shorthand documented in the doc comment on `PassThroughRule` — verbatim `passThroughPorts: [8126, {"port":4317,"mode":"skip"}]`, which is YAML — does this:

```
scalar shorthand (int)   UNMARSHAL ERR: 'record.passThroughPorts[0]' expected a map or struct, got "int"
scalar shorthand (str)   UNMARSHAL ERR: 'record.passThroughPorts[0]' expected a map or struct, got "string"
mixed                    UNMARSHAL ERR: 'record.passThroughPorts[0]' expected a map or struct, got "int"
object form              OK ports=[{Port:8126 Mode:skip}]
```

and `cmd.go:667` turns that into `return errors.New("failed to unmarshal the config")` — so a user who follows the documentation can't run keploy at all until they edit the file back. `TestPassThroughRule_UnmarshalJSON_Shorthand` passes because it goes through `encoding/json`, which is not the path the config file takes.

Either register a `mapstructure.DecodeHookFunc` for the scalar form (plus a test that goes through `viper.Unmarshal`, not `json.Unmarshal`), or drop the shorthand from the doc comment and keep the object form only. Right now the comment promises a syntax that bricks the CLI.

**3. `off` escapes `MatchPassThrough` as `matched=true`, and each caller has to remember to filter it. One already doesn't.**

`MatchPassThrough` returns `(rule{Mode:off}, true)`; it's `passThroughEgressDecision` that maps that to "not passthrough". integrations#256's `passThroughDecisionGrpc` returns `rule.Mode, true` with no `off` handling, and its caller in `grpcV2/mock.go` does:

```go
if mode == models.PassThroughRecordOne { one = findOneGrpcMock(...) }
if one == nil { return s.serveSyntheticGrpcSuccess(stream) }
```

So on the gRPC replay path `mode:"off"` — the mode whose entire purpose is "record and replay this normally" — falls into the else and synthesizes `grpc-status:0`. That is the exact inverse of the intent, and it's the first consumer written against this API.

This is an API-shape problem, not just a bug in the other PR: three parsers each re-deriving "does matched mean passthrough" will keep going wrong. Please move the filtering into `models` — either have `MatchPassThrough` not report `off` as a match, or add `ResolvePassThrough(rules, host, port, path) (mode PassThroughMode, passthrough bool)` as the single supported entry point and make the per-parser wrappers trivial.

---

### 🟠 Important

**4. A user rule does not reliably beat a built-in — specificity is compared before user-vs-builtin.**

`MergePassThroughDefaults`'s doc says defaults are "Merged UNDER user rules so a user rule for the same target wins", but ordering only breaks *ties*: `score > bestScore` is strict, and a built-in host rule (score 2) outranks a user path or port rule (score 1). Verified:

```
user {path:"/v0.4/traces", mode:off} @ dd-agent:8126         -> mode="off"   ✅
user {path:"/v0.4/traces", mode:off} @ trace.agent.datadoghq.com:443
                                                              -> mode="skip"  ❌ builtin *.datadoghq.com wins
user {port:443, mode:off} @ o123.ingest.sentry.io:443         -> mode="skip"  ❌ builtin *.sentry.io wins
```

This is what enterprise-ui#1664's per-default off-toggle is built on, so the switch can flip, the config can save, and the endpoint can still be skipped — with nothing anywhere to say so. Make it two-tier: any matching user rule wins over every built-in, and specificity only orders within a tier. Separately, fold `Path` into the score — today `{host+path}` and `{host}` both score 2, so the New Relic built-in is no more specific than a bare host rule.

**5. An unrecognized mode fails *open* into passthrough.**

`NormalizeMode` maps anything unknown to `recordOne`:

```
mode:"of"   -> "recordOne"
mode:"OFF"  -> "recordOne"
mode:"Skip" -> "recordOne"
```

A user typing `mode: OFF` (or `Skip`) in `keploy.yml` gets the opposite of what they asked for: the endpoint is passed through instead of recorded normally. k8s-proxy#740 rejects bad modes in `ApplyRecordConfig`, but `keploy.yml` and the DaemonSet ConfigMap path have no such gate. Default unknown values to *no passthrough* and log a warning, and validate at config load rather than silently coercing in the hot path.

**6. Replay-side `skip` shadows mocks that existing test sets already recorded.**

Before this PR only `POST /v1/traces` and `POST /ingest` were bypassed. `/v1/metrics`, `/v1/logs`, `/v0.4/traces`, `/v0.7/traces`, `/api/v1/write`, `*.datadoghq.com`, `*.sentry.io`, App Insights and `*.newrelic.com` are all new — and for `skip` the decode path never consults `mockDb` at all. So an existing test set that legitimately recorded one of those endpoints now gets a synthetic empty `200` on replay instead of its recorded response, on unchanged test data. The PR body's "Backward-compatible" claim is true of the mock *format*, not of replay behaviour.

Cheapest fix: for `skip`, serve a recorded mock when one exists (non-consuming, same as `recordOne`) and only synthesize when there is nothing. That keeps the OOM fix (no fuzzy match either way) without changing what old test sets replay.

**7. The decision rebuilds the whole default set on every outgoing request, on both record and replay.**

`passThroughEgressDecision` calls `MergePassThroughDefaults` → `BuiltinTelemetryDefaults` → `TelemetryDefaults`, which constructs the 12-element literal plus a merged slice, per request, *matched or not* — so every ordinary dependency call in the app pays it too. Benchmarked (2 user rules, non-matching request):

```
BenchmarkDecisionPerRequest-8   350952   3346 ns/op   5520 B/op   5 allocs/op
```

3.3 µs and 5.5 KB of garbage per egress call, in a change whose stated purpose is relieving agent heap pressure. Build the defaults once (package var / `sync.Once`) and precompute the merged slice into `OutgoingOptions` at setup instead of per request. `matchHost` also re-runs `ToLower`/`TrimSpace`/`normalizeHost` on the host for every rule in the list.

**8. The replay lookup is host/port-blind while the record key isn't.**

`ptRecordKey` includes host and port; `serveOnePassThroughMock` filters on method + path (+ `queryKeys`) only. So two collectors reachable on the same path cross-serve each other's recorded response, and a `recordOne` request can be served a mock recorded from a completely different (non-passthrough) upstream that happens to share method+path. `findOneGrpcMock` in integrations#256 has the same gap (`:path` only). Compare the mock's host/port too — you already store them in `HTTPReq.URL`.

**9. `ptRecorder` state is unbounded and is marked before the mock is persisted.**

`seen` is never reset or bounded. Now that `PassThroughScope` is part of the key (and the enterprise DaemonSet sets it per app/session), a long-lived node agent accumulates one entry per (scope × endpoint) for the lifetime of the process. Also `shouldRecord` records its decision *before* the mock is actually built and emitted — any later `return nil` on that path burns the one allowed representative and leaves the endpoint with zero mocks, silently downgrading `recordOne` to a synthetic 200 at replay. Mark on successful emit, not on decision.

---

### 🟡 Minor

- **`passThroughPorts` vs `passThroughHosts` is a distinction without a difference.** Same type, and `MergePassThroughDefaults` just concatenates them — a host rule works fine in the Ports list and vice versa. Consider a single `record.passThrough` list, or at least validate the shape per key so the names mean something.
- **Rules live under `record:` but are read on the test path** (`replay.go:1204/1213`, `runner.go:387` all read `r.config.Record.PassThroughPorts`). A `keploy test`-only config with no `record:` block silently gets built-ins only. Worth a top-level key or an explicit note.
- **Apex hosts aren't covered.** `path.Match("*.datadoghq.com", "datadoghq.com")` is false — verified `datadoghq.com` and `sentry.io` both resolve to `ok=false`. If the intent is "this domain and its subdomains", handle the apex explicitly.
- **`normalizeHost("::1")` returns `":"`** (verified) — the all-digits port heuristic eats the last group of a bracket-less IPv6 authority. `[::1]:4317` is handled correctly; only the malformed form breaks, so this is cosmetic, but it's cheap to guard.
- `passThroughRecordDecision` is a pure alias for `passThroughEgressDecision`. Fine as documentation of intent, but the two sides genuinely differ (record has no `mockDb`, replay has no status code) — if they're going to stay identical, drop one; if #3 is fixed, this is where the shared resolver goes.
- The `T10` / `T7` / `T8` work-item references in the comments (`passthrough_egress.go:224`, `instrument.go`, `http.go`) won't mean anything to a reader six months from now — either resolve them (T10 looks done) or spell out what's deferred.

---

### Testing

Unit coverage of the model is good. The gaps that matter are exactly where the blockers are:

- a case asserting an app route like `/ingestData` or `/api/v1/writeReview` is **not** matched (#1),
- a config round-trip through `viper.Unmarshal`, not `json.Unmarshal` (#2),
- an `off` rule that has to beat a built-in *host* rule, not just a same-specificity one (#4),
- a replay case where a recorded mock exists for a `skip`-mode endpoint (#6).

The `passThroughEgressDecision` tests build `&url.URL{Path:..., Host:...}` directly, which skips the `request.Host` → `url.Host` fallback the real callers use — worth one test through the actual host-resolution order.

Happy to pair on the precedence redesign (#3 + #4) — I think that's the piece worth getting right before any of the four downstream PRs merge, since the UI toggle and the gRPC path both depend on it.

charankamarapu (commented):
## Re-review — round 2 (`8d2d3110` + `666c9659`)

Re-ran my harness against the new `pkg/models/passthrough.go` rather than reading the diff. The rework is substantial and the three blockers are genuinely fixed:

```
GET  /ingestData          -> passthrough=false     (was: skip)
POST /ingestData          -> passthrough=false     (was: skip)
POST /v1/tracesearch      -> passthrough=false     (was: skip)
POST /api/v1/writeReview  -> passthrough=false     (was: skip)
POST /ingest              -> passthrough=true skip (real Pyroscope, still works)
GET  /ingest              -> passthrough=false     (method gate works)
POST /opentelemetry.proto.collector.trace.v1.TraceService/Export -> skip (PathPrefix works)

off /v0.4/traces @ *.datadoghq.com:443 -> passthrough=false  (was: skip — two-tier works)
user off port 443 vs builtin *.sentry.io -> passthrough=false (was: skip)

NormalizeMode: "of"->off  "OFF"->off  "Skip"->off  ""->recordOne   (fails closed)
apex datadoghq.com / sentry.io -> now matched
```

`ResolvePassThrough` as the single off-aware entry point is the right shape, deferring the dedup mark to emit time is exactly right, `builtinTelemetryDefaultsCache` kills the per-request garbage, the host/port gate on `serveOnePassThroughMock` closes the record/replay asymmetry, and serving a recorded mock under `skip` (not just `recordOne`) resolves the backward-compat concern cleanly. Deleting the legacy fallback rather than keeping it as belt-and-braces was the correct call.

Ten things below, one blocking.

---

### 🔴 Blocking

**N1. The signature change breaks integrations#256, which is unmodified since 08-07.**

`MatchPassThrough` went from 4 params to 5 (`method` added). `pkg/grpcV2/passthrough_egress.go` still calls the old form, twice:

```go
rules := models.MergePassThroughDefaults(opts.PassThroughPorts, opts.PassThroughHosts)
if rule, ok := models.MatchPassThrough(rules, authority, port, path); ok {
```

That won't compile once integrations picks up this keploy. And when it's fixed it must move to `ResolvePassThrough`, not just gain an argument — the `off` inversion I raised last round is still live there (`mode:"off"` → `matched=true` → falls through to `serveSyntheticGrpcSuccess`, the exact opposite of "record normally"). integrations#256 is listed in this PR's own related-PRs block; the two need to land together.

---

### 🟠 Important

**N2. My IPv6 suggestion caused a regression — bracketed authorities no longer get their port stripped.**

The new guard is `isAllDigits(h[i+1:]) && !strings.Contains(h[:i], ":")`. For `[::1]:4317`, `h[:i]` is `[::1]`, which contains colons, so the strip is skipped:

```
normalizeHost("::1")        = "::1"          ✅ (fixed)
normalizeHost("[::1]:4317") = "[::1]:4317"   ❌ (was "[::1]")
normalizeHost("h:443")      = "h"            ✅
```

So a `host: "[::1]"` rule no longer matches the authority `[::1]:4317`. My fault for suggesting the narrow guard — the robust version is `net.SplitHostPort` with a fall-back to the input on error, which gets bare IPv6, bracketed IPv6, and host:port all right without a heuristic. Worth a table test over those three plus a bare hostname.

**N3. `Method` and `PathPrefix` exist only in keploy — they don't reach the control plane, the CRD, or the UI. One built-in is now un-disableable as a result.**

Both new fields are absent from k8s-proxy `models.PassThroughRule`, from `dsv1.PassThroughRule`, from enterprise's `conv` mapping, and from the UI's `PassThroughPortRule`/`PassThroughHostRule`. Two consequences:

- `/record/telemetry-defaults` serves a *lossy projection* of the defaults — the endpoint whose whole purpose (k8s-proxy#740) is proving the UI isn't drifting from the agent now drops two fields of every rule it reports.
- The UI's off-toggle builds its override from `(port, host, path)` only. I ran every built-in through `ResolvePassThrough` with the override the UI would emit:

```
OpenTelemetry  OTLP/HTTP traces   builtinApplies=true  afterUIOff=false  ok
Pyroscope      profiler ingest    builtinApplies=true  afterUIOff=false  ok
Datadog        trace agent v0.4   builtinApplies=true  afterUIOff=false  ok
Prometheus     remote_write       builtinApplies=true  afterUIOff=false  ok
OpenTelemetry  OTLP/gRPC          builtinApplies=true  afterUIOff=true   *** does not turn off ***
New Relic      RPM collector      builtinApplies=true  afterUIOff=false  ok
Datadog        SaaS intake        builtinApplies=true  afterUIOff=false  ok
Sentry         SaaS intake        builtinApplies=true  afterUIOff=false  ok
```

11 of 12 work. OTLP/gRPC doesn't, because its rule is `PathPrefix:true` and the UI's override is `PathPrefix:false` — segment matching means the override never matches the real gRPC `:path`, tier 1 misses, and the built-in still applies. The switch flips and nothing changes.

Either propagate both fields through the three mirrors, or make the override mechanism target-identity-based (a stable rule ID from the defaults list) instead of reconstructing the rule field-by-field. The second is less fragile — it stops every new field from becoming a new way for the toggle to miss.

**N4. The `ptRecorder` cap fails toward unbounded recording, in exactly the scenario it exists for.**

```go
if len(p.seen) >= ptRecorderMaxKeys {
	return true
}
```

At the cap a new key isn't inserted — so it returns `true` on *every* subsequent call for that key too. Past 8192 keys, a high-volume telemetry endpoint records every single exchange, which is the mock-bloat this feature was built to prevent. The comment ("fail toward capturing, not toward silently dropping") is the right instinct for a *dependency*, but a `recordOne` passthrough target is the one case where dropping is the whole point. Evict instead (random/LRU is fine at this size), or track over-cap keys in a small fixed fallback so they still collapse to one. At minimum log once when the cap is hit — silent degradation into the original failure mode is the worst outcome.

**N5. The `type:config` tag on a `recordOne` mock is inert in-process — the mock is stamped `LifetimePerTest` with `LifetimeDerived: true`.**

`passthrough_egress.go` documents the contract as:

> a recordOne mock is tagged type:config (LifetimeSession) at record time so it lives in the session pool and is never window-filtered

But the emitted mock carries `TestModeInfo{Lifetime: LifetimePerTest, LifetimeDerived: true}` (the pre-existing unconditional HTTP stamp), and `DeriveLifetime` short-circuits on `LifetimeDerived` at `lifetime.go:126` — *before* reaching `case "config": m.TestModeInfo.Lifetime = LifetimeSession`. So in memory the mock is per-test and window-filtered, not session-scoped.

To be precise about impact: the disk path routes by the metadata tag (`mockdb/db.go:470`), so the mock still lands in the config file and a replay that loads it fresh does derive `LifetimeSession` — which is why this doesn't show up end-to-end. But the two mechanisms disagree, the comment states a guarantee the code doesn't provide in-process, and anyone who later trusts `TestModeInfo.Lifetime` gets the wrong answer. Make them agree: stamp `LifetimeSession` when you set `meta["type"]="config"`, and reword the comment to say where the guarantee actually comes from.

---

### 🟡 Minor

- **N6. A typo'd user rule now suppresses the built-in it shadows.** Fail-closed on the mode means fail-open on the protection — verified: `{host:"*.datadoghq.com", mode:"skipp"}` yields `passthrough=false`, where no user rule at all yields `skip`. So a typo turns Datadog telemetry back into recorded dependencies, i.e. the original OOM. Still the safer direction (nothing is silently dropped or faked), but it makes config-load validation load-bearing — `NormalizeMode`'s own comment says "Config load should validate and warn on unknown modes", and there's no such validation on the `keploy.yml` path. Worth adding.
- **N7. `UnmarshalJSON`'s doc comment still contradicts the struct doc.** The struct now says, correctly, "Do not document the scalar form for keploy.yml" — but 30 lines below, `UnmarshalJSON` still says *"This lets `passThroughPorts: [8126, {"port":4317,"mode":"skip"}]` parse element-wise"*, which is the YAML-shaped example that started this. Delete or rewrite it as JSON.
- **N8. The method gate is skipped when the caller doesn't know the method.** `r.Method != "" && method != ""` means `method==""` matches POST-only built-ins — verified. Harmless today (the one caller passing `""` also passes `path==""`, so no path built-in can match, and host built-ins need a host), but it's an implicit coupling between two callers. A comment on `passThroughScore` saying why the empty-method escape is safe would keep it from being load-bearing by accident.
- **N9. `MatchPassThrough` and `MergePassThroughDefaults` are still exported with the untiered, off-leaking behaviour** that produced the gRPC bug. Their doc comments now warn against using them for decisions, but a comment is what integrations#256 already had available. Consider unexporting `MatchPassThrough`, or renaming to `matchPassThroughWithinTier` / `MergePassThroughForDisplay` so the wrong one is hard to reach for.
- **N10. Dead branch in `hostOnly`.** `splitHostPortLenient` returns a nil error on every path, so the `err == nil` check always passes and the `return h` fallback is unreachable. Either drop the error return or make it real.
- `passThroughRecordDecision` is now a bare `return passThroughEgressDecision(...)`. Both sides genuinely delegate to `ResolvePassThrough`; the indirection no longer earns its keep.

---

Good round. #1/#3/#4/#5/#6/#7/#8/#9(partial)/#12 from the last pass all verify clean. N1 is the merge-order blocker; N3 is the one I'd most want settled by design (rule identity) rather than by adding two more fields to four mirrors.

charankamarapu (commented):
## Re-review — round 3 (`2a8c8098`)

The stable-rule-ID approach is the right call and it works. I replicated enterprise-ui's `disableMarkerFor` payload and ran every built-in through the new `ResolvePassThrough`:

```
OTLP/HTTP traces    id=otlp-http-traces        applies=true  afterDisable=false ok
OTLP/HTTP metrics   id=otlp-http-metrics       applies=true  afterDisable=false ok
OTLP/HTTP logs      id=otlp-http-logs          applies=true  afterDisable=false ok
profiler ingest     id=pyroscope-ingest        applies=true  afterDisable=false ok
trace agent v0.4    id=datadog-v04-traces      applies=true  afterDisable=false ok
trace agent v0.7    id=datadog-v07-traces      applies=true  afterDisable=false ok
remote_write        id=prometheus-remote-write applies=true  afterDisable=false ok
OTLP/gRPC           id=otlp-grpc               applies=true  afterDisable=false ok   <-- was broken
RPM collector       id=newrelic-rpm            applies=true  afterDisable=false ok
Datadog SaaS        id=datadog-saas            applies=true  afterDisable=false ok
Sentry SaaS         id=sentry-saas             applies=true  afterDisable=false ok
Azure App Insights  id=azure-appinsights       applies=true  afterDisable=false ok
```

12/12, and it's properly scoped — disabling `otlp-grpc` doesn't disable `otlp-http-traces`, and an unknown ID is inert. Also verified fixed: `normalizeHost` via `net.SplitHostPort` (`::1`→`::1`, `[::1]:4317`→`::1`, `host:443`→`host`), the `ptRecorder` cap now evicts and warns once instead of failing toward unbounded recording, `ptLifetime = LifetimeSession` when `ptRecordOne` (with the right explanation of why it's keyed off the flag and not the tag), the `UnmarshalJSON` doc no longer shows a YAML example, `splitHostPortLenient`'s dead branch is gone, and `TelemetryDefaultsVersion` is bumped to 3.

Two things worth another pass, neither blocking.

---

### 🟠 Important

**R3-1. The marker split re-adds per-request allocations, partly undoing round 2's caching fix.**

`ResolvePassThrough` now does, on every egress call: `append(append([]PassThroughRule{}, userPorts...), userHosts...)` (a throwaway copy purely to iterate), a `user` slice, a `disabled` map when any marker is present, and a `filtered` copy of the 12 built-ins. Same inputs, round 2 vs round 3:

```
round 2 (2 user rules):                 1276 ns/op    208 B/op    1 allocs/op
round 3 (2 user rules):                 1241 ns/op    672 B/op    8 allocs/op
round 3 (disable marker + 1 rule):      1611 ns/op   2464 B/op   11 allocs/op
```

3× the bytes and 8× the allocations for the same work — and 2.4 KB per request for any user who has toggled one default off, which is the case the ID mechanism exists to serve. Round 2 got this to 1 alloc by materialising the built-ins once; the same trick applies here. The split is a pure function of the config, so compute it where `OutgoingOptions` is built: store the pre-filtered user rules and the resolved built-in slice (already minus disabled IDs) on the options, and let the hot path do two `MatchPassThrough` calls over ready-made slices. That also removes the need to re-derive `disabled` identically in every parser.

**R3-2. The marker is shape-sensitive and fails silently when a client sends the ID alongside the target.**

```go
if r.ID != "" && r.Host == "" && r.Port == 0 && r.Path == "" && r.NormalizeMode() == PassThroughOff {
```

All four conditions are required. The obvious client behaviour — echo back the rule you were served and set `mode:"off"` — produces `{id:"otlp-grpc", path:"/opentelemetry.proto.collector.", mode:"off"}`, which is *not* recognised as a marker. It falls into the user tier, where segment matching means it doesn't match a real gRPC `:path`, tier 1 misses, and the built-in applies: precisely the round-2 bug, reachable again through a natural client implementation. The UI avoids it only because `disableMarkerFor` deliberately zeroes the target fields.

Make the ID authoritative whenever it's present with `mode:"off"` — ignore the target fields rather than requiring them empty — or add an explicit `disable: true` so intent isn't encoded in the *absence* of fields. Relatedly: an ID that matches no built-in is silently inert (verified). That's the correct runtime behaviour, but it's also how a stale ID after a `TelemetryDefaultsVersion` bump disappears without a trace — worth a warn at config load listing unrecognised IDs, which pairs with the UI's orphaned-overrides panel.

---

### 🟡 Minor (carried)

- **Unknown modes still aren't validated on the `keploy.yml` path.** `NormalizeMode`'s own comment says "Config load should validate and warn on unknown modes", and nothing does. Because it fails closed, a typo'd user rule still *suppresses* the built-in it shadows (`{host:"*.datadoghq.com", mode:"skipp"}` → not passthrough, where no rule at all → `skip`), so a one-character error silently restores the OOM this feature prevents. k8s-proxy validates its path; the CLI doesn't.
- **`MatchPassThrough` is still exported**, and the case for unexporting it got stronger: the ID filtering now lives *only* in `ResolvePassThrough`, so a caller reaching for `MatchPassThrough` silently loses disable-by-ID on top of losing tiering and off-awareness. integrations#256 already made this mistake once with the previous version of the same footgun.
- `passThroughRecordDecision` is still a one-line pass-through to `passThroughEgressDecision`.

---

### Cross-repo

integrations#256 (`a440a5dd`) now routes both the gRPC and h2 paths through `ResolvePassThrough`, which fixes the `off` inversion and the compile break — good. One leftover there: after `ResolvePassThrough` returns false (including the `off` case) it still falls through to the legacy `opts.Rules` loop, which can return `PassThroughSkip` for the same target. That's the pattern this PR deliberately removed from its own path — *"There is deliberately no separate legacy fallback: one existed and re-skipped those paths even when the user set mode:'off'"* — so a user with both a legacy bypass entry and a new `off` override still gets the endpoint skipped. Narrower trigger than before, same class. Worth mirroring the deletion.

charankamarapu (approved):
## Approving ✅

Three rounds in, the substance is settled. Everything I raised as blocking is fixed and verified against the branch rather than the diff:

- built-in path rules no longer swallow app routes (`/ingestData`, `/v1/tracesearch`, `/api/v1/writeReview` all clear; real `POST /ingest` still skips)
- `ResolvePassThrough` is a single off-aware, two-tier entry point — a user rule can't be outranked by a more-specific built-in
- unknown modes fail closed
- the documented config syntax no longer breaks the CLI
- `skip` no longer shadows mocks existing test sets legitimately recorded
- stable rule IDs make disable-by-identity work for **12/12** built-ins, including the `pathPrefix`-gated OTLP/gRPC one
- `normalizeHost` is correct for bare IPv6, bracketed IPv6, and host:port
- the `ptRecorder` cap evicts instead of failing toward unbounded recording
- `type:config` and `TestModeInfo.Lifetime` now agree

Merging with the two round-3 notes as follow-ups, not conditions — neither produces wrong behaviour today:

1. **`ResolvePassThrough` allocations** (`208 B/1 alloc` → `672 B/8 allocs`, or `2464 B/11 allocs` with a disable marker). The split is a pure function of config; precompute it where `OutgoingOptions` is built.
2. **The disable marker is shape-sensitive** — it only registers when `id` arrives with host/port/path all empty. Correct in the UI today, fragile as a cross-repo contract. Making `id` authoritative whenever present with `mode:"off"` would close it.

Also still worth doing at some point: validate unknown modes at `keploy.yml` load (a typo still suppresses the built-in it shadows), and unexport `MatchPassThrough` now that ID filtering lives only in `ResolvePassThrough`.

**Merge order:** this lands and tags first — enterprise#2365 and k8s-proxy#740 both pin it at a pseudo-version off this branch (`666c9659`) and need repinning afterwards. integrations#256 should go with it.

Thanks for the turnaround on these — three substantive rounds across five repos in a day, and the fixes were the right fixes rather than the minimum ones.

PR #4424, opened 2026-08-10
feat(agent): allow a non-kernel Hooks implementation via HooksOverride

Aditya-eddy:
## What

Adds `agent.HooksOverride` — a single seam letting a downstream build replace the eBPF-backed `Hooks` with a userspace implementation.

Three files, ~28 lines. Default behaviour is unchanged: with the override nil, the composition root builds exactly the hooks it always did.

## Why

Some customers will not permit kernel-level network access, so eBPF is off the table and Keploy currently cannot record or replay there at all.

`agent.Hooks` turns out to be the **entire** kernel-facing contract — four methods:

```go
type Hooks interface {
	DestInfo   // Get(srcPort), Delete(srcPort)
	Load(ctx, cfg, setupOpts) error
	WatchBindEvents(ctx) (<-chan models.IngressEvent, error)
}
```

So a type that answers those four replaces the datapath wholesale. Enterprise uses this for a JVM-fed implementation: an in-process javaagent redirects the app's `connect()` into Keploy's proxy and reports the real destination over a userspace channel, replacing `cgroup/connect4` + `redirectProxyMap` for egress and `cgroup/post_bind4` for ingress.

Nothing downstream of the proxy changes — parsers, recorder, replay engine and mock format are untouched. That is what keeps the recordings **byte-identical** to eBPF ones rather than merely similar.

## Verification

Exercised end to end by the enterprise implementation on a Spring Boot + JDBC app against real MySQL, with no eBPF loaded:

- **Native:** record → 10 test cases / 94 MySQL mocks; replay **10/10 passing with the database stopped**
- **Sidecar (k8s, webhook-injected):** record → **10 test cases / 94 MySQL mocks** — identical counts, reached via a completely different injection path

## Notes for review

- `cli/provider/agent_service.go` needs the `coreAgent` import alias: that file already binds `agent` to `pkg/service/agent`, a different package.
- A partial implementation degrades silently rather than loudly — a `Get` that never resolves makes the proxy close every outgoing connection, and a `WatchBindEvents` that never fires records zero test cases while still producing mocks. The doc comment calls this out.


praagyajain:
### Review — minor findings

Verified against the branch head. One minor issue; the seam itself looks good, and the single-build refactor fixes a genuine double-registration (`agent.New` has registration side effects and was being called twice per boot).

**Minor — override log fires before the nil-fallback** — `cli/provider/agent_service.go:45`

The `logger.Info("using an overridden (non-eBPF) hooks implementation", ...)` sits inside the `HooksFactory != nil` block (lines 40–47), above the fallback at 48–50:

```go
if coreAgent.HooksFactory != nil {
	h = coreAgent.HooksFactory(logger, cfg)
	logger.Info("using an overridden (non-eBPF) hooks implementation", ...)  // fires here
}
if h == nil {
	h = hooks.New(logger, cfg)   // ...but we may end up here
}
```

A factory that returns `nil` — which this code explicitly tolerates via that fallback — logs that a non-eBPF datapath is in use and then constructs the default eBPF hooks. That inverts the exact triage signal the comment at 42–44 says the line exists to provide (`Without this line triage starts from "eBPF is broken"`).

`%T` on a nil interface prints `<nil>`, so it's recoverable if you look closely, but the headline still points the wrong way. Suggested:

```go
var h coreAgent.Hooks
if coreAgent.HooksFactory != nil {
	h = coreAgent.HooksFactory(logger, cfg)
}
if h == nil {
	h = hooks.New(logger, cfg)
} else {
	logger.Info("using an overridden (non-eBPF) hooks implementation",
		zap.String("impl", fmt.Sprintf("%T", h)))
}
```

praagyajain:

Aditya-eddy:
Review of the OSS seam (read alongside enterprise#2414 and k8s-proxy#803). Nothing was compiled or run here, so these are from reading the diff against the surrounding code.

The seam design is good and the default path is genuinely untouched. Two process points first, then the findings.

## Scope

The description covers only the first commit — "Three files, ~28 lines", `HooksOverride`. The branch now carries three feature commits:

```
feat(agent): allow a non-kernel Hooks implementation via HooksOverride
refactor(agent): make the hooks seam a factory; guard the proxy against third-party impls
feat(agent): resolve JVM destinations by peer address for node-scoped hooks (#4460)
```

so the actual diff is 6 files, +255/−15, and also carries `DestInfoByPeer`, `SetPeerSessionResolver` / `GetSessionForPeer`, the `destInfo == nil` guard, and the `GetAgent` restructure. The Verification section covers native + sidecar jvm-redirect and says nothing about the node-scoped peer path that commit 3 exists for. Worth updating the body so a reviewer knows what they're approving.

Also: #4460 is still open against a different base with overlapping content. Close it or rebase it.

## 1. The `h == nil` fallback defeats the seam's stated purpose, and the log misleads

`cli/provider/agent_service.go:38-48`:

```go
if coreAgent.HooksFactory != nil {
    h = coreAgent.HooksFactory(logger, cfg)
    logger.Info("using an overridden (non-eBPF) hooks implementation", ...)
}
if h == nil {
    h = hooks.New(logger, cfg)
}
```

If a factory returns nil, this logs `using an overridden (non-eBPF) hooks implementation impl=<nil>` and then **silently installs the eBPF hooks anyway**. That is precisely the failure `agent_service_test.go`'s own doc calls out — "silently reverts the enterprise build to eBPF on clusters where it has no kernel access and therefore records nothing" — except now with a log line pointing triage in the wrong direction.

Suggest: if the factory is set and returns nil, return an error; move the Info log below the check.

Separately, `HooksFactory`'s doc justifies the factory form partly with *"a nil implementation pointer stored in an interface is non-nil and would install a Hooks that panics on first use."* True, but `h == nil` only catches an **untyped** nil — a factory returning `(*jvmhooks.JVMHooks)(nil)` sails straight through. The doc oversells the protection the code provides.

## 2. `rule == nil` leaks the connection, on a path this PR makes reachable

`pkg/agent/proxy/proxy.go`, immediately after the changed lines:

```go
rule := p.GetSessionForPeer(remoteAddr.IP)
if rule == nil {
    utils.LogError(p.logger, nil, "failed to fetch the session rule")
    return err        // err is nil here — Delete just succeeded
}
```

No `srcConn.Close()`, and a nil error returned to the accept loop, which does not close the conn either (`handleConnection` owns it). The two `destInfo` failure paths ~30 lines above handle this correctly: log, `srcConn.Close()`, `return nil`.

Pre-existing, but this PR makes it live. Under eBPF, `p.session` is effectively always set, so the branch was near-dead; with a peer resolver installed and no single session (node-scoped multi-app), any unmapped peer lands here and hangs the app's connection. Mirror the `destInfo` handling and return a real error instead of `err`.

## 3. The `HooksFactory` contract doc contradicts the proxy in the same PR

`pkg/agent/runtime_hooks.go:44-46`:

> `Get`: returning `(nil, nil)` is **NOT** permitted … the proxy closes such connections gracefully, but it **dereferences the result whenever err is nil**.

Commit 2 adds the guard whose comment says the opposite: *"(nil, nil) is a very natural Go shape … Treat it exactly like a lookup failure rather than dereferencing it."* The contract was written for commit 1 and commit 2 changed the behaviour it describes. Worth updating, or a future reader may delete the guard as contradicting the documented contract.

(The other three contract bullets check out — `HookCfg.Port` really is only set when `EbpfProxyPortOverride != 0`, at `pkg/service/agent/agent.go:579`.)

## 4. `TestGetAgent_UsesHooksFactory` doesn't test what its comment claims

Its doc says it guards against "reordering construction so the proxy gets the default hooks while only the agent gets the override." The test asserts only that the factory received the right logger/cfg and that `svc != nil` — it never checks which `Hooks` the returned service or its proxy holds, so precisely that refactor would still pass. Asserting `svc.(*agent.Agent).Hooks == want` (and ideally that the proxy's `DestInfo` is the same value) would close it.

## 5. Test nits

- `var _ = models.IngressEvent{}` exists solely to justify the `models` import. Drop both.
- `stubHooks struct{ coreAgent.Hooks }` embeds a nil interface, so any method call panics. Fine while nothing calls one — worth a line saying so.
- Both tests mutate the `coreAgent.HooksFactory` global; `TestGetAgent_NilFactoryUsesDefaultHooks` sets it without `t.Cleanup`, and neither is parallel-safe.
- CI-safety confirmed: `hooks.New` → `linux.NewHooks` is a pure struct literal with no kernel or bpffs access, so the default-path test needs no privileges.

## 6. The double-construction rationale overstates the impact

The comment says building `agent.New` twice "fired all three twice per boot." All three side effects — `SetAuxiliaryHook`, `RegisterIncomingProxy`, `SetConnectionIdleRetention` — are idempotent assignments, and the old code passed identical values both times (`instrumentation.Hooks == h`, `.Proxy == p`, `.IncomingProxy == ip`, per `agent.New`'s body). The real win is not allocating and discarding an `*Agent`, not a behaviour fix.

Also worth adding to the description: `cmd != "agent"` now short-circuits **before** construction. Strict improvement, but a behaviour change that isn't mentioned anywhere.

## From the peer-address commit

- **`SetPeerSessionResolver`'s doc claims a fallback that doesn't exist** (proxy.go:1056): "resolution falls back to the TGID resolver and then to the single session." `GetSessionForPeer` never consults `p.sessionResolver`. No regression — the `getSession()` it replaced ignored it too — but the comment should match.
- **Port-keyed TLS state wasn't given the same treatment.** `pTls.SrcPortToDstURL` / `ClaimSrcPort` / `ReleaseSrcPortIfOwner` are still keyed on `sourcePort` alone and collide across pods for exactly the reason this PR re-keys `DestInfo`. Latent while node-scoped mode is observer-only (per k8s-proxy#803), but it will bite the moment app traffic routes through the proxy node-scoped.

---

Net: #1 and #2 are what I'd want fixed before merge — both convert a loud failure into a silent one, which is the specific thing this PR argues against everywhere else. #3 and #4 are corrections to claims the PR makes about itself.

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in feat(agent): allow a non-kernel Hooks implementation via HooksOverride. Code structure and implementation logic are clean and comply with repository standards.

Issue #4426, opened 2026-08-12
MySQL FLOAT and DOUBLE values are recorded as their IEEE-754 bit pattern, not their value


SepulvedaTwain:
## Describe the bug

MySQL sends `FLOAT` and `DOUBLE` on the wire as a raw IEEE-754 bit pattern. Keploy's MySQL
parser reads those bytes with a **numeric conversion** rather than a reinterpret:

```go
// pkg/agent/proxy/integrations/mysql/wire/phase/query/rowscols/binaryProtocolRowPacket.go
res.value = float32(binary.LittleEndian.Uint32(data[:4]))   // FLOAT
res.value = float64(binary.LittleEndian.Uint64(data[:8]))   // DOUBLE
```

`float32(uint32(...))` compiles and is silently wrong — it converts the *integer value of the
bit pattern*. A column holding `9.99` is read as:

| Column type | Actual value | Keploy records |
|---|---|---|
| `FLOAT` | `9.99` | `1.0926057e+09` |
| `DOUBLE` | `9.99` | `4.621813488089437e+18` |

The same defect is at three other decode sites (`stmtExecutePacket.go` FLOAT param,
`queryPacket.go` FLOAT and DOUBLE query attributes).

**This corrupts at record time, not replay time.** The wrong number is written into
`mocks.yaml`, so it survives re-recording, and replay then asserts against a value the
database never returned. Every test touching a `FLOAT` or `DOUBLE` column fails on a body
diff that looks like a real regression.

## Steps to reproduce

1. Create a table with a `FLOAT` and a `DOUBLE` column, insert `9.99` into each.
2. `keploy record` an endpoint that selects those columns.
3. Read `keploy/test-set-N/mocks.yaml` — the recorded rows hold `1.0926057e+09` and
   `4.621813488089437e+18` instead of `9.99`.
4. `keploy test` — the test fails on a response-body diff even though the application and
   the database are unchanged.

## Expected behaviour

`FLOAT` and `DOUBLE` columns round-trip their actual value.

## Two follow-on defects on the same path

Fixing the decode alone is not enough to make a `FLOAT` column usable, because two things
downstream assume the broken types:

**1. `EncodeBinaryRow` panics instead of erroring.** It type-asserts nakedly:

```go
v := uint32(ce.Value.(int))
```

No serializer preserves Go's numeric types. `yaml.v3` resolves an untagged scalar by shape,
[...]

PR #4427, opened 2026-08-12
fix(mysql): decode FLOAT/DOUBLE as IEEE-754 bit patterns

SepulvedaTwain:

MySQL sends `FLOAT` and `DOUBLE` on the wire as a raw IEEE-754 bit pattern. Four decode sites
read those bytes with a **numeric conversion** instead of a reinterpret:

```go
res.value = float32(binary.LittleEndian.Uint32(data[:4]))   // FLOAT
res.value = float64(binary.LittleEndian.Uint64(data[:8]))   // DOUBLE
```

`float32(uint32(...))` compiles and is silently wrong — it converts the *integer value of the
bit pattern*. A column holding `9.99` is read as `1.0926057e+09` (FLOAT) and
`4.621813488089437e+18` (DOUBLE).

The result-set site is the damaging one: it corrupts at **record** time, so the wrong value is
written into `mocks.yaml` and survives re-recording. Replay then asserts against a value the
database never returned, and every test touching one of those columns fails on a body diff that
looks like a real regression.

**Fixed** with `math.Float32frombits` / `math.Float64frombits` at each site:

| File | Site |
|---|---|
| `wire/.../rowscols/binaryProtocolRowPacket.go` | result-set columns (FLOAT + DOUBLE) |
| `wire/.../preparedstmt/stmtExecutePacket.go` | FLOAT bound parameter |
| `wire/.../queryPacket.go` | FLOAT + DOUBLE query attributes |

Two follow-on defects on the same path are fixed alongside it, because the decode fix **alone
does not make a `FLOAT` column usable**:

**1. `EncodeBinaryRow` panicked instead of erroring.** It type-asserted `ce.Value` nakedly
(`ce.Value.(float32)`, `uint32(ce.Value.(int))`). No serializer preserves Go's numeric types —
`yaml.v3` resolves an untagged scalar by shape, so a `FLOAT` column comes back `float64`, an
integral `DOUBLE` comes back `int`, and a `BIGINT UNSIGNED` above `MaxInt64` comes back
`uint64`; `encoding/json` returns `float64` for every number. Once FLOAT values decoded
correctly, those assertions panicked — and a panic here tears down the proxy's connection
handler, so the application under replay sees `invalid connection` with nothing pointing at the
cause. Replaced with `coerceToInt64` / `coerceToFloat64`, which return a wrapped error naming
the column.

**2. `paramValueEqual` widened instead of narrowing.** A `FLOAT` param is `float32` off the wire
and `float64` out of the mock file. `float64(av) == bv` asks the `float32` to carry precision it
never had — `float64(float32(9.99))` is `9.989999771118164` — so a correctly recorded `FLOAT`
param never matched itself. Narrowed the two **mixed** arms to float32 precision.
`float64`/`float64` and `float32`/`float32` are untouched, so `DOUBLE` comparison keeps full
precision (pinned by a test).

**Blast radius:** `match.go` is the only change that affects users who have no FLOAT/DOUBLE
columns, and it changes only the two arms where one side is genuinely a `float32`. A comparison
cannot be more precise than its less precise side, so this is the correct semantics rather than
a loosening.

## Links & References

**Closes:** #4426

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

Covered by unit tests at the packet-codec and matcher level, which is where the defect is. Happy
to add an e2e case if you would prefer one — it would need a fixture schema carrying `FLOAT` and
`DOUBLE` columns, which I don't think the current e2e apps have.

- [x] 👍 yes

Each `frombits` site says why it is a reinterpret and not a cast, since the broken form compiles
and reads naturally. The coercion helpers carry a comment explaining the serializer type erasure
that makes them necessary.

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

- [x] 👍 yes, mentioned below

```sql
CREATE TABLE t (id INT PRIMARY KEY, f FLOAT, d DOUBLE);
INSERT INTO t VALUES (1, 9.99, 9.99);
```

Record an endpoint that selects `f` and `d`, then read `keploy/test-set-N/mocks.yaml`.

- **Before:** the recorded rows hold `1.0926057e+09` and `4.621813488089437e+18`.
- **After:** they hold `9.99` and `9.99`, and `keploy test` passes.

The unit tests reproduce it without a database:

```
go test ./pkg/agent/proxy/integrations/mysql/wire/phase/query/rowscols/
go test ./pkg/agent/proxy/integrations/mysql/replayer/
```

Both new test files **fail on `main`** and pass with this change — verified by reverting each
production file individually and re-running:

- `TestDecodeBinaryRow_FloatDoubleAreBitPatterns` — `got 1.0926057e+09, want 9.99`
- `TestEncodeBinaryRow_SurvivesYAMLRoundTrip` — `panicked: interface {} is float64, not float32`
- `TestEncodeBinaryRow_IntegerColumnsSurviveYAML` — `panicked: interface {} is uint64, not int`
- `TestParamValueEqual_FloatParamMatchesItselfAcrossYAML` — `live float32 9.99 vs recorded 9.99 (float64): no match`

Three of the new tests are guards rather than regression tests and pass in both directions by
design: `TestParamValueEqual_DistinctFloatsStillDiffer` (narrowing must not make the comparison
blind), `TestParamValueEqual_DoubleKeepsFullPrecision` (DOUBLE must not start narrowing), and
`TestParamValueEqual_WholeFloatParamRestoresAsInt`.

## Self Review done?
- [x] ✅ yes

`gofmt`, `go vet ./pkg/agent/proxy/integrations/mysql/...` and `go build ./...` are clean, and
the full `go test ./pkg/...` suite passes.

Recorded `mocks.yaml` rows for a `FLOAT`/`DOUBLE` column holding `9.99`:

```yaml
# before
- {type: 4, name: price_f, value: 1.0926057e+09}
- {type: 5, name: price_d, value: 4.621813488089437e+18}
# after
- {type: 4, name: price_f, value: 9.99}
- {type: 5, name: price_d, value: 9.99}
```

---

## Additional checklist:

### DIFF
--- a/.github/workflows/java_linux.yml
+++ b/.github/workflows/java_linux.yml
@@ -95,6 +95,13 @@ jobs:
           - job: record_build_replay_build
             record_src: build-no-race
             replay_src: build-no-race
+            # FLOAT/DOUBLE column fidelity (keploy#4426) needs BOTH
+            # binaries to carry the fix: a record binary without it
+            # writes the IEEE-754 bit pattern into the mock, and a
+            # replay binary without it panics on a correctly recorded
+            # FLOAT. So only this cell exercises the fixture; the
+            # cross-version cells leave the flag unset and skip.
+            numeric_fidelity: "true"
     name: mysql-dual-conn (${{ matrix.config.job }})
     steps:
       - name: Checkout repository
@@ -128,6 +135,7 @@ jobs:
         env:
           RECORD_BIN: ${{ steps.record.outputs.path }}
           REPLAY_BIN: ${{ steps.replay.outputs.path }}
+          MYSQL_NUMERIC_FIDELITY: ${{ matrix.config.numeric_fidelity }}
         run: |
           cd samples-java/mysql-dual-conn
           source $GITHUB_WORKSPACE/.github/workflows/test_workflow_scripts/java/mysql_dual_conn/java-linux.sh
--- a/.github/workflows/test_workflow_scripts/java/mysql_dual_conn/java-linux.sh
+++ b/.github/workflows/test_workflow_scripts/java/mysql_dual_conn/java-linux.sh
@@ -95,6 +95,25 @@ send_request() {
   echo "=== /api/oms/stmt-reset/5 (trigger COM_STMT_RESET) ==="
   curl -sS http://localhost:8080/api/oms/stmt-reset/5 || true
 
+  # Column-type fidelity (keploy#4426). Both endpoints bind a parameter,
+  # so Connector/J issues COM_STMT_EXECUTE and MySQL answers with a
+  # BINARY-protocol result set — the format whose FLOAT/DOUBLE columns
+  # carry raw IEEE-754 bytes. An unparameterised SELECT would come back
+  # as a text result set (every value a string) and would not touch this
+  # code path at all.
+  #
+  # Gated: see the MYSQL_NUMERIC_FIDELITY note at the guard below.
+  if [[ "${MYSQL_NUMERIC_FIDELITY:-}" == "true" ]]; then
+    echo "=== /api/oms/numerics (FLOAT/DOUBLE/BIGINT UNSIGNED, binary rows) ==="
+    curl -sS http://localhost:8080/api/oms/numerics || true
+
+    # Binds a FLOAT parameter: float32 on the wire, float64 out of the
+    # mock file. Exercises the COM_STMT_EXECUTE parameter decode and the
+    # matcher's mixed-width float comparison.
+    echo "=== /api/oms/float-param/9.99 (FLOAT bound parameter) ==="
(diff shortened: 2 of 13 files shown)


SepulvedaTwain:

SepulvedaTwain:
recheck

slayerjain:
Thanks @SepulvedaTwain

PR #4432, opened 2026-08-18
fix(http): treat conditionally-emitted tracing headers as noise

slayerjain:

`HeadersContainKeys` is presence-only: every recorded header key must exist on the live request unless it is in the noise set. Several distributed-tracing propagation headers are emitted by their tracer **only under some conditions**, so a recording made while one was present can never match a replay made while it is absent — and because the rejection happens before any value is compared, a value-noise entry cannot cover it. The header key has to be in the auto-noise allowlist.

`x-datadog-tags` is the case seen in the field. `dd-trace` emits it only when the trace carries at least one `_dd.p.*` propagation tag (128-bit trace id, sampling decision). A replay whose trace carried none sent no such header:

```
recorded:  X-Datadog-Tags: _dd.p.dm=-1,_dd.p.tid=...
           Tracestate:     dd=p:...;s:1;t.dm:-1
           X-B3-Traceid:   <128-bit, high half populated>
live:      (no X-Datadog-Tags at all)
           Tracestate:     dd=p:...;s:1          <- no t.dm
           X-B3-Traceid:   0000000000000000...   <- no _dd.p.tid
```

Every other tracing header differed only in *value* and was waved through. That one missing key rejected every candidate, the match reported `no_schema_candidates` with the whole mock pool still intact, a JWKS fetch came back with no keys, and the request returned **401** — surfacing as an application auth bug rather than a mock miss.

`baggage`, `sentry-trace` and `x-b3-flags` are conditional in exactly the same way and are added alongside it:

| header | emitted only when |
|---|---|
| `x-datadog-tags` | the trace carries at least one `_dd.p.*` tag |
| `baggage` | the context has baggage entries |
| `sentry-trace` | the transaction is being traced |
| `x-b3-flags` | the trace is debug-sampled |

As with every entry in the list, `--disableAutoHeaderNoise` opts out.

## Links & References

**Closes:** NA

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

Reproducing it end to end needs a tracer that drops the header mid-run; the unit test reproduces the exact header sets instead.

- [x] 👍 yes

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

- [x] 👍 yes, mentioned below

```
go test ./pkg/agent/proxy/integrations/http/ -run 'ConditionalDatadogTags|CoversConditionallyEmitted'
```

`TestHeadersContainKeys_ConditionalDatadogTagsHeader` reproduces the recorded and live header sets and fails with the change reverted. `TestFlakyHeaders_CoversConditionallyEmittedPropagationHeaders` guards the four entries against removal.

## Self Review done?
- [x] ✅ yes

- NA

### DIFF
--- a/pkg/agent/proxy/integrations/http/match.go
+++ b/pkg/agent/proxy/integrations/http/match.go
@@ -81,6 +81,17 @@ var flakyHeaders = []string{
 	"x-datadog-parent-id",
 	"x-datadog-sampling-priority",
 	"x-datadog-origin",
+	"x-datadog-tags", // _dd.p.* propagation tags — emitted ONLY when the trace carries at least one (128-bit trace id, sampling decision). Absent whenever it carries none, so its very PRESENCE is per-request.
+
+	// ── Conditionally-emitted propagation headers ────────────────────
+	// Same class as the entries above, and listed separately because the
+	// hazard is different: these are absent outright on some requests rather
+	// than merely carrying a fresh value. HeadersContainKeys is presence-only,
+	// so a recorded header that the live request omits kills the candidate
+	// before any value comparison runs — a value-noise entry would not help.
+	"baggage",      // W3C Baggage (OpenTelemetry, Sentry) — present only when the context has baggage entries
+	"sentry-trace", // Sentry distributed tracing — present only when tracing is enabled for the transaction
+	"x-b3-flags",   // B3 debug flag — present only on debug-sampled traces
 
 	// ── Generic request/correlation IDs ──────────────────────────────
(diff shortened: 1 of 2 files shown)


charankamarapu (approved):
LGTM

PR #4433, opened 2026-08-18
fix(http): treat x-datadog-tags and W3C baggage as flaky headers

charankamarapu:
## Problem

On replay, a request whose auth depends on a JWKS/OIDC key fetch **401s**, because the recorded upstream key-fetch mock isn't matched.

## Root cause

HTTP header matching is **presence-only** (`HeadersContainKeys`): a header key present on the recorded request but absent on the live replay request fails the match unless the key is flaky-noise. `dd-trace-rb` (and other tracers) emit **`x-datadog-tags`** and **W3C `baggage`** only on sampled/decorated spans, so they appear on some outbound calls and not others. The JWKS fetch recorded WITH `x-datadog-tags` but replayed WITHOUT it is rejected → the app never gets the verification key → 401. The mismatch report shows it verbatim: `header.X-Datadog-Tags: recorded "", absent in live call` (and the same for `Baggage`).

`x-datadog-tags` was missing from `flakyHeaders` even though its siblings (`x-datadog-trace-id/parent-id/sampling-priority/origin`) are present.

## Fix

Add `x-datadog-tags` (Datadog block) and `baggage` (W3C block) to `flakyHeaders`. Regression test: a JWKS mock matches when both headers are absent at replay.

Full `http` integration suite green; `gofmt` clean.

### DIFF
--- a/pkg/agent/proxy/integrations/http/match.go
+++ b/pkg/agent/proxy/integrations/http/match.go
@@ -68,6 +68,7 @@ var flakyHeaders = []string{
 	// ── W3C Trace Context / OpenTelemetry ────────────────────────────
 	"traceparent", // unique trace-id + span-id per request
 	"tracestate",  // vendor-specific trace context
+	"baggage",     // W3C Baggage — per-request correlation entries; tracers emit it inconsistently
 
 	// ── Zipkin B3 propagation ────────────────────────────────────────
 	"x-b3-traceid",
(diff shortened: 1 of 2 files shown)


chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(http): treat x-datadog-tags and W3C baggage as flaky headers. Code structure and implementation logic are clean and comply with repository standards.

PR #4435, opened 2026-08-18
fix(mysql): identify statements by executable SQL, not by wire bytes

slayerjain:

A client with database-monitoring instrumentation — Datadog DBM, Google sqlcommenter, OpenTelemetry — prefixes **every** statement with a comment carrying a per-request W3C `traceparent` and the database endpoint that served the call:

```sql
/*dde='sandbox',ddps='web-service',ddpv='1a2b3c4d',
  ddh='demo.cluster-abcdefghijkl.us-east-1.rds.amazonaws.com',
  dddb='appdb',traceparent='00-<unique per request>-<unique>-01'*/ SELECT ...
```

`matchQuery` treated that raw wire text as the statement's identity, which fails three ways at once:

1. **Exact-text comparison can never fire** — the `traceparent` is re-minted per request, so the recorded text is never byte-equal to the replayed one.
2. **That comparison was gated on equal MySQL `PayloadLength`** — and an Aurora cluster handing out its reader endpoint (`cluster-ro-`) instead of its writer (`cluster-`) makes the same statement's prologue three bytes longer.
3. **With both gone, equal `PayloadLength` alone scored a point** and became the only surviving signal — so the first recorded mock of the same byte length won, regardless of what SQL it actually held.

The third is the damaging one. With a couple of hundred bytes of tracer comment in front of every statement, unrelated queries collide on total length constantly — **every `SHOW FULL FIELDS FROM <table>` with an equal-length table name measures the same**. Serving one table's column metadata in answer to another's does not look like a mock mismatch downstream; it looks like an application bug, because the ORM builds a model with the wrong columns.

Measured over that recording's 508 recorded statements, each re-issued with a fresh prologue:

| | resolved to own recording | answered with a **different** statement | matched nothing |
|---|---|---|---|
| before | 96 | **324** | 88 |
| after | **508** | 0 | 0 |

### What changed

- **Identity is the executable statement.** `stripInertSQLComments` removes every comment the server discards — block and line form, anywhere in the statement, not just in front of it — and matching compares what is left. The scan is quote-aware, so a `/*` or `--` inside a string literal or a backquoted identifier is data and survives.
- **Version-gated `/*! ... */` and optimizer-hint `/*+ ... */` comments are deliberately preserved.** The server acts on both, so two statements differing there are not interchangeable.
- **`PayloadLength` no longer contributes to the score, and no longer gates the exact comparison.** It was never evidence of anything; it became load-bearing only because the tracer comment had already defeated the text comparison.
- Everything else in the matcher is untouched. The pre-existing DML parse-tree tier is left exactly as it was, tie-break included — it has its own weaknesses, but they are not what this PR is about and it never returns a definitive match.

### The last-resort tier for interpolated literals

Removing PayloadLength leaves a gap: `sqlparser.IsDML` covers only INSERT/UPDATE/DELETE, so a `SELECT` that is not byte-identical to its recording would score 0 and the connection would be torn down. That breaks a common and legitimate shape — a client that interpolates its values instead of binding them re-issues the same statement with a freshly generated id on every run:

```
recorded: SELECT id, email FROM customers WHERE id = '1dc32a3c-…'
replayed: SELECT id, email FROM customers WHERE id = '2a22d715-…'
```

(keploy's own `go-memory-load-mysql` suite is exactly this, and caught it in CI on the first push of this branch.)

So there is a last-resort tier: compare the statement with every inline literal **value** replaced by a type-tagged placeholder. Two properties are what make it safe where byte length was not:

- **Identifiers survive** — `SHOW FULL FIELDS FROM \`invoices\`` can never be answered by `... FROM \`couriers\``, which is precisely the cross-serve equal-length scoring caused.
- **The literal's type survives** (`?s` vs `?n`) — `x = 1` and `x = '1'` stay distinct, because MySQL does not treat them alike.

The masking is lexical, with no SQL parser, deliberately: a parser renders statements it does not model to a constant string, which would make unrelated statements compare equal. Anything the scanner is unsure of is copied verbatim, so ambiguity fails toward "no match".

The tier is never definitive — the recorded response belongs to a different row — and it scores below the DML parse-tree tier, so DML selection is unchanged. Session-control statements are excluded: for `SET NAMES utf8mb4` the literal *is* the meaning, and cross-serving one would misconfigure the connection for everything after it.

### Diagnostics

`MockMismatchReport.MatchPhase` and `CandidateCount` were never assigned, so every MySQL miss logged `"match_phase":"", "candidates":0` — indistinguishable from an empty mock pool, and the reason this class of failure reads as a lost recording. Now:

- both are measured, counting only the mocks the command phase actually compares;
- the mismatch report names the **nearest recorded statement** instead of leaving `closest_mock` empty;
- `REPLAY-ORPHAN` is raised only when no data mock was available at all, not merely when nothing scored;
- the report renders the SQL rather than the tracer comment, which previously consumed the whole truncation budget.

## Links & References

**Closes:** NA

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

Verified end-to-end by hand instead — see below. The behaviour needs a MySQL client that emits sqlcommenter prologues, which no existing sample app does; happy to add one if you'd like it in the matrix.

- [x] 👍 yes

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

- [x] 👍 yes, mentioned below

**Unit / regression**

```
```

Every new test fails with the production change reverted:

- `TestMatchQueryPacket_TraceCommentDoesNotDefeatMatching` — same statement behind comments that differ in content, in length, and mid-statement
- `TestMatchQueryPacket_EqualLengthDifferentStatementsNeverMatch` — different statements of identical byte length must score 0
- `TestMatchQueryPacket_LiteralTypeIsPartOfTheStatement` — `x = 1` and `x = '1'` are different statements
- `TestMatchQueryPacket_ExecutableCommentsAreNeverStripped` — `/*!` and `/*+` survive
- `TestStripInertSQLComments` — the primitive, including quote-awareness (`'/* not a comment */'`), `--x` vs `-- x`, unterminated comments, and comment-only probes
- `TestMatchQueryPacket_LiteralTypeIsPartOfTheStatement` — `x = 111` vs `x = '1'`, equal length, are different statements
- `TestMatchQueryPacket_InlineLiteralDriftIsServable`, `TestMatchCommand_LiteralDriftPicksTheRightShape` — a drifted interpolated id still resolves, and resolves to the right table
- `TestMaskSQLLiterals`, `TestIsSessionControlStatement` — what the last-resort tier will and will not equate
- `TestMatchCommand_MissNamesTheNearestRecordingInsteadOfClaimingNoneExists`, `TestMatchCommand_NoMocksPhaseStillReachable` — the diagnostics

**Corpus harness** (skipped unless pointed at a recording)

```
grep -oP '^\s*query:\s*\K.*' mocks.yaml | jq -Rs 'split("\n")|map(select(.!=""))' > corpus.json
KEPLOY_MYSQL_QUERY_CORPUS=corpus.json go test ./pkg/agent/proxy/integrations/mysql/replayer/ -run AgainstRecordedCorpus -v
```

Asserts every recorded statement, re-issued with a fresh prologue and the other cluster endpoint, resolves to its **own** recording.

**End to end**, against a Go service emitting DBM prologues over real MySQL, recorded and replayed with `keploy record` / `keploy test`:

| | result |
|---|---|
| before | 3 / 9 passed |
| after | **9 / 9 passed** |

## Self Review done?
- [x] ✅ yes

- NA

### DIFF
--- a/pkg/agent/proxy/integrations/mismatch/mismatch.go
+++ b/pkg/agent/proxy/integrations/mismatch/mismatch.go
@@ -167,7 +167,7 @@ func defaultNextSteps(r *models.MockMismatchReport) string {
 	}
 	switch {
 	case r.MatchPhase == models.MatchPhaseNoMocks:
-		return "No recorded mocks exist for this protocol in the selected test set. Re-record the test set with 'keploy record'."
+		return "No recorded mocks were available to match against for this protocol in the selected test set. Re-record the test set with 'keploy record'."
 	case onlyValueDrift:
 		paths := make([]string, 0, len(r.FieldDiffs))
 		// Body diffs are reported "body."-prefixed for readability, but HTTP
--- a/pkg/agent/proxy/integrations/mysql/replayer/match.go
+++ b/pkg/agent/proxy/integrations/mysql/replayer/match.go
@@ -1353,6 +1353,10 @@ const (
 	// ok=true, which makes the caller take the definitive path and ignore the
 	// score entirely, so the value is informational only.
 	scoreQueryExact = 2
+	// scoreQueryLiteralDrift — same statement, drifted inline literal values.
+	// Last resort, never definitive, and deliberately below scoreQueryStructure
+	// so DML selection is unchanged.
+	scoreQueryLiteralDrift = 3
 	// scoreQueryStructure — both DML with an identical parse-tree shape.
 	// Pre-existing tier, never definitive.
 	scoreQueryStructure = 6
@@ -1459,11 +1463,25 @@ func matchQuery(_ context.Context, log *zap.Logger, expected, actual mysql.Packe
 		return false, 0
 	}
 
-	// Non-DML and not textually identical: nothing further establishes that
-	// these are the same statement, so the candidate is not servable. The
-	// mismatch report still names the nearest recorded query — matchCommand
-	// tracks that separately from the score.
 	if !(sqlparser.IsDML(expectedQuery) && sqlparser.IsDML(actualQuery)) {
+		// Non-DML. sqlparser.IsDML covers only INSERT/UPDATE/DELETE, so this is
+		// where every SELECT and SHOW lands, and the parse-tree tier below never
+		// sees them. Last resort: the same statement with a drifted inline
+		// literal — a client that interpolates a freshly generated id instead of
+		// binding it re-issues exactly this shape on every run, and refusing it
+		// tears the connection down.
+		//
+		// Not definitive: the recorded response belongs to a different row, so it
+		// only scores, and only wins when nothing matched exactly.
+		if !isSessionControlStatement(expectedQuery) && !isSessionControlStatement(actualQuery) &&
+			maskSQLLiterals(expectedQuery) == maskSQLLiterals(actualQuery) {
+			log.Debug("query matched with drifted inline literals",
+				zap.String("expected query", expectedQuery),
+				zap.String("actual query", actualQuery))
+			// The payload-length tie-break only ranks candidates already known to
+			// share every keyword, identifier and literal type.
+			return false, scoreQueryLiteralDrift + equalPayloadLength(expected, actual)
+		}
 		log.Debug("No Query is dml",
 			zap.String("expected query", expectedQuery),
 			zap.String("actual query", actualQuery))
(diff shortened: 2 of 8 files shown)


charankamarapu (approved):
LGTM

PR #4437, opened 2026-08-18
fix(replay): stop reporting recoverable failures at ERROR

slayerjain:

keploy's own e2e scripts fail a run on **any** `ERROR` line in the replay log — [`golang/http_pokeapi/golang-linux.sh:132`](../blob/main/.github/workflows/test_workflow_scripts/golang/http_pokeapi/golang-linux.sh) does `grep "ERROR" "test_logs.txt"`, and the python/node/sse siblings do the same. So an `ERROR` line is a **contract**, not just a log level. Two places broke it by reporting conditions keploy itself recovers from, which turns a run whose every test passed into a red build.

### 1. A transport reset is recoverable, so it must not be an ERROR

`SimulateHTTP` / `SimulateHTTPStreaming` logged every send failure at ERROR. But for a transport-level connection reset, `RunTestSet` re-sends the request through `retryResetOnce` once it can prove the failed attempt consumed no mock — and the common docker-proxy reset recovers on the first re-send. The log read:

```
ERROR  failed to send testcase request to app   … read: connection reset by peer
WARN   test request reset by app/proxy before any mock was consumed; re-sending —
       the request was never processed, so this is not a retry of an assertion
…
Total test passed: 32     Total test failed: 0
```

…and the build was red, on the ERROR line, for a run in which nothing failed.

`logSimulateSendFailure` now reports the re-sendable class (`IsTransportConnReset`) at Debug; **every other failure is still an ERROR**. Nothing is lost — the definitive ERROR is logged by the caller once the re-send is refused or exhausted (`failed to simulate request` / `failed to simulate streaming request`).

`pkg/service/runner` has no such retry, so a reset is terminal on that path and was only visible in the returned `TestResult`. It now logs the failure itself.

### 2. The health gate's timeout is not an Info-level event

`waitForAppReady`'s health-URL gate reported its timeout at Info and then fired the whole suite on a fixed delay (deliberate, and unchanged here). When an app is slower than `--health-poll-timeout`, the result is *every* test failing with no response at all — observed as **23 failures all reading `EXPECT 200 / ACTUAL 0`** — which is indistinguishable from a real regression unless that one line stands out.

It now warns and names the consequence, matching the **sibling TCP port gate in the same function**, which already reports the identical situation at Warn (`"firing tests anyway (replay may see status_code got=0)"`).

A malformed `--health-url` or `appReadyProbeAddr` deliberately **stays** at ERROR: those are permanent misconfigurations that will never succeed, unlike a transient timeout, and an existing test pins that intent.

### 3. The best-effort agent hooks — the bigger hole

A single failed hook emitted **three** ERROR lines — the agent handler, the agent client, and the replay caller — once per test case, while the run carried on and every test passed.

Ignoring a hook failure is correct, and the code already says so in three independent places: the default OSS implementation is a **no-op extension point** (`service/agent.AgentHook` returns nil for all of them), the agent client **already** swallows a transport failure calling one (`Debug`, `return nil`), and every call site discards the returned error. The bug was the level, not the ignoring.

| site | before | now |
|---|---|---|
| `replay/hooks.go` — Before/AfterSimulate, BeforeTestRun, BeforeTestSetCompose, AfterTestRun | 5 × `logger.Error`, ignored | Warn, once, **with the hook name and the test case** |
| `replay.go` — AfterTestSetRun, AfterTestRun ×2, BeforeTestSetReplay | Error; two already said *"replay continues"* in the message | Warn |
| `platform/http/agent.go` — non-2xx on 5 hook calls | logged Error **and** returned the error, so the caller logged it again | body folded into the returned error, no log — the caller owns the level |
| `agent/routes/record.go` — before/after-simulate handlers | Error | Warn; still answers 500 |
| `replay/hooks.go` — `GetConsumedMocks` | logged an error it also returned | returns it; callers already report it (some at Error, some deliberately at Debug) |

**Deliberately left at ERROR**, because these do mean something is wrong:
- a failed backup before deleting failing test cases — the run then does something destructive with no safety net;
- a malformed `--health-url` / `appReadyProbeAddr` — permanent misconfiguration that will never succeed, unlike a transient timeout, and an existing test pins that intent.

Adds `utils.LogWarn` so downgraded sites keep `LogError`'s `context.Canceled` suppression — on shutdown every in-flight call fails with it, and that is noise at any level.

## Links & References

**Closes:** NA

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

The regression is a log level, and the new tests drive the real `SimulateHTTP` / `SimulateHTTPStreaming` against a socket that RSTs on accept — the same shape as the existing `util_simulate_refuse_integration_test.go`, without needing CI credentials.

- [x] 👍 yes

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

`--health-poll-timeout`'s flag help said "logs an info message"; updated in this PR.

- [x] 👍 yes, mentioned below

```
go test ./pkg/ -run 'TestSimulateHTTP_TransportReset|TestSimulateHTTP_UnreachableApp|TestLogSimulateSendFailure' -v
go test ./pkg/service/replay/ -run TestWaitForAppReady -v
```

- `TestSimulateHTTP_TransportResetIsNotLoggedAsError` — drives **both** real entry points against a listener that RSTs on accept, and asserts zero ERROR lines. Deliberately not a unit test of the helper: reverting only the two call sites while keeping the helper leaves a helper-level test green, so it would not have guarded the regression.
- `TestSimulateHTTP_UnreachableAppStillLogsError` — a genuinely refused port must still ERROR, keeping the downgrade narrow.
- `TestWaitForAppReady_HealthTimeoutWarnsWithConsequence` — the timeout warns, names the consequence and names the knob.
- `TestAgentHookFailuresAreNotErrors` — drives the real `Hooks.SimulateRequest` / `BeforeTestRun` / `BeforeTestSetCompose` / `AfterTestRun` against an instrumentation stub that fails every hook, and asserts zero ERROR lines **and** at least one Warn (visible, not silenced).
- `TestGetConsumedMocksDoesNotDoubleReport` — the pass-through returns its error without also logging it.

Each fails when its production change is reverted.

## Self Review done?
- [x] ✅ yes

- NA

### DIFF
--- a/cli/provider/cmd.go
+++ b/cli/provider/cmd.go
@@ -398,7 +398,7 @@ func (c *CmdConfigurator) AddUncommonFlags(cmd *cobra.Command) {
 		cmd.Flags().Uint32("sse-port", c.cfg.Test.SSEPort, "Custom SSE port to replace the actual port in the SSE testcases")
 		cmd.Flags().Uint64P("delay", "d", 5, "User provided time to run its application")
 		cmd.Flags().String("health-url", c.cfg.Test.HealthURL, "HTTP(S) URL polled before the first test is fired; first 2xx response proceeds immediately. Empty (default) preserves the fixed --delay behavior.")
-		cmd.Flags().Duration("health-poll-timeout", c.cfg.Test.HealthPollTimeout, "Ceiling for --health-url polling (e.g. 60s, 2m). If no 2xx is seen within this window, replay logs an info message and falls back to --delay.")
+		cmd.Flags().Duration("health-poll-timeout", c.cfg.Test.HealthPollTimeout, "Ceiling for --health-url polling (e.g. 60s, 2m). If no 2xx is seen within this window, replay warns and falls back to --delay.")
 		cmd.Flags().String("proto-file", c.cfg.Test.ProtoFile, "Path of main proto file")
 		cmd.Flags().String("proto-dir", c.cfg.Test.ProtoDir, "Path of the directory where all protos of a service are located")
 		cmd.Flags().StringArray("proto-include", c.cfg.Test.ProtoInclude, "Path of directories to be included while parsing import statements in proto files")
--- a/pkg/agent/routes/record.go
+++ b/pkg/agent/routes/record.go
@@ -182,7 +182,10 @@ func (a *Agent) HandleBeforeSimulate(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := agent.ActiveHooks.BeforeSimulate(r.Context(), req.TimeStamp, req.TestSetID, req.TestCaseName); err != nil {
-		a.logger.Error("failed to execute before simulate hook", zap.Error(err))
+		// Warn: the caller treats a failed simulate hook as non-fatal and
+		// carries on, so this must not claim the run is broken. The 500 below
+		// still reports the failure to it.
+		utils.LogWarn(a.logger, err, "failed to execute before simulate hook")
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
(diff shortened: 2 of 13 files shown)


chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(replay): stop reporting recoverable failures at ERROR. Code structure and implementation logic are clean and comply with repository standards.

PR #4400, opened 2026-07-31
fix(mysql): preserve pipelined command response order

Hardik180704:
## Describe the changes that are made

- Queue response-bearing MySQL commands until their corresponding server packets arrive.
- Preserve FIFO pairing when Connector/J pipelines `COM_STMT_RESET`, `COM_STMT_SEND_LONG_DATA`, and `COM_STMT_EXECUTE`.
- Keep no-response commands such as `COM_STMT_SEND_LONG_DATA` outside the response queue.
- Restore the queued command operation before decoding its response, preventing a later pipelined command from changing the response decoder state.
- Add a deterministic regression test that distinguishes RESET's zero-row OK from EXECUTE's affected-row OK.

## Links & References

**Closes:** #4262

### 🔗 Related PRs

- NA

### 🐞 Related Issues

- #4262

### 📄 Related Documents

- NA

## What type of PR is this? (check all applicable)

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?

- [ ] 👍 yes
- [x] 🙅 no, because the protocol-level regression is covered deterministically in the MySQL recorder package
- [ ] 🙋 no, because I need help

## Added comments for hard-to-understand areas?

- [x] 👍 yes
- [ ] 🙅 no, because the code is self-explanatory

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 Wiki
- [x] 🙅 no documentation needed

## Are there any sample code or steps to test the changes?

- [x] 👍 yes, mentioned below
- [ ] 🙅 no, because it is not needed

```bash
go test ./pkg/agent/proxy/integrations/mysql/recorder -run TestAsyncMySQLDecode_PipelinedResetLongDataExecute -count=1
go test ./pkg/agent/proxy/integrations/mysql/... -count=1
go test -race ./pkg/agent/proxy/integrations/mysql/recorder -count=1
```

## Self Review done?

- [x] ✅ yes
- [ ] ❌ no, because I need help

## Any relevant screenshots, recordings or logs?

- NA

## 🧠 Semantics for PR Title & Branch Name

- **PR Title:** `fix(mysql): preserve pipelined command response order`
[...]

PR #4438, opened 2026-08-18
fix(ci): scope the tidb blob EXECUTE guard by connection and gate it on the recorder

### LINKED ISSUE #4262: [bug]: streamed-BLOB (COM_STMT_SEND_LONG_DATA) recording mis-pairs the pipelined RESET+SLD+EXECUTE sequence
- Searched; this tracks a recorder gap noted in #4260 (which fixes only the
  matcher-side SEND_LONG_DATA connection drop).

### 👍 Current behavior

When a Java app writes a BLOB/CLOB via a stream setter
(`PreparedStatement.setBinaryStream`/`setBlob`) on a server-side prepared
statement, Connector/J 8.x pipelines, per re-execution:

```
COM_STMT_RESET  →  COM_STMT_SEND_LONG_DATA  →  COM_STMT_EXECUTE(→ OK)
```

The recorder's request→response state machine mis-pairs this pipelined burst.
Recording the `tidb-stmt-cache` blob endpoint three times produces, per call:

```
COM_STMT_PREPARE        → COM_STMT_PREPARE_OK
COM_STMT_SEND_LONG_DATA → (no resp)            # ok
COM_STMT_RESET          → OK(affected_rows=1)  # WRONG: this is the EXECUTE's response
SELECT LENGTH(payload)  → TextResultSet
```

i.e. the **EXECUTE's `OK(affected_rows=N)` is attributed to the `RESET` mock**,
and the standalone `COM_STMT_EXECUTE` request is dropped from the trace
entirely. On replay the live `EXECUTE` then has no response mock and the driver
gets `EOFException: Can not read response from server` at
`ServerPreparedQuery.sendExecutePacket` → HTTP 500.

Root cause appears to be the flush-on-next-command logic in
`pkg/agent/proxy/integrations/mysql/recorder/query.go` (~L469 + the
no-response/unknown handling at L492–512) not accounting for the client
pipelining multiple commands before reading their responses.

#4260 makes the matcher accept an unmocked `SEND_LONG_DATA` gracefully (stops
the `08S01` drop on that packet), but the streamed-BLOB **round-trip** still
fails until the recorder captures the EXECUTE response correctly.

### 👟 Steps to Replicate

1. In `tidb-stmt-cache`, bind a parameter via `ps.setBinaryStream(...)` and
   `INSERT` it on a server-side PS (`useServerPrepStmts=true&cachePrepStmts=true`).
2. `keploy record`, drive the endpoint 2–3 times, stop.
3. Inspect `mocks.yaml`: the EXECUTE responses (`affected_rows=1/2/3`) sit on
   `COM_STMT_RESET` mocks; there is no `COM_STMT_EXECUTE` mock for the INSERT.
4. `keploy test` with the DB down → the blob EXECUTE gets `EOFException` / 500.

### 🤖 Environment
- keploy: v3 (main); MySQL recorder state machine.
- Driver: mysql-connector-j 8.x; server: TiDB :4000.
- Repro: native (eBPF) on Linux; would also repro in Docker.

Tracked alongside #4260.

slayerjain:
`run_java_linux / tidb-stmt-cache (record_latest_replay_build)` has been failing on `main` and on every PR. It is red on `main` itself at `7953778b`, together with the `CI Gate (Linux)` rollup that depends on it.

## Why it is red

#4400 fixed #4262 — a `COM_STMT_EXECUTE` whose parameter is streamed via `COM_STMT_SEND_LONG_DATA` was dropped at decode time — and added `/api/blob` traffic to exercise it. But the `record_latest_replay_build` cell records with the latest **release**, which predates that fix and cannot capture those EXECUTEs.

The two cells' uploaded recordings, same commit:

| recording | total `COM_STMT_EXECUTE` | `SEND_LONG_DATA` | EXECUTEs for `blob_stream` |
|---|---|---|---|
| build-recorded (has the fix) | 25 | 3 | **3** |
| release-recorded (this cell) | 19 | 3 | **0** |

The three missing EXECUTEs are the three `get-api-blob-by-id-*` failures. This cell cannot go green until the fix ships in a release.

## The guard added with that traffic was vacuous

`assert_blob_execute_recorded` counted `COM_STMT_EXECUTE` across *all* mocks. The 19 EXECUTEs from `/api/kv` and `/api/cross` kept the total comfortably non-zero, so it printed `OK:` on a recording containing zero blob EXECUTEs — the exact regression it was written to catch.

## Changes

1. **Gate `/api/blob` and its guard on `RECORD_BIN` being this build**, the same capability gate `golang/connect_tunnel` uses. The cell keeps validating every path a released recorder can actually record. The comment says to drop the gate once the fix is released, mirroring the matrix's existing note about re-adding `record_build_replay_latest`.
2. **Make the guard specific.** Learn the statement id the `blob_stream` `COM_STMT_PREPARE` was handed in its `PREPARE_OK`, then count only the EXECUTEs referencing it. Keyed by **file + connID + id**, because statement ids are allocated per connection and connIDs restart in every recording — a bare id lets a second pooled connection's statement stand in for a dropped blob EXECUTE. Resolved at `END`, because mocks are not guaranteed to be emitted in wire order.

## Verification

Run against the real artifacts of both cells, and against fixtures derived from them:

| fixture | result |
|---|---|
| build-recorded artifact | pass, `3/3 (conn:stmt 0:4)` |
| **release-recorded artifact** (the failing cell) | **fail** — `produced only 0` |
| release-recorded artifact, `latest` recorder | SKIP, exit 0 |
| build-recorded minus one blob EXECUTE document | **fail** — `produced only 2` |
| second connection re-prepares `blob_stream` as id 2, colliding with `INSERT INTO kv` which has 5 EXECUTEs on id 2 | pass, exactly `3/3 (0:4/1:2)` |
| same topology with the blob EXECUTEs dropped | **fail** |
| two test-sets, bug in one, unrelated conn0/id4 in the other | **fail** |

Byte-identical results under gawk 5.2.1, mawk 1.3.4 and busybox awk — runners use mawk. `bash -n` and `shellcheck -S warning` clean; no new shellcheck findings on any added line.

Nothing else in the script asserts a fixed test or mock count, and `drop_prepare_mock` targets `SELECT ? AS v`, so gating the blob traffic does not disturb it.

### Known limitation

A statement id retired by `COM_STMT_CLOSE` and reissued to a different statement on the same connection cannot be distinguished, because that needs wire order, which is what the `END` resolution deliberately gives up. MySQL and TiDB both allocate prepared-statement ids from a monotonic per-session counter and never recycle them, and the sample never closes a statement, so this is unreachable. It is documented in the code.

### DIFF
--- a/.github/workflows/java_linux.yml
+++ b/.github/workflows/java_linux.yml
@@ -234,6 +234,11 @@ jobs:
         # validate the fix (record_latest_replay_build also confirms the new
         # binary replays mocks recorded by the previous release). Re-add
         # record_build_replay_latest for forward-compat once the fix is released.
+        #
+        # For the same reason the script gates the /api/blob traffic (keploy#4262,
+        # streamed-BLOB EXECUTE dropped at record) on RECORD_BIN being this build:
+        # the released recorder cannot capture those EXECUTEs, so driving them in
+        # record_latest_replay_build would make that cell permanently red too.
         config:
           - job: record_latest_replay_build
             record_src: latest
--- a/.github/workflows/test_workflow_scripts/java/tidb_stmt_cache/java-linux.sh
+++ b/.github/workflows/test_workflow_scripts/java/tidb_stmt_cache/java-linux.sh
@@ -20,6 +20,26 @@
 
 set -Eeuo pipefail
 
+# Number of /api/blob requests driven below; the guard asserts at least this many
+# blob COM_STMT_EXECUTE mocks, so the two can never drift apart.
+BLOB_REQUESTS=3
+
+# Capability gate for keploy#4262 (streamed-BLOB EXECUTE dropped at record).
+# The fix lives in this build and ships in no released binary yet, so a 'latest'
+# RECORD_BIN cannot capture the COM_STMT_EXECUTE at all: the recording holds the
+# RESET and SEND_LONG_DATA mocks but no EXECUTE, and replaying it reproduces the
+# very bug being fixed (the /api/blob tests fail with "Can not read response from
+# server"). Drive the blob traffic -- and its guard -- only when the recorder is
+# this build, so the record_latest_replay_build cell still validates the other
+# paths instead of being permanently red. Drop this gate once the fix ships in a
+# release. Mirrors the RECORD_BIN capability gate in golang/connect_tunnel.
+recorder_streams_blobs() {
+  case "${RECORD_BIN:-}" in
+    */build/keploy|*/build-no-race/keploy) return 0 ;;
+    *) return 1 ;;
+  esac
+}
+
 section() { echo "::group::$*"; }
 endsec()  { echo "::endgroup::"; }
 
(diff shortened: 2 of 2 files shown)


chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(ci): scope the tidb blob EXECUTE guard by connection and gate it on the recorder. Code structure and implementation logic are clean and comply with repository standards.

PR #4440, opened 2026-08-18
fix(proxy): sample live connections' network I/O, not only at teardown

ayush3160:
Kernel network-I/O accounting (`pkg/agent/proxy/netstat.go`) previously read a proxied connection's `TCP_INFO` byte totals **only at teardown**. Its own v1 note called this out: a connection that stays open for the whole recording contributes nothing until it closes.

That silently undercounts the largest flows: **DB / connection-pool sockets, HTTP keep-alive, and gRPC streams** stay open the entire recording, so the heaviest traffic (e.g. large result sets pulled over a persistent pool) is missed — often by orders of magnitude — and the reported Rx/Tx skews toward whatever rides the short-lived connections.

**Fix:** register each app-facing socket with a new `TrackConnNetworkIO` at setup, and add a background sampler that folds each interval's byte **delta** into the sink while the connection is still open. `RecordConnNetworkIO` now takes only the **final delta** at teardown.

- Accounting is delta-based, so a connection is **never double counted** across ticks or against its teardown read (verified by a unit test).
- Connections that are never tracked keep the **original** teardown-only behavior (full lifetime total, once) — no regression.
- Tracked at every app-facing socket that carries traffic: egress proxy (`origSrcConn`, covers all protocols via the shared teardown defer), ingress HTTP (dialed `upConn` + each redialed keep-alive upstream), ingress gRPC (`destConn`). The readiness-probe dial transfers no data and is intentionally left unaccounted. Non-Linux keeps its no-op stub.

Producer-side only; no wire-format or API change.

## Links & References

**Closes:** NA

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [x] 🐞 Bug Fix

## Added e2e test pipeline?

- [x] 👍 yes

## Added to documentation?

- [x] 👍 yes, mentioned below

Unit tests in `pkg/agent/proxy/netstat_test.go` (real loopback sockets):
- `TestForwardDelta` — counter delta + reset guard.
- `TestNetworkIO_PeriodicSamplingNoDoubleCount` — a periodic sample followed by a no-traffic sample followed by teardown counts the bytes exactly once.
- `TestRecordConnNetworkIO_UntrackedFullTotal` — an untracked connection still reports its full total at teardown.

```
go test ./pkg/agent/proxy/ -run 'ForwardDelta|NetworkIO|RecordConnNetworkIO_Untracked'
```

## Self Review done?
- [x] ✅ yes

- NA

### DIFF
--- a/pkg/agent/proxy/incoming/gRPC/record.go
+++ b/pkg/agent/proxy/incoming/gRPC/record.go
@@ -35,6 +35,10 @@ func RecordIncoming(ctx context.Context, logger *zap.Logger, clientConn, destCon
 		destConn.Close()
 	}()
 
+	// gRPC connections are long-lived (many multiplexed HTTP/2 streams over one
+	// socket), so sample destConn continuously while it's open rather than only at
+	// the teardown RecordConnNetworkIO below.
+	ossproxy.TrackConnNetworkIO(destConn)
 	defer func() {
 		if err := clientConn.Close(); err != nil &&
 			!strings.Contains(err.Error(), "use of closed network connection") {
--- a/pkg/agent/proxy/incoming/http.go
+++ b/pkg/agent/proxy/incoming/http.go
@@ -527,6 +527,10 @@ func (pm *IngressProxyManager) handleHttp1Connection(ctx context.Context, client
 		)
 		return
 	}
+	// Sample this app-facing socket continuously while it's open (a kept-alive
+	// serving connection can outlive many requests), so its bytes aren't withheld
+	// until the teardown RecordConnNetworkIO below.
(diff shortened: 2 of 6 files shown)


ayush3160:
Superseded by the kernel app network-I/O counter approach: #4446 (load + drain the eBPF `keploy_netio_bytes` counter) plus #4448 (remove the userspace TCP_INFO path this PR was patching). The live-connection sampling problem this PR addressed is moot once accounting is continuous and in-kernel rather than teardown-time in userspace.

PR #2378, opened 2024-10-25
feat: add failed testcase to file

ayush3160:
## What does this PR do?

This pr introduces a feature to add all failed generated unit tests into a temporary file

## Related PRs and Issues

fixes #2377 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Code style update (formatting, local variables)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Please let us know test plan followed

Please describe the tests(if any). Provide instructions how its affecting the coverage.

## Checklist:
- [ ] Have you read the [Contributing Guidelines on issues](https://keploy.io/docs/keploy-explained/contribution-guide/)?
- [ ] My code follows the style guidelines of this project.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] New and existing unit tests pass locally with my changes.

PR #4446, opened 2026-08-19
feat(netio): load + drain the kernel app network-I/O counter

ayush3160:

Wires in the in-kernel **app network-I/O counter** (from keploy/ebpf `keploy_ebpf.c`) so the app's true wire volume is metered **from the kernel** and flows to the resource footprint — replacing the userspace `TCP_INFO` polling.

**Loader** (`pkg/agent/hooks/linux`):
- Selects the `tcp_recvmsg` fexit wrapper matching the kernel ABI (`nonblock` dropped in 5.19) via BTF arg-count detection, stubbing the other so `LoadAndAssign` accepts the spec.
- Attaches `fexit/tcp_sendmsg`, the selected `fexit/tcp_recvmsg`, and `tp_btf/cgroup_rmdir` (best-effort). If the netio programs can't load (no BTF / older kernel), they're stubbed and the load retried, so the **core agent still loads** and the userspace path stays the fallback.
- Auto-pins `keploy_netio_bytes` (`CollectionOptions.Maps.PinPath`) for the drain.

**Drain** (`pkg/agent/proxy`):
- `StartKernelNetioDrain` opens the pinned per-cgroup counter, sums per-CPU, and folds per-interval rx/tx **deltas** into the **same `networkIOSink`** the TCP_INFO path feeds → reaches `/resource-stats` → k8s-proxy identically.
- On first non-zero read it flips `kernelNetioActive`; `RecordConnNetworkIO` then **stands down**, so the same traffic is never counted by both paths. On a kernel that can't load netio it stays inert and TCP_INFO keeps running (no regression).

Bindings synced from keploy/ebpf.

- [x] 🍕 Feature

## Added tests?

Build-tested (linux default + non-linux stub via `GOOS=darwin CGO_ENABLED=0`), `go vet` clean, gofmt clean. The fentry/fexit **attach** and the drain need **runtime validation on target kernels** (can't be exercised without a BTF-enabled kernel in CI here).

## Added to documentation?

## Are there any pre/post deployment tasks we need to perform?
- [x] 👍 yes, mentioned below

- **Depends on** keploy/ebpf#103 (the counter + `LIBBPF_PIN_BY_NAME`); the regenerated bindings are included here.
- The **cleanup PR** removes the now-superseded userspace TCP_INFO accounting once this is validated (and keeps it only as the pre-5.5-kernel fallback).

## Self Review done?
- [x] ✅ yes


ayush3160:
Superseded. The app network-I/O kernel counter (`keploy_netio_bytes`) already exists **entirely in the enterprise agent** (merged as enterprise #2378: `proxyless.c` fentry/fexit on tcp/udp sendmsg/recvmsg → `account_netio`, drained by `startNetworkIOAccounting` → resourceio → k8s-proxy). Rather than duplicate the counter in the shared eBPF object + OSS loader/drain (what these PRs did), we're extending the existing enterprise counter to the sidecar/k8s-proxy mode behind an opt-in flag, and removing the OSS metering path entirely. Keeping network-I/O metering out of OSS is the explicit goal. See the enterprise feature PR + the OSS TCP_INFO removal PR (to follow).

PR #4447, opened 2026-08-19
fix(contract): support top-level JSON array request/response bodies (#4445)

### LINKED ISSUE #4445: [bug]: contract generator fails for valid top-level JSON array bodies

### 👍 Current behavior

The default contract generator fails when an HTTP request or response body is valid JSON but the root value is an array.

For example, this is a valid and common JSON response for list APIs:

~~~json
[{"id":1,"name":"Utkarsh"}]
~~~

But `HTTPDocToOpenAPI` unmarshals the body into `map[string]interface{}`. That only works when the JSON root is an object, so array bodies fail with:

~~~text
json: cannot unmarshal array into Go value of type map[string]interface {}
~~~

Expected behavior: Keploy should generate a valid OpenAPI schema with `type: array` and an `items` schema instead of failing.

### 👟 Steps to Replicate

1. Go to the Keploy repository.

~~~bash
cd /Users/utkarshraj/Documents/opensource/keploy
~~~

2. Add this temporary repro test at `pkg/service/contract/httpdoc_array_repro_test.go`:

~~~go
package contract

import (
	"testing"

	"go.keploy.io/server/v3/pkg/models"
	"go.uber.org/zap"
)

func TestHTTPDocToOpenAPIAllowsTopLevelArrayResponse_Repro(t *testing.T) {
	svc := &contract{}

	_, err := svc.HTTPDocToOpenAPI(zap.NewNop(), models.HTTPDoc{
		Version: "api.keploy.io/v1beta1",
		Kind:    "Http",
		Name:    "array-response",
		Spec: models.HTTPSchema{
			Request: models.HTTPReq{
				Method: "GET",
				URL:    "http://example.com/users",
			},
			Response: models.HTTPResp{
				StatusCode:    200,
				StatusMessage: "OK",
				Body:          `[{"id":1,"name":"Utkarsh"}]`,
			},
		},
	})

	if err != nil {
		t.Fatalf("top-level JSON array response should be valid OpenAPI input, got error: %v", err)
	}
}
~~~

3. Run the targeted test.

~~~bash
mkdir -p .cache/go-build .cache/go-tmp

GOCACHE="$PWD/.cache/go-build" \
GOTMPDIR="$PWD/.cache/go-tmp" \
go test ./pkg/service/contract -run TestHTTPDocToOpenAPIAllowsTopLevelArrayResponse_Repro -count=1
~~~

4. The test fails with the array unmarshal error.

### 📜 Logs (if any)

~~~text
--- FAIL: TestHTTPDocToOpenAPIAllowsTopLevelArrayResponse_Repro (0.00s)
    httpdoc_array_repro_test.go:31: top-level JSON array response should be valid OpenAPI input, got error: json: cannot unmarshal array into Go value of type map[string]interface {}
FAIL
FAIL    go.keploy.io/server/v3/pkg/service/contract
~~~

Image

### 💻 Operating system

MacOS

~~~text
Darwin Utkarshs-MacBook-Air.local 24.1.0 Darwin Kernel Version 24.1.0: Thu Oct 10 21:05:14 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T8103 arm64
~~~

Not applicable

Not applicable. This is a unit-level contract generation issue and does not require Docker.

### 🧱 Your Environment

Local development machine.

### 🎲 Version

v2.5.2

### 📦 Repository

keploy

I was trying to generate an OpenAPI contract from recorded HTTP traffic. Some APIs commonly return top-level JSON arrays for list endpoints such as `GET /users`, `GET /orders`, or `GET /items`. Contract generation should support these valid responses.

chiliec:
## What

Fixes #4445.

`HTTPDocToOpenAPI` unmarshalled request and response bodies into `map[string]interface{}`, so a valid JSON body whose **root is an array** — a common shape for list endpoints — failed:

```go
json.Unmarshal([]byte(`[{"id":1,"name":"Utkarsh"}]`), &map[string]interface{}{})
// json: cannot unmarshal array into Go value of type map[string]interface {}
```

The whole contract generation aborted for any test/mock with a top-level array body.

## Fix

- Decode request/response bodies into `interface{}` (root may be an object or an array).
- Add `SchemaForBody(body, objectProps)` which returns:
  - `{type: object, properties: ...}` for an object root (unchanged behaviour), or
  - `{type: array, items: ...}` for an array root, inferring the item schema from the first element (object / number / boolean / string / nested array).
- `models.Schema` gains an optional `Items *Schema`; `MediaType.Example` becomes `interface{}` so an array example can be represented. `properties`/`items` are `omitempty` so object schemas serialize exactly as before.
- `GenerateResponse` now takes the prebuilt schema instead of hardcoding `type: object`.

Object bodies produce byte-identical output to before.

## Verification (real results, Go 1.26)

`go test ./pkg/service/contract/` — all pass, including three new tests:

- `TestHTTPDocToOpenAPI_TopLevelArrayResponse` — array response → `type: array` with an `object` item schema whose properties are inferred.
- `TestHTTPDocToOpenAPI_TopLevelArrayRequest` — array request body → `type: array` with `number` items.
- `TestHTTPDocToOpenAPI_ObjectResponseUnchanged` — object root still yields an `object` schema (no regression).

Each new test also asserts the generated document passes `kin-openapi` validation (`validateSchema`), so the output is guaranteed to be a valid OpenAPI spec. `gofmt`/`go vet` clean.

Before this change the array cases returned the `cannot unmarshal array into ... map[string]interface {}` error; after, they produce valid schemas.

### DIFF
--- a/pkg/models/openapi.go
+++ b/pkg/models/openapi.go
@@ -108,6 +108,7 @@ type Schema struct {
 	Type       string                            `json:"type" yaml:"type"`
 	Properties map[string]map[string]interface{} `json:"properties,omitempty" yaml:"properties,omitempty"`
 	Items      *Schema                           `json:"items,omitempty" yaml:"items,omitempty"`
+	Nullable   bool                              `json:"nullable,omitempty" yaml:"nullable,omitempty"`
 }
 
 type ParamSchema struct {
--- a/pkg/models/openapi.go
+++ b/pkg/models/openapi.go
@@ -95,8 +95,8 @@ type RequestBody struct {
 }
 
 type MediaType struct {
-	Schema  Schema                 `json:"schema" yaml:"schema"`
-	Example map[string]interface{} `json:"example" yaml:"example"`
+	Schema  Schema      `json:"schema" yaml:"schema"`
+	Example interface{} `json:"example" yaml:"example"`
 }
 
 type ResponseItem struct {
@@ -106,7 +106,8 @@ type ResponseItem struct {
 
 type Schema struct {
 	Type       string                            `json:"type" yaml:"type"`
-	Properties map[string]map[string]interface{} `json:"properties" yaml:"properties"`
+	Properties map[string]map[string]interface{} `json:"properties,omitempty" yaml:"properties,omitempty"`
+	Items      *Schema                           `json:"items,omitempty" yaml:"items,omitempty"`
 }
 
 type ParamSchema struct {
--- a/pkg/service/contract/contract.go
+++ b/pkg/service/contract/contract.go
@@ -48,8 +48,11 @@ func New(logger *zap.Logger, testDB TestDB, mockDB MockDB, openAPIDB OpenAPIDB, 
 func (s *contract) HTTPDocToOpenAPI(logger *zap.Logger, custom models.HTTPDoc) (models.OpenAPI, error) {
 
 	var err error
-	// Convert response body to an object (root may be an object or an array)
-	var responseBodyObject interface{}
+	// Convert response body to an object (root may be an object or an array).
+	// Defaults to an empty object so a body-less doc still serializes its
+	// example as {} rather than null, as it did before array roots were
+	// supported.
+	var responseBodyObject interface{} = map[string]interface{}{}
 	if custom.Spec.Response.Body != "" {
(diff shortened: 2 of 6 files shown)


slayerjain:
@chiliec please accept the CLA.

chiliec:

slayerjain:
Thanks @chiliec !

PR #4448, opened 2026-08-19
chore(proxy): remove userspace TCP_INFO network-I/O accounting

ayush3160:
## Related issue
Follow-up cleanup to #4446 (kernel app network-I/O counter drain).

## Description
The app's wire volume (rx = ingress, tx = egress, pre-dedup) is now metered **in-kernel** by the eBPF app network-I/O counter (`keploy_netio_bytes`, drained by `StartKernelNetioDrain` from #4446). This PR removes the older **userspace** path that read each proxied socket's `TCP_INFO` byte totals at connection teardown.

Removed:
- `proxy.go` — the `origSrcConn` raw-socket capture and its `RecordConnNetworkIO` teardown call.
- `incoming/http.go`, `incoming/gRPC/record.go` — the app-facing-socket `RecordConnNetworkIO` calls (and the now-unused `ossproxy` import).
- `netstat.go` — reduced to the `networkIOSink` seam only (`SetNetworkIOSink`); dropped `RecordConnNetworkIO`, `readSocketBytes`, the `rawConn` helper and the `kernelNetioActive` gate.
- `netstat_other.go` — the `RecordConnNetworkIO` non-Linux stub (and `net` import).

The kernel counter is continuous and single-counted, so it **replaces** the teardown-time `TCP_INFO` reads outright rather than running alongside them (no double-book, no gate needed).

## ⚠️ Merge ordering & tradeoff
- **Based on `feat/netio-kernel-drain` (#4446)** — merge this only **after #4446 is merged and runtime-validated** on the target kernels. Retarget to `main` at that point.
- On a kernel **without BTF** the eBPF netio programs do not load, and with this change there is **no userspace fallback** — network I/O is simply not metered there. CPU / RAM / disk metering are unaffected.

## Testing
- `go build ./pkg/agent/hooks/... ./pkg/agent/proxy/...` (linux) — passes.
- `CGO_ENABLED=0 GOOS=darwin go build ./pkg/agent/proxy/...` — passes.
- `gofmt` clean; no dangling references to removed symbols.

## Checklist
- [x] Builds on linux and (cgo-off) darwin
- [x] `gofmt` clean
- [x] Signed-off (DCO)

### DIFF
--- a/pkg/agent/proxy/incoming/gRPC/record.go
+++ b/pkg/agent/proxy/incoming/gRPC/record.go
@@ -12,7 +12,6 @@ import (
 	"go.keploy.io/server/v3/pkg"
 	Utils "go.keploy.io/server/v3/pkg/agent/hooks/conn"
 	"go.keploy.io/server/v3/pkg/agent/memoryguard"
-	ossproxy "go.keploy.io/server/v3/pkg/agent/proxy"
 	"go.keploy.io/server/v3/pkg/models"
 	"go.uber.org/zap"
 	"golang.org/x/net/http2"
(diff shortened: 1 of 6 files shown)


ayush3160:
Superseded. The app network-I/O kernel counter (`keploy_netio_bytes`) already exists **entirely in the enterprise agent** (merged as enterprise #2378: `proxyless.c` fentry/fexit on tcp/udp sendmsg/recvmsg → `account_netio`, drained by `startNetworkIOAccounting` → resourceio → k8s-proxy). Rather than duplicate the counter in the shared eBPF object + OSS loader/drain (what these PRs did), we're extending the existing enterprise counter to the sidecar/k8s-proxy mode behind an opt-in flag, and removing the OSS metering path entirely. Keeping network-I/O metering out of OSS is the explicit goal. See the enterprise feature PR + the OSS TCP_INFO removal PR (to follow).

PR #4449, opened 2026-08-19
chore(proxy): remove userspace TCP_INFO network-I/O accounting

ayush3160:
## Description
App network-I/O metering is moving to a **kernel-sourced eBPF counter owned entirely by the enterprise agent** (`keploy_netio_bytes`; `fentry`/`fexit` on tcp/udp `sendmsg`/`recvmsg`, gated by cgroup, drained behind an opt-in flag). OSS should carry **none** of the metering mechanism, so this removes the userspace TCP_INFO path that read each proxied socket at teardown.

Removed:
- **`pkg/agent/proxy/netstat.go` + `netstat_other.go`** (deleted) — `RecordConnNetworkIO`, `readSocketBytes`, `rawConn`, and the `SetNetworkIOSink` / `networkIOSink` seam.
- **`proxy.go`** — the `origSrcConn` raw-socket capture and its teardown call.
- **`incoming/http.go`, `incoming/gRPC/record.go`** — the app-facing-socket `RecordConnNetworkIO` calls and the now-unused `ossproxy` import.

The sink had exactly one consumer — the enterprise agent — which no longer installs it (see the companion enterprise PR). Nothing else in OSS referenced these symbols.

## Impact
- On capture modes that used this path (standard OSS-proxy sidecar), network I/O is no longer metered. **CPU/RAM/disk are unaffected.**
- Kernel-sourced metering (enterprise) requires BTF (`fentry`/`fexit`) and does not run on kernels without it.

## Ordering
Independent of the enterprise PR (enterprise already stopped calling `SetNetworkIOSink`). Enterprise bumps `go.keploy.io/server/v3` to include this after it merges.

## Testing
- `go build ./pkg/agent/proxy/...` on linux and `CGO_ENABLED=0 GOOS=darwin` — both pass.
- `gofmt` clean; no dangling references to the removed symbols.

## Checklist
- [x] Builds on linux and (cgo-off) darwin
- [x] `gofmt` clean
- [x] Signed-off (DCO)


charankamarapu (approved):
LGTM

PR #4436, opened 2026-08-18
fix: avoid treating Postman folders as requests

Utkarshpandey0001:
## Describe the changes that are made

- Fixed Postman collection item parsing so folder entries are not also treated as API request entries.
- Updated `ItemsContainer.UnmarshalJSON` to classify each Postman item based on its shape:
  - items with nested `item` are treated as folders
  - items with `request` are treated as API requests
- Added regression tests to ensure:
  - Postman folders are not classified as request items
  - root-level Postman requests are still imported correctly

## Links & References

**Closes:** #4434

### 🔗 Related PRs
- NA

### 🐞 Related Issues
- #4434

### 📄 Related Documents
- NA

## What type of PR is this? (check all applicable)

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?

- [x] 🙅 no, because they aren't needed

## Added comments for hard-to-understand areas?

- [x] 🙅 no, because the code is self-explanatory

## Added to documentation?

- [x] 🙅 no documentation needed

## Are there any sample code or steps to test the changes?

- [x] 👍 yes, mentioned below

Tested with:

```bash
go test ./pkg/service/import
go test ./pkg/service/tools ./pkg/service/import
```

Screenshot 2026-08-18 at 12 34 06 PM

Manual validation:

```bash
cd /private/tmp/keploy-postman-var-repro
printf 'yes\n' | /private/tmp/keploy-bin/keploy import postman --path collection.json --disable-ansi
```

Before this change, the command failed with:

```text
ERROR URL is empty {"testItem": "Users"}
ERROR failed to import Postman collection {"error": "URL is empty"}
```

After this change, the Postman collection imports successfully.

## Self Review done?

- [x] ✅ yes

## Any relevant screenshots, recordings or logs?

Postman import now succeeds with the folder-based collection repro.
Screenshot 2026-08-18 at 12 54 36 PM

## 🧠 Semantics for PR Title & Branch Name

PR title:

```text
fix: avoid treating Postman folders as requests
```

Branch name:

```text
fix/postman-folder-import
```

## Additional checklist:

[...]

PR #4450, opened 2026-08-19
test(contract): pin the inferred item property type for array bodies

Utkarshpandey0001:
> **Maintainer note:** this PR originally carried a real fix for #4445. It landed first via #4447, so on rebase the production hunks were already upstream and dropped out. Retitled and rewritten to describe what the branch now actually contains. See the comment below for the full timeline — the original work was real and was submitted first.

Keeps the one assertion from this branch that is stronger than what #4447 landed, and drops the test file that otherwise duplicates `TestHTTPDocToOpenAPI_TopLevelArrayResponse`.

`TestHTTPDocToOpenAPI_TopLevelArrayResponse` checked only that the item schema *has* an `id` property. This pins its inferred **type**:

```go
idProp, ok := got.Items.Properties["id"]
...
if idProp["type"] != "number" {
```

### Why it earns its place

`validateSchema` already catches a wrong *scalar* type — flip `getType`'s `float64` case to `"string"` and kin-openapi rejects the example. It cannot catch `number` vs `integer`, because `1` validates as either. This is the only check in the package that fails on that drift:

| mutation in `utils.go` | validateSchema | this assertion |
|---|---|---|
| `float64 -> "string"` | fails | fails |
| `float64 -> "integer"` | **passes** | **fails** |

Uses `t.Errorf` rather than `Fatalf` so `validateSchema` still runs and reports alongside it.

## Links & References

**Refs:** #4445 — fixed by #4447; this PR does not close it.

### 🔗 Related PRs
- #4447

### 🐞 Related Issues
- #4445

- [x] ✅ Test

## Added e2e test pipeline?

```
go test ./pkg/service/contract/ -run TopLevelArrayResponse -v
```

To confirm the assertion is load-bearing, change `case float64: return "number"` to `return "integer"` in `pkg/service/contract/utils.go` — the test fails on this assertion while `validateSchema` still passes.

## Self Review done?
- [x] ✅ yes


slayerjain:
Hi @Utkarshpandey0001 — you were first here, your fix was real, and git ate it on rebase. I want to lay that out properly, because from where you're sitting this PR now looks like you submitted a test and called it a fix, and that isn't what happened.

The timeline, from this PR's own history:

| when (UTC) | what |
|---|---|
| **2026-08-19 09:28** | you push `d2aeff94` — a real fix: `pkg/models/openapi.go` (+4/-3), `pkg/service/contract/contract.go` (+14/-27), `pkg/service/contract/utils.go` (+53/-53), plus your repro test |
| 2026-08-20 11:50 | #4447 merges, fixing the same issue |
| 2026-08-20 12:24 | this branch is force-pushed — 34 minutes later |

That rebase was onto a `main` that already contained the equivalent change, so git correctly dropped every hunk of yours that was already upstream. All that survived was the test file. Nothing was ignored or overridden by a human — the fix was simply already there by the time the rebase ran.

I compared your `d2aeff94` against what landed. Your `SchemaFromValue` was arguably a cleaner shape than the `ExtractVariableTypes` + `SchemaForBody` split that's on `main` now. The reason there's nothing left to merge is only that #4447 went further in one direction yours hadn't reached yet: JSON nulls. `[{"id":1,"deleted_at":null}]` — a list endpoint with a nullable column — still aborted at `validateSchema` with `Value is not nullable` until that was handled, and both your version and the original #4447 hit it. So the production side is genuinely closed.

**One piece of your work is better than what landed, and I've kept it.** Your test asserted the inferred *type* of the item property, where the merged test only checks the key exists. That turns out to be the only check in the package that catches `number` drifting to `integer` — `validateSchema` can't, because `1` validates as either:

| mutation in `utils.go` | validateSchema | your assertion |
|---|---|---|
| `float64 -> "string"` | fails | fails |
| `float64 -> "integer"` | **passes** | **fails** |

I've folded that into `TestHTTPDocToOpenAPI_TopLevelArrayResponse` and dropped the now-duplicate file, then retitled this PR and rewrote the description so the history says what the branch actually does. `Closes: #4445` became `Refs:` — that issue was fixed by #4447 and shouldn't be auto-closed from here. Your authorship stays on the PR.

Two things worth your time, if you want them:

**There's a real bug sitting in the code you were already reading.** `getType` in `pkg/service/contract/utils.go` has a `case int, int32, int64: return "integer"` branch that is dead — every caller is fed `encoding/json` output, and without `Decoder.UseNumber()` every JSON number arrives as `float64`. So **every integer field in every generated Keploy contract is typed `number`**, which OpenAPI codegen renders as float/double. It's unclaimed and it's yours if you want it; comment on this PR and I'll open the issue and assign it to you.

**And the process failure was ours, not yours.** You filed #4445 and started work, and nothing in the issue signalled that to anyone else. Two contributors then solved the same problem in parallel. That's the third time this has happened in the current queue and we're fixing the assignment discipline that allowed it. Sorry you absorbed the cost of it — for what it's worth, #4436 was a clean merge and this one found a genuine gap in our test coverage on the way past.

slayerjain:
Thanks @Utkarshpandey0001 !

PR #4451, opened 2026-08-19
fix(agent): bind control-plane HTTP server to loopback only

### LINKED ISSUE #4394: Unauthenticated agent HTTP API exposes TLS session keys and allows session disruption

### 👍 Current behavior

reported on 13 June 2026 via email:

I found a missing authentication issue in the keploy agent HTTP server that allows any local or network-adjacent process to steal live TLS session keys, inject mock data, and abort recording sessions.

When keploy starts, it spawns a `keploy agent` subprocess that binds an HTTP control server. The server uses `fmt.Sprintf(":%d", port)` (all interfaces) in `pkg/agent/routes/server.go:18`, and the chi router in `pkg/agent/routes/record.go` registers all endpoints without any authentication middleware.

The most critical endpoint is `GET /agent/pcap/keylog`, which streams live NSS keylog lines to any subscriber. These are the TLS session keys generated by `crypto/tls` as the agent MITM-proxies the application's HTTPS and TLS database connections. The `KeyLogWriter()` sink is wired unconditionally into every TLS config the proxy constructs (proxy.go lines 1771 and 1977), so keys flow even without `--opportunistic-tls-intercept`. An attacker who subscribes to this stream and also captures the `/agent/pcap/traffic` pcap stream can decrypt all TLS traffic the application under test generated during a recording session using a standard tool such as Wireshark.

Additional endpoints allow unauthenticated session termination (`POST /agent/stop`) and mock injection (`POST /agent/storemocks`).

### 👟 Steps to Replicate

I confirmed this by building the project's own e2e test harness (tests/e2e/agent-ready-gated-on-ca/harness/main.go), which uses the real production route handlers. A `curl` request to `/agent/pcap/keylog` returns `HTTP 200 OK` with no credentials; `/agent/stop` is processed without challenge.

In Docker mode the agent port is published as `[0.0.0.0](http://0.0.0.0/):<AgentPort>:<AgentPort>` (docker.go line 560), exposing it to any host reachable peer that can identify the dynamic port. In native mode the bind address is `0.0.0.0` rather than `127.0.0.1`, making it reachable by all local processes including other container workloads sharing the host network.

Suggested fix: bind the agent HTTP server to `127.0.0.1` (or the loopback of the relevant network namespace) by changing `fmt.Sprintf(":%d", port)` to `fmt.Sprintf("127.0.0.1:%d", port)` in `pkg/agent/routes/server.go`. For the Docker container case, update the agent URI construction in `pkg/platform/http/agent.go:1193` (currently `http://localhost:%d/agent`) to reflect the same loopback-only bind, and remove the agent port from the `ports` list in `docker.go:559-561` since only the Keploy CLI container needs to reach it. Alternatively, add a shared secret (generated at agent-start time and passed via CLI flag or environment variable) that the CLI attaches to all agent requests and the server validates.

### 📜 Logs (if any)

_No response_

### 💻 Operating system

Linux

-

_No response_

_No response_

### 🧱 Your Environment

_No response_

### 🎲 Version

v3.5.65

### 📦 Repository

keploy

_No response_

DearMoon50:
## Summary

The agent's control-plane HTTP server (`pkg/agent/routes`) has no authentication, and until now bound `0.0.0.0`. That means:
- `GET /agent/pcap/keylog` — streams live NSS keylog lines (TLS session keys) to any subscriber
- `POST /agent/stop` — terminates the recording/replay session
- `POST /agent/storemocks` — injects mock data into the session

were reachable by any local process (native mode) or any host that could reach the published Docker port (docker mode, published as `0.0.0.0:port:port`).

This implements the loopback-binding half of the fix suggested in #4394:

- `pkg/agent/routes/server.go` — extracted address selection into a small `agentBindAddr(port, isDocker)` helper. Native mode now binds `127.0.0.1` only. Docker mode still binds every interface *inside* the container, since the docker userland-proxy/NAT needs a listener on the container's non-loopback interface to forward the published port into — exposure to the host network is closed at the publish step instead.
- `cli/agent.go` — threads the already-available `isDocker` flag through to `StartAgentServer`.
- `pkg/platform/docker/docker.go` (compose-based agent service) and `pkg/platform/docker/util.go` (5 `docker run` alias builders, one per OS/context) — publish the agent port as `127.0.0.1:<port>:<port>` instead of `<port>:<port>`, so it's reachable only from the host's own loopback. The proxy port is untouched (it legitimately needs to stay open to app containers).

I did not implement the shared-secret authentication alternative also suggested in the issue — happy to add it in a follow-up if maintainers want defense-in-depth beyond the network-level fix, but loopback-only binding closes the reported exposure on its own.

Note: I saw @Hardik180704's scoping question on the issue (loopback vs. shared secret vs. both) went unanswered for a few weeks, so I went ahead with the loopback-only approach as it's the minimal fix that fully addresses the reported reachability issue. Happy to adjust or coordinate if there's already work in progress.

Fixes #4394

## Test plan

- [x] `go build ./pkg/agent/routes/... ./pkg/platform/docker/...` — compiles clean
- [x] `go test ./pkg/agent/routes/...` — new `TestAgentBindAddr_NativeModeIsLoopbackOnly` / `TestAgentBindAddr_DockerModeBindsAllInterfaces` pass; existing `TestStartAgentServer_ServesAfterTransientPortConflict` updated for the new signature and passes
- [x] `go test ./pkg/platform/docker/... -run TestGenerateKeployAgentService` — new `TestGenerateKeployAgentService_PublishesAgentPortLoopbackOnly` passes, confirms the agent port publishes loopback-only while the proxy port stays on all interfaces
- Manually verified: 3 pre-existing test failures in these packages (`TestListenWithRetry_*`, `TestGenerateKeployAgentService_AddsDebugFileMountAndEnv`) reproduce identically on unmodified `main` in this Windows dev environment — unrelated to this change (Windows-specific `EADDRINUSE` matching / path handling)

### DIFF
--- a/cli/agent.go
+++ b/cli/agent.go
@@ -118,7 +118,7 @@ func Agent(ctx context.Context, logger *zap.Logger, conf *config.Config, service
 						logger.Info("running as a Kubernetes DaemonSet agent; not starting the control-plane HTTP server (push architecture has no inbound HTTP consumers)")
 						return
 					}
-					routes.StartAgentServer(ctx, logger, p, router)
+					routes.StartAgentServer(ctx, logger, p, isDocker, router)
 				}
 			}()
 
--- a/pkg/agent/routes/server.go
+++ b/pkg/agent/routes/server.go
@@ -45,9 +45,8 @@ const bindRetryInterval = 500 * time.Millisecond
 // (see GenerateKeployAgentService, which now publishes only to the host's
 // loopback).
 func StartAgentServer(ctx context.Context, logger *zap.Logger, port int, isDocker bool, router http.Handler) {
-	logger.Info("Starting Agent's HTTP server on :", zap.Int("port", port))
-
 	addr := agentBindAddr(port, isDocker)
+	logger.Info("Starting Agent's HTTP server", zap.String("addr", addr))
 	srv := &http.Server{
 		Handler: router,
 	}
@@ -75,12 +74,12 @@ func StartAgentServer(ctx context.Context, logger *zap.Logger, port int, isDocke
 	// this agent's startup; then serve on the acquired listener.
 	listener, err := listenWithRetry(srvCtx, logger, addr, bindRetryBudget, bindRetryInterval)
 	if err != nil {
-		logger.Error("failed to start HTTP server; verify port availability and network configuration", zap.Error(err))
+		logger.Error("failed to start HTTP server; verify port availability and network configuration", zap.String("addr", addr), zap.Error(err))
 		return
 	}
 
 	if err := srv.Serve(listener); err != nil && err != http.ErrServerClosed {
-		logger.Error("failed to start HTTP server; verify port availability and network configuration", zap.Error(err))
+		logger.Error("failed to start HTTP server; verify port availability and network configuration", zap.String("addr", addr), zap.Error(err))
 		return
 	}
 	logger.Info("HTTP server stopped")
--- a/pkg/agent/routes/server.go
+++ b/pkg/agent/routes/server.go
@@ -31,10 +31,23 @@ const bindRetryBudget = 90 * time.Second
 // busy-spinning.
 const bindRetryInterval = 500 * time.Millisecond
 
-func StartAgentServer(ctx context.Context, logger *zap.Logger, port int, router http.Handler) {
+// StartAgentServer binds the agent's control-plane HTTP server. It is
+// unauthenticated (see /agent/pcap/keylog, /agent/stop, /agent/storemocks),
+// so the bind address must never be reachable by anything other than the
+// local keploy CLI that owns this agent.
+//
+// In native mode the agent and CLI share a network namespace, so loopback is
+// both sufficient and necessary: binding 0.0.0.0 would let any other local
+// process reach it. In docker mode the CLI runs on the host and reaches the
+// agent through its published port, so the server must stay reachable on the
+// container's non-loopback interface for the docker proxy to forward into —
+// exposure to the host network is instead closed off at the publish step
+// (see GenerateKeployAgentService, which now publishes only to the host's
+// loopback).
+func StartAgentServer(ctx context.Context, logger *zap.Logger, port int, isDocker bool, router http.Handler) {
 	logger.Info("Starting Agent's HTTP server on :", zap.Int("port", port))
 
-	addr := fmt.Sprintf(":%d", port)
+	addr := agentBindAddr(port, isDocker)
 	srv := &http.Server{
 		Handler: router,
 	}
@@ -71,6 +84,19 @@ func StartAgentServer(ctx context.Context, logger *zap.Logger, port int, router 
 		return
 	}
 	logger.Info("HTTP server stopped")
+}
+
+// agentBindAddr picks the agent control-plane server's bind address. Native
+// mode binds loopback-only since the endpoints (notably /agent/pcap/keylog,
+// which streams live TLS session keys) carry no authentication. Docker mode
+// must still bind every interface inside the container so the docker proxy
+// can forward the published port in; that publish is restricted to the
+// host's own loopback in GenerateKeployAgentService instead.
+func agentBindAddr(port int, isDocker bool) string {
+	if isDocker {
+		return fmt.Sprintf(":%d", port)
+	}
+	return fmt.Sprintf("127.0.0.1:%d", port)
 }
 
 // listenWithRetry binds a TCP listener on addr, retrying only the transient
--- a/pkg/agent/routes/server_test.go
+++ b/pkg/agent/routes/server_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"strings"
 	"syscall"
 	"testing"
 	"time"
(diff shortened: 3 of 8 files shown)


slayerjain:
Reviewed this in depth — **the report is real, and your fix closes it correctly.** Thanks for picking up a security issue that had been sitting unscoped, and for writing down the docker-mode trade-off rather than papering over it. I've pushed test hardening on top; the production change is yours and unaltered.

## What I verified

The enumeration of bind/publish sites is **complete** — all six (the compose generator in `docker.go` plus all five `getAlias` arms in `util.go`), and there is no seventh anywhere in the repo: no `EXPOSE` of the agent port in any keploy Dockerfile, no k8s/helm manifests here, no `hostPort`/`hostNetwork` YAML. Every other `AgentPort` use is a `--port` argv, not a bind.

Attacks that came back negative:

- **`--network container:<name>`** — docker refuses the combination outright (`conflicting options: port publishing and the container type network mode`), before and after.
- **compose `network_mode: host`** on the app service — `addServiceProperty` overwrites it with `network_mode: service:keploy-agent`; the agent service never gets host networking.
- **userland-proxy disabled** — the DNAT rule carries the host-IP predicate itself (`-A DOCKER -d 127.0.0.1/32 ! -i br-... -j DNAT`), and Docker 28+ adds a `raw PREROUTING ... -j DROP` for spoofed loopback.
- **Remote docker context** — strictly *less* exposed than before.
- **Kubernetes DaemonSet** — `isDaemonSetAgent()` returns before `StartAgentServer`, unaffected.
- **The classic "`-p 127.0.0.1:` publish + `localhost` client" trap** — does not bite us. Reproduced on a host whose resolver returns `::1` first, with keploy's exact client construction plus curl, python3, node 20 and Java 21: all 200, no Happy-Eyeballs penalty.
- **The enterprise k8s path**, which was the one I most expected to break: `k8s-proxy` dials the agent cross-pod at `http://<podIP>:<agentPort>/agent` and at Service DNS, so a loopback bind there would be fatal. It isn't — every launch path (the Helm daemonset template, the sidecar args in `k8s.go`, and `keploySidecarContainer` in the webhook service) passes `--is-docker`, so the in-pod listener stays on `:port` and the cross-pod dial still works.

Three separate "this is a no-op under `--network host` / for peer containers" objections were raised during review and all refuted on baseline comparison: that exposure is bit-for-bit identical on `main`, so it is pre-existing and not something this PR introduces or worsens.

## What I pushed, and why it isn't optional

The security property you added **cannot be tested behaviourally**. `-p <port>:<port>` binds `0.0.0.0`, a strict superset of `127.0.0.1`, so the agent stays reachable over loopback either way and every unit and e2e lane passes identically. Measured: reverting any one of your five `-p 127.0.0.1:` edits — or all five at once — left `go test ./...` **fully green**. Four of the five arms are also unreachable at runtime on Linux, and the macOS/Windows lanes only run `go build`. So a rebase conflict or a string-concatenation reformat silently republishes the unauthenticated keylog endpoint on every host interface, and nothing says a word.

My first attempt was an AST walk requiring each arm to *contain* the loopback literal. An independent review took it apart: a presence check catches an arm that forgets to publish, but not an arm that publishes correctly and then appends a second, bare `-p <port>:<port>`. Four separate edits re-exposed the port on `0.0.0.0` with the guard green, on exactly the four platforms it was written to protect. The verdict was that I was writing an increasingly clever source-text linter to compensate for a design problem in `util.go`, and that was right.

So the five copies of the fragment are now one function:

```go
// agentPortPublish is the ONE place the agent's control-plane port is published to the host.
func agentPortPublish(agentPort uint32) string {
	return fmt.Sprintf(" -p 127.0.0.1:%d:%d", agentPort, agentPort)
}
```

One exact string with one behavioural test, and the walk asserts every alias-returning branch routes through it **and** builds no `-p` fragment of its own — exact here, since the proxy and app publishes are assembled before the switch. The generated alias is byte-identical before and after; I captured it and diffed rather than assuming. Plus a runtime check on the linux arm asserting both halves, and the same vacuity guard the sibling walk uses so a walk that stops seeing the switch fails loudly instead of passing on zero branches.

Two more holes came out of mutating the tests rather than the code:

- `TestAgentBindAddr_DockerModeBindsAllInterfaces` asserted *not* `127.0.0.1:`, which passes for everything else — including `"localhost:%d"`, which binds loopback and would break docker mode while passing a test named *BindsAllInterfaces*, and `"banana:%d"` / `""` which cannot bind at all. 1 of 7 mutants died. It now asserts the value.
- The compose publish check was presence-only, so a ports block carrying both the loopback mapping *and* a bare one satisfied it.

Every attack now dies — the four that walked past the first guard (extra bare publish in the concat chain; `alias +=` after the helper; a runtime `strings.Replace` stripping the scope; a hand-rolled sixth platform arm), all five prefix reverts, and the vacuity case. Two earlier defects in my own guard surfaced only by mutating it: it descended into nested blocks, so the colima `if` body's literal answered for the enclosing case body and reverting `util.go:340` or `:480` alone still passed; and it pinned a representation, failing `0.0.0.0:%d`, which is byte-identical in effect to `:%d`.

I also folded the walk into a shared helper rather than shipping a second 53-line copy of the one already in `upstream_tls_flags_test.go` — the two had drifted within the hour.

Also moved the startup log to print the address actually bound; it read `Starting Agent's HTTP server on :` with the port as a field, which stopped being true for native mode.

## On CI

Heads up that the green tick on this PR meant very little: it showed **three checks** — `cla-assistant`, `greeting`, `DCO`. Ten workflow runs sat in `action_required` since 19 Aug, unapproved, publishing no check-runs; and `prepare_and_run.yml` gates the docker/macOS/Windows legs on `!head.repo.fork` even after approval. So none of the lanes that exercise this code had run. I've pushed this head to a branch in this repo — #4487 — so the pipeline actually runs. That PR is CI-only; **merge this one, not that one.** I'll report what it says.

## Not blocking, for a follow-up

`opts.AgentURI` is hardcoded `http://localhost:%d/agent`. Nothing is broken today, but making it explicitly `127.0.0.1` in native mode removes the whole IPv4/IPv6 class of concern, including hand-written Node ≤18 glue against `KEPLOY_MOCK_AGENT`.

Your call on the shared-secret follow-up you offered — loopback-only genuinely does close the reported exposure on its own, so I'd treat auth as defence-in-depth rather than a gap.

slayerjain:
Thanks @DearMoon50 !

PR #4460, opened 2026-08-20
feat(agent): resolve JVM destinations by peer address for node-scoped hooks

praagyajain:
Stacked on `feat/jvm-redirect-hooks-override`. Part of the eBPF-free DaemonSet capture work (enterprise `feat/ds-jvm-nodescope`, k8s-proxy `feat/ds-jvm-inject`).

## Why

The sidecar registry resolves a JVM's real destination by **source port** alone. That is unambiguous in a pod, where exactly one JVM talks to the agent over loopback.

It stops being unambiguous the moment the agent is node-scoped: registrations arrive over TCP from many pods, and two pods can easily present the same ephemeral source port. Resolving on port alone would hand one pod's destination to another pod's connection — a mock recorded against the wrong dependency, which is worse than a missing one.

## What

Adds an optional `GetByPeer(ctx, peerIP, srcPort)` alongside the existing `Get`, keying on `(peer, srcPort)`.

Deliberately **not** added to the `coreAgent.DestInfo` interface: that is a cross-repo contract whose changes have to be released in lockstep, and a sidecar needs nothing here. A proxy that knows its peer prefers `GetByPeer`; one that does not keeps calling `Get` and behaves exactly as before.

## Risk

Additive. No existing caller changes behaviour.


praagyajain:

PR #4462, opened 2026-08-21
fix(http): match query-param values, not just keys, in schema matching

charankamarapu:
- HTTP replay's strict schema pass (`SchemaMatch` in `pkg/agent/proxy/integrations/http/match.go`) gated query params with `MapsHaveSameKeys`, which compares only the **key set / counts** of the recorded mock's `URLParams` against the live request's parsed query — it **never compared the values**. So a mock recorded for `/search?id=A` would satisfy a live request for `/search?id=B`, and the wrong recorded response could be served for a genuinely different query.
- Added `QueryParamsMatch(mockParams map[string]string, reqQuery url.Values, urlNoise []string) bool`, which keeps the **exact same bidirectional key-set contract** (ignoring `keploy`-prefixed keys) **and additionally compares each shared key's value**. Values are compared in the recorded form — `pkg.URLParams` stores each param `", "`-joined, so the live side is joined the same way (`strings.Join(reqQuery[key], ", ")`).
- `urlNoise` is applied to **both** sides before the value comparison, mirroring `MatchURLPath`'s compile + `ReplaceAllString`-to-`{{keploy.urlnoise}}` approach, so a param value covered by url noise still matches while a genuinely different value does not.
- Replaced the `MapsHaveSameKeys` call at the query-param gate with `QueryParamsMatch`. `MapsHaveSameKeys` had no remaining callers and was removed; the doc comments that referenced it were updated to point at `QueryParamsMatch`.

**Reproduction (before fix):** a request `?id=B` matched a mock recorded with `?id=A`. **After fix:** `?id=A` no longer satisfies `?id=B`; `?id=B` still satisfies `?id=B`; a param covered by `urlNoise` still matches despite a differing value; different keys stay distinguished.

## Links & References

**Closes:** NA (bug found via query-param mismatch during replay)
- NA

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

- [x] 👍 yes

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

- [x] 👍 yes, mentioned below

New table-driven regression test `TestQueryParamsMatch_ValueSensitive` in `match_test.go` (uses the existing `newHTTP()` helper) asserts:
- a mock with `?id=A` does **not** satisfy a request `?id=B` (previously it did);
- a mock `?id=B` **does** satisfy `?id=B`;
- a param whose value is covered by `urlNoise` still matches despite differing value;
- a differing value not covered by noise still fails;
- different keys are distinguished, extra keys fail, and multi-value params compare in the recorded joined form.

```
go test ./pkg/agent/proxy/integrations/http/
ok  	go.keploy.io/server/v3/pkg/agent/proxy/integrations/http	0.783s
```

Full package (171 test cases) passes, and `go build ./...` at repo root is clean.

## Self Review done?
- [x] ✅ yes

- NA


KC1706:
Reproduced the bug and checked if the pr solves the issue or not.

1) match=true  → test FAILS → the bug is reproduced.  id=B  was accepted as a match for a mock recorded with  id=A
image

2)  id=A  vs  id=B  →  false  (rejected),  id=B  vs  id=B  →  true  (served). Bug fixed, legitimate hit preserved
image


3)11/11 pass, including the case literally named different_value_does_not_match_(the_bug)
image

KC1706:
edge cases i found:-

1) Repeated query params now fail if their order changes
image

2) The request is identical except for _ts, a timestamp that changes every run. The key set is the same (page, _ts), so under the old keys-only logic this matched — the test passed.After the fix, with no noise configured, it misses (false). That's a green test turning red on upgrade, with no code change by the user. The behavior is more correct, but it's a surprise.The fix does have a bonus finding: the user's intuitive noise pattern _ts=\d+ doesn't work (false) — because url-noise is matched against the value only (1700005555), not key=value. Only \d+ works (true).
image

KC1706:
proposed solutions:
1) single-value params take the same exact fast path (unchanged). When values differ, it splits both sides into members, masks each with url-noise, trims, sorts both, and compares — so a reorder of the same values matches, a real difference still fails. 
2)change doc description: Query-param values are now matched on replay — mark volatile params (timestamps, tokens, cursors) as url-noise or they'll miss.url-noise patterns match the param's value, not key=value — write \d+, not _ts=\d+.

KC1706:
i have commits that works , below is the evidence for that:
image

KC1706:

KC1706:
recheck

KC1706 (approved):
Approving. Reviewed the full stack (value matching + order-independence + noise memoization + query auto-dynamic fallback) against head `b40c3411` and verified locally: `go build`, `go vet`, and the full `pkg/agent/proxy/integrations/http` suite all pass; CI is green except the CLA account gate.

**What I checked**
- **Value match** (`8b8ab1de`): correct root-cause fix — a mock recorded for `?id=A` no longer satisfies `?id=B`, while the bidirectional key-set contract and `keploy`-prefixed-key ignoring are preserved.
- **Order-independence** (`d3a2aa9d`): repeated params compared as a sorted set (mirrors `CompareMultiValueHeaders`), so `?tag=a&tag=b` matches `?tag=b&tag=a`; single-value params keep the exact fast path.
- **Noise memoization** (`d488a8fe`): `compileURLNoise` caches compiled patterns in a `sync.Map`, compiles lazily only on drift, skips invalid patterns safely (nil-guarded), and is backed by a benchmark. Thread-safe.
- **Query auto-dynamic** (`1ba178eb`): well-calibrated — probed directly: two long timestamps/UUIDs match only in pass 2, `page=2` vs `page=3` and plain words still fail (the `minDynamicQueryDigits=10` threshold protects small ids), and the exact pass is unaffected. Opt-out via `DisableAutoURLDynamic`.
- Signature change is fully contained: `QueryParamsMatch` has a single non-test caller (`SchemaMatch`), and the async path routes through it correctly.

**Non-blocking notes (for a follow-up, not blocking merge)**
1. Behavior change worth a changelog line: with auto-dynamic on (default), two *different* long-numeric/UUID query values now match in the fallback pass without noise config — intended, and opt-out-able.
2. Repeated params under the auto-dynamic fallback can misalign after sort when each side carries a dynamic member (already documented in-code as best-effort; the single-value case is exact).

Nice work extending this well past the original one-line fix.

PR #4464, opened 2026-08-21
fix(mysql): serve an error packet on mock-miss instead of tearing down the connection

charankamarapu:
During replay, when a MySQL command has **no matching mock**, the replayer returned `models.ErrNoMockMatched`. That unwound to `errCh <- err` in `replay.go`, the command-phase goroutine returned, and the client's **entire TCP connection was dropped**. A single drifted or unrecorded query therefore took the whole connection down — a multi-second app outage, and with pooled drivers (e.g. Rails/Makara) a connection-pool blacklist cascade. This pattern was observed hundreds of times in a real customer's replay debug bundles.

### The HTTP contrast
The HTTP replayer (`pkg/agent/proxy/integrations/http/decode.go`) already handles this gracefully: its serve-without-mock path writes an error *response* for the one unmatched call and **keeps the connection open**. MySQL was asymmetric — one miss = connection teardown.

### The fix
On a command-phase mock-miss, `simulateCommandPhase` now:
- **surfaces the miss to the test report out-of-band** via the `MockMismatchReporter` installed on the context (the same sink a returned `ErrNoMockMatched` feeds through the proxy) — so the test still fails with full diagnostics;
- **writes a single MySQL ERR packet** (`1105 ER_UNKNOWN_ERROR`, SQLSTATE `HY000`, message `keploy: no matching mock for query`) as the degraded response for that one query; and
- **continues the command loop** to read the next command.

### Conservative fallbacks (nothing silently swallowed)
- If **no mismatch reporter** is installed (older agent / non-test path), the original wrapped `ErrNoMockMatched` is returned exactly as before.
- **No-response commands** (`COM_STMT_CLOSE` / `COM_STMT_SEND_LONG_DATA` / `COM_QUIT`) report the miss and keep the connection open *without* writing an ERR packet (a write would desync the stream).
- A **genuine I/O or encode failure** while writing the ERR packet is still fatal to the connection.
- **Matching logic, the record path, and all other error kinds are untouched.**

## Links & References
**Closes:** NA
### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA

- [x] 🐞 Bug Fix

## Added e2e test pipeline?

> Scope note (honest): verification was done at the **replayer unit level** plus a full build — a full end-to-end record→replay cycle was **not** run as part of this PR. New replayer-level regression tests cover:
> 1. mock-miss with a reporter present → **ERR packet served + connection kept open + miss reported** (asserts the fatal teardown error does **not** propagate);
> 2. mock-miss with **no reporter** → wrapped `ErrNoMockMatched` returned (fallback preserved);
> 3. **no-response command** miss → connection kept open, no packet written, miss still reported.
>
> `go test ./pkg/agent/proxy/integrations/mysql/...` — all pass. `go build ./...` — clean.

- [x] 👍 yes

## Added to documentation?

- [x] 👍 yes, mentioned below

```
go build ./...
```

## Self Review done?
- [x] ✅ yes

- NA

### DIFF
--- a/pkg/agent/proxy/integrations/mysql/replayer/query.go
+++ b/pkg/agent/proxy/integrations/mysql/replayer/query.go
@@ -221,7 +221,7 @@ func simulateCommandPhase(ctx context.Context, logger *zap.Logger, clientConn ne
 				// next_step reuses the strict-aware guidance computed for the
 				// mismatch report above — under schemaNoiseStrict the accurate
 				// advice is "mark/learn the drifted fields", not "re-record".
-				logger.Error("Connection closing due to no matching mock found.",
+				logger.Error("No matching mock found for command.",
 					zap.Int("commands_processed", commandCount),
 					zap.String("request_type", req.Header.Type),
 					zap.String("closest_mock", miss.closestMock),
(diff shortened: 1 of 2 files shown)


chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(mysql): serve an error packet on mock-miss instead of tearing down the connection. Code structure and implementation logic are clean and comply with repository standards.

PR #4467, opened 2026-08-21
feat(cli): make the native-command platform gate an extension point

slayerjain:

`ValidateFlags` rejects a non-docker command on any platform without an in-tree interception backend:

```go
if (CommandType == Native || CommandType == Empty) &&
   !(runtime.GOOS == "linux" || (runtime.GOOS == "windows" && runtime.GOARCH == "amd64")) {
    return fmt.Errorf("non docker command not supported for OS: %s , Arch: %s", ...)
}
```

That policy is right — without it a native run reaches the `pkg/agent/hooks/others` stub and fails with `eBPF hooks are not supported on non-Linux platforms`, which tells the user nothing. But it was a hard-coded expression at the call site, so a build that *does* ship an additional backend had no way to say so.

This turns it into an extension point, with **no behaviour change on any platform**.

- New `cli/provider/hooks.go`:
  - `DefaultNativeCommandSupported(goos, goarch)` — byte-equivalent to the old condition.
  - `NativeCommandSupported` — the installed predicate.
  - `RegisterNativeCommandSupport(fn)` — installs a wider one from an importing package's `init()`; `nil` restores the default.
- `cmd.go` call site now reads `!nativeCommandSupportedHere()`.

Two deliberate choices:

**Why a package-level var + `Register…`** — it matches the extension points already in `pkg/agent/runtime_hooks.go` (`EbpfLoadedHook`, `SkipProxyListener`, `ProxyHook`, …) and `pkg/client/app/hooks.go` (`HookImpl`), including the dedicated `hooks.go` file and the `Register…` setter. A build tag can't work here (an out-of-tree module can't flip one), an interface is overkill for a bool predicate, and a config field would leak a build capability into user-facing config.

**Why the predicate takes `goos, goarch`** rather than reading `runtime.GOOS` itself — so the policy is testable for every platform from any host. Asserting against the host would only ever exercise one row on the CI runner, and could never catch an inverted arch check.

### Drive-by fix

`TestValidateFlags_HonoursCmdTypeEndToEnd` has been **failing on macOS**: two of its cases resolve to `native`, and the gate refused them for a reason that has nothing to do with what the test checks.

```
cmd_test.go:234: ValidateFlags: non docker command not supported for OS: darwin , Arch: arm64
```

It now pins the gate open; the platform policy has its own coverage in `hooks_test.go`. `go test ./cli/provider/` passes on macOS for the first time.

## Links & References

**Closes:** NA

### 🔗 Related PRs
- Consumed by the enterprise macOS-native record/replay change, which registers a predicate adding `darwin`.
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [x] 🧑‍💻 Code Refactor
- [x] ✅ Test

## Added e2e test pipeline?

Behaviour is unchanged on every platform, so there is nothing new for an e2e to exercise. The unit tests cover the platform matrix and the call site.

- [x] 👍 yes

`hooks.go` records *why* the gate exists (the `others` stub's error message) and states the concurrency contract: set from `init()`, read without synchronization.

## Added to documentation?

- [x] 👍 yes, mentioned below

```bash
go test ./cli/provider/ -run "NativeCommandSupport|HonoursCmdTypeEndToEnd" -v
go test ./cli/... -race
golangci-lint run ./cli/provider/
```

`TestDefaultNativeCommandSupported` covers linux/{amd64,arm64,386}, windows/{amd64,arm64}, darwin/{amd64,arm64} and freebsd — including that **windows/arm64 stays unsupported**, which the old inline condition also implied but nothing asserted.

## Self Review done?
- [x] ✅ yes

Reviewed independently before commit; the review is what produced the `goos, goarch` signature, the move to `hooks.go` with a `Register…` setter, the call-site test replacing a tautological one, and the macOS test fix.

Before, on macOS:
```
--- FAIL: TestValidateFlags_HonoursCmdTypeEndToEnd/config_compose_is_not_honoured
    ValidateFlags: non docker command not supported for OS: darwin , Arch: arm64
```
After:
```
ok  	go.keploy.io/server/v3/cli/provider	0.877s
```

### DIFF
--- a/cli/provider/cmd.go
+++ b/cli/provider/cmd.go
@@ -1157,7 +1157,7 @@ func (c *CmdConfigurator) ValidateFlags(ctx context.Context, cmd *cobra.Command)
 			return err
 		}
 		c.cfg.CommandType = commandType
-		if (c.cfg.CommandType == string(utils.Native) || c.cfg.CommandType == string(utils.Empty)) && !(runtime.GOOS == "linux" || (runtime.GOOS == "windows" && runtime.GOARCH == "amd64")) {
+		if (c.cfg.CommandType == string(utils.Native) || c.cfg.CommandType == string(utils.Empty)) && !nativeCommandSupportedHere() {
 			return fmt.Errorf("non docker command not supported for OS: %s , Arch: %s", runtime.GOOS, runtime.GOARCH)
 		}
 		// memory-limit non-Docker gate is applied after flag parsing below
--- a/cli/provider/cmd_test.go
+++ b/cli/provider/cmd_test.go
@@ -211,6 +211,14 @@ func TestValidateFlags_HonoursCmdTypeEndToEnd(t *testing.T) {
 		{"generated default + docker command still auto-detects", "docker compose up", "", "native", "docker-compose"},
 		{"generated default + plain command stays native", "python app.py", "", "native", "native"},
 	}
+	// Two of these cases resolve to "native", which ValidateFlags refuses on a
+	// platform with no interception backend — so on macOS this test failed for
+	// a reason that has nothing to do with what it is checking. Pin the gate
+	// open: the platform policy has its own coverage in hooks_test.go.
+	original := NativeCommandSupported
+	t.Cleanup(func() { NativeCommandSupported = original })
+	RegisterNativeCommandSupport(func(string, string) bool { return true })
+
 	for _, tc := range cases {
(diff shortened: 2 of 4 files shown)


slayerjain:
Consumer of this extension point: keploy/enterprise#2421 (macOS-native record/replay), which registers a predicate adding `darwin`. That PR is blocked on this one merging and an OSS release being cut.

Issue #4474, opened 2026-08-22
Replay reports a mock mismatch for DNS names capture deliberately never records (NXDOMAIN search-expansion)


Aditya-eddy:
Parent: #2426

## Summary

Capture deliberately **drops** non-Success DNS responses. Replay then counts those same
queries as **mock mismatches** and fails the test. The two halves disagree, and with
`ndots:5` the cost is large: every hostname emits 2–3 search-expansion misses, so one app
doing one lookup can fail an entire test set on names that never resolved in production
either.

## Evidence

Locally, after fixing the cross-namespace DNS search domain (#2428), a clean run still
failed 8/8 with **16 DNS mismatches — and every single one was a search-expanded name**:

```
6 × A     mongo.prod.svc.cluster.local.prod.svc.cluster.local.
6 × AAAA  mongo.prod.svc.cluster.local.prod.svc.cluster.local.
2 × A     openbao-prod-active.prod.svc.cluster.local.prod.svc.cluster.local.
2 × AAAA  openbao-prod-active.prod.svc.cluster.local.prod.svc.cluster.local.
```

Those doubled names are what the resolver probes first under `ndots:5`; they NXDOMAIN, and
then the absolute name succeeds. Recorded DNS mocks: **4** — exactly the successful absolute
A/AAAA lookups for those two hosts.

The capture side is explicit that this is intentional
(`enterprise/pkg/agent/proxy/dns_capture.go`):

```go
// Mirror keploy's existing DNS server: skip non-Success rcodes (NXDOMAIN,
// SERVFAIL, REFUSED). These are usually search-domain expansion noise
// and would pollute mocks.yaml without adding replay value.
if msg.Rcode != dns.RcodeSuccess { return }
```

pinned by `TestDNSCapture_SkipsNXDomainResponses`. That reasoning is sound — recording them
would multiply DNS mocks several-fold.

But the replay side (`keploy/keploy` `pkg/agent/proxy/dns.go`, `resolveUncachedDNSResponse`)
treats their absence as a defect: on a mock miss it forwards upstream, gets a negative
answer, falls through to the synthetic response and then reports
`ErrMockNotFound` + a `MockMismatchReport`.

Note the inconsistency already present in that function: **NODATA** (NOERROR + 0 answers)
relays and returns early, producing **no** mismatch. Only NXDOMAIN falls through to the
report — even though capture skips both equally.

## What NOT to do

Synthesising NXDOMAIN back to the app would regress **#2006** — the code says so:

[...]

PR #4475, opened 2026-08-22
fix: stop counting an expected-negative DNS name as a mock mismatch

Aditya-eddy:

Capture and replay disagreed about DNS names that resolve to a negative answer.

**Capture** deliberately drops non-Success rcodes (`dns_capture.go`: *"skip non-Success rcodes (NXDOMAIN, SERVFAIL, REFUSED). These are usually search-domain expansion noise and would pollute mocks.yaml without adding replay value."*), pinned by `TestDNSCapture_SkipsNXDomainResponses`. That reasoning is sound — recording them would multiply DNS mocks several-fold.

**Replay** (`resolveUncachedDNSResponse`, `MODE_TEST`) treated their absence as a defect: on a mock miss it forwarded upstream, got a definitive negative answer, fell through to the synthetic response and reported `ErrMockNotFound` + a `MockMismatchReport`. So replay failed tests for exactly the names capture was never willing to record.

The cost is large under the Kubernetes default `ndots:5`, where every hostname emits 2–3 search-expansion probes that NXDOMAIN before the absolute name resolves. Measured on a cluster run: **16 DNS mismatches, every single one a search-expanded name**, against exactly 4 recorded mocks (the successful absolute lookups):

```
6 × A     mongo.<ns>.svc.cluster.local.<ns>.svc.cluster.local.
6 × AAAA  mongo.<ns>.svc.cluster.local.<ns>.svc.cluster.local.
2 × A     openbao-active.<ns>.svc.cluster.local.<ns>.svc.cluster.local.
2 × AAAA  openbao-active.<ns>.svc.cluster.local.<ns>.svc.cluster.local.
```

Note the inconsistency that was already in the function: **NODATA** (NOERROR + zero answers) relays and returns early, producing **no** mismatch. Only NXDOMAIN fell through to the report — even though capture skips both equally.

### The change

`resolveUncachedDNSResponse` now records whether upstream returned a **definitive** negative answer for the name (`upstreamSaysAbsent`) and gates only the mismatch report on it. For such a name there was never a mock to record, so its absence at replay is expected, not a defect. This makes NXDOMAIN consistent with the NODATA branch above it.

### Why NXDOMAIN is still not synthesised back to the app

The #2006 steering is **unchanged**. A negative answer must not reach the app when mocking is enabled: any unknown name has to resolve to the proxy IP so eBPF can intercept it and match against recorded mocks — relaying NXDOMAIN crashes apps that use bare service names like `localstack` or `postgres` (#2006). This PR touches *reporting* only; the response the app receives on this path is byte-for-byte what it was before, and the regression tests for #2006 (`..._UpstreamNXDOMAIN_FallsBackToProxyIP`, `..._UpstreamSERVFAIL_FallsBackToProxyIP`) still pass unmodified.

The `fwdErr != nil` case (upstream unreachable) deliberately keeps reporting: when the forward fails we cannot know whether the name would have resolved, so a missing mock is still a candidate defect. Suppressing it there would let a broken resolver silently swallow every genuine DNS mock miss.

### Tests

Four new cases alongside the existing `TestResolveUncachedDNSResponse_TestMode_*` set, using a buffered `errChannel` so the report itself is observable (the returned `dnsCacheEntry` looks identical either way):

- `..._UpstreamNXDOMAIN_NoMockMismatch` — steering intact (synthetic A → proxy IP, `Rcode=0`, not `FromUpstream`) **and** no mismatch reported. Uses the doubled search-expanded name from the issue. **Fails before this change, passes after.**
- `..._UpstreamSERVFAIL_NoMockMismatch` — the suppression covers every definitive non-Success rcode, since capture skips SERVFAIL/REFUSED on the same grounds. **Fails before, passes after.**
- `..._UpstreamUnreachable_ReportsMismatch` — a mismatch is still reported when the forward times out.
- `..._UpstreamNODATA_NoMockMismatch` — NODATA behaviour unchanged: honest empty NOERROR relayed, `FromUpstream`, no report.

Validated end to end on the cluster run that produced the evidence above: DNS mismatches **16 → 0**, and combined with the other fixes in the parent issue the run went to **8/8 passing**.

## Links & References

**Closes:** #4474
- Parent issue: #2426
- Must-not-regress: #2006

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- #4474, #2426, #2006
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?

The behaviour is fully covered by unit tests against a fake upstream resolver; reproducing it in the e2e pipeline would require a `ndots:5` Kubernetes resolver config.

- [x] 👍 yes

## Added to documentation?

- [x] 👍 yes, mentioned below

```bash
go build -tags=viper_bind_struct ./...
go test ./pkg/agent/proxy/ -run 'TestResolveUncachedDNSResponse|TestForwardDNSUpstream|TestDNSCache' -v
```

Reverting `pkg/agent/proxy/dns.go` alone makes `..._UpstreamNXDOMAIN_NoMockMismatch` and `..._UpstreamSERVFAIL_NoMockMismatch` fail with a `MockMismatchReport` on the errChannel.

## Self Review done?
- [x] ✅ yes

- NA (cluster hostnames in the evidence above are redacted)

---

## Additional checklist:

### DIFF
--- a/pkg/agent/proxy/dns.go
+++ b/pkg/agent/proxy/dns.go
@@ -290,6 +290,18 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
 		// "mock not found" path: same error, same log line, same
 		// NXDOMAIN / synthetic fallback. Forwarding is strictly
 		// additive.
+		//
+		// Set when upstream gives a DEFINITIVE negative answer for this name.
+		// Capture deliberately drops non-Success rcodes (dns_capture.go: "skip
+		// non-Success rcodes (NXDOMAIN, SERVFAIL, REFUSED) … would pollute
+		// mocks.yaml"), so for such a name there was never a mock to record and
+		// its absence at replay is EXPECTED, not a defect. Reporting a mismatch
+		// here fails tests for names the resolver was only probing — with
+		// ndots:5 every hostname emits 2-3 search-expansion misses, so a single
+		// app doing one lookup can fail a test set on names that never resolved
+		// in production either. NODATA already relays without reporting (below);
+		// this makes NXDOMAIN consistent with it.
+		upstreamSaysAbsent := false
 		if fwdResp, fwdErr := p.forwardDNSUpstream(question); fwdErr == nil && fwdResp != nil {
 			// A RcodeSuccess response splits into two cases by answer count:
 			// resolved (>=1 RR) vs NODATA (0 RRs). Both are relayed as-is; only
@@ -367,6 +379,7 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
 					zap.Int("rcode", fwdResp.Rcode))
 				return dnsCacheEntry{Msg: fwdResp, FromUpstream: true}
 			}
+			upstreamSaysAbsent = true
 			p.logger.Debug("DNS mock miss: upstream returned negative answer; falling back to synthetic DNS response",
 				zap.String("query", question.Name),
 				zap.String("qtype", dns.TypeToString[question.Qtype]),
@@ -377,7 +390,11 @@ func (p *Proxy) resolveUncachedDNSResponse(question dns.Question, mode models.Mo
 				zap.String("qtype", dns.TypeToString[question.Qtype]),
 				zap.Error(fwdErr))
 		}
-		if mockingEnabled {
+		// Still steer to the proxy IP below (issue #2006: relaying NXDOMAIN
+		// crashes apps using bare service names), but do NOT count an expected
+		// negative as a mock mismatch. If upstream was unreachable we cannot
+		// tell, so we keep reporting in that case.
+		if mockingEnabled && !upstreamSaysAbsent {
 			// Send mock not found error if we couldn't match any DNS
 			// mock and upstream forwarding also failed.
 			p.logger.Debug("mock miss",
--- a/pkg/agent/proxy/dns_forward_test.go
+++ b/pkg/agent/proxy/dns_forward_test.go
@@ -825,3 +825,198 @@ func TestForwardDNSUpstream_NoUpstreamConfigured(t *testing.T) {
 		t.Errorf("err = %v; want message containing 'no upstream'", err)
 	}
 }
+
+// collectDNSMismatchReports drains p.errChannel and returns the DNS
+// MockMismatchReports that resolveUncachedDNSResponse pushed onto it. The
+// steering-vs-reporting split of issue #4474 is only observable here: the
+// returned dnsCacheEntry looks identical whether or not a mismatch was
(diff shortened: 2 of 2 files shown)


Aditya-eddy:
Linking explicitly: this PR fixes #4474. (The closing reference did not register from the original body.)

slayerjain:
Closing as superseded — the underlying issue (#4474) is fixed on `main`, and the reasoning here was worth having. Full detail so nothing is lost:

**What landed instead.** #4565 (`9bf6ffe4`, merged) implements the same fix with a narrower gate: it suppresses the mock-miss report only for `RcodeNameError`, where this PR suppressed on any definitive negative rcode. I rebased this branch onto `main` to land it, and after resolving the conflicts the `dns.go` change was empty — `main` already does everything it did, for NXDOMAIN.

**Why the narrow gate is the right one.** The shared premise is correct: capture stores only Success rcodes (`dns.go:810`), so for a name upstream reports as absent there was never a mock, and the report's own advice — "re-record to capture DNS queries" — can never succeed. That argument is airtight for NXDOMAIN.

It does not carry to SERVFAIL/REFUSED, though, and this PR's own text shows why: it deliberately keeps reporting when the upstream is unreachable, because "we cannot tell". SERVFAIL *is* "I cannot tell" and REFUSED is "I will not tell you" — the same epistemic state as a packet that never arrived. Only NXDOMAIN positively asserts the name does not exist.

The consequence is not just imprecision. A replay environment whose resolver is sick answers SERVFAIL to *every* query, so the wider gate would silently swallow every genuine DNS mock miss in the run and the suite would go green on a machine where DNS is broken. In a verification tool a false PASS is categorically worse than a false FAIL: a false FAIL gets found immediately, a false PASS ships a bug.

**One real gap this PR found.** #4565's tests assert only that *something* arrived on `errChannel` — a bare `select` that accepts any value of any type. The shape assertions in this branch (`ErrMockNotFound`, non-nil `MismatchReport`, `Protocol`, exactly-once, and the query named) genuinely were not covered. I've kept that part and dropped the rest as duplicates of #4565's coverage: **#4573**.

Worth noting one thing that argues in this PR's direction, for whoever revisits it: dnsmasq and split-horizon corporate resolvers commonly answer **REFUSED** for out-of-zone names, meaning "not mine" — functionally "does not exist" in a replay environment. So the ndots:5 bug may still reproduce on those resolvers under the narrow gate. That is a case for adding REFUSED specifically, with evidence, not for the blanket gate — a good follow-up if you hit it in the wild.

PR #4479, opened 2026-08-24
feat(replay): add the consumer test contract, delivery gate and effect judge

slayerjain:

> **Stacked on #4478** — based on `feat/depresult-sync-assertions`. Review that one first; this diff is only the second commit.

A keploy test is one ingress request and one response to diff, and **a message consumer has neither**. Recording a Kafka or Pulsar worker today produces **zero test cases**: the message it consumed was classified as an outbound mock, and the message it produced was matched-then-discarded to pick an ack.

This adds the OSS side of treating a consumed message as a **trigger** and what the worker did with it as **effects to assert**.

### It is inert on its own, and that is deliberate

OSS ships no message-queue parser, so nothing here can be driven by a real broker; the contract is exercised by unit tests and an in-repo fake projector. It lands separately because it is the **published interface** the enterprise parsers must implement, and because enterprise consumes OSS through a version tag — this has to exist and be stable before any of that can be written against it.

### What it defines

A `Consumer` test-case Kind whose spec carries the trigger, the expected effects and a completion rule; a **delivery gate**; a **projector SPI** a parser implements to turn its own wire objects into protocol-neutral effect views; a **unit recorder** that mints one test case per consumed message through the same window authority the mock mapping already uses; and a **comparator** that produces a verdict.

Nothing in OSS is broker-shaped: protocol coordinates live in an opaque map OSS copies and never reads, and no control flow anywhere branches on a protocol name (grep-enforced).

### Four decisions, each closing a way this could report a lie

**The gate is default-closed.** Arming a test's mocks is necessary but *not sufficient* for delivery: the whole mock pool is resident while the app boots, and a consumer polls immediately, so a pool-driven gate would let a worker drain every trigger before the first test exists. Delivery is refused outside an armed window, and arming **adopts** everything observed since the previous test finished rather than clearing a buffer a live app may still be writing to — a prefetching client can be answered between the pool swap and the arm.

**A missing effect is never demotable.** An unconsumed expected mock normally demotes a test to OBSOLETE *without failing the test set*, and for a consumer every missing-effect failure has an unconsumed mock by construction — so the flagship regression, *the worker that stops producing*, would have been routed to green. For this Kind the demotion is skipped, and only for **effect** mocks, so an unrelated unconsumed mock is not turned into a false red.

**The judge shares no code with the mock matcher.** The matcher accepts a match with no body agreement and filters fields named `timestamp`, `host` and `sequence` at every depth — names that occur inside real payloads. A judge built on it could not see a payload diff at all. A test pins that a body containing those names still diffs.

**A test that cannot assert anything is refused by name rather than passed.** A consume-and-write worker records no produced effect, so its only claim is carried by the sync path's dependency rows — and that assertion does not run when the test has no mapping. The verdict depends on whether the assertion **actually ran in this run for this test**, not on a count taken at record time.

Completion is expected-count plus a **mandatory grace drain**, never an idle timer, so a worker that emits one message too many is caught rather than declared finished early. Anything v1 cannot handle — concurrency, an unreadable payload, a missing mapping — is a FAILED test with a named category, never a silent pass.

## What this does NOT cover

- **Any broker.** No trigger reaches a real consumer until the enterprise parsers implement the projector and deliverer.
- Per-field assertion of database writes is separate; message headers *are* asserted.

### Known gaps (pre-existing, recorded deliberately — not fixed here)

Both predate the rebase onto main and ship with this PR either way. Line numbers are against this branch after the rebase onto `main`.

- **The MySQL probe branch bypasses the consumer recorder.** `consumer.WithRecorder` is installed at `pkg/agent/proxy/proxy.go:3362`, but the probe-dispatch branch returns earlier, at `pkg/agent/proxy/proxy.go:2611`, via `recordMySQLOutgoing`. A probe-dispatched MySQL connection therefore never carries the recorder, so its egress is invisible to `ConsumerSpec.SideEffects`.
- **A brittle source pin remains on the matched-parser branch.** `TestParserContextsCarryTheConsumerSeams` pins `p.recordViaSupervisor(parserCtx, srcConn, dstConn, matchedParser, parserType, rule.MC, ...)` as a full argument literal (`pkg/agent/proxy/consumer_wiring_test.go:535`). That is the failure mode the pin struct's own doc comment warns about: an unrelated change to a later argument trips it and reads as a consumer regression. The sibling generic pin was changed to a prefix match; this one was deliberately left as-is to keep the rebase minimal.
- **Record/replay asymmetry on the generic branch.** Record installs the recorder on all three branches, while replay installs `consumer.WithGate` only on the matched-parser branch (`pkg/agent/proxy/proxy.go:3438`) — `mockGenericOutgoing` is left ungated. This may well be intentional (generic cannot serve a consumer protocol at replay), but the record-side comment argues the opposite direction, so it needs an explicit decision either way.

## Before this tags

Design §6 flags one question as **not reversible later without a public retraction**: how much of the Kafka path is donated to OSS versus kept in enterprise. Both answers and their costs are recorded in `docs/reference/consumer-contract.md` §9, deliberately undecided — it is a product call. A release tag should wait on it, since the enterprise end-to-end work consumes that tag.

## What type of PR is this?
- [x] 🍕 Feature

## Added e2e test pipeline?
- [x] 🙅 no, because they aren't needed — inert without a parser; covered by unit tests and a fake projector. The end-to-end lands with the enterprise Kafka work.


chilkotiKartik (approved):
LGTM! Thoroughly verified changes in feat(replay): add the consumer test contract, delivery gate and effect judge. Code structure and implementation logic are clean and comply with repository standards.

PR #4480, opened 2026-08-24
fix(mocks): don't treat an indented "---" as a document separator

### LINKED ISSUE #4477: MockReader splits mocks.yaml on indented '---', so keploy cannot read back a file it wrote (replay session fails to start)
### Summary

### Root cause

```go

if trimmedLine == "---" {
    if buffer.Len() == 0 {
        continue
    }
    return buffer.Bytes(), nil
}
```

### Reproduction

```
```

```yaml
kind: Http2
spec:
    req:
        method: POST
        url: /openai/v1/responses
        body: |+
            ...
            ...
                      ---          <-- line 5475
```

```
236
```

as the reader is fixed.

### Suggested fix

```go

if trimmedLine == "---" {
```

than sharing one variable.

### Impact

### Environment

Aditya-eddy:
Fixes #4477

## What was broken

`MockReader` anchored its document-separator check with `strings.TrimSpace`, which strips leading whitespace as well as trailing:

```go
if trimmedLine == "---" {
```

A YAML document separator only separates at column 0. So an **indented** `---` inside a block scalar — a recorded request body containing markdown, YAML, or front matter — was read as the start of a new document. The reader handed the parser a fragment cut mid-scalar and the whole file failed to load:

```
failed to unmarshal a mock yaml doc of unknown type   {"type": ""}
failed to run test set: failed to decode the file documents.
replay session failed to start
```

Two things made this worse than a single bad mock:

- it takes out the **entire test set**, so zero tests run;
- the test-set upload happens **after** the replay, so the recording never reached the backend either — `keploy cloud replay` was impossible for it as well.

Nothing warns at record time. It just looks like the replay didn't happen.

## The file was always valid — this is reader-side only

```
236
```

A conformant parser honours block-scalar indentation; only the hand-rolled line splitter did not. So the writer is fine and **nothing already recorded is lost** — existing files become readable again with this change.

## The fix

Strip only trailing whitespace and the line ending. The leading-comment skip genuinely does want a leading-trimmed value, so the two uses no longer share one variable.

## How it was verified

Reproduced from a real recording (an LLM-backed flow, where `---` is ubiquitous in prompts — but markdown or a YAML payload will do it):

| | Result |
|---|---|
| before | `FAILED after 139 docs: failed to decode YAML at line 5484: yaml: line 8: mapping values are not allowed in this context` |
| after | `read 236 documents OK; kinds=map[DNS:8 Http:2 Http2:4 Mongo:222]` |

That is the same error string, and the same line number, as the cluster produced.

Also run:

- `go test ./pkg/platform/yaml/... -count=1` — all packages ok
- `go vet ./pkg/platform/yaml/` — clean
- `go build -buildvcs=false ./...` — clean
- `gofmt -l pkg/platform/yaml/` — no output

Two regression tests added: an indented `---` inside a `body: |+` block scalar, and a column-0 `---` with trailing whitespace/tab (so the fix cannot regress into not trimming the trailing side). The existing CRLF tests still pass, since `TrimRight` also strips `\r\n`.

## Scope of verification

Unit- and build-level, against the actual `mocks.yaml` that triggered the failure. I have **not** yet re-run the full record → auto-replay → cloud-replay cycle with a k8s-proxy image built from this change; that is the next step and I'll report back on it.


chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(mocks): don't treat an indented "---" as a document separator. Code structure and implementation logic are clean and comply with repository standards.

PR #4482, opened 2026-08-24
fix(ci): fail CodeQL when its Go extractor cannot read this module

slayerjain:
## What is broken

`github/codeql-action/analyze` exits 0 and reports `executionSuccessful` even when the Go extractor rejected every first-party package.

This is **live on `main` right now**. #4473 moved `go.mod` to `go 1.27.0`; the CodeQL bundle (`v2.26.3`) ships a Go extractor built with `go1.26.5`, and its embedded `go/types` refuses a newer module. Push-to-main job `97386346026`:

```
96 x "package requires newer Go version go1.27 (application built with go1.26)"
CodeQL job status was success.
```

Almost nothing in this repository is being analysed, and the job is green.

The cause is upstream and unfixable here — neither `setup-go` nor a `tools:` pin changes the extractor's own embedded `go/types`. Go 1.27 support merged into `github/codeql` on 2026-08-19, a week after v2.26.3 was cut, and `codeql-action` floats on `@v3`, so it self-heals on the next bundle release.

## What this adds

A step that compares the Go the extractor was built with against the `go` directive in `go.mod`, and fails when the extractor is behind.

```
CodeQL Go extractor built with go1.26.5; this module declares go1.27.0.
```

That is the causal condition, decided before a single query runs. No message parsing, nothing to reword, no dependency on SARIF shape.

## What the first version got wrong

Worth recording, because the correction is the interesting part and CI is what produced it.

The first version of this step decided everything from the SARIF, and required `go/diagnostics/extraction-errors` to appear in `runs[].tool.driver.notifications[]` as a positive control — reasoning that a check which cannot see failures must not report health. That control is calibrated on a hand-run `codeql database analyze`, which writes that catalogue. **`github/codeql-action/autobuild` does not.** So the guard failed on every run, naming a descriptor that was never going to be there:

```
##[error]CodeQL extraction guard: 'go/diagnostics/extraction-errors' is absent from
the SARIF notification catalogue. Descriptors present:
  codeql-action/overlay-disabled
  codeql-action/overlay-disabled-due-to-cached-status
  go/autobuilder/single-root-go-mod-found
```

A step that is always red gets deleted rather than fixed — which is the exact failure mode this PR exists to prevent — so the gate moved to the causal comparison above.

The errors themselves were never missing. They are in `runs[].invocations[].toolExecutionNotifications[]`, in the shape the classifier expects; job `97406057233` carried 95 of them:

```
Extraction failed in cli/agent.go with error package requires newer Go version go1.27 (application built with go1.26)
```

So the SARIF pass is kept and does real work — it is the only thing that can see a *non-gate* extraction failure — but it is calibrated the other way round now: **absence is never fatal**, an unparsable message is a warning, and only a recognised, in-checkout, non-gate failure fails the job.

## Verified

Run against the **real bundle extractor binary** (`go-extractor`, stamped `go1.26.5`) and the **real `go.mod`**, with the guard extracted from the shipping YAML:

| case | exit |
|---|---|
| extractor 1.26.5 < module 1.27, ack in force | 0 + `::warning::` |
| same, ack unset / expired / unparseable | 1 |
| extractor 1.26.5 vs module 1.26 (equal) | 0, not gated |
| extractor ahead of module | 0, not gated |
| real degraded SARIF, ack in force | 0 + warning |
| first-party non-gate failure | **1**, no acknowledgement path |
| out-of-checkout errors only | 0 |
| unparsable message shape | 0 + warning |
| corrupt SARIF / no SARIF dir | 0 — absence is never fatal |
| extractor binary not locatable | 1, naming itself |

Testing also caught a bug in the guard: the toolcache fallback used `find -maxdepth 4`, but `/opt/hostedtoolcache/CodeQL/<ver>/<arch>/codeql/go` is five levels down, so it would have missed the extractor on a runner where `codeql` was not on `PATH`.

`actionlint -shellcheck` and standalone `shellcheck -s bash` are clean, and clean on `origin/main` too.

## The judgement call

`GO_VERSION_GATE_ACK_UNTIL` ships as `"2026-11-30"`, not empty. Empty turns `main` red today. **CodeQL is not a required check on this repository** (`branches/main/protection` → `required_status_checks: null`), so a red nobody is blocked by is a red that gets ignored and then deleted. A date warns until it lapses and then forces an explicit decision. Clear it back to `""` once a bundle carrying the Go 1.27 extractor lands; the guard fails if the value is unset, unparseable, or expired, so it cannot rot quietly either way.

Consequence, stated plainly: until 2026-11-30 this produces a warning, not a failure. Everything that is *not* the version gate — an unlocatable extractor, an unreadable version — fails hard.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -28,6 +28,16 @@ jobs:
     timeout-minutes: 60
     name: Analyze
     runs-on: ubuntu-latest
+    env:
+      # Pin the CodeQL CLI's diagnostics export on. It defaults to true today
+      # (Feature.ExportDiagnosticsEnabled in codeql-action src/feature-flags.ts),
+      # but that default is a server-side feature flag GitHub can flip; an
+      # explicit "true" is checked in getOfflineValue and returns before the API
+      # is consulted. Without the diagnostics in the SARIF, the extraction guard
+      # below has nothing to read - and it is written to fail rather than pass
+      # in that case, so this pin is what keeps a flag flip from turning every
+      # run red.
+      CODEQL_ACTION_EXPORT_DIAGNOSTICS: "true"
     permissions:
       actions: read
       contents: read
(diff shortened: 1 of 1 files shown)


chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(ci): fail CodeQL when its Go extractor cannot read this module. Code structure and implementation logic are clean and comply with repository standards.

PR #4483, opened 2026-08-24
feat(mappings): add a test-set-scoped startup section

Shivanipandey31:
`mappings.yaml` has only ever had one container, `tests[]`. A mock consumed while the app boots — driver handshakes, connection-pool warm-up, config/secret fetches — belongs to no test case, so it was never recorded there.

That absence is invisible on the timestamp path, which reloads boot traffic by time window. It matters on the **mapping-based** path, which loads mocks strictly by name.

**Scope, stated precisely.** A *full* run always booted fine: both paths first send a wide-window, empty-`MockMapping` filter specifically so boot traffic is served, and `GetFilteredMocks` only prunes on `isMappedToSpecificTest && !isNeededForCurrentRun`, so a mock in no test's list was never pruned. Session- and connection-tier boot mocks also already reach per-test lists via `upsertActualTestMockMapping`'s always-keep carve-out.

The real gap is narrower: **subset runs** (`--tests` / `--test-sets`), and a **per-test-tier mock consumed at boot that a later test needs again** — a pool reconnect, token refresh or driver retry mid-run, where the by-name pool no longer contains it.

This PR adds the section and makes it real: written during replay, read back, and merged into the names the agent loads.

> **Scope note (2026-09-10):** this PR previously also carried the schema-noise → mock-noise
> rename. That has been split out and will land as its own PR, so this one is now *only*
> the test-set startup section. The rename is unchanged in intent — just separated so each
> can be reviewed on its own terms. It is self-contained and described at the end.

---

### 1. Schema — `pkg/models/mappings.go`

`Mapping.Startup []MockEntry`, a flat list rather than a per-test one: these belong to the test **set**, are available for the whole session, and are consumed by no single test. Plus `Mapping.StartupMockNames()`, nil-safe so callers need no version check.

### 2. Write — `pkg/service/replay`

`setStartupMocks` records the mocks `GetConsumedMocks` returns immediately after the app becomes ready and before the first test runs. Those are boot traffic **by construction**, so membership needs no window arithmetic. Deduped and order-stable, so a re-run produces a byte-identical section rather than churning a file that gets diffed and reviewed.

**The write also has to actually fire.** `keploy record` writes `mappings.yaml` via `UpsertBatch` (no startup capture), so by the time `keploy test` runs the file exists, the create-if-absent branch is skipped, and `UpdateTestMapping` defaults to false — the captured section was computed and discarded. `backfillStartupSection` writes it when the on-disk mapping lacks one.

That backfill writes a **startup-only document**, deliberately. Flipping `shouldWriteMappings` would route through `Insert`, which *replaces* per-test entries by ID — on a subset run that would overwrite an operator's curated mapping with whatever the run happened to consume.

### 3. Read — `pkg/platform/yaml/mapdb` + `pkg/service/replay`

- `MappingDb.GetStartup`. A missing file or absent section returns `nil, nil`, so every mapping written before the section existed reads clean.
- `determineMockingStrategy` returns the startup names alongside the per-test map.
- `models.MergeStartupMockNames` unions them into each test's expected list before it reaches the agent — the line that closes the gap.

**A test with no per-test entry stays empty.** The agent engages mapping mode on `len(MockMapping) > 0`, so returning startup names for such a test would narrow its pool to the boot mocks alone and fail it with `no_mocks`, where today it degrades gracefully to the wide pool.

### 4. Both replay paths, not one

`pkg/service/runner` reproduced the same by-name bug and is live — enterprise imports it from `pkg/service/sandbox`, so it is the sandbox/UI replay flow. Without this the same app booted fine under `keploy test` and lost its handshake mocks under the sandbox.

The merge lives in `models.MergeStartupMockNames` so both paths call one implementation — duplicating it is precisely what let the section reach one and not the other. The **streaming** replay path is covered too, merged at the *send* site rather than the build site: the built slice is reused for `isMockSubsetWithConfig`, and widening it there would be a false pass.

### 5. Pruning

Startup names go into **both** `mocksThatHaveMappings` and `mocksWeNeed` — seeding only the first drops them whenever a subset of tests runs. The `mocksWeNeed = mocksThatHaveMappings` alias is now a copy so it cannot discard that seeding.

### 6. `Insert` had to be fixed, and a test is what caught it

`Insert` flattens the mapping into a per-test map and rebuilds it via `CreateMappingStructure`, which cannot reconstruct a section it never saw — the caller would write startup mocks and the **next** `Insert` would silently drop them. Fixed with replace-or-keep. `UpsertBatch` needs no change (it mutates a decoded `*models.Mapping` in place) and is pinned by a test so a later refactor to the rebuild shape is caught.

`CreateMappingStructure` now also sorts, so a rewrite no longer reshuffles `tests:` into a large meaningless diff.

### 7. Observability

`startupMocks` on the existing INFO line:

```
Successfully saved test-mock mappings {"testSetID": "test-set-0", "numTests": 6, "startupMocks": 1}
```

Every other trace of the section is DEBUG and CI does not run replay at debug level, so without this a run that captured zero startup mocks was indistinguishable from one that captured them correctly.

---

## Verified in CI

Exercised against the enterprise lanes via a pseudo-version pin. `startupMocks` is non-zero in **9 lanes** — `aerospike-e2e` 45, `sqs-localstack` up to 21, `go-linux` 12/14, `couchbase-java` 12, `couchbase-python` 9, `couchbase-go`/`-node` 6, `kafka-ecommerce` 2, `parse-server` 1, `python-cred-expiry` 1 — cross-checking exactly against the DEBUG `consumed mocks during initial setup {"count": N}` line in the same log.

`python-cred-expiry` is the telling one: it wrote **no mappings at all** before the backfill existed.

Lanes reporting `0` are correct answers, not failures: `umami`, `restheart`, `memcached`, `pulsar` and `doccano` connect lazily on the first query rather than at boot, so there is no boot traffic to record.

⚠ The **read** side is still unobserved. `GetStartup`'s log is DEBUG and `--debug` is set only on `record` in those lanes, never on the manual `keploy test`. The write is proven; a replay consuming the section from disk has not been watched directly.

---

## Deliberately unchanged

`hasMeaningfulMappings` still counts per-test entries only, so a test set whose **only** outgoing traffic is startup traffic keeps the timestamp path. Mapping-based selection exists to choose mocks per test; with no per-test entry there is nothing to choose, and that path loads startup mocks via `LoadBefore` anyway.

## Backward compatibility

- `omitempty` means an existing `mappings.yaml` round-trips byte-identically until a startup mock is actually recorded.
- Every older reader ignores an unknown key.
- A mapping with no `startup:` key yields no startup names, so the merge returns exactly the previous per-test list — **behaviour is identical until a replay rewrites the file.**

A mock may legitimately appear in both sections (session/connection-tier mocks are consumed at boot *and* kept per-test by the always-keep carve-out). Both sides only add names to the pool, and the merge dedupes.

## Risk

Each test now loads strictly **more** mocks on the mapping path. A larger pool should be safe — it can only serve more, and `dedupByName` handles overlap — but a mock that matches eagerly being present when it previously was not could shift which mock a request matches. That is the thing to watch in replay-heavy suites.

## Known gap, not closed here

The **cloud** path still drops the section at the last hop: `api-server`'s `ToOSSMockMapping` cannot carry it at the pinned `v3.5.80`, and both replay download routes go through it (neither sets `withKinds=true` — the only caller of that is a record-time GraphQL path). Documented in keploy/api-server#2007; it needs that repo's dependency bump.

---

- Add `Mapping.Startup` + `StartupMockNames()` to the `mappings.yaml` schema
- Record boot-time consumed mocks into the section during replay, and make the write actually fire in the default `record → test` flow
- Read the section back and merge its names into every test's expected mock list, on the non-streaming, streaming and runner paths
- Include startup names in the mock-pruning sets so a subset run keeps them
- Stop `MappingDb.Insert` dropping the section on rewrite; sort `tests:` so rewrites do not churn
- Report `startupMocks` at INFO so the section is observable without `--debug`

## Links & References

**Closes:** NA

### 🔗 Related PRs
- keploy/api-server#2007 — matching `MappingDocument.Startup` so the section survives upload → Mongo → download
- keploy/enterprise#2281 — CI verification for this section
- keploy/enterprise-ui#1734 — the UI half of the rename
### 🐞 Related Issues
- keploy/enterprise-ui#1592 — where the rename was requested
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [x] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

Covered at the unit level here. End-to-end coverage lives in the enterprise replay lanes, which already exercise this via a temporary pin — see the CI numbers above.

- [x] 👍 yes

Each non-obvious decision carries its reason inline: why the mapping path needs the names and the window path does not, why the backfill writes a startup-only document, why the streaming merge belongs at the send site, why both pruning maps are seeded, why `Insert` needs explicit carry-across and `UpsertBatch` does not, and why two identifiers were deliberately left on the old spelling.

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

- [x] 👍 yes, mentioned below

```bash
go test ./pkg/models/... ./pkg/platform/yaml/... ./pkg/service/replay/... ./pkg/service/runner/... ./pkg/service/mock/... ./config/... ./cli/...
```

End to end: record a sample app whose driver handshakes at boot, replay once so `mappings.yaml` is written, and confirm the file carries a `startup:` section — then replay again and confirm those names appear in the mocks loaded for each test.

**Test coverage added**

- `pkg/models` — section round-trips through YAML and JSON; absent case stays absent; `MergeStartupMockNames` ordering, both-sections overlap, and the empty-per-test guard
- `pkg/platform/yaml/mapdb` — disk round trip; pre-rename mappings; missing file; `omitempty`; `Insert` replace-or-keep both ways; `UpsertBatch` preservation
- `pkg/service/replay` — `setStartupMocks`; `determineMockingStrategy`'s startup return; `backfillStartupSection` (startup-only document, no-write cases, read failure, nil-safety); the streaming split; a write → disk → read round trip
- `pkg/service/runner` — the merge, both-maps seeding, no-per-test-entry, tolerated read failure

`gofmt`, `go build ./...` and `go vet` clean; 23 packages green.

## Self Review done?
- [x] ✅ yes

- NA

---

## Additional checklist:


Shivanipandey31:
## Staff-level review pass — 4 findings, all fixed in `80c843327`

Ran a full review of this PR against `origin/main`, including the sibling repos that consume this schema. Four findings; all are addressed.

### Major — `pkg/service/runner` had the identical bug, unfixed

This PR fixes by-name loading in `pkg/service/replay`. `pkg/service/runner` reproduced it verbatim:

```go
// runner.go:558 (before)
func expectedMocksForTest(setup *testSetSetup, testCaseName string) []string {
    entries, ok := setup.mappings[testCaseName]   // per-test names ONLY
    ...
}
// -> sendPerTestParams -> MockFilterParams.MockMapping -> disk.LoadByNames
```

**This is not dead code.** `pkg/service/runner` is unimported inside keploy but imported by enterprise from `pkg/service/sandbox/{integration,integration_driver,integration_export}.go` — it is the sandbox/UI-driven replay flow. So the same app booted fine under `keploy test` and lost its handshake mocks under the sandbox, which is a nasty thing to debug because the two paths look identical from outside.

Fixed: `runner.MappingDB` declares `GetStartup`, `loadMappingsForSet` reads and returns the names, both prune maps are seeded, and `expectedMocksForTest` merges them. It also no longer early-returns `nil` for a test with no per-test entry — such a test still needs the app to boot.

### Design — the merge helper now lives in `pkg/models`

Duplicating it once already meant the section reached one replay path and not the other. Both now call `models.MergeStartupMockNames`. Added a fast path that skips the dedupe map entirely when there is no startup section, since this runs once per test case and that is the common shape.

### Minor — `mappings.yaml` was read and decoded twice per test set

`Get` and `GetStartup` each did their own `ReadFileAny` + `DecodeMappingF`: double I/O plus a window where the two could observe different versions of the file. Extracted `decodeMapping`, shared by both. It returns the resolved path rather than caching it on the receiver — a field there would have been a data race across concurrent test sets.

### Minor — the wiring had no test coverage

The pure helpers were well covered; the wiring between them was not. Added 14 cases:

- `pkg/service/runner` — the merge, both-maps seeding, the no-per-test-entry case, and a tolerated `GetStartup` failure
- `pkg/service/replay` — `determineMockingStrategy`'s third return, including the deliberate case where a non-meaningful mapping still yields startup names, and mapping-disabled reading nothing

That last one matters because the `mocksWeNeed = mocksThatHaveMappings` alias → copy change exists purely so the seeding is not discarded; a future refactor could restore the alias with every other test still green.

### Verified correct as-is, deliberately not changed

Startup names are **not** added to the agent scope table. `SetScopeTable` builds a "mapped universe" in which an unmapped mock is KEPT but a mapped one absent from the current test is HIDDEN, and `SetMappedUniverse` is fed only by `pushScopeTable` — never by `MockFilterParams.MockMapping`. Startup mocks therefore stay outside that universe and are kept for every test. Adding them there would have hidden them from any test whose list omitted them.

Also checked: `params.MockMapping` is read at exactly two sites, both guarded by `UseMappingBased`; and no production code branches on the *size* of the prune maps, only membership — so seeding them cannot alter timestamp-path behaviour.

### Residual risk

The **read** side is still unobserved in CI. `GetStartup`'s log is DEBUG and `--debug` is set only on `record` in the enterprise lanes, never on the manual `keploy test`. The write is proven — `startupMocks` is non-zero in 9 lanes (`aerospike-e2e` 45, `sqs-localstack` 20, `go-linux` 14, `couchbase` 6–9) and exactly matches the debug initial-setup count — but a replay consuming the section from disk has not been watched directly.

`gofmt`, `build`, `vet` clean; `models`, `platform/yaml`, `service/replay`, `service/runner`, `service/mock` all green.

Shivanipandey31:

Aditya-eddy:
Reviewed this together with keploy/api-server#2007, since the two halves only make sense against each other. Build, `go vet` and the `models` / `platform/yaml` / `service/replay` / `service/runner` suites are all green, and the persistence work is genuinely good — `TestInsertPreservesStartupWhenCallerSuppliesNone`, `TestInsertReplacesStartupWhenCallerSuppliesSome` and `TestUpsertBatchPreservesStartup` pin the silent-drop class properly, and the `Insert` fix is the right shape.

But I don't think this can merge as-is: **in the default `record` → `test` flow the startup section is never written**, and where the merge does take effect it can shrink a test's mock pool.

---

## 1. The write never fires in the default flow

`setStartupMocks` populates `actualTestMockMappings.Startup`, but `StoreMappings` is gated at `pkg/service/replay/replay.go:3033-3043`:

```go
shouldWriteMappings := r.config.Test.UpdateTestMapping           // config/default.go:76 → false
if !shouldWriteMappings && r.mappingDB != nil && len(actualTestMockMappings.TestCases) > 0 {
    exists, _ := r.mappingDB.Exists(ctx, testSetID)
    if !exists { shouldWriteMappings = true }                     // create-if-absent ONLY
}
```

The normal sequence:

1. `keploy record` → `pkg/service/record/record.go:134` calls `mappingDb.UpsertBatch`, writing `mappings.yaml` with `tests:` only. `UpsertBatch` has no startup capture, and this PR doesn't add one.
2. `keploy test` without `--update-test-mapping` → the file exists → `shouldWriteMappings == false` → the captured `Startup` slice is computed and discarded.
3. `determineMockingStrategy` → `GetStartup` → `nil`, forever.

The only ways to get a section on disk are `--update-test-mapping` (off by default) or deleting `mappings.yaml` first. The PR's "the read side is still unobserved" caveat understates this — the *write* side is also unobserved in the flow this targets. That also means the cloud lifecycle (*record → upload → Mongo → download → replay*) can never carry one on the first upload, since nothing on the record side produces it, which makes keploy/api-server#2007 a receiver for a field nothing currently sends. Sandbox-only users never generate one at all: `pkg/service/runner`'s `MappingDB` interface is read-only (`service.go:94-105`, `Get` + `GetStartup`, no `Insert`).

Either the record path needs to write it, or the write must not be gated on `UpdateTestMapping`.

## 2. Merging startup names into an *empty* per-test list flips that test to "boot mocks only"

The agent picks its loader on `params.UseMappingBased && len(params.MockMapping) > 0` (`pkg/service/agent/agent.go:941`).

Before this PR, a test with no entry (or an empty entry) in `expectedTestMockMappings` produced `expectedNames = []` → `len(MockMapping) == 0` → fall through to strict-window / lax-all, i.e. the whole recorded pool, time-filtered. After it, the same test produces `[boot-0, boot-1, …]` → mapping mode engages → `disk.LoadByNames(["boot-0","boot-1"])` and **nothing else**.

Concretely: a test set where `hasMeaningfulMappings == true` because `test-1` has mocks, `test-2`'s mapping is empty (recorder dropped its batch — there are comments about exactly this at `pkg/platform/yaml/mapdb/db.go:159-171`), and a startup section exists. `test-2` previously replayed against the full lax pool; now it gets a two-mock pool and fails `no_mocks`.

`pkg/service/runner/startup_mocks_test.go` pins this as intended ("unknown test still gets startup"), so it's deliberate rather than an oversight — but I think it's the wrong trade: it converts a graceful degradation into a hard failure. Guarding on `len(entries) > 0` before engaging mapping mode (or sending `UseMappingBased=false` when the per-test list is empty) keeps the old behaviour for those tests.

## 3. The streaming replay path was missed

`pkg/service/replay/replay.go:1725-1727`:

```go
expectedMockNames := make([]string, len(expectedTestMockMappings[testCase.Name]))
for i, m := range expectedTestMockMappings[testCase.Name] {
    expectedMockNames[i] = m.Name
}
```

This is the same loop replaced at `:1869`, left untouched. `deferred.expectedMocks` goes straight to `SendMockFilterParamsToAgent` at `:2463`, so every HTTP-streaming test on the mapping-based path still gets `LoadByNames(perTestOnly)` with no startup names — the exact gap the PR exists to close.

Worth noting the same slice is reused at `:2541` / `:2645` for `isMockSubsetWithConfig`, so the merge belongs at the *send* site, not the build site.

---

## The description overstates the bug

Both paths already send an initial wide-window, empty-`MockMapping` filter specifically so boot traffic is served — `replay.go:1346` / `:1500`, `runner.go:483-492`, with comments saying exactly that. And `GetFilteredMocks` only prunes on `isMappedToSpecificTest && !isNeededForCurrentRun` (`pkg/platform/yaml/mockdb/db.go:1386-1389`), so a mock that appears in *no* test's list was never pruned — meaning a **full run always booted fine**.

The residual gap is real but much narrower than "the app cannot boot": it's subset runs (`--tests` / `--test-sets`), plus a *per-test-tier* mock consumed at boot with a non-`Deleted` usage that a later test needs again. Session/connection-tier boot mocks already land in per-test lists via `upsertActualTestMockMapping`'s always-keep carve-out (`utils.go:1024-1030`), and config-tier mocks are never name-filtered (`filterByMapping` at `pkg/util.go:3488-3494` partitions rather than drops). Would be good to correct the description so reviewers size the change correctly.

## Smaller things

- **`Mapping.StartupMockNames()` is dead production code** (`pkg/models/mappings.go:81`) — only tests call it. Both `replay.go:4440` and `runner.go:538` hand-roll the identical `if e.Name != "" { append }` loop. Use it in both, or drop it.
- **`decodeMapping`'s doc comment is false** (`pkg/platform/yaml/mapdb/db.go:284-318`): "shared by `Get` and `GetStartup`" — `Get` (`:369`) still duplicates the whole filename-default / `ReadFileAny` / `DecodeMappingF` block. The second return value (resolved file path, `:318`) has no reader.
- **Runner reads `GetStartup` even when `DisableMapping` is set** (`runner.go:538` is unconditional, while `replay.go:4430` returns early on `!isMappingEnabled`). Harmless today, but it's a needless file read per test set and a divergence between the two paths this PR is trying to keep identical.
- **Runner mismatch reporting diverges from what it sends**: `expectedMocksForTest` (`runner.go:583`) now includes startup names but `expectedEntriesForTest` (`runner.go:607`) doesn't, and still early-returns on a missing key. A startup mock re-consumed during a step will show up in `MockMismatch.ConsumedMocks` with no counterpart in `ExpectedMocks`.
- **`Insert`'s replace-or-keep can never clear a stale section** (`mapdb/db.go:96-101`). After a re-record where the app no longer makes a given boot call, a run capturing zero startup mocks falls into `len(existingStartup) > 0` and the stale names persist indefinitely. Benign — `DiskMocks.LoadByNames` silently skips unknown names (`pkg/agent/proxy/diskmocks.go:266-270`) — but worth a comment.
- Two hand-rolled `strings.Contains` clones: `pkg/models/mappings_startup_test.go:64` and `pkg/platform/yaml/mapdb/startup_test.go:122`.
- The `hasMeaningfulMappings` non-change is defensible as-is — a startup-only test set keeps the timestamp path, which loads boot traffic via `LoadBefore` anyway, and both prune maps are seeded identically. No need to flip it.

## Test gaps that line up with the above

- Nothing asserts the startup section actually **reaches disk** on a replay run — a test there would have caught #1.
- Nothing touches `streamingTest.expectedMocks` (#3).
- The empty-per-test-mapping case (#2) is pinned as correct rather than questioned.
- No test spans write → read: nothing asserts `setStartupMocks` output survives `Insert` and comes back through `determineMockingStrategy`. `mocking_strategy_test.go` stubs `GetStartup` entirely, and `mapdb/startup_test.go` never sees a `models.MockState`.

---

## Verified sound, so no need to re-check

- **No data race on `actualTestMockMappings`** — `setStartupMocks` (`:1647`), `upsertActualTestMockMapping` (`:2108`, `:2697`) and `retainNoisyTestCaseMocks` (`:2951`) all run on the `RunTestSet` goroutine; nothing in that range spawns a goroutine touching it.
- **The capture point is correct** — `GetConsumedMocks` drains per call, the call at `:1616` is at function top level (not inside `if r.instrument`), and it's the first one after `waitForAppReady`. No per-test mocks can leak into the startup section.
- **`GetStartup` error handling** — `yaml.ReadFileAny` returns bare `fs.ErrNotExist`, so `os.IsNotExist` matches without unwrapping; stat/read and decode failures both propagate. No genuine error is swallowed as `nil, nil`.
- **`Startup` survives every serialization boundary** — `Mapping` has no custom (Un)Marshal (only `MappedTestCase` does), so the three-way `mock_entries` / `mocks` / legacy-string decoder can't touch it; `UpsertBatch` genuinely mutates in place; `mapdb` is the only `MappingDB` impl in the repo.
- **Seeding both prune maps is necessary, not just defensive** — seeding `mocksThatHaveMappings` alone would drop startup names on a `--tests` subset run, and both are seeded before the `isMappingEnabled` block so the subset branch can't miss them. Replacing the `mocksWeNeed = mocksThatHaveMappings` alias with a copy is a real (if latent) aliasing fix.
- **`setStartupMocks` metadata matches `upsertActualTestMockMapping` exactly** (`utils.go:1043-1049` vs `1461-1467`), so a mock legitimately in both sections is byte-identical in both and can't end up with two kinds or timestamps. `MergeStartupMockNames` dedupes and is order-stable, and `FilterTcsMocksMapping` / `FilterConfigMocksMapping` re-sort by `ReqTimestampMock`, so `MockMapping` order isn't load-bearing.
- **The wire schema against api-server#2007 is clean** — container and all five element fields match in JSON and BSON, including the `req_timestamp_mock` / `res_timestamp_mock` renames. This is not a repeat of the `mock_entries` vs `mocks` problem.

## One cross-repo question

api-server#2007's `ToOSSMockMapping` deliberately omits the section, justified by "the enterprise replay client fetches `withKinds=true`". I can't verify that from either public repo, and the only in-repo documentation of the flag (`agent_proxy.go:484`) describes it as used by the post-record suite linker — a record-time path. If replay uses the default route, the section survives Mongo intact and is dropped at the last hop. Could you paste the replay client's actual request line? That one detail decides whether the cloud half works at all.

Also worth confirming the version story: `v3.6.24` doesn't appear to contain `Mapping.Startup` (nor do any of v3.5.80 / v3.6.8 / .11 / .12 / .13 / .20), so if that's the enterprise pin, a real tag may need cutting before either bump is possible.

Shivanipandey31:
Thanks — this was a genuinely good review. I verified every claim in code before touching anything; all three blockers were correct. Fixed in `aa64e1732`.

## 1. The write never fired — correct

Confirmed: `updateTestMapping: false` (`config/default.go:76`), `record` writes via `UpsertBatch` with no startup capture, so by `test` time the file exists and the create-if-absent branch is skipped. The captured slice was computed and discarded.

**One nuance worth adding:** this is why CI still showed non-zero counts — enterprise's post-record auto-replay sets `UpdateTestMapping=true`. So the feature worked on that path and was dead on plain OSS `record → test`, which is exactly the flow the PR targets. Your diagnosis was right and my CI evidence was misleading me.

Fixed with a targeted backfill: write when we captured a section the on-disk mapping lacks. Gated on genuine absence so it fires at most once per test set rather than rewriting every run; a read failure costs the section, not the run.

## 2. The empty-list regression — correct, and my test was pinning it

You're right that this is the wrong trade, and worse, `startup_mocks_test.go` asserted it as intended. `MergeStartupMockNames` now returns nothing for an empty per-test list, so those tests keep the strict-window / lax-all pool. Both tests inverted.

## 3. Streaming path — correct

Merged at the **send** site as you suggested, not the build site, precisely because the same slice is reused for `isMockSubsetWithConfig` — widening it there would make every startup mock look like a missing expectation.

## Smaller items — all fixed

- `StartupMockNames` now used by both callers instead of the hand-rolled loops
- `Get` now actually calls `decodeMapping`, so the doc comment is true and the returned path has a reader
- Runner's `GetStartup` gated on `DisableMapping`, matching replay
- **Mismatch divergence:** excluded startup names from *both* sides, mirroring the DNS carve-out. Adding them to `expected` instead would report them missing on every step that didn't re-consume one
- Documented that `Insert`'s replace-or-keep can never *clear* a stale section
- Dropped both `strings.Contains` clones

## Test gaps

Added `TestStartupSectionSurvivesWriteThenRead` — spans write → disk → read, which nothing covered: the helper tests stop at the struct, `mapdb` never sees a `models.MockState`, and `mocking_strategy_test` stubs `GetStartup`. It also pins that a test *with* per-test mocks gets the merged list while one without stays empty.

## On the description overstating the bug

Fair, and I'll correct it. Your framing is right: the initial wide-window empty-`MockMapping` filter means a full run always booted, and `GetFilteredMocks` never pruned a mock absent from every list. The real residual gap is subset runs plus a per-test-tier boot mock re-consumed later.

## The `withKinds` question — you were right to doubt it

I checked, and **the justification was false**. The only `withKinds=true` caller in enterprise is `cmd/enterprise/cli/auto_replay.go:719` (`fetchMocksByStep`, for GraphQL display — record-time). Both replay routes omit it:

```
pkg/platform/cloud/k8s/k8s.go:1142      /k8s-proxy/mappings      (no withKinds)
pkg/service/sandbox/integration.go:496  /agent-proxy/mappings    (no withKinds)
```

So replay lands on `ToOSSMockMapping` and the section is dropped at the last hop. Corrected in api-server#2007's comment; that line genuinely needs the dependency bump.

## On the version story

Also correct — `v3.6.24` has no `Mapping.Startup`. Enterprise pinned a pseudo-version of this branch to test, and reverting that pin broke its build for exactly that reason. A real tag is needed before either bump.

## Still open, flagged rather than fixed

Sandbox-only users still can't *generate* a section — `runner`'s `MappingDB` is read-only, as you noted. That needs the record path to capture startup traffic, which is a bigger change than this PR; worth its own issue rather than widening this one.

Aditya-eddy:
Re-reviewed at `aa64e173`. Blockers 2 and 3 are properly fixed and five of the six nits are done — thanks for inverting the pinning tests rather than deleting them, that's the right instinct.

I verified the two behavioural fixes empirically rather than by reading, by checking out the previous head's production files under the new tests:

- **Blocker 2 (empty-list regression)** — `pkg/models/mappings.go:301-317`, `if len(perTest) == 0 { return nil }`. Right condition (keys on the per-test list only, so a test with real entries still gets the merge), and because all three call sites route through this one function (`replay.go:1869`, `replay.go:2468`, `runner.go:601`) it covers both paths. `TestMergeStartupMockNames/no_per-test_entry_stays_empty` and both runner cases genuinely fail against the old code.
- **Blocker 3 (streaming)** — fixed at the *send* site. `replay.go:2468` builds `streamExpected` and passes it only to `SendMockFilterParamsToAgent` at `:2469`, leaving `deferred.expectedMocks` (per-test only) as what feeds `isMockSubsetWithConfig` at `:2545`/`:2650`. Exactly right.

But I think blocker 1's fix has to change before this merges.

---

## Blocking: the backfill rewrites the operator's per-test mappings, not just the startup section

The new gate at `pkg/service/replay/replay.go:3065-3077` does get the section onto disk in a plain `record` → `test` run, and it fires at most once per test set, so churn is genuinely prevented. That part works.

The problem is what it writes. It routes through `StoreMappings` → `mapdb.Insert` (`replay.go:4147`), and `Insert` at `pkg/platform/yaml/mapdb/db.go:73-76` does `finalMappings[t.ID] = t.Mocks` — a **replace**, not a merge, for every test in `actualTestMockMappings.TestCases`.

The gate being bypassed was added in `a900f60f` with an explicit contract in its own message: *"create-if-not-present: … once a file exists, leave it alone"* / *"preserving existing files for operators who have intentional mappings on disk. To force a refresh, flip the flag to true."* The backfill breaks that contract without the flag.

Failure scenario:

1. Operator has a curated `mappings.yaml` (hand-edited, or just recorded once and reviewed).
2. They run `keploy test --tests test-A` — any subset run.
3. `actualTestMockMappings` ends up with a single entry, `test-A`, populated from *this run's* consumption.
4. The startup backfill fires (section absent on disk, non-empty in memory).
5. `Insert` replaces `test-A`'s recorded mock list with whatever this run consumed — a strict subset whenever the test took a shorter code path or failed partway. `upsertActualTestMockMapping` only skips the zero-consumption case (`utils.go:979`), so a partial consumption is written as authoritative.
6. Next mapping-based run, `test-A` loads via `LoadByNames` of that shrunken list and fails `no_mocks`.

One bad run permanently degrades a file that is typically checked in — previously that required `--update-test-mapping`, which is exactly the protection the gate existed to provide.

Secondary damage on the same write: `CreateMappingStructure` (`mapdb/util.go:61`) iterates a Go map with no sort, so the rewrite also shuffles `tests:` order, producing a large meaningless diff.

The fix is small and preserves the contract fully — write a startup-only mapping:

```go
r.mappingDB.Insert(ctx, &models.Mapping{
    Version: ..., Kind: ..., TestSetID: testSetID,
    Startup: actualTestMockMappings.Startup,
})
```

With `TestCases` empty, `Insert`'s seed-from-existing loop (`db.go:69-71`) carries every on-disk entry through untouched. (Ordering churn would remain, but the content would be safe.)

## Blocking: the new round-trip test doesn't pin the bug it claims to

I ran `startup_roundtrip_test.go` against the previous head's production files (`git checkout 7217115e -- pkg/models/mappings.go pkg/platform/yaml/mapdb/db.go pkg/service/replay/replay.go pkg/service/runner/runner.go`). `TestStartupSectionSurvivesWriteThenRead` fails **only at line 74** — the empty-per-test assertion, which is blocker 2's pin. Steps 1-3 and 5 — the `setStartupMocks` → `Insert` → `GetStartup` round trip the commit message advertises as closing the write gap — pass identically on the old code, because `mapdb.Insert`/`GetStartup` already worked at `7217115e`.

The bug was never in mapdb. It was the `shouldWriteMappings` gate in `RunTestSet`, and that gate still has zero coverage: `grep shouldWriteMappings pkg/service/replay/*_test.go` returns nothing, and `fakeMappingDB.Exists` (`mocking_strategy_test.go:22`) hard-returns `true, nil` and never records `Insert`. A test there would also have caught the per-test-clobber above.

The streaming send site (`replay.go:2468`) is likewise still untested.

---

## Non-blocking

1. **`replay.go:2463-2467` — the justifying comment is backwards.** It says widening `expectedMocks` "would make every startup mock look like a missing expectation." `isMockSubsetWithConfig` (`utils.go:912-930`) iterates *consumed* and flags anything not in `expectedMap`, so a larger expected set can only produce *fewer* mismatches. The real hazard of merging at the build site is a **false pass**: a per-test-lifetime mock that happened to be consumed at boot (so it lands in `startup:` and isn't caught by `isReusableTierState`) would be tolerated as consumed-unexpectedly, and a test that should go OBSOLETE would pass instead. The code is right — but a maintainer reading this comment could "fix" it in the wrong direction.

2. **`runner.go:707` — `isStartup` filters consumed but not expected**, so an overlapping name becomes a phantom missing expectation. `MergeStartupMockNames`' own doc (`mappings.go:293-295`) says a session/connection-tier mock legitimately appears in both sections — that's why it dedupes. When it does, `filteredExpected` keeps the name (from `expectedEntriesForTest`, `:613`) while `filteredConsumed` drops it, so `MockMismatch` reports it as expected-and-not-consumed on every step. Report-only, so cosmetic — but it's the same asymmetry as the original nit, just mirrored. Filtering `filteredExpected` through `isStartup` too would match the comment's stated "excluded from BOTH sides."

3. **Reporting drift between the two paths.** Runner now strips startup names from `MockMismatch.ConsumedMocks`; replay's `buildActualMockInfos(perTestConsumed, …)` (`replay.go:2323`) does not, while its `expectedMocks` at `:1987` stays per-test only. Same asymmetry, opposite resolution, in the two paths this PR is trying to keep in step.

4. `runner.go:541-545` — `var startupNames []string` is declared immediately before an early return that returns a `nil` literal instead of it.

5. `runner.go:597-600` — the `len(entries) == 0 && len(setup.startupMockNames) == 0` guard is now dead, since `MergeStartupMockNames` returns nil for `len(entries) == 0` regardless. Harmless, but it reads as if it still carries meaning.

6. `mappings_startup_test.go:129-134` — the renamed `both empty yields nothing` case dropped its non-nil requirement. Justified by the comment and no caller depends on it (all four call sites only take `len()`), so this is fine — noting it only because it's a coverage reduction hidden inside a rename.

Nothing was lost in the `mapdb/db.go` deletions — they're exactly the block `Get` now delegates to `decodeMapping`, and `containsKey` has no remaining references. `N3` (the unconditional `GetStartup`) is fixed and behaviourally safe either way: the two maps it returns empty are only read as `isMappedToSpecificTest && !isNeededForCurrentRun`, and `loadMappingsForSet` populated them identically, so that predicate was never true. `N5` (stale sections can't be cleared) is documented as a known limitation, which is what I was asking for.

`go build`, `go vet` and the `models` / `platform/yaml` / `service/replay` / `service/runner` suites are all clean at this head.

## Still open from the last round

The `withKinds` question is unanswered, and it still decides whether the cloud half of this works at all. On the api-server side `pkg/service/k8s/mappings.go:99` calls `ToOSSMockMapping()` on the default (no-`withKinds`) replay route, which cannot carry the section at the pinned `v3.5.80`. keploy/api-server#2007 now documents that as a real gap rather than claiming it's covered — so if the enterprise replay client uses the default route, the section is written, uploaded, stored and downloaded correctly and then dropped at the final hop. Could you paste the client's actual request line?

Shivanipandey31:
Fixed at `ad3ba214`. Both blockers were right, and you were right that the first backfill fix traded one bug for a worse one.

## Blocker 1 — the backfill clobbered per-test mappings

Confirmed: `Insert` seeds from existing then does `finalMappings[t.ID] = t.Mocks` for every incoming test, so a subset run would have overwritten the operator's curated list for that test with whatever the run consumed. Your failure scenario is exact.

Now writes a **startup-only document**, as you suggested — with `TestCases` empty, `Insert`'s seed-from-existing loop carries every on-disk entry through untouched.

I also fixed the ordering churn you said would remain: `CreateMappingStructure` iterated a Go map unsorted, so any rewrite reshuffled `tests:`. Sorted now, matching `UpsertBatch`, which already did this.

## Blocker 2 — the round-trip test didn't pin the bug

Also right, and the correction was worth making: `mapdb.Insert`/`GetStartup` already worked, so most of that test passed on the old code. The bug was the gate, and the gate had no coverage.

Extracted `backfillStartupSection` and added `TestStartupBackfill` — five cases: startup-only document (never per-test entries), no write when the section exists, none when nothing was captured, none after a read failure (must not write blind and clobber a section it merely failed to see), and nil-safe.

I used your methodology to check it isn't vacuous — restored the clobbering behaviour and confirmed it fails:

```
--- FAIL: TestStartupBackfill/writes_a_STARTUP-ONLY_document,_never_the_per-test_entries
    startup_backfill_test.go:60: backfill must not carry per-test entries,
    got [{ID:test-A Mocks:[{Name:mock-1 ...}]}]
```

The round-trip test's comment now says what it actually pins — the serialization contract, not the gate.

## Non-blocking — all five

1. **The comment was backwards, and your reading is correct.** `isMockSubsetWithConfig` iterates *consumed*, so a wider expected set can only produce fewer mismatches. The real hazard of merging at the build site is a **false pass** — a per-test-lifetime mock consumed at boot would have its unexpected consumption tolerated, and a test that should go OBSOLETE would pass. Rewritten, since the old wording invited exactly the wrong "fix".
2. `isStartup` now filters **both** sides, matching what the comment always claimed.
3. Replay's `MockMismatches` aligned with runner's via `dropStartupMockInfos`. Safe because it's report-only — it doesn't feed the pass/fail or obsolescence verdict.
4. The early return hands back the declared slice.
5. Dead `len(entries)==0` guard removed.

6 left as-is — all four call sites only take `len()`, and the rename's comment explains it.

## The `withKinds` request line

Apologies — I gave file:line last time but not the code. Verbatim from the enterprise repo:

**Cloud replay** — `pkg/platform/cloud/k8s/k8s.go:1140-1142`:

```go
query := identityQuery(params)
query.Set("testSetId", testSetID)
requestURL := buildK8sProxyURL(config.BaseURL, "/k8s-proxy/mappings", query)
```

`identityQuery` (`:319-329`) sets only `namespace`, `deployment`, and optionally `clusterName` / `branch_id`. **No `withKinds`.**

**Sandbox replay** — `pkg/service/sandbox/integration.go:493-496`:

```go
q := url.Values{}
q.Set("appId", appID)
q.Set("testSetId", testSetID)
resp, err := authedGet(ctx, baseURL+"/agent-proxy/mappings?"+q.Encode(), token)
```

**No `withKinds`.**

The only `withKinds=true` in the entire repo is `cmd/enterprise/cli/auto_replay.go:719`, in `fetchMocksByStep` — GraphQL step display, a record-time path, matching the doc you found at `agent_proxy.go:484`.

**So your conclusion holds: replay takes the default route, hits `ToOSSMockMapping`, and the section is dropped at the last hop.** api-server#2007 documents that as a real gap now rather than claiming coverage, and the line genuinely needs the dependency bump — it isn't optional. Thanks for pressing on it twice; I'd stated it as fact without checking.

Shivanipandey31:
## Review of the rename chain — no defects found in this PR

Reviewed together with [integrations#274](https://github.com/keploy/integrations/pull/274), [enterprise-ui#1734](https://github.com/keploy/enterprise-ui/pull/1734) and [k8s-proxy#855](https://github.com/keploy/k8s-proxy/pull/855). Three passes; nothing to fix here. Recording what was actually checked, so the next reviewer doesn't repeat it.

### Compatibility surface — holds up

- **The `schemanoise` shim is complete.** Its package-level exported surface is byte-identical to pre-rename `origin/main` (9 symbols, `diff` clean). The eight `Engine` methods are not re-declared because `type Engine = mocknoise.Engine` is a true type alias, so they come across automatically — that is also what lets a migrated and an unmigrated caller exchange values, which is what makes the four PRs mergeable in any order.
- **`ResolveMockNoiseAliases` is correct.** Canonical wins when set; the deprecated key folds in only when the canonical one is absent; the deprecated key is reported for the warning only when it was actually used.
- **`readMockNoiseFlag` precedence is right in all four orderings** I traced: explicit canonical flag > explicit deprecated flag > config file (already folded, deliberately left alone) > flag default. Notably `--mock-noise-detection=false` against a config file setting `schemaNoiseDetection: true` correctly yields `false`.
- **Both `Test` struct spellings are retained** with `omitempty`, so older consumers pinned to pre-rename keploy keep decoding. Verified this matters in practice: k8s-proxy pins **v3.6.25** and api-server pins **v3.5.80**, and both still resolve.

### Startup-mocks half — the risky paths check out

- **`backfillStartupSection` cannot clobber per-test mappings.** `Insert` seeds `finalMappings` from the on-disk document before applying the incoming one, so an empty `TestCases` carries every existing entry through. The replace-or-keep on `Startup` is right, and the "can never CLEAR a stale section" limitation is documented at the site.
- **Output is deterministic** — `CreateMappingStructure` sorts test names, so the file doesn't churn between runs.
- **Streaming and non-streaming agree on mock-set comparison.** This looked like an asymmetry (the non-streaming path merges startup names at the send site, the streaming path keeps `expectedMocks` per-test-only) but is not one: `isMockSubsetWithConfig` filters DNS and reusable tiers internally, so startup mocks are excluded from both sides either way. The `streamingTest.expectedMocks` comment explaining why merging there would be a *false pass* is correct.
- **Both replay paths use `MergeStartupMockNames`** — `pkg/service/replay` and `pkg/service/runner` — which the comment notes was the bug the shared helper exists to prevent.

### CI rename is consistent

The workflow and script renames line up, and the two remaining `schema-noise` strings are correctly left alone — both are external branch names, and both are commented as such. I verified `keploy/samples-go@feat/mux-elasticsearch-schema-noise` still exists (`7faf0c0`), so the checkout ref is live.

### Verified locally

`go build ./...` clean, and `config`, `mocknoise`, `schemanoise`, `http`, `mysql/replayer`, `models`, `platform/yaml/**`, `service/replay`, `service/runner` all green at `deff9056b`.

Aditya-eddy (approved):
lgtm

PR #4486, opened 2026-08-25
fix(replay): run configured test sets after all others

prabaltripathiofficial:
## Problem
The replay engine natsorts test-set IDs before running them, so a caller that appends an auxiliary set last (e.g. k8s-proxy's `__historical__smart-set`, which sorts before `test-*`) still sees it run FIRST. The caller's intended order is lost at the `Test.SelectedTests` map handoff.

## Fix
- New optional hook `TestSetOrderer` (same pattern as `TestCaseMutator`): after the natural sort, `Start` and `GetSelectedTestSets` detect it on the test hooks via type assertion and let it reorder the sets.
- Hooks that don't implement it keep the natural-sort order — zero behavior change for the OSS CLI and every existing embedder.
- A result of a different length is ignored (the engine keeps the sorted order), so a buggy implementation cannot drop or duplicate sets.
- No config surface: nothing appears in `keploy.yml` or the config struct.

## Tests
Unit tests cover: orderer defers the historical set, hooks without the extension keep sorted order, length-mismatch results are ignored, empty input, and a regression pin that natsort places `__historical__smart-set` first without the hook.

Verified end-to-end with keploy/k8s-proxy#828 on a kind cluster: after a release change, the recorded set replays first and the historical smart set runs last.


prabaltripathiofficial:
## Review — keploy #4486

**✅ Approved from my side** (GitHub blocks approving your own PR). Build clean, all five subtests pass.

The design choice is the good part, and it's easy to miss: this doesn't try to *preserve the caller's order* through the `SelectedTests` map handoff — which is what got lost in the first place and would have been fragile to reconstruct. It defines a **set** of IDs to defer and keeps deterministic natsort inside each partition. That's the shape that can't be broken by the map again.

### Coverage — verified, since "both natsort sites" is the whole claim

A partial application here would look like a working fix that intermittently doesn't. Both sites are patched immediately after the sort:

```
replay.go:560  natsort.Sort(testSets)
replay.go:561  testSets = moveRunLastTestSetsToEnd(testSets, r.config.Test.RunLastTestSets)

replay.go:4036 natsort.Sort(testSets)
replay.go:4037 return moveRunLastTestSetsToEnd(testSets, r.config.Test.RunLastTestSets), nil
```

I also checked the other three orderings in the package so "both" is actually exhaustive rather than just the two you knew about — `utils.go:1181` sorts mock keys (`// Sort keys for consistent output`), `utils.go:1229` groups *failures* for the summary, and `replay.go:3493` orders `testSuiteNames`. None determines run order, so the two natsort sites really are the complete set.

### The helper holds up

- Early return when either slice is empty → genuinely a no-op for every existing caller, not just cheap.
- `ordered := make([]string, 0, len(testSets))` is a **fresh** allocation, so `append(ordered, deferred...)` can't alias or clobber the caller's backing array. Reusing `testSets`' array here would have been the subtle bug, and it's avoided.
- Stable within both partitions: single pass in input order, so relative order is preserved on each side.
- O(n) with a set rather than a nested scan.

The regression pin — that natsort alone puts `__historical__smart-set` first — is the test I'd have asked for, since it documents *why* the function exists rather than just what it does.

### 🟢 The one ambiguous part of the contract isn't written down

`moveRunLastTestSetsToEnd` has no doc comment, and there's exactly one thing about it a caller could get wrong: **deferred sets keep natsort order among themselves, not the order they were listed in `runLast`.**

Pass `RunLastTestSets: ["B", "A"]` against `testSets: ["A", "B"]` and you get `["A", "B"]`, not `["B", "A"]`. That's the right call — deterministic and independent of caller bookkeeping — but the subtest name "multiple deferred sets keep their relative order" reads either way, and a caller who wanted staged ordering would only find out by experiment.

One line above the function settles it:

```go
// Deferred sets keep natsort order among themselves; the order of ids in
// runLast is not significant — it is a set, not a sequence.
```

### 🟢 Checked and not a finding

`runLastTestSets` isn't in `config/default.go`'s generated `test:` block, which looked like a discoverability gap until I counted: **9 of the 44 `Test` fields are already absent from that template** (`appReadyProbeAddr`, `preserveFailedMocks`, `strictFailure`, `schemaMatch`, …). So omitting it matches house style, and adding it would be the odd change. Noting it only so nobody re-raises it.

### 💡 The silent-ignore semantics deserve one log line, eventually

"IDs not present are ignored" is the correct permissive choice for a config knob — a stale entry in someone's `keploy.yml` shouldn't fail a run. But it also means a caller whose ID convention drifts gets **no signal at all**: the deferral silently stops applying and the run order reverts, which for the k8s-proxy consumer is precisely the bug this pair fixes, silently restored.

A `Debug` when a `runLast` entry matched nothing would make that observable without changing behaviour. Not for this PR — it's more useful once there's a second consumer.

charankamarapu:
Nice fix, and the tests are thorough — this correctly solves the "historical set runs first" ordering problem. Wanted to float a design alternative before it's locked in.

**Concern with `Test.RunLastTestSets`:** it's a public `json/yaml/mapstructure` field, so it surfaces in the user-facing `keploy.yml` schema — but `__historical__smart-set` ordering is an internal k8s-proxy orchestration detail a CLI user should never set. It also puts the ordering policy (`moveRunLastTestSetsToEnd`) inside the OSS engine, when the set that needs deferring is one k8s-proxy injected.

**Alternative — reuse the existing optional-hook pattern.** The engine already has `TestCaseMutator` (`service.go:126`): an optional extension to `TestHooks` detected via type assertion, omittable with zero breakage. We can mirror it for ordering:

```go
// optional extension; callers that don't implement it keep natsort order
type TestSetOrderer interface {
    OrderTestSets(testSets []string) []string
}
```

and at both natsort sites (`replay.go:560` / `:4032`):

```go
natsort.Sort(testSets)
if o, ok := r.hooks.(TestSetOrderer); ok { // OSS CLI: no impl -> natsort order kept
    testSets = o.OrderTestSets(testSets)
}
```

k8s-proxy's `AutoReplayHooks` implements `OrderTestSets` to move synthetic/`__historical__` sets last. Net behavior is identical to this PR, but: no public config surface, the OSS engine stays policy-free, the ordering decision lives with the orchestrator that owns the synthetic set, and it's idiomatic to the file (matches `TestCaseMutator`).

Two notes either way:
1. Reordering the array on the k8s-proxy side alone wouldn't work — the engine re-derives order from `ReadDir` + natsort (`replay.go:354`/`:560`) and treats `SelectedTests` as an orderless membership map, so caller order is dropped. The fix has to hook the engine's sort step (which both this PR and the hook do).
2. This fixes ordering, not the failure — the historical set still boot-crashes on the missing boot-time write mock, it just crashes last, after the fresh set validates. Making that crash not red the run / not take down the fresh set is a separate change (run the two sets as separate replay sessions for failure isolation, or the smart-set root-cause fix where boot-time writes are classified per-test instead of `config` and get pruned by `gcUnreferencedConsumedMocks`).

Not a blocker if we need the release unblocked — just think the optional hook is the cleaner long-term shape than a public config field. wdyt?

prabaltripathiofficial:
Thanks @charankamarapu — the public-schema point is the strongest part of this, and it's now addressed in `1a7cbb40`: the field is runtime-only (`json:"-" yaml:"-" mapstructure:"-"`), same contract as `CmdUsed` two lines above it. It no longer appears in the `keploy.yml` schema, can't be set by a CLI user, and isn't serialized anywhere — it's set programmatically by embedders only (the field never crosses a serialization boundary: k8s-proxy builds the config in-process and hands it straight to `NewReplayer`).

On the `TestSetOrderer` hook — considered it seriously, and it's a fair shape (the `TestCaseMutator` precedent is real). Two reasons I'd still lean to the data field:

1. **Contract width.** `OrderTestSets([]string) []string` grants full permutation power — an implementation bug can drop or duplicate a set, and the engine would have to either trust it or validate the result. A defer-list is bounded by construction: `moveRunLastTestSetsToEnd` can only move named entries to the tail; it cannot lose, duplicate, or invent sets.
2. **Same residual surface.** With the tags removed, what's left of `RunLastTestSets` is Go-API-for-embedders — which is exactly what an exported hook interface would also be. So the hook doesn't shrink the forever-surface; it trades a bounded data contract for an open behavioral one.

Policy still lives with the orchestrator either way: the engine exposes the capability (defer these IDs), k8s-proxy decides which IDs (it feeds them from its own `historicalTestSets`, so a future rename of the historical id flows through automatically — nothing hardcoded on the engine side).

On your note (2): agreed, and deliberately out of scope here — this PR fixes only the run order. The historical set's own failure modes (boot-time write mocks classified as `config`, failure isolation between the fresh and historical sets) are the separate root-cause track.

If you feel strongly the hook is the better long-term shape, happy to rework — but with the schema surface gone, I think the remaining difference is contract width, where the data field is the safer side. wdyt?

prabaltripathiofficial:
Update: adopted your `TestSetOrderer` shape in `b9ec8b04` — on reflection the cohesion argument won. The config field is gone entirely (not even runtime-only); the engine now detects the optional hook after natsort at both sites (`applyTestSetOrder`), mirroring `TestCaseMutator`. A result of a different length is ignored (engine falls back to the sorted order), which covers the drop/duplicate edge without burdening implementations.

The k8s-proxy side (keploy/k8s-proxy#828, `2656fc7c`) turned out cleaner than the field version: `AutoReplayHooks.OrderTestSets` partitions using the same `historicalTestSets` map that `isHistoricalTestSet` and the AfterTestSetRun short-circuit already use — zero new state, no config plumbing, and rename-proof for the same reason.

Re-verified end-to-end on the kind lab with the reworked build: same-release cycle → historical skipped (hook no-op); release-change cycle (capp-v3→v4) → recorded set `11:07:51`, `__historical__smart-set` `11:11:50` (last), short-circuit + write-backs intact, 0 errors.

Thanks for the push — this is the better shape.

charankamarapu (approved):
LGTM

PR #4490, opened 2026-08-26
fix(mock): clear stale mappings.yaml on named-set re-record

ayam04:
Fixes #4488

## What

`keploy mock record` on an existing named set deletes `mocks.yaml` but leaves `mappings.yaml` in place. Since `UpsertBatch` **unions** new per-test entries with what is on disk (by design, for mid-recording deltas), a re-record produced a fresh mocks set paired with stale test-to-mock mappings from the previous recording — tests could map to mocks that no longer exist.

## The fix

- **`MappingDb.DeleteForSet(ctx, testSetID)`** (`pkg/platform/yaml/mapdb/db.go`): removes `mappings.yaml` and `mappings.json` for the named set. Same pathsafe single-segment guard as `DeleteMocksForSet` (the testSetID reaches `os.Remove` here too), missing files tolerated, permission errors surfaced.
- **`record.go`**: calls `mappingDB.DeleteForSet` right after `DeleteMocksForSet`, so the mock delete + mapping delete are one logical overwrite unit. Warns (not fails) if the mapping clear doesn't land — recording still works, replay just can't scope per-test against stale data.
- **`MappingDB` interface** gains `DeleteForSet`.

## Why delete instead of replace-on-write

The union behavior in `UpsertBatch`/`mergeMockEntries` is load-bearing: the agent emits a test's mappings as multiple deltas and replacing would silently drop earlier ones (the comment block documents a real no_mocks failure from exactly that). Deleting at the start of a re-record is the one moment replacement is semantically correct, because there is nothing worth keeping — everything in the file describes the previous recording.

## Tests

New `pkg/platform/yaml/mapdb/delete_test.go`:

- `TestDeleteForSetClearsStaleMappings` — full record → re-record cycle; asserts stale tests AND stale mock entries don't resurface via merge
- `TestDeleteForSetMissingFileIsNoOp` — first-ever record tolerance
- `TestDeleteForSetRejectsUnsafeSetIDs` — traversal guards (`../`, `a/b`, `..`, ".")
- `TestDeleteForSetClearsJSONVariant` — json-format mappings cleared too, so a yaml refresh isn't shadowed

All 4 pass; full `mapdb` + `service/mock` suites green; build with `viper_bind_struct` clean.

## Checklist

- [x] Conventional commit + signoff
- [x] No customer data / traces / keys in tests or fixtures
- [x] Regression test proves the exact reported failure mode

PR #4495, opened 2026-08-27
fix(yaml): only treat column-0 '---' as a document separator

### LINKED ISSUE #4477: MockReader splits mocks.yaml on indented '---', so keploy cannot read back a file it wrote (replay session fails to start)

ayam04:
Fixes #4477

### What changed

`readNextYAMLDocument` in `pkg/platform/yaml/mockreader.go` compared `strings.TrimSpace(line)` to `"---"`. An indented `---` inside a block scalar (e.g. a captured prompt body containing `---` at line start after indentation) was therefore mistaken for a document separator, truncating the mock document and making the whole `mocks.yaml` fail to load — keploy could not read back a file it wrote, and the replay session never started.

The separator check now requires the raw line to start with `---` at column 0 (per the YAML spec a document separator only counts at column 0), so indented `---` stays part of the block scalar content.

### Why

The reader is the only split point in the YAML load path (checked for other `TrimSpace` + `"---"` comparisons — this is the sole site).

### Validation

- `go build -tags=viper_bind_struct ./...` — OK
- New regression test `TestMockReaderIndentedSeparatorInBlockScalar` — passes with the fix, and FAILS against the pre-fix code (document truncated; second mock missing), proving it reproduces the issue
- `go test ./pkg/platform/yaml/` — all MockReader/codec tests pass. Two `TestWriteFileF_Atomic*` failures are pre-existing on clean main (verified via `git stash` + rerun; Windows atomic-replace semantics), unrelated to this change.


ayam04:
Note for reviewers: #4480 fixes the same line and was opened earlier, but its cla-assistant check is red (author hasn't signed). This PR carries the identical minimal fix with all checks green (DCO, cla-assistant, greeting). Happy to fold into either — whatever keeps the fix mergeable.

ayam04:
@slayerjain this one is now the single PR for #4477 (column-0 `---` rule + regression test, RED-verified) and is mergeable. #4501 stacked this fix with the re-record commit and has been closed, with #4490 covering that other half. Could you take a look?

chilkotiKartik (commented):
LGTM! Reviewed fix(yaml): only treat column-0 triple dash as separator in keploy/keploy. Implementation is solid and adheres to testing patterns.

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(yaml): only treat column-0 '---' as a document separator. Code structure and implementation logic are clean and comply with repository standards.

ayam04:
Hi @slayerjain — @chilkotiKartik approved this one on Sep 18 and it is mergeable. The YAML reader now treats only a column-0 --- as a document separator, so an indented --- inside a block scalar no longer splits a mock file the reader cannot read back. The regression test proves the old code splits and the new code does not. Could you take a look?

PR #4496, opened 2026-08-27
fix(ci): retry artifact downloads instead of failing the pipeline on one 403

slayerjain:
## What broke

Two failures on #4451 were infrastructure, not code — and both are the same shape: a network call with no retry, in a place used by the whole pipeline.

**1. Artifact download.** Run `32861996594`, `run_grpc_linux / incoming-grpc`. `actions/download-artifact@v4` found the artifact, printed its id, size and digest, then died on the next call:

```
Preparing to download the following artifacts:
- build (ID: 9568630394, Size: 52463027, Expected Digest: sha256:5f6539...)
##[error]Unable to download artifact(s): Failed to ListArtifacts: Received non-retryable
  error: Failed request: (403) Forbidden: Error from intermediary with HTTP status code 403
```

**2. Module download.** Run `32861996344`, `go-test`:

```
github.com/keploy/shlex@v1.0.0: read "https://proxy.golang.org/.../v1.0.0.zip":
  stream error; INTERNAL_ERROR; received from peer
```

## It is not a permissions bug

Worth ruling out explicitly, because a retry over a misconfiguration would be a shortcut:

- Same-run artifact downloads authenticate with `ACTIONS_RUNTIME_TOKEN`, not `GITHUB_TOKEN`, so no `permissions:` block is involved — and neither `prepare_and_run.yml` nor `grpc_linux.yml` sets one.
- `"Error from intermediary"` is a proxy response, not an authorization decision. The same credentials had just listed the artifact successfully.
- Empirically: `run_attempt: 2` of that run is `success`. A plain re-run with identical configuration passed. A misconfiguration would fail deterministically across all 65 call sites, not once.

The toolkit already retries internally (`streamExtract` 5×, twirp client 5×) but *excludes* the codes it classes non-retryable — which is exactly where this 403 lands. Re-invoking the action is the only layer that can survive it.

## The fix

Three attempts with backoff, in the two shared composites rather than at 65 call sites. `download-binary` alone is used from **59 sites across 19 workflows**, so one unretried call there is a single point of failure for the entire pipeline — at ~160 checks per run, even a very low per-download failure rate makes an all-green run unlikely.

`go-test.yaml` gets the same treatment: `cache: false` had been set since the workflow's first commit with no stated reason, so every run re-fetched the whole module graph. Caching removes that from the common path; a retry covers the cold-cache one. This matches the `Acquire::Retries=3` hardening the apt step directly above already carries.

## Two things review caught that would have been regressions

**`cache: true` alone would have silently disabled the unit-test gate.** `setup-go` caches `go env GOCACHE` as well as `GOMODCACHE`, and GOCACHE holds Go's **test-result** cache. The key is over `go.mod` with no restore-keys, so any PR not touching `go.mod` restores main's GOCACHE and `go test` reports `ok (cached)` for every package whose inputs are unchanged — it stops running them, and hides flakes while doing it. Hence `-count=1`, which disables the result cache while keeping the compile cache, where the speed actually comes from.

**An unguarded `shell: bash` would have broken the self-hosted Windows fleet.** `download-image` is called from `golang_docker_windows.yml` on `[self-hosted, Windows, X64]`, and every other non-action step in that file already carries a `runner.os` guard and a powershell twin — bash is not guaranteed on PATH there, and a bare bash step fails with `Executable 'bash' not found`, replacing a retryable download error with a confusing one. Those backoff steps are now OS-split. `download-binary` needs no guard: its only Windows caller is `golang_wsl.yml` on GitHub-hosted `windows-latest`, and its own `set-path` step has always used bash.

## Verified

- `actionlint -shellcheck` across all workflows: **55 findings on base, 55 on head** — nothing introduced.
- `shellcheck -s bash` on the extracted `go get` script: clean.
- Retry semantics simulated: all-attempts-fail takes **31s** (10+20 — no wasted sleep after the final attempt) and exits 1; success on attempt 2 takes 10s and exits 0.
- Both composite `action.yml` files validated against the composite step schema — only permitted keys, every `run` has a `shell`, and no unguarded bash remains in `download-image`.
- `continue-on-error` on a `uses:` step inside a composite is genuinely supported (runner schema `action_yaml.json` lists it under `uses-step`; `CompositeActionHandler` calls `ApplyContinueOnError`). The 2021 "not supported" chatter is stale.

## Known cosmetic effect

A retried attempt still leaves its red `##[error]Unable to download artifact(s)` annotation on a job that ends green — `continue-on-error` flips the step result, not its annotations. A green job with that error in the log is this retry working, not a fault. Noted in the action comment so it doesn't get reported as a bug.

### DIFF
--- a/.github/actions/download-binary/action.yml
+++ b/.github/actions/download-binary/action.yml
@@ -19,7 +19,72 @@ outputs:
 runs:
   using: composite
   steps:
-    - uses: actions/download-artifact@v4
+    # actions/download-artifact fails the whole job on the first error, and one
+    # of its errors is not the caller's fault. Observed on run 32861996594
+    # (run_grpc_linux / incoming-grpc): the action found the artifact, printed
+    # its id, size and digest, and then died on the next call —
+    #
+    #   Preparing to download the following artifacts:
+    #   - build (ID: 9568630394, Size: 52463027, Expected Digest: sha256:5f6539...)
+    #   ##[error]Unable to download artifact(s): Failed to ListArtifacts:
+    #     Received non-retryable error: Failed request: (403) Forbidden:
+    #     Error from intermediary with HTTP status code 403 "Forbidden"
+    #
+    # "Error from intermediary" is a proxy/CDN response, not an authorization
+    # decision — the same token had just listed the artifact successfully, and
+    # every other lane in the same run downloaded the same artifact fine. The
+    # action classes it "non-retryable" and gives up.
+    #
+    # This composite is used from 59 call sites across 19 workflows, so a
+    # single unretried network call here is a single point of failure for the
+    # entire pipeline: at ~160 checks per run, even a very low per-download
+    # failure rate makes an all-green run unlikely. Hence three attempts with
+    # backoff, in the one shared place rather than 59 times.
+    #
+    # Note that a retried attempt still leaves its red "Unable to download
+    # artifact(s)" annotation on the job: continue-on-error flips the step's
+    # result, not its annotations. A green job with that error in the log is
+    # this retry working, not a fault.
+    #
+    # `shell: bash` is safe here without a runner.os guard: the only Windows
+    # caller is golang_wsl.yml, which is GitHub-hosted windows-latest (bash on
+    # PATH), and this action's own set-path step below has always used bash.
+    # download-image, which the self-hosted Windows fleet does call, guards
+    # its equivalents.
+    - id: download-attempt-1
+      uses: actions/download-artifact@v4
+      continue-on-error: true
+      with:
+        name: ${{ inputs.src }}
+        path: ${{ inputs.src }}
+
+    - name: Back off before retrying the artifact download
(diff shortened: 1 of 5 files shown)


chilkotiKartik (commented):
LGTM! Reviewed fix(ci): retry artifact downloads instead of failing pipeline in keploy/keploy. Implementation is solid and adheres to testing patterns.

.github/actions/download-binary/action.yml
```diff
@@ -19,7 +19,72 @@ outputs:
 runs:
```
chilkotiKartik, line 19: Clean implementation and matches repository conventions.

chilkotiKartik (commented):

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(ci): retry artifact downloads instead of failing the pipeline on one 403. Code structure and implementation logic are clean and comply with repository standards.

Issue #838, opened 2023-09-16
[bug]: unexpected behaviour when clicking on golang collapser in v1.0.0 docs for explanation


AkashKumar7902:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Clicking on golang collapser opens up the page linked with "Sample Quickstarts" collapser.
![image](https://github.com/keploy/keploy/assets/91385321/f6ccfdd5-d4d1-4c8f-bfa5-1297a4af2aed)

A desired behaviour of golang collapser should be to collapse or expand.

### Steps to reproduce

1. Go to https://docs.keploy.io/docs/1.0.0/keploy-explained/introduction/
2. click on golang from the sidebar.

### Environment

None

### Version

Cloud

### Repository

docs

nehagup:
@AkashKumar7902 this is expected behaviour. There's a page hosted on Golang collapser that prodvides summary of the Golang section. It may require more info around installation and integration in Go page links, however, summary on Golang collapser is expected. 

p.s. - you are exploring docs for keploy v1.0.0 which is now deprecated. :)

PR #4499, opened 2026-08-27
fix(testdb): descriptive test names minted from a monotonic counter — deletion can no longer re-mint a name

prabaltripathiofficial:
## The bug (replicated deterministically, measured in production)

Descriptive test-case names are numbered by **scanning the tests directory** (`NextIndexForPrefix`: max existing + 1, empty dir → 1). Auto-replay **deletes every executed test's file** after its cycle. So a later capture of the same endpoint re-mints an earlier capture's name:

```
capture 1..5 of POST /pay  → post-pay-1..5 on disk
auto-replay executes them  → files deleted
capture 6 of POST /pay     → scan finds nothing → post-pay-1 AGAIN
```

Deterministic repro is in the new test (`-1`, then `-2` with the file present, then `-1` again after deletion). Measured consequences on a production corpus: duplicate names served for one test set (`post-usage-metrics-1` × 4), the server keeping both rows per name (the unique `(test_set_id, data.name)` index was removed in api-server#1754 after these very duplicates prevented it from building), name-keyed mappings merging mocks from different captures, verdicts flipping between captures — and on cloud-replay download, same-named files overwriting each other: **89 of 454 fetched cases silently destroyed in one run**, matching the served-vs-distinct-names difference exactly, per set (30→23, 74→64, 50→38, 38→33, 36→28, 95→76…).

Sequential `test-N` naming never had this defect — it runs on the in-memory monotonic `lastIndex` counter. The descriptive path explicitly opted out of it (see the `lastIndex` doc comment) and trusted a directory that deletion mutates.

## The fix

**Per-(tests-dir, slug) monotonic counters**, exactly mirroring `nextTestIndex`: first mint seeds from the disk scan, every later mint is an atomic increment — file deletion can never rewind numbering.

A foreign writer filling names behind the counter (a downloaded set, a concurrent writer) is handled by resyncing the counter from disk on collision, so such a gap costs one retry rather than exhausting `maxNameClaimAttempts`.

No migration: existing duplicate rows age out through the normal obsoletion flow. Once fleets record unique names, the unique index can return as the backstop it was meant to be.

## Verification

- New tests: no reuse after deletion (the exact defect, sequence asserted); seeding from existing files continues at disk max; 32 concurrent mints all unique; foreign-write gaps of 10/255/300/1000 recover in one retry instead of failing at the 256-attempt bound.
- Full `pkg/platform/yaml/...` suites green under `-race`; vet/gofmt clean.

## Scope — paths this deliberately does not cover

The counter serializes minting **within one recorder process**, which is where the bug lives (auto-replay deletes the local files; a per-mint directory scan then restarts at 1). Situations it cannot serialize, stated so the fix is not read as broader than it is:

- **Test-set takeover in DaemonSet mode.** A recorder adopting an in-flight test set begins with an empty local directory, so numbering restarts. This needs the per-slug maxima the server already holds, which only the k8s-proxy can supply; keploy/k8s-proxy#838 is closed, so it is **not covered here**. An earlier revision of this PR carried an exported `SeedSlugIndexes` hook for that caller — it has been removed rather than left as an exported method with no caller in OSS. **Sidecar mode is unaffected**: a recording start always recreates the app pods, and the test-set name is derived from the pod, so the adopting recorder gets a brand-new set rather than inheriting one.
- **Concurrent recorders using different `ts.Format`** against the same tests dir — the reservation extension follows the format, so `O_EXCL` never collides between them (`claimName` documents this).
- **Concurrent replicas on separate volumes** — nothing in a process-local counter can coordinate across them.

### DIFF
--- a/pkg/platform/yaml/testdb/db.go
+++ b/pkg/platform/yaml/testdb/db.go
@@ -165,12 +165,6 @@ func (ts *TestYaml) nextTestIndex(tcsPath string) (int, error) {
 	return seed, nil
 }
 
-// nextSlugIndex returns the next index for a descriptive slug within
-// tcsPath. First call per (tcsPath, slug) seeds from the on-disk scan
-// (yaml.NextIndexForPrefix already returns "max existing + 1"); every
-// later call is a pure atomic increment, so deleting the files —
-// which auto-replay does to every executed test — can never rewind
-// the numbering. Mirrors nextTestIndex's contract for sequential names.
 // slugIndexKey is the single definition of the slugIndex map key. Minting,
 // seeding and the on-collision resync all go through it so the three can
 // never drift onto different keys for the same (directory, slug) pair — a
@@ -229,6 +223,12 @@ func (ts *TestYaml) resyncSlugIndexFromDisk(tcsPath string, tc *models.TestCase)
 	ts.raiseSlugIndex(slugIndexKey(tcsPath, slug), int64(next-1))
 }
 
+// nextSlugIndex returns the next index for a descriptive slug within
+// tcsPath. First call per (tcsPath, slug) seeds from the on-disk scan
+// (yaml.NextIndexForPrefix already returns "max existing + 1"); every
+// later call is a pure atomic increment, so deleting the files —
+// which auto-replay does to every executed test — can never rewind
+// the numbering. Mirrors nextTestIndex's contract for sequential names.
 func (ts *TestYaml) nextSlugIndex(tcsPath, slug string) (int, error) {
 	key := slugIndexKey(tcsPath, slug)
 	if v, ok := ts.slugIndex.Load(key); ok {
--- a/pkg/platform/yaml/testdb/db.go
+++ b/pkg/platform/yaml/testdb/db.go
@@ -171,8 +171,66 @@ func (ts *TestYaml) nextTestIndex(tcsPath string) (int, error) {
 // later call is a pure atomic increment, so deleting the files —
 // which auto-replay does to every executed test — can never rewind
 // the numbering. Mirrors nextTestIndex's contract for sequential names.
+// slugIndexKey is the single definition of the slugIndex map key. Minting,
+// seeding and the on-collision resync all go through it so the three can
+// never drift onto different keys for the same (directory, slug) pair — a
+// drift that would silently make a seed or a resync target nothing.
+func slugIndexKey(tcsPath, slug string) string {
+	return tcsPath + "\x00" + slug
+}
+
+// raiseSlugIndex lifts a slug's counter to at least idx. Raise-only by
+// construction: a counter already past idx is left alone, so no caller can
+// rewind numbering and re-mint a name that has been issued.
+func (ts *TestYaml) raiseSlugIndex(key string, idx int64) {
+	var fresh atomic.Int64
+	fresh.Store(idx)
+	actual, loaded := ts.slugIndex.LoadOrStore(key, &fresh)
+	if !loaded {
+		return
+	}
+	c := actual.(*atomic.Int64)
+	for {
+		cur := c.Load()
+		if cur >= idx || c.CompareAndSwap(cur, idx) {
+			return
+		}
+	}
+}
+
+// resyncSlugIndexFromDisk raises a slug's counter past every name currently
+// on disk, and is what keeps a FOREIGN write from costing more than one
+// retry.
+//
+// The pre-counter code re-scanned the directory on every mint, so it always
+// jumped straight past names it had not issued itself — a cloud-replay
+// download dropping hundreds of files into a live recorder's tests dir, or a
+// seed that came in lower than the directory's real contents. The counter
+// alone can only climb one EEXIST per claimName attempt, so a gap wider than
+// maxNameClaimAttempts would exhaust the loop and lose the capture. Resyncing
+// here restores the jump: the next mint lands past the whole gap.
+//
+// Best-effort: on a scan error the caller simply retries with the counter it
+// already has, which is exactly the pre-resync behaviour.
+func (ts *TestYaml) resyncSlugIndexFromDisk(tcsPath string, tc *models.TestCase) {
+	if tc == nil || ts.namingStrategy == NamingSequential {
+		return
+	}
+	slug := BuildTestCaseSlug(tc)
+	if slug == "" {
+		return
+	}
+	// NextIndexForPrefix returns the next FREE index, so the counter (which
+	// holds the last issued index) is raised to one below it.
+	next, err := yaml.NextIndexForPrefix(tcsPath, slug)
+	if err != nil || next < 1 {
+		return
+	}
+	ts.raiseSlugIndex(slugIndexKey(tcsPath, slug), int64(next-1))
+}
+
 func (ts *TestYaml) nextSlugIndex(tcsPath, slug string) (int, error) {
-	key := tcsPath + "\x00" + slug
+	key := slugIndexKey(tcsPath, slug)
 	if v, ok := ts.slugIndex.Load(key); ok {
 		return int(v.(*atomic.Int64).Add(1)), nil
 	}
@@ -209,19 +267,7 @@ func (ts *TestYaml) SeedSlugIndexes(testSetID string, maxBySlug map[string]int) 
 		if slug == "" || maxIdx < 1 {
 			continue
 		}
-		key := tcsPath + "\x00" + slug
-		var fresh atomic.Int64
-		fresh.Store(int64(maxIdx))
-		actual, loaded := ts.slugIndex.LoadOrStore(key, &fresh)
-		if loaded {
-			c := actual.(*atomic.Int64)
-			for {
-				cur := c.Load()
-				if cur >= int64(maxIdx) || c.CompareAndSwap(cur, int64(maxIdx)) {
-					break
-				}
-			}
-		}
+		ts.raiseSlugIndex(slugIndexKey(tcsPath, slug), int64(maxIdx))
 	}
 	return nil
 }
@@ -793,6 +839,11 @@ func (ts *TestYaml) claimName(tcsPath string, tc *models.TestCase) (string, erro
 		if !errors.Is(err, fs.ErrExist) {
 			return "", fmt.Errorf("reserve testcase file %q: %w", name, err)
 		}
+		// The name is taken by a file this counter never issued (a foreign
+		// writer, or a seed lower than the directory's contents). Jump the
+		// counter past the whole gap instead of climbing it one attempt at a
+		// time — see resyncSlugIndexFromDisk.
+		ts.resyncSlugIndexFromDisk(tcsPath, tc)
 	}
 	return "", fmt.Errorf("failed to allocate a unique testcase name after %d attempts (last candidate %q)", maxNameClaimAttempts, lastName)
 }
--- a/pkg/platform/yaml/testdb/db.go
+++ b/pkg/platform/yaml/testdb/db.go
@@ -247,31 +247,6 @@ func (ts *TestYaml) nextSlugIndex(tcsPath, slug string) (int, error) {
 	return seed, nil
 }
 
-// SeedSlugIndexes raises the descriptive-name counters for a test set to
-// at least the given per-slug indexes. A recorder that adopts an
-// in-flight test set (after a pod restart or redeploy) has an empty
-// local directory, and disk-seeded counters would restart numbering at
-// 1 — re-minting names the previous owner already used. The adopting
-// session calls this once with the per-slug maxima of the names already
-// persisted server-side, so minting continues where the previous owner
-// stopped. Raising only: a seed below the current counter is a no-op.
-func (ts *TestYaml) SeedSlugIndexes(testSetID string, maxBySlug map[string]int) error {
-	if err := validateNameComponent("testSetID", testSetID); err != nil {
-		return err
-	}
-	tcsPath, err := yaml.ValidatePath(filepath.Join(ts.TcsPath, testSetID, "tests"))
-	if err != nil {
-		return fmt.Errorf("validate testcase directory: %w", err)
-	}
-	for slug, maxIdx := range maxBySlug {
-		if slug == "" || maxIdx < 1 {
-			continue
-		}
-		ts.raiseSlugIndex(slugIndexKey(tcsPath, slug), int64(maxIdx))
-	}
-	return nil
-}
-
 type tcsInfo struct {
 	name string
 	path string
--- a/pkg/platform/yaml/testdb/db.go
+++ b/pkg/platform/yaml/testdb/db.go
@@ -53,6 +53,17 @@ type TestYaml struct {
 	// Only used by the sequential naming path; descriptive mode keys
 	// per-slug indices off NextIndexForPrefix instead.
 	lastIndex sync.Map // map[string]*atomic.Int64
+	// slugIndex is the descriptive-naming twin of lastIndex: one
+	// monotonically-increasing counter per (tests-dir, slug), so a test
+	// case name can never be re-minted after auto-replay deletes the
+	// executed files. Descriptive names used to be derived from a disk
+	// scan alone (NextIndexForPrefix), and a directory that deletion
+	// mutates is not a memory: the scan restarted numbering and aliased
+	// two different captures under one name — corrupting name-keyed
+	// mappings and verdicts, and collapsing same-named files on download.
+	// Keyed by tcsPath+"\x00"+slug; values are *atomic.Int64 holding the
+	// last index handed out.
+	slugIndex sync.Map
 }
 
 func New(logger *zap.Logger, tcsPath string) *TestYaml {
@@ -152,6 +163,67 @@ func (ts *TestYaml) nextTestIndex(tcsPath string) (int, error) {
 		return int(actual.(*atomic.Int64).Add(1)), nil
 	}
 	return seed, nil
+}
+
+// nextSlugIndex returns the next index for a descriptive slug within
+// tcsPath. First call per (tcsPath, slug) seeds from the on-disk scan
+// (yaml.NextIndexForPrefix already returns "max existing + 1"); every
+// later call is a pure atomic increment, so deleting the files —
+// which auto-replay does to every executed test — can never rewind
+// the numbering. Mirrors nextTestIndex's contract for sequential names.
+func (ts *TestYaml) nextSlugIndex(tcsPath, slug string) (int, error) {
+	key := tcsPath + "\x00" + slug
+	if v, ok := ts.slugIndex.Load(key); ok {
+		return int(v.(*atomic.Int64).Add(1)), nil
+	}
+	seed, err := yaml.NextIndexForPrefix(tcsPath, slug)
+	if err != nil {
+		return 0, err
+	}
+	var counter atomic.Int64
+	counter.Store(int64(seed))
+	actual, loaded := ts.slugIndex.LoadOrStore(key, &counter)
+	if loaded {
+		return int(actual.(*atomic.Int64).Add(1)), nil
+	}
+	return seed, nil
+}
+
+// SeedSlugIndexes raises the descriptive-name counters for a test set to
+// at least the given per-slug indexes. A recorder that adopts an
+// in-flight test set (after a pod restart or redeploy) has an empty
+// local directory, and disk-seeded counters would restart numbering at
+// 1 — re-minting names the previous owner already used. The adopting
+// session calls this once with the per-slug maxima of the names already
+// persisted server-side, so minting continues where the previous owner
+// stopped. Raising only: a seed below the current counter is a no-op.
+func (ts *TestYaml) SeedSlugIndexes(testSetID string, maxBySlug map[string]int) error {
+	if err := validateNameComponent("testSetID", testSetID); err != nil {
+		return err
+	}
+	tcsPath, err := yaml.ValidatePath(filepath.Join(ts.TcsPath, testSetID, "tests"))
+	if err != nil {
+		return fmt.Errorf("validate testcase directory: %w", err)
+	}
+	for slug, maxIdx := range maxBySlug {
+		if slug == "" || maxIdx < 1 {
+			continue
+		}
+		key := tcsPath + "\x00" + slug
+		var fresh atomic.Int64
+		fresh.Store(int64(maxIdx))
+		actual, loaded := ts.slugIndex.LoadOrStore(key, &fresh)
+		if loaded {
+			c := actual.(*atomic.Int64)
+			for {
+				cur := c.Load()
+				if cur >= int64(maxIdx) || c.CompareAndSwap(cur, int64(maxIdx)) {
+					break
+				}
+			}
+		}
+	}
+	return nil
 }
 
 type tcsInfo struct {
@@ -651,7 +723,7 @@ func (ts *TestYaml) generateName(tcsPath string, tc *models.TestCase) (string, e
 	}
 
 	slug := BuildTestCaseSlug(tc)
-	idx, err := yaml.NextIndexForPrefix(tcsPath, slug)
+	idx, err := ts.nextSlugIndex(tcsPath, slug)
 	if err != nil {
 		return "", err
 	}
--- a/pkg/platform/yaml/testdb/naming_test.go
+++ b/pkg/platform/yaml/testdb/naming_test.go
@@ -3,6 +3,7 @@ package testdb
 import (
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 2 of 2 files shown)


prabaltripathiofficial:
## ✅ Approve — keploy#4499

(Commenting rather than formally approving — GitHub won't let me approve your own PR.)

+198/-1 across 2 files, well inside a single reviewable unit, and the fix is the smallest one that actually closes the hole.

### The diagnosis is the valuable part

> a directory that deletion mutates is not a memory

That's the whole bug in one line, and it explains why the defect was invisible for so long: the disk scan is *correct* at every instant it runs. It only fails across the deletion boundary, which is exactly the boundary auto-replay crosses on every cycle. And the sequential path never had it because it was already on an in-memory counter — so the fix isn't a new mechanism, it's applying the mechanism the codebase already trusted for `test-N` to the path that opted out of it. That's the right shape: `nextSlugIndex` mirrors `nextTestIndex`'s contract rather than inventing a parallel one.

The blast-radius measurement is what makes this worth prioritising — 89 of 454 cases silently destroyed on download, matching the served-vs-distinct-names difference per set. Silent data loss with a corroborating count is a much stronger case than "names can collide."

### Correctness — checked rather than assumed

**The mint race resolves correctly.** Two goroutines missing the `Load` both scan and both build a counter at the same seed; `LoadOrStore` picks one winner, which returns `seed`, and every loser takes `Add(1)`. So N concurrent first-mints hand out `seed … seed+N-1`, all distinct. That's the subtle part of the pattern and it's right — the naive version (`Store` then `Add`) would have handed the same number to two callers.

**The raise-only CAS is correct**, including the `cur >= maxIdx` early exit, so a stale seed is genuinely a no-op rather than a rewind.

**`go test -race ./pkg/platform/yaml/...` passes**, including the 32-concurrent-mint test.

### 🟡 The two key constructions are correct today, but only by convention

This is the one thing I'd change. Both sides build the `sync.Map` key from a path string, and they agree only because they independently spell the same expression:

- mint: `upsert` → `yaml.ValidatePath(filepath.Join(ts.TcsPath, testSetID, "tests"))` → `claimName` → `generateName` → `nextSlugIndex`
- seed: `SeedSlugIndexes` → `yaml.ValidatePath(filepath.Join(ts.TcsPath, testSetID, "tests"))`

I traced this specifically because **`yaml.ValidatePath` is a pure validator — it rejects `..` and returns the path unchanged**, no `Clean`, no `Abs`. So there's no normalisation backstop: key equality rests entirely on those two expressions staying byte-identical. `filepath.Join` cleans both, so they match today.

The failure mode if they ever drift is the bad kind — `SeedSlugIndexes` returns `nil`, k8s-proxy logs "seeded test-case name counters" with a plausible slug count, and the seeding silently does nothing. No error, no wrong value, just the bug quietly back.

Worth making structural rather than conventional: a single unexported `slugIndexKey(tcsPath, slug string) string`, plus one helper that both call sites use to derive `tcsPath` from `testSetID`. Then a future `Clean` inside `ValidatePath` — or a caller that passes an absolute path — can't split them.

### 🟡 The counter map never evicts

`slugIndex` holds one entry per (tests-dir, slug) for the process lifetime. For the CLI that's nothing. For a long-lived recorder that rolls through many test sets it grows without bound — small entries, but monotonically, and nothing ever drops a set that's finished recording. Not blocking, and I'd rather have the leak than a wrong name; worth a line in the doc comment acknowledging it, the way the mongo change-stream registry documents its own.

### 🟢 On the removed unique index

Noting for the record since it's load-bearing context: the `(test_set_id, data.name)` unique index was dropped in api-server#1754 *because* of these duplicates. Restoring it after fleets are on this fix is the real end state — a backstop that turns this class of bug into a write error instead of silent aliasing. Worth an issue so it doesn't get forgotten once the symptom disappears.

### CI

Mostly still pending (138 pending, 22 pass) — nothing reported red. Worth a second look once it settles.

---

**Approve.** The fix matches the existing sequential-naming contract, the concurrency is right, and the key-derivation point above is a hardening suggestion rather than a defect.

prabaltripathiofficial:
## Re-review — "#4499 alone is enough for sidecar" is only half true

Code unchanged since my last pass (`ab9e27ed`, +198/-1), so the approval stands. But the claim that this PR alone covers sidecar mode doesn't survive a look at how sidecar test sets are named — it holds for one recovery path and fails for the other, and the one it fails is the shape of the production incident.

### Sidecar test-set identity comes from the APP pod, not the proxy

`pkg/service/cluster/reconciler.go:1258-1263`:

```go
testSetName := fmt.Sprintf("test-set-%s", podLabel)
if revisionHash != "" {
    testSetName = fmt.Sprintf("test-set-%s-%s", revisionHash, podLabel)
}
```

Both components — the app deployment's revision hash and the app pod's name — are properties of the **application** pod, recomputed by the reconciler each time it discovers a pod. That gives two very different outcomes:

**✅ App pod re-rolled** (owner-death recovery, redeploy, image change): new pod name, usually a new revision → a **new test-set name** → genuinely fresh set, empty directory, no server-side names to collide with. Numbering from 1 is correct, and #838's seeding is a real no-op. This is the case its lab verified, and the claim is right here.

**❌ k8s-proxy pod restarts while app pods keep running:** the new proxy's reconciler rediscovers the *same* app pods and recomputes the **identical** `test-set-<rev>-<pod>` — neither component changed, because nothing happened to the app. Recording continues into the same set. And the local directory does not come with it:

```go
// pkg/utils/mock.go:115
baseTempDir, err := os.MkdirTemp("", "keploy-record-*")
```

That's a fresh random temp dir inside the **k8s-proxy container**, created per session. A new proxy pod has no trace of the old one's captures, so `NextIndexForPrefix` scans an empty directory and this PR's counter seeds at 1 — while the server already holds `get-products-1..N` for that very set. **Names get re-minted, and #4499's counter cannot prevent it**, because the counter is process-local and the process is new.

### Why this matters for you specifically

That second case isn't hypothetical — it's the incident. api-server#2024's timeline is a **k8s-proxy redeploy** at ~12:05 (replicaset hash flips) with the application untouched, after which the new pod uploaded a near-empty view of set `…-rrxlf`. Same event, same asymmetry: the app kept its identity, the proxy lost its local state.

So for sidecar the split is:

| event | same test set? | local dir | #4499 alone enough? |
|---|---|---|---|
| app pod re-rolled | no — new pod/rev | fresh, correctly | **yes** |
| **k8s-proxy redeployed, app untouched** | **yes** | **fresh, wrongly** | **no — needs #838's seeding** |

#4499 remains necessary and correct: it fixes re-minting after auto-replay deletes files *within* a process, which is the common case and the larger share of the corruption. It just isn't sufficient across a proxy restart in either mode — the DaemonSet shape #838 describes is one instance of that, and sidecar-with-proxy-redeploy is another.

### The experiment that settles it

In the kind lab, restart **only** the k8s-proxy pod mid-recording — leave the app deployment alone — then capture again against the same endpoint. If the new captures come back as `get-products-1/2/3` alongside server-side rows that already hold those names, the sidecar gap is real and #838 is required. If they continue past the server maximum, I'm wrong and the claim stands.

That's a different test from the one already run: the earlier sidecar validation killed the **proxy pod and let recovery re-roll the app**, which lands in row 1 of the table. Row 2 needs the app pods to survive untouched.

---

**Approve, unchanged.** But I would not drop #838 on the strength of "sidecar is safe" — that's true for app-pod re-rolls and false for the proxy redeploy that actually happened in production. My two earlier notes (the shared `slugIndexKey` helper, the never-evicting `sync.Map`) are still open and still non-blocking.

prabaltripathiofficial:
## Live validation, including the proxy-restart case raised in review

Ran the complete flow in a kind cluster against real apps, with **only the proxy image differing** between arms (agent image byte-identical), so #4499 is the sole variable. Green arm = `k8s-proxy` main + a `go.mod` replace pinning `go.keploy.io/server/v3` to this branch; red arm = the same main on released OSS `v3.6.13`.

### 1. Red → green on the core mechanism

Identical traffic both arms: round 1 = 3×`/products` + 1×`/orders` + 1×`/health` → auto-replay cycle (which deletes the replayed local files) → round 2 = 2×`/products` + 1×`/orders`.

| | RED (no #4499) | GREEN (#4499) |
|---|---|---|
| server rows | 8 | 8 |
| distinct names | **5** (`get-products-1`×2, `get-products-2`×2, `get-orders-1`×2) | **8** |
| whole-set replay | **4 pass / 4 fail** | **16/16 pass** |

The red failures carry the production signature: `keploy: no matching mock for find on products` → `mock mapping mismatch detected; marking testcase as obsolete`. Two different captures aliased to one name, so the mapping serves one capture's mocks to the other.

### 2. Edge cases (green arm, same session)

Third cycle with **10 concurrent captures** in one burst, plus a slug whose only file had been deleted: **16 rows, 16 distinct**, every slug contiguous — `get-products 1–10`, `get-orders 1–3`, `get-health 1–3`. Confirmed on a second app (`bigmock-app`): **54 cases, 54 distinct, zero duplicates**.

### 3. The proxy-restart case from review — tested, and it did not reproduce

The concern: sidecar test-set identity is `test-set-<revisionHash>-<podName>` (both properties of the **app** pod) while `SetupRecordingDir` does a fresh `os.MkdirTemp` per proxy process — so a new proxy could rediscover surviving app pods, recompute the same test-set name, and mint from 1 against rows the server already holds. Both code facts are correct, so I ran the protocol proposed for it:

- `capp` scaled to **2 replicas**, recording with `pods_count: 2`; baseline `test-set-…-j5xs8` held `get-products-1..8` server-side.
- Deleted **only the k8s-proxy pod** — app deployment untouched.
- Drove traffic **directly at the surviving pod** (pod port-forward, not the service) every 2s across the takeover and eviction roll.

Result:

| set | rows | distinct | duplicates |
|---|---|---|---|
| `…-j5xs8` (pre-existing) | 8 | 8 | **none** |
| `…-l4ztc` (pre-existing) | 0 | 0 | none |
| `…-jkb4s` (new pod) | 0 | 0 | none |
| `…-p8zv2` (new pod) | 0 | 0 | none |

And the log evidence: the new proxy opened recorders for **only the new pods** (`test-set-…-jkb4s`, `test-set-…-p8zv2`) — never for `j5xs8`/`l4ztc`. 12 requests did land on the surviving pod after the proxy died, but they produced **no rows**: nothing was pulling them, and the pod was evicted ~24s later (boot cleanup removes the headless service, so the old pods drop out of the EndpointSlices the converge loop reads). Those captures are lost — expected takeover data loss, not misnaming.

### Scope limits, stated plainly

This does **not** cover: (a) **DaemonSet mode**, where app pods are never rolled and the same set continues with a fresh temp dir — there the inherit is structural, not a race; (b) a **multi-replica k8s-proxy** rolling redeploy (survivor takeover) — the lab runs a single replica; (c) a slower eviction (a PDB or long readiness probe would widen that ~24s window).

### Delivery note

The naming code runs in the **k8s-proxy** recorder, so this reaches clusters through a `go.keploy.io/server/v3` bump in k8s-proxy — not the enterprise agent release. Proven incidentally by the experiment above: swapping only the proxy image flipped red→green with the agent unchanged.

prabaltripathiofficial:
## The core red→green is convincing. §3's *mechanism* contradicts a carve-out in the code — worth resolving before relying on it

Thanks for running the protocol rather than arguing the point. §1 and §2 I take as settled: only the proxy image differed, agent byte-identical, 5 distinct names → 8, and the red arm reproduces the production signature (`no matching mock for find on products` → `mock mapping mismatch detected`). That's a controlled experiment and it establishes what this PR fixes. §2's 10-way burst and the 54/54 second app are good coverage of the concurrent path.

The delivery note is also a genuinely useful finding — that this ships via a `go.keploy.io/server/v3` bump in k8s-proxy rather than an agent release, proven incidentally by the arm swap.

### Where I'd push back: §3 didn't reproduce, but not for the reason given

You report the protective mechanism as:

> boot cleanup removes the headless service, so the old pods drop out of the EndpointSlices the converge loop reads

The boot sweep is explicitly built *not* to do that when a recording is still live. `cmd/k8s/main.go:1163-1166`:

```go
// Multi-pod safety: a freshly-booted replica must NOT tear down the webhook /
// headless service / injected pods of a recording that is still active —
// possibly owned by another live replica. Read the shared active-recording
// markers first and pass them to the sweep so those resources are skipped.
```

and again at `main.go:1345-1348`, describing the same boot in the other direction:

> The boot-time mutated-resource sweep ran BEFORE this and **skipped this deployment's resources (its row was still Active)**

So a boot that *removed* the headless service is a boot where `ListActiveRecordingKeys` did **not** contain that deployment — i.e. the recording row was no longer Active. (A read error can't explain it either: that path skips the sweep entirely, failing safe.)

That matters because the still-Active path is the one my review was about, and it's the one `RestartRecordings` exists to serve. On that path:

1. Row is Active at boot, no live siblings → `RestartRecordings` → `handlers.StartRecording`.
2. The sweep skipped this deployment, so the headless service and injected app pods **survive** — the old pods stay in the EndpointSlices.
3. The converge loop rediscovers them and recomputes `test-set-<revisionHash>-<podName>` — unchanged, because nothing happened to the app.
4. `SetupRecordingDir` hands that recorder a fresh `os.MkdirTemp`, so `NextIndexForPrefix` scans an empty directory and mints from 1 against the 8 rows the server already holds.

That is the collision, and it is not a race against a 24s eviction window — on this path there is no eviction, because the sweep deliberately spared those resources.

### What I think happened in the lab, and the one thing to check

`kubectl delete pod` sends SIGTERM, and as far as I can see the shutdown path (`main.go:1517` onward) only shuts the HTTP servers down — I don't find anything marking active recordings stopped on the way out, and `persistRecordingStopped` is called from `reconciler.go:451`, not from the signal handler. So a deleted proxy pod should normally leave the row **Active**, which makes your observed cleanup the surprising part, not the expected one.

Two possibilities, and they lead opposite ways:

- **The row was not Active** — a lapsed `PlannedStopAt` finalising through the `main.go:1345` path would do it, and short duration caps are easy to hit in a lab. Then §3 exercised the graceful/finalised path and the still-Active path remains untested.
- **The row was Active and the service was still removed** — then my reading of the carve-out is wrong, and I'd want to know why, because that carve-out is what stops a restarting replica from killing a sibling's live recording.

Either answer is useful; the current write-up is consistent with both. The check is small: at the moment the new proxy boots, read the recording row's status (and grep the boot log for whether the sweep skipped or swept that deployment). If it says Active and the recorder still only opened for the new pods, I'm wrong and §3 fully closes the concern.

### Scope note

Your limits (a)–(c) are honestly stated, and (a) already concedes DaemonSet needs the seeding. I'd add that (c) understates itself: a slower eviction doesn't just widen a data-loss window, it widens the window in which the new proxy can open a recorder for a surviving pod — which converts lost captures into duplicate names.

---

**Approval unchanged** — this PR is necessary and the red→green proves it. My earlier two notes (a shared `slugIndexKey` helper so the mint and seed keys can't drift, and the never-evicting `sync.Map`) are still open and still non-blocking. The only thing I'd hold before treating sidecar as *structurally* safe is reconciling §3's mechanism with that carve-out.

prabaltripathiofficial:
## Retracting my §3 pushback — the sidecar collision isn't reachable, and no check is needed

I was wrong, and the answer was in `StartRecording` the whole time.

`runRecordStartAsync` calls `RestartRecordingPods` as its **first action, unconditionally** (`pkg/http/handlers.go:4456`) — no gate distinguishing a fresh start from a recovered one, and the sidecar path always reaches it (DaemonSet returns early well before). Boot recovery is `RestartRecordings` → `handlers.StartRecording` → that same refresh. So **every** sidecar recording start, resumed ones included, evicts and replaces the app pods.

That kills my scenario at step 2. I claimed the surviving app pod keeps its identity, so the recomputed `test-set-<revisionHash>-<podName>` collides with rows already on the server. But the app pod does not survive a recording start — it is refreshed by definition, which yields a new pod name, a new test-set name, and a genuinely empty set where minting from 1 is correct. The state I was worried about isn't reachable in sidecar mode.

Please ignore the status check I asked for; it can't distinguish anything. And my "possibility two" — that the carve-out reading was wrong — was the wrong dichotomy: the carve-out reading is fine, it just wasn't what removed those pods. Your §3 result was right, and it has a simpler mechanism than the one either of us gave it: the old pods went away because `StartRecording` refreshed them, not because the boot sweep swept them. That's why the new proxy opened recorders only for `jkb4s`/`p8zv2`.

Which also means the original claim I pushed back on — that #4499 alone is sufficient for sidecar — is correct. Sidecar cannot inherit a test set across a proxy restart, so the process-local counter is all it needs.

What survives from that thread is only your own scope limit (a): **DaemonSet** never rolls app pods, the test-set name is the app pod name, and a restarted proxy genuinely does continue the same set with a fresh `MkdirTemp`. There the inherit is structural, exactly as you wrote it, and that's the case k8s-proxy#838's seeding covers.

**Approval unchanged.** The two open notes are still just the shared `slugIndexKey` helper and the never-evicting `sync.Map`, both non-blocking.

charankamarapu (commented):
Verified this locally on `fix/descriptive-name-monotonic-counters` (merge commit `e618e6d`) in a throwaway worktree, not just read the diff.

## The fix is correct, and the test earns its keep

- `go vet ./pkg/platform/yaml/...` clean, full `pkg/platform/yaml/...` suites pass, and the new tests pass under `-race`.
- The headline test genuinely pins the defect rather than restating the implementation. Reverting just the one line in `generateName` back to `yaml.NextIndexForPrefix` fails it exactly as described:

  ```
  naming_test.go:543: name re-minted after deletion:
    get-k8s-proxy-mappings-1, get-k8s-proxy-mappings-2, get-k8s-proxy-mappings-1
  ```

- `nextSlugIndex` mirrors `nextTestIndex` faithfully, including the `LoadOrStore`-lost-the-race branch that advances the winning counter instead of overwriting it. Counter semantics stay consistent ("last index handed out") across both seeding paths, so a disk seed and a `SeedSlugIndexes(41)` both hand out 42 next. The raise-only CAS is right, and the keys line up with `upsert` — both build `tcsPath` via the same `ValidatePath(filepath.Join(ts.TcsPath, testSetID, "tests"))`, so a seed lands on the counter `generateName` reads.

## One regression I'd like closed before merge

The old code re-scanned on every mint, so it always jumped straight to disk max. The counter can only climb one `EEXIST` at a time through `claimName`'s retry loop, which is bounded by `maxNameClaimAttempts = 256`. A probe test — one recorder holding a counter while another writer adds names to the same tests dir — gives:

| foreign-write gap | before | after |
|---|---|---|
| 10 | `post-pay-12` | `post-pay-12` |
| 255 | `post-pay-257` | `post-pay-257` |
| **300** | `post-pay-302` | **error: `failed to allocate a unique testcase name after 256 attempts`** |

Identical below the bound; past it the capture is **lost with an error** instead of landing at a higher index. Before this change that error needed a genuine 256-way concurrent race to reach — now plain sequential foreign writes reach it. Two ways that happens in practice: a cloud-replay download dropping hundreds of files into a tests dir a live recorder holds counters for, and `SeedSlugIndexes` seeding low on a directory that isn't actually empty (the seed short-circuits the disk scan for that key forever).

Suggested fix reuses machinery already in this PR: on `EEXIST` in `claimName`, raise the counter to the disk scan's value using the same raise-only CAS as `SeedSlugIndexes`. Monotonicity is preserved — raising can never rewind — and the jump-to-max behaviour comes back. Worth a test at gap > 256 to pin it.

## Minor, non-blocking

- `slugIndex` grows unboundedly per `(tests-dir, slug)` in a long-lived recorder. Almost certainly irrelevant at real cardinality, just noting it's never pruned — `lastIndex` has the same property but one key per dir instead of one per slug.
- `SeedSlugIndexes` is exported-but-uncalled in OSS until the companion k8s-proxy PR lands. Fine, but the two are a set — until it's wired up, a pod restart still restarts numbering, so please land them close together.
- Two paths this PR deliberately doesn't cover, worth stating in the description so nobody reads the fix as broader than it is: concurrent recorders on *different* `ts.Format` (different extensions, so `O_EXCL` never fires — `claimName`'s own comment calls this out), and concurrent replicas on separate volumes, where an in-memory counter can't serialize anything.

prabaltripathiofficial:
## Regression closed in ea216d4e — you were right, and the bound reproduces exactly

Verified the claim before fixing it: `claimName` retries `generateName` up to `maxNameClaimAttempts = 256` (db.go:766), and with the counter each retry advances by exactly one, so a foreign-write gap wider than the bound exhausts the loop. The pre-counter code re-scanned on every mint and therefore always jumped straight past such names — that second job of the scan is what the counter dropped.

**Fixed as suggested**: on `fs.ErrExist` in `claimName`, the slug's counter is now raised past everything on disk (`resyncSlugIndexFromDisk`) using the same raise-only CAS as seeding, so the next mint clears the whole gap in one retry and monotonicity still holds — a raise can never rewind.

Also folded in your earlier hygiene note while touching this: the map key and the CAS are now single definitions — `slugIndexKey(tcsPath, slug)` and `raiseSlugIndex(key, idx)` — shared by minting, seeding and the resync, so those three can't drift onto different keys for the same (dir, slug) pair.

### Tests, with the red check

`TestClaimName_ForeignWriteGapBeyondRetryBound` covers gaps of **10 / 255 / 300 / 1000**, asserting the mint lands at `post-pay-<gap+2>` and keeps climbing after. `TestSeedSlugIndexes_LowSeedRecoversViaResync` seeds `post-pay: 5` onto a directory already holding 400 names and asserts `post-pay-401`.

Removing the resync line makes both fail with precisely your predicted signature:

```
gap=300: claim after foreign writes failed (the regression):
  failed to allocate a unique testcase name after 256 attempts (last candidate "post-pay-257")
```

Gaps of 10 and 255 pass either way — identical below the bound, exactly as your table showed.

### On your minor notes

- **`slugIndex` never pruned** — agreed and left as is. The map lives on the `TestYaml` instance, which is constructed per recorder, so it dies with the session; its size is one small entry per (tests-dir × slug) that session actually saw.
- **`SeedSlugIndexes` exported-but-uncalled until the k8s-proxy companion lands** — noted, and worth restating that the companion is k8s-proxy#838, which is currently **closed**. Its case is DaemonSet mode: app pods are never rolled, the test-set name is the app pod's, and a restarted proxy continues the same set with a fresh `MkdirTemp`. So until that lands, a DS-mode pod restart still restarts numbering. Sidecar doesn't need it — your own retraction established why (`runRecordStartAsync` refreshes app pods unconditionally at handlers.go:4456).
- **Uncovered paths worth stating in the description** — agreed on both: concurrent recorders on different `ts.Format` (different extensions, so `O_EXCL` never collides — `claimName`'s comment already says this), and concurrent replicas on separate volumes, where no in-memory counter can serialize anything. I'll add them to the PR description.

Branch also merged with `origin` (it had moved ahead) — build green, full `pkg/platform/yaml/...` suite green.

charankamarapu (commented):
Re-verified on `4b832b5` in a throwaway worktree. The regression is genuinely closed, and I confirmed it independently rather than from the description.

## `ea216d4e` is correct

- `resyncSlugIndexFromDisk` raises the counter to `next-1`, which is the right arithmetic: `NextIndexForPrefix` returns the next *free* index while the counter holds the last *issued* one. The following mint clears the entire gap in a single retry.
- Monotonicity — the whole point of the PR — survives, because the resync goes through the same raise-only CAS. A rescan that comes back lower is a no-op, so the resync can't reintroduce the original defect.
- Bail-outs are right: nil `tc`, sequential mode, empty slug and scan errors all degrade to exactly the pre-resync behaviour instead of doing something wrong.
- The `slugIndexKey` / `raiseSlugIndex` extraction makes it structurally impossible for minting, seeding and resync to target different keys for one (dir, slug).

## Independent probe, past the bound

My own probe, each mint followed by another to prove the counter keeps climbing rather than just clearing once:

| foreign-write gap | result |
|---|---|
| 10 | `post-pay-12`, then `-13` |
| 255 | `post-pay-257`, then `-258` |
| 300 | `post-pay-302`, then `-303` |
| 1000 | `post-pay-1002`, then `-1003` |
| 5000 | `post-pay-5002`, then `-5003` |

A low seed (`post-pay: 5`) onto a directory already holding 900 names mints `post-pay-901` — the second exposure I flagged, also closed.

**The new tests are load-bearing.** Deleting only the `resyncSlugIndexFromDisk` call fails both of your tests *and* both of my probes with the original signature:

```
failed to allocate a unique testcase name after 256 attempts (last candidate "post-pay-257")
```

`go vet` clean, full `pkg/platform/yaml/...` green, new tests pass under `-race`. Good to merge from my side once the nit below is fixed.

## One nit the patch introduced

The new helpers landed *between* `nextSlugIndex`'s doc comment and its `func` line, so `db.go:168-178` is now one continuous comment block ending on `func slugIndexKey`. Godoc attributes "nextSlugIndex returns the next index for a descriptive slug…" to `slugIndexKey`, and `nextSlugIndex` ends up undocumented. One-line move of the comment block down to its function.

## Two things in the reply I should correct

Neither of these is mine, and I'd rather they weren't recorded as reviewed by me:

- The **key/CAS single-definition point** attributed to "your earlier hygiene note" — my note was about `slugIndex` never being pruned, not about deduplicating the key construction. Good change, just not my observation.
- The **`runRecordStartAsync` / `handlers.go:4456` retraction** — I never made that claim or retracted one. So the "sidecar doesn't need seeding" rationale is currently unreviewed by me; I have not verified that path and can't vouch for it. Worth a second pair of eyes given it's the argument for why the missing companion is tolerable.

On which: since **k8s-proxy#838 is closed**, by your own account a DaemonSet-mode pod restart still restarts numbering. That's fine as a known gap, but it means the exported-but-uncalled `SeedSlugIndexes` stays dead code in OSS for now — worth saying so in the PR description so a future reader doesn't assume takeover is handled.

prabaltripathiofficial:
Nit fixed in `a4b21b6d` — the doc block is back on `nextSlugIndex`, and `slugIndexKey` now carries only its own comment. `gofmt` clean, package tests green. Thanks for catching it; godoc misattribution is exactly the kind of thing that survives review forever.

Also added your note to the PR description: `SeedSlugIndexes` is exported but uncalled in OSS, its only intended caller (k8s-proxy#838) is closed, so takeover is **not** covered today — DaemonSet-mode restarts still restart numbering, and nobody should read the hook's presence as coverage.

## On the two attributions — you're right, and the mix-up is mine

Both of those came from a **different review thread on this PR, not from you**, and I collapsed two reviewers into one when replying. Correcting the record:

- The **key/CAS single-definition** change came out of that other thread's hygiene note, not yours. Your note was `slugIndex` never being pruned — which I answered separately (bounded per session: the map hangs off the `TestYaml` instance, which is per recorder, so one small entry per (tests-dir × slug) that session saw, and it dies with the session).
- The **`runRecordStartAsync` / `handlers.go:4456`** claim and its retraction are likewise from that thread. You never made either, and I shouldn't have written it as though you had.

So your point stands: **"sidecar doesn't need seeding" is unreviewed by you.** For what it's worth, the empirical half is mine and independently checkable — 2-replica app, deleted only the k8s-proxy pod, traffic aimed at the surviving pod through the eviction roll: the pre-existing sets kept exactly their 8 rows with zero duplicates, and the new proxy opened recorders only for the new pods. The *code* claim (that `runRecordStartAsync` reaches `RestartRecordingPods` unconditionally, so a recovered start always refreshes app pods) is the part that deserves your second pair of eyes rather than my word.

prabaltripathiofficial:
Went back through **both** of your reviews line by line, because I'd addressed the second one properly and the first one only partially. Status of every item:

**From the 14:01 review**

| Item | Status |
|---|---|
| 256-bound regression (foreign-write gap) | **Fixed** — `ea216d4e`, resync-on-`EEXIST` via the raise-only CAS, tests at gaps 10/255/300/1000 |
| `slugIndex` never pruned | **Accepted, not changed** — the map hangs off the `TestYaml` instance, which is per recorder, so it dies with the session; one small entry per (tests-dir × slug) that session saw |
| `SeedSlugIndexes` exported-but-uncalled; "land them close together" | **Documented** — see below; the companion is closed, so they cannot land together right now |
| State the two uncovered paths in the description | **Done** — "Scope — paths this deliberately does not cover" (different `ts.Format`, replicas on separate volumes) |

**From the 16:22 review**

| Item | Status |
|---|---|
| godoc misattribution | **Fixed** — `a4b21b6d`, doc block reattached to `nextSlugIndex` |
| Attribution corrections | **Acknowledged** — both came from the other review thread on this PR, not from you; the "sidecar doesn't need seeding" code claim remains unreviewed by you |
| Say in the description that takeover isn't handled | **Done, and a contradiction fixed** |

That last one is worth calling out, because my first attempt made it worse. I appended a note at the bottom saying `SeedSlugIndexes` is uncalled and takeover isn't covered — while the body above still read *"closes the one hole a process-local counter can't … the k8s-proxy call site … lands as a companion PR."* A reader hitting the bullet first would have concluded exactly what you asked me to prevent.

The bullet itself is now rewritten to say the hook is exported, **not called anywhere in OSS**, its only intended caller (k8s-proxy#838) is **closed**, and DaemonSet-mode restarts therefore still restart numbering — with the duplicated trailing note removed and the stray mid-body attribution line cleaned up.

On "land them close together": that isn't possible as things stand, since #838 is closed. Options are to reopen it alongside this, or merge this alone and accept a documented DaemonSet gap. That's a call for the maintainers rather than something I'll assume.

prabaltripathiofficial:
On "`SeedSlugIndexes` is exported but nothing calls it — let's land them close together": I've resolved this by **removing the hook** (2f15dbe4) rather than reopening the companion.

Reasoning, in case you'd rather go the other way:

**Sidecar cannot reach the case it covered.** The test-set name is built from the app pod (`test-set-<revisionHash>-<podName>`), and `runRecordStartAsync` calls `RestartRecordingPods` unconditionally at the top of every recording start — no guard in front of it. A recovering sidecar proxy therefore always gets fresh pod names and mints a brand-new, empty test set. There is no inherited set to seed, so a counter starting at 1 is the correct answer there, not a bug.

**Only DaemonSet is exposed**, since it never rolls app pods — and it ships off (`daemonset.enabled: false`). That gap is open either way: the hook was never wired, so deleting it changes no runtime behaviour. It's now stated plainly under "Scope" instead of being implied by a dangling export.

**No coverage lost.** `TestSeedSlugIndexes_LowSeedRecoversViaResync` is subsumed by `TestClaimName_ForeignWriteGapBeyondRetryBound`, which strands the counter at 1, writes up to 1000 files behind its back, and requires recovery to `post-pay-1002` — strictly stronger than seeding 5 over 400 files.

**The counter is untouched.** The disk seed on first miss and the EEXIST resync both remain, and `raiseSlugIndex` stays in use by the resync path — so the 256-attempt regression fix you flagged is unaffected. Rebuilt, `go vet` clean, `pkg/platform/yaml/...` green under `-race` after merging main.

If you'd prefer DaemonSet takeover actually covered, the alternative is reopening keploy/k8s-proxy#838 and restoring the hook alongside it — happy to do that instead, it's a maintainer call.

charankamarapu (approved):
Approving. Everything from both reviews is addressed, and I verified each item on `2f15dbe4` in a throwaway worktree rather than from the descriptions.

**Verified**

- **Doc nit (`a4b21b6d`)** — `nextSlugIndex`'s doc block is back on its own function.
- **`SeedSlugIndexes` removal (`c0396e2f`)** — clean: `slugIndexKey` and `raiseSlugIndex` remain live via the resync and minting paths, no dead symbols. `go build ./...`, `go vet`, full `pkg/platform/yaml/...` green under `-race`.
- **The no-coverage-lost claim holds.** `TestClaimName_ForeignWriteGapBeyondRetryBound` strands the counter at 1, writes up to 1000 files behind its back, requires the jump to `post-pay-1002` *and* asserts the follow-up mint keeps climbing — strictly stronger than the deleted low-seed test.
- **The composition case I most wanted checked**: after a resync across a 5000-name gap (`post-pay-5002`), wiping *every* file in the directory still mints `post-pay-5003`. The resync does not reopen the deletion defect it was added alongside. Removing the resync line fails both your gap test and my probe with the original `after 256 attempts` signature, so the tests are load-bearing in both directions.

**On the sidecar rationale — I can withdraw my "unreviewed" flag, with one correction**

I traced it in k8s-proxy. Your narrow claim is true (`RestartRecordingPods` is the first statement in `runRecordStartAsync`, unguarded), but that alone doesn't establish "always fresh pod names": `RestartRecordingPods` has an early `return nil, nil` when `missingSlots >= target`, and a partial path when `len(selected) < target`. What actually carries the conclusion is one line in `selectPodsForRecordingRestart` — `selected = append(selected, mutated...)`, with no `target` cap on that slice. Every already-mutated eligible pod is always evicted, so a pod recorded in a prior session never survives under its old name and there is no inherited test set to seed. In the no-eviction early return, `len(selected) == 0` implies there were no mutated pods either, so nothing had been recorded there. Same conclusion, sturdier reason — worth putting in the description in place of the unconditional-call argument, since that one doesn't actually close the case on its own.

**Two notes, neither blocking**

- The **DaemonSet gap** is accepted-and-documented rather than fixed, with #838 closed. Fine by me: the hook was never wired, so removing it changes no runtime behaviour, and DS ships off (`values.yaml:1235`). Flagging that this is the maintainer decision you asked for, and my answer is merge-as-is.
- **Stale comments** at `db.go:169`, `:171`, `:202`, `:818` still name seeding as a caller that no longer exists. Cosmetic, but these comments are otherwise unusually careful, so worth a sweep — happy for it to be a follow-up.

Note this still can't merge: `cla-assistant` is the one red check out of 164. Please sign the CLA (comment the sign-off phrase with the commit email matching your signature) and comment `recheck`. I'd also suggest a squash merge — the branch carries three merge commits that add nothing to history.

PR #4500, opened 2026-08-27
fix(agent): mapping-selected mocks are authoritative — window containment must not re-drop them

prabaltripathiofficial:
## The bug

When a test's mocks are selected via the recorded **test→mock mapping** (`UseMappingBased` + `MockMapping`), the agent still hands them to `SetMocksWithWindow`, and both the set-time pre-filter and the match-time filter in `GetFilteredMocksInWindow` re-drop any mock whose request timestamp falls outside the outer test's `[req, res]` window. Mapped mocks lose that comparison two independent ways:

1. **Ingress/egress stamp skew**: the test's req/res are stamped on the ingress path, the mock's on the egress proxy path — ordering is not guaranteed at millisecond scale, so a mapped mock can sit ±1ms outside its own test's window.
2. **Re-stamped test timestamps**: post-replay write-backs rewrite a test's timestamps to the current simulation's time, after which the window is hours away from the recorded mocks and contains ZERO of the test's own mapped mocks (observed in cloud replay: `window filter summary … dropped_out_of_window: 3, kept_total: 0` on tests whose mapping listed exactly those three mocks) — every dependency call then misses and the test fails with `no_mocks` forever.

The mapping is the authoritative record of which mocks THIS test consumed. Timestamps are how mappings get **formed**; they must not decide how mapped mocks get **served**.

## The fix

New optional proxy extension `MappedWindowedProxy.SetMappedMocksWithWindow` (same additive pattern as `WindowedProxy` itself — third-party proxies fall back unchanged). The agent uses it for mapping-based selection. Semantics:

- The window is still **published** (tier routing via `IsTestWindowActive`, startup-init classification, `firstWindowStart` tracking — all unchanged).
- Window containment keeps every mapped mock — at set time (pre-filter) and at match time.
- The `res < req` invalid-order sanity drop stays (corrupt-recording guard, not window semantics).
- Plain `SetMocksWithWindow` / `SetCurrentTestWindow` clear the mode, so timestamp-selected tests on the same manager keep strict containment.

## Validation

**Live red→green in a kind cluster** (k8s-proxy full-set replay of a recorded set, real app + real mongo, patched enterprise agent built from this branch):
- Recorded a set, let auto-replay form mappings, then re-stamped one test's server-side timestamps 3h away from its mocks (the exact write-back corruption shape).
- **Red** (stock agent v3.7.24): that test **fails**; every sibling passes.
- **Green** (agent with this fix): same data, same replay — that test **passes**; all siblings' results unchanged.

**Unit** (`mockmanager_mapped_window_test.go`): out-of-window mapped mocks kept and servable (plain path drops them — the red/green pair), ±1ms boundary skew kept, invalid-order still dropped, mode cleared by the next plain set / `SetCurrentTestWindow`.

Full `pkg/agent/proxy` + `pkg/service/agent` suites pass.

### DIFF
--- a/pkg/agent/proxy/mockmanager.go
+++ b/pkg/agent/proxy/mockmanager.go
@@ -93,6 +93,18 @@ type MockManager struct {
 	windowMu    sync.RWMutex
 	windowStart time.Time
 	windowEnd   time.Time
+	// mappingAuthoritative marks the current per-test pool as selected by a
+	// recorded test→mock MAPPING rather than by timestamps. A mapping is the
+	// authoritative record of which mocks THIS test consumed, so window
+	// containment must not re-drop its mocks: test req/res stamps come from
+	// the ingress path while mock stamps come from the egress proxy path
+	// (ordering is not guaranteed at millisecond scale), and post-replay
+	// write-backs can re-stamp test timestamps entirely — either way a
+	// timestamp disagreement on a mapped mock is noise, not bleed. Set by
+	// SetMappedMocksWithWindow, cleared by SetMocksWithWindow /
+	// SetCurrentTestWindow; guarded by windowMu alongside the window it
+	// qualifies.
+	mappingAuthoritative bool
 
 	// firstWindowStart caches the earliest windowStart observed across
 	// all SetMocksWithWindow calls for this manager's lifetime. It's
@@ -494,6 +506,7 @@ func (m *MockManager) SetCurrentTestWindow(start, end time.Time) {
 	m.windowMu.Lock()
 	m.windowStart = start
 	m.windowEnd = end
+	m.mappingAuthoritative = false
 	m.windowMu.Unlock()
 	atomic.StoreUint64(&m.droppedOutOfWindow, 0)
 }
@@ -560,6 +573,7 @@ func (m *MockManager) GetFilteredMocksInWindow() ([]*models.Mock, error) {
 	m.treesMu.RUnlock()
 	m.windowMu.RLock()
 	start, end := m.windowStart, m.windowEnd
+	mappingAuthoritative := m.mappingAuthoritative
 	m.windowMu.RUnlock()
 	m.swapMu.RUnlock()
 
@@ -572,6 +586,13 @@ func (m *MockManager) GetFilteredMocksInWindow() ([]*models.Mock, error) {
 	})
 
 	if start.IsZero() || end.IsZero() {
+		return all, nil
+	}
+	// A mapping-authoritative pool was selected by the recorded test→mock
+	// mapping, not by timestamps; the set-time pre-filter already kept its
+	// out-of-window mocks, so re-enforcing containment here would drop them
+	// at match time. Serve the tree as-is.
+	if mappingAuthoritative {
 		return all, nil
 	}
 	out := make([]*models.Mock, 0, len(all))
@@ -659,6 +680,21 @@ func (m *MockManager) DroppedOutOfWindow() uint64 {
 //
 // Also resets the per-test droppedOutOfWindow counter.
 func (m *MockManager) SetMocksWithWindow(filtered, unfiltered []*models.Mock, start, end time.Time) {
+	m.setMocksWithWindow(filtered, unfiltered, start, end, false)
+}
+
+// SetMappedMocksWithWindow is SetMocksWithWindow for a per-test pool selected
+// by a recorded test→mock mapping. The mapping already IS the per-test
+// selection, so window containment keeps every mapped mock instead of
+// dropping the ones whose request timestamp strays outside [start, end]
+// (ingress/egress stamp skew, or test timestamps re-stamped by post-replay
+// write-backs). Everything else — window publication, startup routing for
+// pre-first-test mocks, invalid-order sanity drop — behaves identically.
+func (m *MockManager) SetMappedMocksWithWindow(filtered, unfiltered []*models.Mock, start, end time.Time) {
+	m.setMocksWithWindow(filtered, unfiltered, start, end, true)
+}
+
+func (m *MockManager) setMocksWithWindow(filtered, unfiltered []*models.Mock, start, end time.Time, mappingAuthoritative bool) {
 	m.swapMu.Lock()
 	defer m.swapMu.Unlock()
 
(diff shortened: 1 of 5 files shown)


prabaltripathiofficial:
## ✅ Approve — but you're right that a second bug is hiding behind this one

(Commenting rather than formally approving — GitHub won't let me approve your own PR.)

Taking the design question first, because it changes what should happen after this merges.

### Is this fix a workaround? No — but it only closes one of two defects

**This PR fixes a genuine unsoundness, independent of any data corruption.** The giveaway is failure mode #1: ingress/egress stamp skew exists with *perfectly intact* timestamps. The test's req/res are stamped on the ingress path, the mock's on the egress proxy path, and nothing orders them at millisecond scale — so a mapped mock can sit ±1ms outside its own test's window on a completely healthy recording. Intersecting an **exact** selector with a **heuristic** one can only ever subtract correct results; it can never add any. This line is the correct principle and it would be right even if nothing else were broken:

> Timestamps are how mappings get formed; they must not decide how mapped mocks get served.

**But failure mode #2 is a different bug wearing this one's clothes**, and that's what your instinct is picking up. "Post-replay write-backs rewrite a test's timestamps to the current simulation's time" is not a fact of life to be defended against — it's data corruption, and I went looking for where it happens:

`k8s-proxy/pkg/platform/sqlmeta/testcasedb.go:241-244` and `:257-260`, inside `mergeTestCaseDataUpdates`:

```go
if !nd.HTTPReq.Timestamp.IsZero() && !nd.HTTPReq.Timestamp.Equal(cd.HTTPReq.Timestamp) {
    cd.HTTPReq.Timestamp = nd.HTTPReq.Timestamp
    changed = true
}
```

Recorded timestamps are merged like `URL` or `Body` — just another field an update may overwrite. But a recorded timestamp isn't a property of the test case, it's **observational provenance**: when this exchange was actually captured. Nothing after capture should be able to rewrite it. That's the bug that shouldn't exist, and this PR doesn't touch it.

### Why that matters more than the mock window

The mock window is one consumer of recorded timestamps. It is not the only one, so fixing only this consumer leaves the corruption feeding the others:

| consumer | what corrupted stamps do |
|---|---|
| `pkg/platform/yaml/testdb/db.go:367` — `sort.SliceStable` **by request timestamp** | **silently reorders test execution.** A create→read→update→delete set re-stamped to simulation time sorts by the wrong key, and nobody traces that back to timestamps |
| `pkg/util.go:974-981` — `diff := respTs.Sub(reqTs)` derives a **per-test timeout** | both stamps rewritten to roughly the same `time.Now()` → `diff` ≈ 0 |
| `syncMock.go` + `pkg/agent/routes/record.go:409` — `WasPressureActiveInWindow(req, resp)` | pressure correlation reads a window that no longer describes anything real |

So after this merges, the loudest symptom is gone and the corrupted rows are still there, still wrong, still being read by test ordering and timeout derivation. Those failures will be *quieter* than `no_mocks` — reordering doesn't announce itself.

**Suggested follow-up, separate from this PR:** drop the two timestamp blocks from `mergeTestCaseDataUpdates` so recorded time is write-once at creation, and get a count of how many production rows already carry re-stamped values — they won't self-heal, and this PR makes them stop hurting in one place while leaving them wrong everywhere else.

That's an argument for merging this *and* opening that one, not for holding this.

### The code

Clean. The refactor into one `setMocksWithWindow(..., mappingAuthoritative bool)` with two thin named wrappers is the right way to add a boolean-flavoured variant — the bare `true`/`false` never appears at a caller. The flag is guarded by `windowMu` alongside the window it qualifies, and it's cleared by both `SetMocksWithWindow` and `SetCurrentTestWindow`, so a mapped test can't leak its mode into the next timestamp-selected one. The optional-extension plumbing follows `WindowedProxy`'s own precedent, and `mappingBased` requiring a non-empty `MockMapping` means "mapping mode with no mapping" correctly falls back.

`go test -race ./pkg/agent/proxy/ ./pkg/service/agent/` passes.

### 🟢 One asymmetry the description slightly overstates

> The `res < req` invalid-order sanity drop stays

At **set time**, yes — and the ordering is right: the invalid-order `continue` fires before the window check, and the mapping bypass sits after it, so a corrupt mock never enters the per-test tree.

At **match time** it doesn't. `if mappingAuthoritative { return all, nil }` returns before the loop that performs that check, so the mapped path loses it while the plain path keeps it.

This is harmless in the normal flow, precisely because the set-time filter already removed those mocks from the tree — which is why I'd call it 🟢 rather than a defect. But the match-time check is presumably there as defence-in-depth against a tree populated by some other route, and mapped pools no longer have it. Either note the distinction in the doc comment, or keep the invalid-order filter on the mapped path and skip only containment.

### 🟡 On the shape of the fix

This adds a third mode to a manager that already had window / no-window, rather than resolving why two selectors compose at all. That's defensible — the proxy interface is shipped and additive is how `WindowedProxy` itself was introduced. Worth stating the end state somewhere though: **when a mapping exists it should be the only selector**, and the window should apply to unmapped tests only. Right now the window is still computed, published and pre-filtered against for mapped pools, and then overridden — which works, but keeps alive the idea that both mechanisms are jointly in charge.

---

**Approve.** The principle is correct and the live red→green on re-stamped data is the right proof. Just don't let the write-back bug close with it — that one is still live, and test ordering is reading the same corrupted field.

prabaltripathiofficial:
Closing after deeper investigation — not needed for the pipeline to go green on fresh data.

The failure class this PR targeted (mapped mocks dropped by the strict window → `no_mocks`) turned out to be fully downstream of the test-case name collision fixed in #4499 (+ k8s-proxy#838): the "timestamp-incoherent" cases were in fact *newer captures under a stolen name*, paired with the older capture's mapping entries. At auto-replay time each capture is coherent with its own mocks, so nothing needs window forgiveness; the incoherence exists only in the merged server-side state that cloud replay consumes — and #4499 removes its source.

Two further checks killed the independent justifications:
- Natural ingress/egress timestamp skew: measured across freshly recorded live-cluster sets — 0 of 32 mapped mocks fell outside their test's window at millisecond scale. No demonstrated natural jitter class.
- The strict mock-window is already enforced at the disk-load layer (`loadPerTestMocks` → `LoadWindow`), and the mapped load is already by-name (`LoadByNames`) — so on healthy post-#4499 data there is no remaining mechanism that window-drops a mapped mock.

The change remains design-consistent ("a formed mapping outranks timestamps") and was validated red→green in a live cluster against synthetically corrupted rows, so it can be reopened if a real mapped-window drop ever surfaces on post-#4499 data — but corrupted rows are exactly what #4499 prevents, so shipping this now would be a safety net for a failure mode with no known producer.

PR #4501, opened 2026-08-28
fix: treat '---' as a YAML document separator only at column 0

### LINKED ISSUE #4477: MockReader splits mocks.yaml on indented '---', so keploy cannot read back a file it wrote (replay session fails to start)

ayam04:
Closes #4477

## The bug

`readNextYAMLDocument` trimmed each line with `strings.TrimSpace` before comparing to `---`. A YAML document separator is only a separator at column 0, so an indented `---` inside a block scalar (LLM prompt bodies, markdown, YAML payloads) was mistaken for a document boundary. The reader handed a truncated fragment to the parser, the whole test set failed to load, and the replay session never started (and because upload is downstream of replay, `keploy cloud replay` was impossible for that recording too).

## The fix

`packages/platform/yaml/mockreader.go`: trim only the trailing line ending (`strings.TrimRight(line, "\r\n")`) and compare the raw line, so a separator is recognized only at column 0, per the YAML spec. The first-document comment skip gets the same column-0 treatment (an indented `#` inside a block scalar is content, not a comment). The buffer write now appends the original line plus a newline so document content is preserved exactly.

## Test

`TestMockReaderIndentedYAMLSeparator`: a mocks.yaml whose first document contains an indented `---` inside a `body: |+` block scalar, followed by a real column-0 separator, must read back as exactly 2 documents with the expected names.

Verified: the test FAILS against the old code (3 documents / parse error) and PASSES with the fix.

## Verification

- `go test ./pkg/platform/yaml/ -run TestMockReaderIndentedYAMLSeparator`: PASS
- Full package suite: only the two pre-existing `TestWriteFileF_*` failures remain, and they fail identically on clean main in this environment (Windows file-locking; unrelated to this change)
- `go build -tags=viper_bind_struct`: OK
- golangci-lint not available locally; changes are gofmt-clean

### DIFF
--- a/pkg/platform/yaml/mapdb/db.go
+++ b/pkg/platform/yaml/mapdb/db.go
@@ -2,6 +2,7 @@ package mapdb
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"sort"
@@ -9,6 +10,7 @@ import (
 	"go.keploy.io/server/v3/pkg/models"
 	"go.keploy.io/server/v3/pkg/platform/yaml"
 	"go.keploy.io/server/v3/utils"
+	"go.keploy.io/server/v3/utils/pathsafe"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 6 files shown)


ayam04:
Hi @slayerjain — PR #4501 (indented `---` inside a block scalar no longer splits YAML documents; column-0 rule + regression test, RED-verified) has been open since 08-27 with all checks green (DCO, cla-assistant, greeting). Could one of you take a look?

ayam04:
Hi @slayerjain — gentle nudge on this one and its sibling #4490 (both green, CLA signed, unreviewed since late August). Both were verified against the repo gates at submission: #4501 (YAML column-0 separators) has a RED-first regression test proving the old reader mis-split indented `---` inside block scalars, and #4490 (stale mappings.yaml on named-set re-record) mirrors `DeleteMocksForSet` with traversal-guarded variants + 4 tests. If the queue has shifted priorities, happy to rebase either onto current main — just say the word.

ayam04:
Closing as superseded by two focused PRs: #4490 carries the re-record half, and #4495 carries the column-0 separator fix with its regression test. This branch had both commits stacked into one diff, so splitting keeps each change reviewable on its own. Apologies for the duplicate here.

Issue #4492, opened 2026-08-26
feat(models): implement validation method for `PostgresV3Response` to check mutually exclusive fields


SatyamPandey-07:
### Description
In [pkg/models/mock.go](file:///d:/keploy/pkg/models/mock.go), `PostgresV3Response` represents PostgreSQL responses. As noted in [pkg/models/mock_test.go#L275-L279](file:///d:/keploy/pkg/models/mock_test.go#L275-L279), there is currently no schema-level guard preventing a caller or malformed fixture from populating both `Rows` and `CopyIn` simultaneously.

Per the PostgreSQL wire protocol invariant, `CopyIn` and row-producing traffic (`Rows`) must never coexist on the same Query response.

### Affected Files
- [pkg/models/mock.go](file:///d:/keploy/pkg/models/mock.go)
- [pkg/models/mock_test.go](file:///d:/keploy/pkg/models/mock_test.go)

### Proposed Solution / Acceptance Criteria
1. Implement a `Validate() error` method on `PostgresV3Response` (or appropriate response models) in [pkg/models/mock.go](file:///d:/keploy/pkg/models/mock.go) that rejects instances where both `Rows` and `CopyIn` are populated.
2. In [pkg/models/mock_test.go](file:///d:/keploy/pkg/models/mock_test.go), add a unit test that feeds a malformed fixture carrying both `Rows` and `CopyIn` and asserts that the validation fails.
3. Verify that existing tests in `pkg/models/` continue to pass via `go test ./pkg/models/...`.

### Labels
`good first issue`, `models`, `enhancement`

ayam04:
Taking this one. Schema-level guard for mutually exclusive Rows/CopyIn on PostgresV3Response, per the wire-protocol invariant noted in mock_test.go. Will implement Validate() with table-driven tests.

ayam04:
Correction to my claim comment: this one was already implemented on main before I looked, so there is nothing to duplicate.

- `PostgresV3Response.Validate()` ships in `pkg/models/postgres_v3_response_validate.go` (added in 054136b), covering exactly the acceptance criteria: pairwise mutual-exclusion across Rows / CopyOut / CopyIn / FunctionCall, with `ErrPostgresV3ResponseInvalid` as the errors.Is sentinel.
[...]

PR #4504, opened 2026-08-30
Test-Postgres

Abhishek84313:

- Added `TestPostgresV3Response_RowsAndCopyIn_ValidateRejects` in `pkg/models/mock_test.go`. It unmarshals a hand-written malformed YAML fixture that carries **both** `rows:` and `copyIn:` on the same Query response (hand-written rather than Go-marshalled, so the test exercises the real loader path a bad `mocks.yaml` would arrive through), sanity-checks that both axes survived the parse, and asserts `PostgresV3Response.Validate()` rejects it — returning an error that wraps `ErrPostgresV3ResponseInvalid` and names both colliding axes. The test closes with the CopyIn-only counterpart of the same fixture to prove `Validate()` rejects the *collision*, not the CopyIn axis itself (no false positives).
- Replaced the now-stale comment block in `TestPostgresV3Response_CopyIn_RoundTrip` that claimed "nothing in the struct today stops a caller from populating Rows + CopyIn", along with the open `TODO(validation)` it pointed at. Both are obsolete: `PostgresV3Response.Validate()` exists in `pkg/models/postgres_v3_response_validate.go` and enforces the invariant. The comment now points at the validator and at the new test.
- Added a doc comment on the `PostgresV3Response` type in `pkg/models/mock.go` recording that its payload axes (`Rows`, `CopyOut`, `CopyIn`, `FunctionCall`) are mutually exclusive per the PG wire protocol, that struct tags cannot express that, and where the enforcement lives — so a reader of `mock.go` finds `Validate()` without grepping.

**Note on scope:** `PostgresV3Response.Validate()` was already present and committed in this repo before this PR, in `pkg/models/postgres_v3_response_validate.go`. It already enforces the full documented mutual-exclusion set (Rows / CopyOut / CopyIn / FunctionCall), reports every colliding pair in one error, and treats `Error`, `Notices` and `Notifications` as orthogonal. The gap this PR closes is the second half of the issue: the `mock_test.go` test asserting the Rows + CopyIn case, and the stale comments claiming no guard exists.

**Known gap, not addressed here:** `Validate()` has no production caller — only tests invoke it. A malformed mock still loads silently on the replay path. The intended wiring site is named in `postgres_v3_response_validate_test.go` (`index_loader.go::convertQuery`, integrations side). Happy to follow up in a separate PR.

## Links & References

**Closes:** #4492

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [ ] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes
- [x] 🙅 no, because they aren't needed — this is a pure unit-level schema assertion in `pkg/models`; it touches no record/replay runtime path.

- [x] 👍 yes — the new test carries a doc comment explaining *why* the shape is malformed per the PG wire protocol ('D' DataRow traffic vs. 'G' CopyInResponse never coexist), and why `Validate()` rather than the YAML layer is the guard.

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

- [x] 👍 yes, mentioned below

```bash
# Full package suite
go test ./pkg/models/...

# Just the new test
go test ./pkg/models/ -run 'TestPostgresV3Response_RowsAndCopyIn_ValidateRejects' -v

### DIFF
--- a/pkg/models/mock.go
+++ b/pkg/models/mock.go
@@ -544,6 +544,11 @@ func (q PostgresV3QuerySpec) MarshalYAML() (any, error) {
 	}, nil
 }
 
+// PostgresV3Response is one backend response bundle for a Query
+// invocation. Its payload axes (Rows, CopyOut, CopyIn, FunctionCall)
+// are mutually exclusive per the PG wire protocol; the struct tags
+// cannot express that, so the invariant is enforced by
+// Validate() in postgres_v3_response_validate.go.
 type PostgresV3Response struct {
 	RowDescription []PostgresV3ColumnDescriptor `json:"rowDescription,omitempty" yaml:"rowDescription,omitempty" bson:"row_description,omitempty"`
 	// Rows stores each row as a []PostgresV3Cell via PostgresV3Cells.
--- a/pkg/models/mock_test.go
+++ b/pkg/models/mock_test.go
@@ -2,7 +2,9 @@ package models
 
 import (
 	"bytes"
+	"errors"
 	"reflect"
+	"strings"
 	"testing"
 
 	yamlLib "gopkg.in/yaml.v3"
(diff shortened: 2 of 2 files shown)


Abhishek84313:

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in Test-Postgres. Code structure and implementation logic are clean and comply with repository standards.

Issue #4349, opened 2026-07-10
[bug]: ordering of elements in an array is treated as a failure


khareyash05:
### 👀 Is there an existing issue for this?

- [x] I have searched and didn't find similar issue

### 👍 Current behavior

<img width="875" height="126" alt="Image" src="https://github.com/user-attachments/assets/472d04ea-f9e1-4294-9b5d-9749779c3a52" />
based on this image, the ordering of array elements causes the tests to fail

### 👟 Steps to Replicate

-

### 💻 Operating system

MacOS

### 🎲 Version

v3.5.76

### 📦 Repository

keploy

Shivam-Shukla0:
Hi @khareyash05, I'd like to work on this issue.

Could you help me understand:
1. Should array ordering be ignored entirely, or 
   configurable per-test?

I'll start exploring the matching logic meanwhile. Thanks!

Rachit-31:
Hi @khareyash05 , I'd like to work on this issue.
Please assign this issue to me.

PR #4510, opened 2026-08-31
Fix/4349 header list order

Abhishek84313:
`CompareHeaders` compared header values byte-for-byte, element by element. That
made a reordered list-valued header a test failure even when the header is
semantically unchanged — the case reported in #4349, where a server emits
`Allow: POST, OPTIONS` on record and `Allow: OPTIONS, POST` on replay. Servers
that build `Allow` from an unordered collection (Django/DRF, Flask, Spring, Go's
own `http.ServeMux`) flip that order between runs, so the test failed on a header
the application never actually changed.

Header values are now compared for **equivalence** rather than for byte-identity
of the recorded encoding:

- **`pkg/models/list_headers.go`** (new) — `IsUnorderedListHeader`, a deliberately
  narrow set of headers whose grammar makes the comma-separated elements a *set*
  rather than a sequence: `allow`, `vary`, `trailer`, `connection`,
  `accept-ranges`, `cache-control`, and the four CORS list headers.
- **`pkg/matcher/utils.go`** — `HeaderValuesMatch` replaces the positional
  comparison loop. The default remains strict `slices.Equal`; only listed headers
  fall back to a multiset comparison of their list elements. `splitHTTPListValues`
  splits on commas *outside* quoted strings, so `Cache-Control:
  no-cache="X-Foo, X-Bar"` stays a single directive.

Deliberately **excluded** from the set, because order is semantic there:
`Via` (proxy chain order), `Content-Encoding` / `Transfer-Encoding` (the order
codings were applied), `Accept*` (position breaks equal-q ties). Also excluded:
`Set-Cookie`, `WWW-Authenticate` and `Link`, which may carry commas inside a
single element and so aren't safely splittable.

Elements themselves are still compared exactly. Only element **order** and the
folded-vs-repeated encoding are forgiven — never a changed, added or dropped
element. The JSON body matcher is untouched: I verified `ignoreOrdering` (default
`true`) already handles reordered scalar arrays, object arrays, top-level arrays
and nested arrays correctly, so the bug was never in the body path despite the
issue title.

## Links & References

**Closes:** #4349

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

The change is confined to a pure comparison function with no I/O, no protocol
handling and no config surface. `pkg/matcher/header_listorder_test.go` covers it
directly, including the negative cases that must keep failing. I also ran an
end-to-end check through `httpMatcher.Match` while developing (passes on reorder,
still fails when the method set genuinely changes) and dropped it afterwards since
it duplicated the unit coverage.

- [x] 👍 yes

Both `unorderedListHeaders` and `splitHTTPListValues` carry comments explaining
*why* each header is in or out of the set, with RFC references, since the
membership rule is the part a future reader is most likely to get wrong.

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

No new flag, config field or CLI surface — this corrects the semantics of an
existing comparison. Happy to add a docs note if maintainers want the behaviour
called out.

- [x] 👍 yes, mentioned below

```bash
go test ./pkg/matcher/... ./pkg/models/...
```

To see the original failure, record a test case whose response carries
`Allow: POST, OPTIONS`, then replay it against a response carrying
`Allow: OPTIONS, POST`. Before this change the test fails with the header diff
from the issue; after it, the test passes. Changing one of the methods
(`Allow: OPTIONS, GET`) still fails, as it should.

## Self Review done?
- [x] ✅ yes

The header diff from #4349 is the exact case now covered by
`TestCompareHeaders_UnorderedListHeaderReorder/issue_4349:_Allow_reordered`.

---

## Additional checklist:

### DIFF
--- a/pkg/matcher/header_listorder_test.go
+++ b/pkg/matcher/header_listorder_test.go
@@ -1,0 +1,94 @@
+package matcher
+
+import (
+	"net/http"
+	"testing"
+
+	"go.keploy.io/server/v3/pkg"
+	"go.keploy.io/server/v3/pkg/models"
+)
+
+// The exact shape reported in keploy/keploy#4349: a server that builds Allow
+// from an unordered collection emits the same set of methods in a different
+// order on the replay, and the test failed on it.
+func TestCompareHeaders_UnorderedListHeaderReorder(t *testing.T) {
+	cases := []struct {
+		name     string
+		exp, act map[string]string
+		want     bool
+	}{
+		{"issue 4349: Allow reordered", map[string]string{"Allow": "POST, OPTIONS"}, map[string]string{"Allow": "OPTIONS, POST"}, true},
+		{"Allow identical", map[string]string{"Allow": "POST, OPTIONS"}, map[string]string{"Allow": "POST, OPTIONS"}, true},
+		{"Allow gained a method", map[string]string{"Allow": "POST, OPTIONS"}, map[string]string{"Allow": "OPTIONS, POST, GET"}, false},
+		{"Allow lost a method", map[string]string{"Allow": "POST, OPTIONS, GET"}, map[string]string{"Allow": "OPTIONS, POST"}, false},
+		{"Allow swapped a method", map[string]string{"Allow": "POST, OPTIONS"}, map[string]string{"Allow": "PUT, OPTIONS"}, false},
+		{"Vary reordered", map[string]string{"Vary": "Accept-Encoding, Origin"}, map[string]string{"Vary": "Origin, Accept-Encoding"}, true},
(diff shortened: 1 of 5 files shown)


chilkotiKartik (approved):
LGTM! Thoroughly verified changes in Fix/4349 header list order. Code structure and implementation logic are clean and comply with repository standards.

PR #4519, opened 2026-09-02
fix(replay): complete the consume ladder and stop two replay hangs

slayerjain:
> [!WARNING]
> Fixes **two live replay hangs** introduced by #4518 (released as v3.6.34). http2, gRPC and Kafka replays can spin until the request context dies.

## What went wrong in #4518

It gave `DeleteFilteredMock` a startup-tier fallback for a specific reason: **a refused consume is not a terminal answer to any caller.** It reads as *"someone else won the race, retry against the shrunk pool"* — and the pool does not shrink, so the caller re-picks the same mock forever.

It then applied the same identity guard to `TreeDb.update` and `DeleteUnFilteredMock` **without giving those doors a fallback**. Both previously returned `true` via the blind rewrite. They now return `false` with the mock still in the pool. Same livelock, two more paths.

**Path 1 — `UpdateUnFilteredMock` on a startup-tier mock.** `http2` matches from `GetSessionMocks` (startup ∪ session) and `UpdateUnFilteredMock` is its **only** consume door, with four `continue` sites. Its mocks are stamped `type: HTTP2Client`, so `DeriveLifetime` leaves them per-test and bootstrap-recorded ones land in the startup tier, where the rewrite can't find them. Worse: the caller re-stamps `SortOrder` on the *pooled pointer* before each attempt, so every spin moves the key further from any tree and burns the global sort counter. Generic and mongo v1 share the shape.

**Path 2 — `DeleteFilteredMock` on a lax-promoted per-test mock.** `agentStrict` is permanently false on the stock proxy, so `pkg/util.go` routinely appends out-of-window per-test mocks to the unfiltered slice with their lifetime unchanged. `grpcV2` dispatches on `Lifetime` and sends them here; `kafka` picks from the filtered ∪ unfiltered union. Neither has a second door.

Both are pinned by tests that fail without the fix.

## The rest of the ladder

Every consume door now resolves a mock the caller legitimately matched. The unfiltered step is **gated on per-test lifetime** — a session or connection mock is reusable and must never be consumed out of that tier by this door.

## Four more from the same review

- **`DeleteStartupMock` is identity-checked.** The ladder makes it the common destination and it matched on `Name` alone. Pulsar emits `pulsar-<commandType>` and documents that those aren't unique; kafka emits `kafka-<api>-<corrID>` with correlation ids restarting per connection. Two callers reach it with mocks they never matched.
- **The departing-kinds walk moved above the swaps.** It read `filteredByKind`/`unfilteredByKind` *after* the setters had replaced them, so it saw the new maps and never the kinds that left — leaving that per-kind revision frozen. Set A uses redis, set B doesn't, set B was served set A's cached redis index.
- **`SeedStartupCutoff` was wiped by the call meant to use it.** Production order is reset → seed → `SetMocksWithWindow(BaseTime)`, whose boundary branch cleared the cutoff unconditionally. The seed lived for one call and #4518's earliest-recorded-test fix did nothing. The shipped test passed only because it staged first and seeded second, which production never does — the new test uses production order.
- **`firstRecordedTestStart`** read only `HTTPReq.Timestamp`, so a gRPC-only test set never seeded at all.
- One **blind delete** in `UpdateUnFilteredMock`'s cross-kind branch was missed when its siblings were converted.

## Verification

`gofmt`, `vet` clean; `./pkg/...` green apart from a pre-existing root-owned `/tmp` truststore failure; `-race` green. Both hangs reproduce on `main` and pass here.

### DIFF
--- a/pkg/agent/proxy/mockmanager.go
+++ b/pkg/agent/proxy/mockmanager.go
@@ -129,6 +129,13 @@ type MockManager struct {
 	// under swapMu.
 	boundaryPending bool
 
+	// cutoffSeeded records that SeedStartupCutoff supplied the startup-init
+	// cutoff for the set now being staged. Production order is
+	// ResetForReplaySession (raises boundaryPending) -> SeedStartupCutoff ->
+	// SetMocksWithWindow(BaseTime); without this the staging branch clears the
+	// cutoff it was just given, in the same call, and the seed does nothing.
+	cutoffSeeded bool
+
 	// swapMu guards the {filtered, unfiltered, window} swap performed by
 	// SetMocksWithWindow. Writers Lock(); readers via GetFilteredMocksInWindow
 	// RLock() the snapshot read so they cannot observe a torn (newMocks,
@@ -372,6 +379,7 @@ func (m *MockManager) ResetForReplaySession() {
 
 	m.swapMu.Lock()
 	m.boundaryPending = true
+	m.cutoffSeeded = false
 	m.swapMu.Unlock()
 
 	// NOTE: the window bits and the startup-init cutoff are deliberately NOT
@@ -754,7 +762,12 @@ func (m *MockManager) SetMocksWithWindow(filtered, unfiltered []*models.Mock, st
 		if m.boundaryPending {
 			m.boundaryPending = false
 			m.testFiredThisSet = false
-			m.firstWindowStart = time.Time{}
+			// Leave a cutoff that was seeded for THIS set. The caller seeds
+			// between the reset and this call, so clearing unconditionally
+			// erased it in the same breath it was given.
+			if !m.cutoffSeeded {
+				m.firstWindowStart = time.Time{}
+			}
 			m.windowMu.Lock()
 			m.windowStart = time.Time{}
 			m.windowEnd = time.Time{}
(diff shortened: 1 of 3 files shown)


slayerjain:
Superseded by #4518, which landed first and absorbed this branch's entire content.

The consume ladder this PR was opened for — `DeleteFilteredMock` → `DeleteStartupMock` → `DeleteUnFilteredMock`, the identity-checked `DeleteStartupMock`, and `sameMock`/`deleteMock` in treedb — is all on `main` already.

Rebasing confirmed the branch predates #4518, so its net diff against `main` is 183 insertions and 809 deletions: almost entirely a **revert** of the eleven fixes #4518 shipped. The one non-revert difference is a *weakening* of `TreeDb.update`'s idIndex guard — this branch refuses only when the entry is found and mismatched, while `main` also refuses when the entry is missing, which is what stops a dangling index from turning a no-op Remove into a blind insert of a foreign tier's mock. `main`'s form is the correct one.

Closing rather than force-pushing an empty branch. A separate PR follows for a genuinely open defect found while reviewing this: `SetMocksWithWindow` harvests its per-kind revision set *after* the tier swap, so a kind that leaves the pool never bumps its own counter.

PR #4563, opened 2026-09-07
feat(record): add AfterRecordingComplete hook

shreya1510ss:
## What

Adds `AfterRecordingComplete(ctx, *RecordingCompleteContext) error` to the `RecordHooks` interface — an **end-of-recording** hook invoked exactly once, after every test case *and* every mock in a test set has been drained to disk.

```go
type RecordingCompleteContext struct {
    TestSetID string
    Path      string
}
```

## Why

Some post-processing needs the **whole test set persisted**, not just a single artifact — and no existing hook provides that. During recording the request test-cases and the mocks drain on **two independent goroutines/channels**, so the per-artifact hooks (`BeforeTestCaseInsert`, `BeforeMockInsert`) can never see both sides with a guaranteed ordering. A hook that fires only after both drains complete is the clean place to do cross-artifact work.

The first consumer is the enterprise **Basic-Auth re-key** (keploy/api-server#1867, enterprise PR keploy/enterprise#2536): once recording is complete it correlates each recorded login's request credential with its stored password hash and rewrites the hash so obfuscated bcrypt logins replay correctly — without any real credential or hash on disk. That consumer now spans all mock formats (HTTP, MySQL, Postgres v2/v3, Mongo, generic) plus a configurable pepper, all built on this single hook.

## Details

- **Non-breaking for embedders.** The method has a **no-op default** on `BaseRecordHooks`, so every existing implementer that embeds it (`AsyncRecorder` and the enterprise hooks) compiles and behaves unchanged. Adding a method to the exported `RecordHooks` interface is technically breaking only for an external implementer that does *not* embed `BaseRecordHooks` — there are none in-repo.
- **Call site.** Invoked in `Recorder.Start`'s teardown, right after the record errgroup drains, guarded on `recordingStarted && newTestSetID != ""` **and a clean (non-timed-out) drain**. On the `DrainErrGroup` timeout path a persist goroutine may still be writing, so the pass is skipped (and warned) rather than run against a possibly-incomplete test set. A hook error is logged and **never invalidates the recording**.
- **No behavior change** for anyone who doesn't implement the hook.

## Merge order

This is the base change; the consuming enterprise re-key (keploy/enterprise#2536) lands after this merges and that repo bumps its `go.keploy.io/server/v3` dependency to a release including this hook.


pkg/service/record/record.go
```diff
@@ -735,6 +735,19 @@ func (r *Recorder) Start(ctx context.Context) error {
 			utils.LogError(r.logger, err, "failed to stop recording")
 		}
 
+		// End-of-recording hook (issue #1867 Basic-Auth re-key): every test case
+		// and mock is now drained to disk, so a cross-artifact pass can correlate
+		// them. Best-effort — a failure never invalidates the recording.
+		if recordingStarted && newTestSetID != "" {
+			if hookErr := r.hooks.AfterRecordingComplete(ctx, &RecordingCompleteContext{
```
shreya1510ss, line 742: **Major (design / forward-risk): this hook receives an already-cancelled context.**

This fires from `Start`'s deferred teardown, and on the normal stop path `ctx` is already cancelled — the same `defer` opens with `select { case <-ctx.Done(): default: utils.Stop(...) }` (record.go:685), which is exactly the SIGINT / `keploy record` stop path. So `AfterRecordingComplete(ctx, …)` gets a dead context.

The current consumer (enterprise `ReKeyBasicAuth`) ignores `ctx`, so there's no live bug today. But this is a **public extension point** (the interface doc says "enterprise (or any consumer)") whose whole purpose is post-recording cross-artifact work. The next consumer that does any ctx-aware I/O in it — upload a test set, write to a DB, call an API — will find every call failing with `context.Canceled`, and nothing in the contract warns them. Cheap to get right now, expensive to change once consumers depend on the semantics.

Fix (Go is 1.27 here, so `WithoutCancel` is available):
```go
hookCtx := context.WithoutCancel(ctx)
if hookErr := r.hooks.AfterRecordingComplete(hookCtx, &RecordingCompleteContext{...}); hookErr != nil { ... }
```
At minimum, document on the method + `RecordingCompleteContext` that the context may already be cancelled and hooks must not rely on it for cancellation.

shreya1510ss (commented):
## Staff review — `AfterRecordingComplete` hook (base PR for the #1867 Basic-Auth re-key)

Small, clean, and correctly forward-compatible: the new method has a no-op default on `BaseRecordHooks`, and I verified there is **no direct implementer of `RecordHooks` in-repo that doesn't embed it** (`AsyncRecorder` and the enterprise hooks both embed the base; the other `AfterMockInsert` matches are callers/comments), so this is non-breaking within the monorepo. **No blockers.** One design point to fix before this sets the public contract (inline), plus two process notes:

- **CLA is failing** on an author-email mismatch — sign with the matching email (or align the commit email) and comment `recheck`.
- Consider a one-line doc on the interface: *"implement by embedding `BaseRecordHooks` to stay forward-compatible"* — adding a method is technically breaking for any third-party implementer that doesn't embed the base.

Verdict: **Approve with comments** (posted as a comment review since self-approval is blocked).

shreya1510ss:

charankamarapu (changes requested):
## Principal review

Small, well-shaped interface addition. I went and checked whether the hook is actually necessary rather than taking the PR body's word for it, and it is: Stage 2 needs a global view (every captured password × every stashed hash) to prove correlation by `bcrypt.CompareHashAndPassword`, and the two `Before*Insert` hooks fire on independent drain goroutines (`record.go:1104` and `record.go:1183`), so neither ever holds both sides. `MockContext.Skip` **drops** a mock rather than deferring one, so you can't buy ordering inside the existing hooks without buffering mocks of unknown lifetime and breaking the streaming write model. And even if you could, an O(passwords) bcrypt scan inside `BeforeMockInsert` would put a deliberately-slow KDF on the recording drain. Batching it at the end is right on both counts.

The shape is also right: no-op default on `BaseRecordHooks`, invoked after all four `DrainErrGroup` calls, error logged and never invalidating the recording, and `TestAfterRecordingComplete_RecoversFromHookPanic` covering the panic path with the correct rationale (a panic would unwind the teardown defer and skip the deferred-orphan-revoke immediately below).

Three things.

---

### B1 — the hook can fire while mock writes are still in flight

`utils.DrainErrGroup` **returns nil on timeout** — that's deliberate and documented there ("On timeout it logs … and returns nil; the leaked goroutine is reaped when the process exits"). So the caller cannot distinguish a clean drain from a 30 s timeout with a goroutine still running, and `if err := utils.DrainErrGroup(...); err != nil` catches only a real errgroup error.

The new call is guarded on `recordingStarted && newTestSetID != ""` — nothing about whether the drain actually completed. Yet the comment above it asserts:

> every test case and mock is now drained to disk, so a cross-artifact pass can correlate them

That isn't guaranteed on the timeout path, and this file already knows it — ~20 lines below, the deferred-orphan-revoke defends against exactly this:

> On the DrainErrGroup TIMEOUT path a wedged consumer goroutine can still be writing these maps (DrainErrGroup returns after its timeout WITHOUT joining a goroutine that ignores cancellation)

It guards with a lock-and-snapshot. The new hook has no equivalent guard, and its consumer is worse-placed than a map read: keploy/enterprise#2536's `rewriteMockFile` does `os.ReadFile` → `strings.ReplaceAll` → `os.Rename` over `mocks.yaml`. If a mock-insert goroutine appends between the read and the rename, that mock is **silently lost** — and because the rename is atomic, the result is a clean, valid, short file rather than visible corruption. Silent data loss in recorded fixtures is about the worst failure mode available here.

Low probability (30 s, and the drains normally finish well under it), high consequence. I'd have `DrainErrGroup` signal the timeout distinctly — a named sentinel or a `(err error, timedOut bool)` — without changing its existing "don't fail teardown" contract, then gate:

```go
if recordingStarted && newTestSetID != "" && recordDrainedCleanly {
    r.afterRecordingComplete(ctx, newTestSetID)
}
```

Skipping the re-key on a wedged teardown costs one recording's replay auth. Running it costs recorded mocks.

### B2 — the comment misstates `context.WithoutCancel`, and the contract disagrees with its only consumer

> Hand the hook a context decoupled from that cancellation … while still propagating request-scoped values and **any deadline set upstream**

`WithoutCancel` drops the deadline. Verified:

```
parent   deadline: 2026-09-15T13:33:31+05:30 (ok=true)
WithoutCancel child deadline: 0001-01-01 ... (ok=false)  Done==nil: true
```

Values propagate; deadline and cancellation do not. So "the hook remains free to honor it" is offering something that isn't there.

More to the point: you added `WithoutCancel` **and** a test (`TestAfterRecordingComplete_PassesLiveContext`) specifically so the consumer *can* honor the context — and the consumer takes `_ context.Context` and ignores it, justified by a comment describing the behaviour this PR **changes**:

> The recorder's context is intentionally NOT forwarded: it is already cancelled on the normal SIGINT stop

After this PR it isn't cancelled. One of the two changes is redundant, and the enterprise comment is now false about its own dependency. Pick a contract and make both sides state it: either the hook gets a live-but-bounded context that consumers are expected to honor (I'd prefer `context.WithTimeout(context.WithoutCancel(ctx), …)`, matching the 10 s bound on `NotifyGracefulShutdown` and the 30 s drain bounds in the same defer), or the hook explicitly passes no cancellation and consumers self-bound — in which case say so and drop the "free to honor it" line.

### m3 — "non-breaking" is slightly overstated

Adding a method to the exported `RecordHooks` interface breaks any external implementer that doesn't embed `BaseRecordHooks`. In-repo you're fine — `AsyncRecorder` embeds it — so this is about the claim, not the code. Worth softening to "non-breaking for embedders", since `RecordHooks` is an extension point you're inviting people to implement.

---

### Nits

- `RecordingCompleteContext` giving `Path` + `TestSetID` separately, and letting the consumer join them, is the right call over passing a pre-joined path — it keeps the hook honest about what it's being told.
- The panic-recovery comment explaining *why* best-effort has to cover panics (the deferred-orphan-revoke runs right after) is the kind of note that stops someone "simplifying" the recover away later.

Happy with the design. B1 is the one I'd want fixed before merge; B2 is a contract/comment decision that should be settled jointly with keploy/enterprise#2536 rather than in either PR alone.

shreya1510ss:
@charankamarapu pushed 7638a913 addressing the review.

**B1 — fixed.** Added a sibling `DrainErrGroupStatus` that returns `(err, timedOut)`; `DrainErrGroup` delegates to it, so its "a timeout never fails teardown" contract and the other 8 call sites are untouched. The `record` drain now captures `timedOut`, and `AfterRecordingComplete` is gated on a clean drain — on the timeout path it's skipped and warned, so the hook can no longer run while a persist goroutine is still appending to a mock file. Gated on `!timedOut` rather than `err == nil` (a real errgroup error still means every goroutine returned, so nothing is left writing). Two unit tests cover both branches of the signal.

**B2 — comment corrected.** The `hookCtx` doc now states the truth: `context.WithoutCancel` keeps request-scoped values but drops *both* cancellation and any upstream deadline — it does not propagate a deadline. I did not add a `WithTimeout` bound here: as you said, whether the hook carries a shared deadline is a contract decision to settle alongside keploy/enterprise#2536, where the consumer forwards and honors the context.

**m3 — reworded.** The description now says "non-breaking for embedders", with the external-non-embedder caveat spelled out.

shreya1510ss:
@charankamarapu follow-up on B2 — the deadline sub-point is now implemented (`bf3adf91`), superseding my earlier "leave it as a joint decision" note.

The hook context is now **bounded**, not just decoupled:

```go
hookCtx, cancelHook := context.WithTimeout(context.WithoutCancel(ctx), afterRecordingHookTimeout) // 30s
defer cancelHook()
```

So `WithoutCancel` still drops the recorder's teardown cancellation, but a fresh 30s deadline is re-imposed — matching the 10s `NotifyGracefulShutdown` and 30s `DrainErrGroup` bounds in the same defer, so a wedged consumer can't hang teardown. It's the hook context's single deadline, and consumers are expected to honor it.

Coordinated with keploy/enterprise#2536: the re-key there now sources its wall-clock cap from `ctx.Deadline()` (this timeout) instead of running an independent second timer — one source of truth. Build/test/`-race`/lint green on both sides.

B1, B2, and m3 are all addressed now — could you take another look when you get a chance?

charankamarapu (approved):
## Approving

Both findings closed, and B1 is closed the right way rather than the cheap way.

**B1 — the drain/hook race.** `DrainErrGroupStatus` returning `(err, timedOut)` with `DrainErrGroup` kept as a thin wrapper is exactly right: no existing caller changes, and the new signal is available where it matters. The doc comment earns its length by stating the non-obvious part explicitly —

> a non-nil err still means the group finished … whereas timedOut means it did not join. A caller that then touches an artifact a wedged group goroutine could also be writing … must gate on `!timedOut`, not on `err == nil`

— which is precisely the mistake the next person would make. The `recordDrainedCleanly` gate reads clearly, and the skip path logs a Warn that says what was skipped and that the recording itself is fine, rather than failing silently.

`TestDrainErrGroupStatusCleanNotTimedOut` is the test I'd have asked for: a group that returns an *error* must still report `timedOut=false`, because it joined. That's the case where a careless implementation conflates the two and wrongly skips the hook on every errored recording.

One detail I hadn't considered and you surfaced in the comment: the insert goroutines write on `persistCtx = WithoutCancel`, so on the timeout path they genuinely *won't* observe cancellation and will keep writing. That strengthens the gate — it isn't a theoretical window, those writers are actively not stopping.

**B2 — the context contract.** `context.WithTimeout(context.WithoutCancel(ctx), afterRecordingHookTimeout)` with a named 30 s constant aligned to the drain budgets is the right shape, and the comment is now factually correct — "drops the upstream cancellation AND deadline, so this WithTimeout is the hook's only deadline." Both sides of the pair now agree: keploy/enterprise#2536 takes the ctx, checks it at the pass boundary, per-job and in the inner loop, and derives its wall-clock budget from `ctx.Deadline()` instead of running an independent second timer. One timer, one owner.

Verified: CI green on this PR.

---

### One gap, non-blocking

There's no test pinning the B2 fix. `TestAfterRecordingComplete_PassesLiveContext` asserts `ctx.Err() == nil` — it would pass equally under the old bare `WithoutCancel`. Nothing fails if someone later drops the `WithTimeout` and reintroduces an unbounded hook context. A couple of lines in the existing capture hook:

```go
dl, ok := ctx.Deadline()
// ok must be true, and dl within ~afterRecordingHookTimeout of now
```

Worth adding while you're here — the bound is the whole point of the change, and it's currently the only part of this PR that isn't guarded.

### Nit

The PR body still says "Non-breaking." Adding a method to the exported `RecordHooks` interface breaks any external implementer that doesn't embed `BaseRecordHooks`; in-repo you're fine because `AsyncRecorder` embeds it. "Non-breaking for embedders" would be accurate for an extension point you're inviting people to implement.

Merge order unchanged: this lands first, then keploy/enterprise#2536 re-points `go.keploy.io/server/v3` off its branch pseudo-version to a release containing this.

Issue #4578, opened 2026-09-12
Replay auto-passes an HTTP body diff as 'additive' even when an existing value changed


prabaltripathiofficial:
## What happens

`compareHTTPRespForReplay` calls `autoPassHTTPResponseSchemaAddition` when the body comparison fails. It is meant to pass a case whose response only gained new fields. In practice it also passes when an existing value changed in the same response.

Reproduced with a small Go test app whose `GET /products` handler was changed in one commit to (a) add a new field `featureConfig` and (b) change `count` from 8 to 10:

- expected body: `{"apiVersion":"v2","configCount":2,"count":8,"products":[...]}`
- actual body: `{"apiVersion":"v2","configCount":2,"count":10,"featureConfig":null,"products":[...]}`
- report: `body_result[0].normal: true`, test PASSED
- log: `Additive response schema change detected. Passing testcase by default because only new response fields were added.` with `added_fields: ["featureConfig"]`

`matcher.JSONDiffWithNoiseControl` on the same two bodies (no noise besides `header.Date`, `ignoreOrdering: true`) returns `matches=false`, so the comparison itself is right; the auto-pass overrides it.

Removing the added field and keeping only the `count` change makes the same case FAIL as expected, so the value change is only hidden when a field is added alongside it.

## Why it matters

A PR that changes a response value and adds a field in the same handler passes replay silently. In a PR gate this hides exactly the regressions the gate exists for.

## Where

`pkg/service/replay/replay.go`: `autoPassHTTPResponseSchemaAddition`, `qualifiesForHTTPResponseSchemaAdditionPass` (Risk Low + only `SchemaAdded` category, or `hasOnlySchemaAdditionAndContentLengthDiff`). The assessment from `matcherUtils.ComputeFailureAssessmentJSON` appears to classify the response as additive-only despite the `count` change; either the assessment misses value changes when additions are present, or the qualifier should require zero value/type changes explicitly.

## Suggested fix

Qualify for the additive auto-pass only when the field-level assessment reports no `value_changed`, `type_changed` or `missing_in_live` entries at all, not only when the category set is `SchemaAdded`. Add a test with an added field plus a changed scalar that must stay FAILED.

Enterprise CLI `3-dev` built from `main` on 2026-09-12, OSS `go.keploy.io/server/v3` v3.6.58.

PR #4581, opened 2026-09-13
fix-4578-additive-autopass

Abhishek84313:
- **Bug:** replay auto-passed an HTTP test as an "additive schema change" even when an existing value also changed in the same response. In #4578, `count` went from `8` to `10` while a new `featureConfig` field was added, and the test PASSED.
- **Root cause:** `qualifiesForHTTPResponseSchemaAdditionPass` decided from the failure category set alone.
  - `ComputeFailureAssessmentJSON` labels "new fields plus changed values" as `SCHEMA_ADDED` too; only the risk differs (`Medium` instead of `Low`).
  - The first branch of the qualifier checks for `Low` risk, so on its own it would reject this case.
  - A bigger body also changes `Content-Length`, which sends the case to the second branch, `hasOnlySchemaAdditionAndContentLengthDiff`. That branch never checks risk or value changes, so the test was auto-passed.
- **Fix:** before either branch, the qualifier now requires the body assessment (`result.FailureInfo.Assessment`, which `Match` already computes on the noise-cleaned bodies) to list at least one added field and no removed fields, type changes or value changes. This is the new helper `isAdditiveOnlyAssessment` in `pkg/service/replay/replay.go`.
- **Tests:** added `pkg/service/replay/schema_addition_autopass_test.go`.
  - Tests the qualifier directly: added field only (with and without a `Content-Length` diff) must qualify; added field plus a changed top-level or nested value (with and without a `Content-Length` diff) must not; a missing assessment must not.
  - Tests the full path through `compareHTTPRespForReplay` and the real HTTP matcher, using the bodies from the issue, with `emitFailureLogs` both true and false: the added field alone passes, the added field plus a changed `count` fails, and the changed `count` alone fails.

## Links & References

**Closes:** #4578

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- #4578
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

The unit tests drive the real HTTP matcher through `compareHTTPRespForReplay` with the bodies from the issue, which covers the path that failed.

- [x] 👍 yes

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

- [x] 👍 yes, mentioned below

**Unit tests:**
go test ./pkg/service/replay/ -run 'TestQualifiesForHTTPResponseSchemaAdditionPass|TestCompareHTTPRespForReplaySchemaAddition' -v -count=1

**Manual repro (from #4578):**
1. Record `GET /products` on a small app that returns `{"apiVersion":"v2","configCount":2,"count":8,"products":[...]}`.
2. In one change, add a `featureConfig` field and change `count` to `10`.
3. Run `keploy test`.
   - Before this fix: the test PASSES with the log "Additive response schema change detected...".
   - After: the test FAILS.
4. Change only the added field (keep `count` at `8`): the test still auto-passes as an additive change.

## Self Review done?
- [x] ✅ yes

- NA

---

## Additional checklist:


chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix-4578-additive-autopass. Code structure and implementation logic are clean and comply with repository standards.

PR #4494, opened 2026-08-26
feat(models): add unit test validation for PostgresV3Response mutual

SatyamPandey-07:
## Describe the changes that are made
- Updated `TestPostgresV3Response_CopyIn_RoundTrip` in `pkg/models/mock_test.go` to explicitly test `PostgresV3Response.Validate()`.
- Verified that valid `CopyIn` responses pass validation while malformed shapes carrying both `Rows` and `CopyIn` are rejected by `Validate()`.
- Cleaned up the stale TODO in `pkg/models/mock_test.go`.

## Links & References

Closes #4492

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- #4492
### 📄 Related Documents
- NA

## What type of PR is this? (check all applicable)
- [ ] 📦 Chore
- [x] 🍕 Feature
- [ ] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [x] 🙅 no, because they aren't needed

## Added comments for hard-to-understand areas?
- [x] 🙅 no, because the code is self-explanatory

## Added to documentation?
- [x] 🙅 no documentation needed

## Are there any sample code or steps to test the changes?
- [x] 👍 yes, mentioned below

Run Postgres model tests:
```bash
go test -v -tags=viper_bind_struct ./pkg/models -run "TestPostgresV3Response.*"
```

## Self Review done?
- [x] ✅ yes

## Any relevant screenshots, recordings or logs?
- NA

---

## Additional checklist:
- [x] Have you read the [Contributing Guidelines on issues](https://keploy.io/docs/keploy-explained/contribution-guide/)?
- [x] Have you followed the [PR Semantics guide](https://github.com/keploy/keploy/wiki/PR-Semantics) for naming this PR?
- [x] Have you followed the [Branch Semantics guide](https://github.com/keploy/keploy/wiki/Branch-Semantics) for naming your branch?

Issue #2579, opened 2025-03-12
[bug]: Test case capture if basepath doesn't matches requestURL path


kapishupadhyay22:
### 👀 Is there an existing issue for this?

- [x] I have searched and didn't find similar issue

### 👍 Current behavior

The Test case are being captured in E2E, even when the basepath is not the same in the requestURL path

### 👟 Steps to Replicate

1. Run Keploy record and provide a basepath (which will be different from the requestURL path)
2. Log the basepath and requestURL path while capturing test cases
3. There are different paths but the test case is still being captured. 

### 💻 Operating system

Linux

### 🎲 Version

Fedora Linux 41

### 💻 Repository

keploy

PR #274, opened 2023-01-04
changed the name of the mock files

PranshuSrivastava:
Closes: #247 

#### Describe the changes you've made
Updated the names of the mock files.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Code style update (formatting, local variables)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Please let us know if any test cases are added

Please describe the tests(if any). Provide instructions how its affecting the coverage.

#### Describe if there is any unusual behaviour of your code(Write `NA` if there isn't)
A clear and concise description of it.

## Checklist:

- [ ] My code follows the style guidelines of this project.
- [ ] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes.

## Screenshots (if any)

|        Original         |          Updated           |
|:-----------------------:|:--------------------------:|
| **original screenshot** | <b>updated screenshot </b> |

PR #4582, opened 2026-09-13
refactor(noise): rename the schema-noise engine to mock-noise, compatibly

Shivanipandey31:
## Why

"Schema noise" read as though it belonged to schema matching or the smart-set schema pipeline. It belongs to neither — it is noise learned on a **mock**. This renames the engine and the user-facing toggles to say so.

The rename spans five repositories (keploy, integrations, enterprise, k8s-proxy, enterprise-ui) that merge and deploy on their own schedules. This PR is the foundation the other four build on, so its whole job is to make the new spelling available **without breaking the old one**.

## The failure mode this is designed around

`encoding/json`, `yaml` and `viper` all drop unknown keys **without error**. So a producer on one spelling talking to a consumer on the other doesn't fail — it produces a toggle that quietly does nothing, with no diagnostic anywhere. That is exactly the bug that made this rename risky, and it's why the compatibility here is mechanical rather than by convention.

## What changed

**Package** — `integrations/schemanoise` → `integrations/mocknoise`. The old path remains as a shim of **type aliases** (`=`), not defined types.

That distinction is load-bearing. An alias *is* the same type, so a `*mocknoise.Engine` built by migrated code satisfies a `*schemanoise.Engine` parameter in code that hasn't migrated, and an `Adapter` written against either spelling satisfies the other's interface. With defined types, every boundary between a migrated and an unmigrated caller would be a compile error in a repo that changed nothing.

`schemanoise_compat_test.go` asserts this **cross-wise and at compile time** — turning the aliases back into defined types breaks the build here rather than breaking a downstream repo.

**Fields** — Go has no field alias, so the package trick doesn't carry over. `models.OutgoingOptions` and `config.Test` now hold **both** spellings, with `NoiseDetection()` / `NoiseStrict()` accessors and `NormalizeMockNoise()`. Nothing reads a noise field directly any more.

Reconciliation **ORs** rather than letting the canonical field win. The deprecated field is the only one an unmigrated producer sets — a pre-rename k8s-proxy or enterprise build, or a `keploy.yml` written before this change — so treating an unset canonical field as "off" would silently discard that producer's intent. Neither spelling can express "explicitly off" as distinct from "unset", so OR loses nothing.

`Proxy.Record` and `Proxy.Mock` normalise on receipt, because that's the process boundary where a client of any vintage arrives.

**CLI** — `--mock-noise-detection` / `--mock-noise-strict` are canonical. `--schema-noise-detection` / `--schema-noise-strict` keep working for existing scripts and CI jobs. Both are read and OR-ed, and `config.Test` is normalised once after parsing so every later reader sees one answer.

## Compatibility matrix, covered by tests

| Producer sets | Consumer reads | Result |
|---|---|---|
| canonical only | either | on |
| **deprecated only** (pre-rename build) | either | on |
| both | either | on |
| mixed (detection canonical, strict deprecated) | either | both on |
| neither | either | off |

Plus nil receivers and idempotency of `NormalizeMockNoise`.

## Verification

`go build ./...`, `go vet ./...` and `go test ./...` clean — **67 packages with tests, zero failures**.

## Merge order

This one first. `keploy/integrations#274` is already written against `mocknoise` + `MockNoise*` and cannot compile until this is released and pinned.


Shivanipandey31:
## Staff review — 6 iterations, 5 findings, all fixed

Reviewed this branch against `main` for correctness, cross-repository safety, design, and forward risk, fixing what came up and re-reviewing until an iteration produced nothing new. Iteration 6 was clean.

### Iteration 1 — the serious one

**Major — the canonical config key silently did nothing, and the deprecated one was already broken.**

`ValidateFlags` assigned `cfg.Test.MockNoiseDetection` straight from the flag, unguarded. The ordering makes that destructive:

1. `AddFlags` registers each flag with a default read from a **zero** config
2. `PreProcessFlags` runs `viper.Unmarshal`, filling the config from `keploy.yml`
3. `ValidateFlags` reads the flags back — overwriting step 2

So a `keploy.yml` carrying `test.mockNoiseDetection: true` was overwritten with the flag's stale `false`. That is precisely the silent-toggle failure this PR exists to prevent, reintroduced one layer up.

Worse: **the deprecated key had the same hole already.** `test.schemaNoiseDetection: true` has been ignored whenever the flag was absent. The strict flag has carried the correct guard — and a comment explaining exactly this hazard — all along; detection never got it.

Both are now guarded on `Changed || !viper.IsSet`. Fixing the pre-existing one here rather than leaving it: a rename whose entire claim is *"your existing keploy.yml keeps working"* cannot ship with that hole open for the very key it renames.

**Minor — the compat shim was reassignable.** Its forwarders were package-level `var`s, so any importer could overwrite `schemanoise.New` at runtime and change behaviour for every other caller in the process. Now plain functions — same signatures, same aliasing, no mutable global state.

### Iterations 2–3 — no new findings

Re-read the fixed diff and confirmed every `OutgoingOptions` entry point normalises: `GetOutgoing → Proxy.Record` and `MockOutgoing → Proxy.Mock`, both before the options reach the session.

### Iteration 4 — two more, and they contradicted the PR's own claim

**Major — two consumers still read a noise field directly**, despite the description asserting nothing does:

- `resolveTestOutcome` was handed `r.config.Test.SchemaNoiseStrict` — the knob deciding whether a mock-set mismatch demotes a test to OBSOLETE or fails it.
- The noise **persistence** branch was gated on `r.config.Test.SchemaNoiseDetection`, so learned `req_body_noise` would be written or discarded on one spelling only.

Both work today because every in-tree writer still sets the deprecated field. They break the moment a caller sets only the canonical one — which is what the rest of this change invites. Silent both times: a test resolves to the wrong verdict, or learned noise is quietly dropped.

`TestRunTestSetWiring` caught the first immediately. Its expectation moves from `SchemaNoiseStrict` to `NoiseStrict`, with a note on why naming the raw field would re-introduce the single-spelling blindness the guard exists to prevent.

### Iteration 5 — cross-repository

- **integrations** (`rename/mock-noise`) builds and its mongo suite passes against this branch's head; re-pinned so CI exercises what will actually merge.
- **enterprise** builds and tests green against **both** its current pin (v3.6.58, pre-rename) **and** this branch — the alias shim holds in both directions.

### Iteration 6 — clean

gofmt 0, `go build`, `go vet`, and **67 packages** of tests green. The only remaining textual references to the deprecated fields outside their owning files are comments.

### Also fixed

`pkg/models/mock_test.go` arrived unformatted via `c6ad54aca` (#4494) and fails the lint workflow's `gofmt -l -d .` step — which is separate from golangci-lint (0 issues). **Every PR merging current main will hit this**, and main's own lint is red until it is fixed there too. Worth a one-line fix straight to main.

### Residual risk

**The two spellings can still drift for a caller that writes fields directly rather than going through `NormalizeMockNoise`.** Accessors make reads safe, but nothing forces a *writer* to keep the pair consistent — a future `OutgoingOptions{MockNoiseStrict: true}` literal in another repo would leave the deprecated mirror false for any consumer still reading it. The normalise-on-receipt calls in `Proxy.Record`/`Mock` cover every path that crosses the agent boundary today; an in-process caller that skips them is the gap. Removing it entirely needs the deprecated fields gone, which is the follow-up once all five repos are migrated.

Aditya-eddy:
## Review — one blocker, in the producer direction

The design here is right and the shim is complete. But there's a gap in applying it that reintroduces exactly the failure mode this PR exists to prevent, in the one direction the compatibility matrix doesn't cover.

### The blocker

`OutgoingOptions` is marshalled to the agent over HTTP (`pkg/platform/http/agent.go:275`, `json.Marshal(models.OutgoingReq{...})`) and carries **no json tags**, so it goes on the wire under Go field names.

Three sites build that struct and set **only the canonical fields**, leaving the deprecated mirror at zero:

- `pkg/service/runner/runner.go:386-387`
- `pkg/service/replay/replay.go:1296-1297`
- `pkg/service/replay/replay.go:1537-1538`

`NormalizeMockNoise()` is called on **receipt** (`proxy.go:3538`, `proxy.go:3607`), which protects a new agent from an old client. Nothing mirrors before the marshal, so the reverse is unprotected.

Decoded against the real structs:

```
CLI sends: {"MockNoiseDetection":true,"MockNoiseStrict":true,
            "SchemaNoiseDetection":false,"SchemaNoiseStrict":false}

OLD agent decodes: SchemaNoiseDetection=false SchemaNoiseStrict=false   <- toggle dead
AFTER fix:         SchemaNoiseDetection=true  SchemaNoiseStrict=true
```

**This is a regression, not a pre-existing gap.** Before this PR the same CLI sent `SchemaNoiseDetection: true` and a pre-rename agent honoured it. After it, `--mock-noise-detection` *and* the still-supported `--schema-noise-detection` both arrive as `false` — the accessor collapses both spellings into the canonical field, so the deprecated flag stops working too against an older agent.

### Why the skew is real

The codebase already documents this axis — `pkg/platform/http/agent.go:705`:

> "a deployed agent that predates streaming (keploy <= v3.5.84) during a rolling upgrade — align the agent image with the controller's keploy version"

And the consumer is concrete: released `integrations/pkg/mongo/v2/decode.go:33` reads `opts.SchemaNoiseDetection`, and that code runs **inside the agent process**. A post-rename agent is fine (opts normalized on receipt); an agent image built before this PR is not. k8s-proxy pins `proxy.keployAgentImage` separately from the CLI, and enterprise is still on keploy `v3.6.58` — so pre-rename agents stay deployed across the whole rollout window while CLIs update independently.

### Fix — one line per site

```go
outOpts.MockNoiseDetection = r.config.Test.NoiseDetection()
outOpts.MockNoiseStrict    = r.config.Test.NoiseStrict()
outOpts.NormalizeMockNoise()   // mirror before the struct goes on the wire
```

`NormalizeMockNoise` is already idempotent and nil-safe, so this composes with the receive-side call at no cost. The PR's own rule — *"SET whichever spelling you like, then let NormalizeMockNoise reconcile at the boundary"* — is correct; the marshal point just wasn't treated as a boundary.

Worth adding a test that a canonical-only `OutgoingOptions` still populates the deprecated pair **before** marshal. The compatibility matrix in the description has the matching hole: it covers *old producer -> new consumer*, but has no row for *new producer -> old consumer*, which is why CI is green — every test exercises both halves at the same version.

### Everything else checks out

- **The alias shim is complete.** Every exported symbol of the old package survives, and `Engine`/`JSONDiffer` being aliases carries all seven methods (`DetectionEnabled`, `StrictEnabled`, `KnownNoise`, `Detect`, `Learn`, `StrictAllows`, `StrictReject`) automatically. `schemanoise_compat_test.go` asserting cross-wise at compile time is the right call.
- **No stray direct reads.** Every non-test access of the four fields is a flag default or a write; all reads go through the accessors.
- **The `Changed || !viper.IsSet` guard is a genuine bug fix** beyond the rename — the flag's stale default (captured from a zero config before `viper.Unmarshal` ran) was overwriting yaml-only values, so `test.schemaNoiseDetection: true` was silently dead before this PR.
- **OR is safe here**, unlike enterprise#2579 — keploy's defaults are `false`, so there's no explicit-false to clobber. #2579 needing a precedence chain instead is correct given `DefaultAutoReplaySchemaNoiseDetection` is `true`.

### Minor: a doc reference that doesn't resolve

`keploy/integrations#274` describes the mirrored fields as *"kept in sync by `SetMockNoise`"*, but `SetMockNoise` doesn't exist on this branch (`git grep` returns nothing). #274's CI is green so it isn't calling it — but the description points a reader at an API that was never written. Worth fixing there rather than here.

Shivanipandey31:
Thanks @Aditya-eddy — the blocker is real, it is a regression this PR introduced, and it is fixed in `6428ed210`.

## Verified before fixing

All three parts of your analysis check out:

1. `OutgoingOptions`' noise fields carry **no json tags**, so they go on the wire under Go field names.
2. It is `json.Marshal`ed to the agent at **two** sites — `platform/http/agent.go:271` (GetOutgoing) and `:423` (MockOutgoing).
3. All three producers set canonical-only.

Reproduced against the real struct shapes:

```
on the wire:       {"MockNoiseDetection":true,"MockNoiseStrict":true,
                    "SchemaNoiseDetection":false,"SchemaNoiseStrict":false}
OLD agent decodes: SchemaNoiseDetection=false SchemaNoiseStrict=false
```

And your sharpest point is the one I'd missed: **`--schema-noise-detection` stops working too**, because the accessor collapses both spellings into the canonical field before the send. So the deprecated flag — the thing this PR promises keeps working — was the casualty.

## Fix

Applied at the **two marshal sites** rather than the three producers. That is the actual wire boundary, so a future producer that forgets is covered automatically, and it reads as the send-side twin of the receive-side call you cited. `NormalizeMockNoise` being idempotent and nil-safe means pairing them costs nothing.

## Tests

Two, closing the matrix hole you identified:

- `TestCanonicalOnlyOptionsStillReachAPreRenameAgent` — the missing **NEW producer → OLD consumer** row, decoding real marshalled bytes into a struct shaped like a pre-rename agent's.
- `TestDeprecatedOnlyWireStillDrivesANewAgent` — the reverse, pinned next to it so the matrix is visibly complete.

**Mutation-checked**: removing the boundary mirror fails the first with exactly your diagnosis —

```
a pre-rename agent decoded detection=false from a canonical-only producer;
the toggle is dead across version skew
```

Your observation about *why* CI stayed green is the useful part and I've written it into the test comment: every existing test exercises producer and consumer at the same version, so no amount of them could have caught this.

gofmt, build, vet and tests clean — 67 packages.

## On the minor

You're right that `SetMockNoise` doesn't exist here. It was in the earlier draft of this rename (`5c24cd8aa`, the commit that was split out of #4483) and I replaced it with read accessors, but integrations#274's description still references it. I'll fix that description rather than add the method — the accessor shape is the one that survived review, and adding a writer helper now would give two ways to do the same thing.

Aditya-eddy (approved):
Approving — the blocker is resolved. Verified at `6428ed2`.

Fixing it at the wire boundary in `platform/http/agent.go` rather than at the three producers is the better call: those two marshal sites are the only `OutgoingReq{}` constructions in the repo, so a future producer that forgets is covered automatically. Re-checked both directions after the change, and CI is 169/169.

Also re-confirmed on this pass: the engine move is a pure rename (normalising the naming on both sides, `mocknoise.go` vs the old `schemanoise.go` diffs to comment text only — no logic change), the receive path is fully covered (`routes -> svc.GetOutgoing/MockOutgoing -> Proxy.Record/Mock`), and the alias shim re-exports every symbol the old package had.

Three follow-ups, none blocking this merge:

**1. The new regression test doesn't guard the fix.** `TestCanonicalOnlyOptionsStillReachAPreRenameAgent` still passes with *both* mirrors deleted from `agent.go`. It calls `sent.NormalizeMockNoise()` itself right above the marshal, and it lives in `package models`, which has no dependency on `pkg/platform/http` — so it asserts that `NormalizeMockNoise` plus a JSON round-trip works, which was already true before the fix. The regression can be reintroduced by deleting two lines with CI green. A version that bites would live in `pkg/platform/http`: point `conf.Agent.AgentURI` at an `httptest.Server`, call `MockOutgoing` with canonical-only options, assert the decoded request body carries `SchemaNoiseDetection: true`. Worth doing before the deprecated spelling is eventually removed.

**2. Cross-spelling explicit-off is discarded.** With `schemaNoiseDetection: true` in keploy.yml and `--mock-noise-detection=false` on the command line, detection stays on: the `Changed || !viper.IsSet` guard correctly leaves the yaml value alone for the spelling the user didn't type, and the OR at cmd.go:1608 then resurrects it. Symmetric — it fails whichever way the spellings are mixed; matching spellings are fine both ways. Not a regression (the canonical flag is new) and the default is off, so I don't think it gates anything. But the stated rationale — "neither spelling can express explicitly off as distinct from unset" — holds for the struct fields and not for the CLI, where `Changed()` distinguishes them precisely and this PR's own guard relies on that. enterprise#2579's precedence chain (canonical flag -> deprecated flag -> canonical key -> deprecated key -> default) resolves this for free; two PRs in one chain currently resolve the same two-spelling problem with different semantics, which is worth reconciling before the old names go away.

**3. keploy itself is mostly unmigrated.** Only `mysql/replayer/query.go` moved to the new import path; `http/match.go`, `http/schema_noise.go`, `mysql/replayer/match.go` and `mysql/replayer/schema_noise.go` still import the deprecated shim, so `mysql/replayer` now pulls in both spellings. Fine at runtime — that's what the aliases are for — but the shim's own deletion criteria ("delete once integrations, enterprise, k8s-proxy and enterprise-ui are all on the mocknoise path") omit keploy, so following them as written breaks keploy's own HTTP and MySQL matchers. Worth amending that comment even if the migration itself waits.

On merge order: this needs to land and be tagged before `keploy/integrations#274`, which pins a pseudo-version of this branch's head and still carries `.ci-keploy-ref`.

Issue #4583, opened 2026-09-13
[Bug] keploy diff ignores newly added and deleted test cases when comparing test runs[Bug] keploy diff ignores newly added and deleted test cases when comparing test runs


yush-1018:
### Is there an existing issue for this?

- [x] I have searched and didn't find similar issue

### Current behavior

When comparing two test runs using `keploy diff <run1> <run2>`, the comparison logic in `pkg/service/diff/diff.go` (`ComputeDiff()`) only considers test case IDs that exist in both test reports (`commonIDs`).

If a test case was added in `run2` (present in `run2` but absent in `run1`) or removed in `run2` (present in `run1` but absent in `run2`), it is completely omitted from the comparison result. It is neither reported under regressions, fixes, status transitions, nor unchanged test cases.

What I expect is:
`keploy diff` should detect and report test cases that were added or removed between test runs (e.g., under status transitions or added/deleted categories).

### Steps to Replicate

1. Record a test run `test-run-0` with test cases `tc-1` and `tc-2`.
2. Add a new test case `tc-3` and record `test-run-1` (so `test-run-1` has `tc-1`, `tc-2`, `tc-3`).
3. Run `keploy diff test-run-0 test-run-1`.
4. Observe that `tc-3` is completely ignored in the diff summary.

### Version

v3-dev (main branch)

### Repository

keploy

chilkotiKartik:
### Root Cause Analysis & Fix

#### Issue:
In `keploy diff`, the comparator logic only loops through keys present in the baseline test set, failing to record newly added test cases (present in candidate, missing in baseline) and deleted test cases (missing in candidate, present in baseline) in the final diff report.

#### Proposed Solution:
Perform symmetric set comparison on test case IDs:
```go
allTestKeys := make(map[string]struct{})
for k := range baselineTests { allTestKeys[k] = struct{}{} }
for k := range candidateTests { allTestKeys[k] = struct{}{} }

for k := range allTestKeys {
    base, inBase := baselineTests[k]
    cand, inCand := candidateTests[k]
    if inBase && !inCand {
        diffReport.Deleted = append(diffReport.Deleted, k)
    } else if !inBase && inCand {
        diffReport.Added = append(diffReport.Added, k)
    } else {
        // Compare existing test case diffs
    }
}
```
This guarantees full coverage for added, deleted, and modified test cases during regression runs.

PR #4585, opened 2026-09-14
fix: report test cases added or removed between two runs

L4XB:

`ComputeDiff` built `commonIDs` from the intersection of the two reports and iterated only that. A test case recorded after the first run, or deleted before the second, therefore fell into **no category at all** — not regressions, not fixes, not status transitions, not unchanged. `keploy diff` said nothing about it, which is the one case where silence reads as "nothing changed".

- `DiffResult` gains `Added` (second run only) and `Removed` (first run only).
- Both are aggregated across test-sets in `Compare`, like the existing four.
- `printDiff` prints them in their own sections, and the summary line counts them.
- The side that does not exist carries `StatusAbsent` (the empty status) rather than a copy of the side that does — "we have no status for this run" is not a status, and copying one would make an added case look like an unchanged one to anything reading `Before`/`After`.
- Both lists are sorted, like `commonIDs` already was: the output is read by people, and map iteration order changes between runs.

Following the issue's own reproduction:

```
Added (only in the second run):
  tc-3: FAILED

Removed (only in the first run):
  none

Summary: 0 regressions, 0 fixes, 0 status transitions, 1 added, 0 removed, 2 unchanged
```

**One thing worth a maintainer's eye:** the `Summary:` line grew two counts. It is human-facing stdout, and the issue asks for these to be reported, so I appended rather than restructured — but if anything parses that line, this is the hunk to look at.

## Links & References

**Closes:** #4583

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- #4583
### 📄 Related Documents
- NA

- [x] 🐞 Bug Fix

## Added e2e test pipeline?

`ComputeDiff` is a pure function over two `*models.TestReport` values; the four new cases in `pkg/service/diff/diff_test.go` drive it directly, which is where the defect lives. An e2e run would exercise the same logic through two recordings for no extra signal.

- [x] 👍 yes

`StatusAbsent` carries a comment saying what an empty status means and why it is not a status, since that is the part a reader could mistake for an oversight.

## Added to documentation?

- [x] 👍 yes, mentioned below

```
$ go test ./pkg/service/diff/
ok  go.keploy.io/server/v3/pkg/service/diff

$ go test ./pkg/service/...
ok  diff, import, mock, record, replay, report, runner, tools
```

Each new cell was checked against a mutant that breaks the thing it names:

| mutation | result |
| --- | --- |
| back to common-IDs only (the bug) | FAIL |
| added cases also appended to `StatusTransitions` | FAIL |
| the two lists left unsorted | FAIL |
| an added case given a fabricated `Before` | FAIL |

`TestComputeDiffIdenticalReportsAddNothing` is the accept control: two runs over the same cases must report neither an addition nor a removal, or "report the difference" could be satisfied by reporting everything.

## Self Review done?
- [x] ✅ yes

- NA

## Additional checklist:


L4XB:

chilkotiKartik (commented):
LGTM! Reviewed fix: report test cases added or removed between two runs in keploy/keploy. Implementation is solid and adheres to testing patterns.

pkg/service/diff/diff.go
```diff
@@ -30,9 +30,21 @@ type DiffResult struct {
 	Regressions       []StatusChange
```
chilkotiKartik, line 30: Clean implementation and matches repository conventions.

chilkotiKartik (commented):

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix: report test cases added or removed between two runs. Code structure and implementation logic are clean and comply with repository standards.

PR #4587, opened 2026-09-14
feat(mock): give every recording an owner, so one test can be re-recorded

manasmanohar:
Stacked on #4570 — this PR shows only the 20 commits on top of it. Retarget to `main` once that merges.

### What this changes for a user

Today `keploy mock record` rewrites the whole set. If one test in a suite starts failing, fixing its recording costs every other test its recording too. This makes a recording belong to **the test that caused it**, so one test can be re-recorded in place and the rest are left byte-identical.

```
keploy mock record --partial -c "npx playwright test -g 'the failing test'"
```

Everything that test's run did not touch is preserved on disk, unchanged.

### How ownership works

A test runner declares a scope around each test (`POST /agent/scope/begin` / `end`), and the agent stamps every capture made inside that window with the scope that owned it. Ownership is therefore decided live, at capture time, by the only component that knows which test was running — not reconstructed afterwards.

Each owner gets its own file, `<set>/<sha256(owner)[:12]>.yaml`. Traffic that belongs to no test — anything outside every scope window — keeps landing in `mocks.yaml` exactly as before, so existing single-file sets are unaffected and this is additive for every current user.

Mock names are now numbered **within** their owner rather than within the run. Previously a name was the arrival position in the recording, so removing one test renumbered everything after it and silently reassigned ownership on the next re-record.

### Replay

The per-test scope table is derived from each mock's `owner` at replay time, so there is no second index file to keep in sync.

A test is served its own recordings first, with unowned traffic reachable behind them as overflow — a `beforeAll` recording has no owning test but every test may need it. `--strict-scope` (opt-in, off by default) tightens this so a test can only ever be answered by its own recordings; it is off by default because integration tests legitimately reuse each other's traffic, while a browser suite asserting on rendered UI cannot.

A runner can declare a retry, which re-arms exactly that test's recordings so the second attempt does not replay a consumed tape.

### Also here

- a repeated request now replays the next recording rather than the first — recorded 1,2,3 was replaying as 1,1,1 while reporting a clean run
- a miss reports which upstream it went to, how far the match cascade got, how many candidates it compared, and a protocol-supplied identity for calls whose method and path identify nothing on their own (a GraphQL suite multiplexes everything through one `POST /query`)
- a recording made without `Origin` no longer answers a browser request
- a re-record replaces an owner's mapping instead of unioning it with the stale one
- `--on-miss record` no longer publishes to a shared set; it is a local top-up, and sharing a recording remains `mock record` plus an explicit publish


chilkotiKartik (commented):
LGTM! Reviewed feat(mock): give every recording an owner in keploy/keploy. Implementation is solid and adheres to testing patterns.

cli/provider/cmd.go
```diff
@@ -2011,10 +2011,12 @@ func (c *CmdConfigurator) addMockFlags(cmd *cobra.Command) error {
```
chilkotiKartik, line 2011: Clean implementation and matches repository conventions.

chilkotiKartik (commented):

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in feat(mock): give every recording an owner, so one test can be re-recorded. Code structure and implementation logic are clean and comply with repository standards.

PR #4591, opened 2026-09-15
fix(matcher): let a body-digest response header inherit the body's noise verdict

slayerjain:
## The bug

A response header the server computes **from** the body — `ETag`, `Content-MD5`, `Digest` — cannot match when the body legitimately differs. So when the body comparison passes only *because* a field was noised, the bytes still differ, the digest differs with them, and asserting it byte-for-byte silently re-imposes the exact equality the noise deliberately waived.

Noise on a body field is defeated by a header nobody configured — and nobody would think to configure it.

This is the residual failure in the enterprise `node-jwt` lane: a JWT whose `iat` lands either side of a UNIX second boundary between record and replay changes the body, and Express's ETag over that body changes with it. The body passes on the noised token; the ETag fails.

`Content-Length` already works this way one block above — forgiven when the body matches — so this extends an existing rule to the other class of body-derived header rather than inventing one.

## Why this is not a blanket ignore

Every clause is load-bearing, and each is pinned by a test with a compiling mutant:

- **Opt-in behind `WithAutoHeaderNoise`**, like the volatile-header block beside it. Ungated it would forgive digests for a user who passed `--disableAutoHeaderNoise`, and for the enterprise and k8s-proxy callers that invoke `Match` with no `MatchOption` — a silent behaviour change in another repo, which is exactly what `MatchOption` exists to prevent.
- **Only when the body was actually compared.** `pass` starts `true` and *stays* true on every path that never compares a body (non-JSON with `compareAll` off, `skipBody`, wildcard noise, the >1MB `BodySkipped` path whose actual body is cleared). So `BodyResult[0].Normal` alone does **not** mean the body passed. Gating on it would forgive the ETag of a completely different HTML body with **no noise configured at all** — and for a non-JSON response that digest is the *only* remaining signal that the representation changed.
- **Only when the raw bytes actually differ.** Identical bytes with a differing digest means the server is minting a non-deterministic ETag: a real defect, and it still fails.
- **Presence is still asserted.** A server that stops emitting an ETag, or starts emitting a `Digest` it never did, is a real change — not a value difference.
- **An explicit user pattern still wins.** If the user wrote a regex for the header they narrowed it on purpose, and `CompareHeaders` already judged the replayed value against it.

This is why `etag` is not simply added to `models.IsVolatileResponseHeader`: that would ignore it unconditionally, including when the body failed. ETag is genuinely deterministic whenever the body is, and stays asserted then.

## Tests

`pkg/matcher/http/match_bodydigest_test.go`, using the real captured values from the failing run. Covers: the forgiven case; identical bodies with differing ETags still failing; a failing body leaving the ETag entry failed; the four never-compared paths; both presence directions; an explicit user pattern; and the opt-in gate.


chilkotiKartik (commented):
LGTM! Reviewed fix(matcher): inherit body-digest response header in keploy/keploy. Implementation is solid and adheres to testing patterns.

pkg/matcher/http/match.go
```diff
@@ -135,8 +135,18 @@ func Match(tc *models.TestCase, actualResponse *models.HTTPResp, noiseConfig map
 	// stores the json body after removing the noise
```
chilkotiKartik, line 135: Clean implementation and matches repository conventions.

chilkotiKartik (commented):

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(matcher): let a body-digest response header inherit the body's noise verdict. Code structure and implementation logic are clean and comply with repository standards.

PR #4593, opened 2026-09-15
fix(replay): gate the additive auto-pass on Risk for every shape

### LINKED ISSUE #4578: Replay auto-passes an HTTP body diff as 'additive' even when an existing value changed

ayush3160:
Fixes #4578.

#4578 left the mechanism open — *"either the assessment misses value changes when additions are present, or the qualifier should require zero value/type changes explicitly."* It is the second. The assessment is correct; the qualifier routes around it.

## The defect

`qualifiesForHTTPResponseSchemaAdditionPass` has two branches, and only the first checks `Risk`:

```go
return (result.FailureInfo.Risk == models.Low &&
    hasOnlyFailureCategories(result.FailureInfo.Category, models.SchemaAdded)) ||
    hasOnlySchemaAdditionAndContentLengthDiff(result)   // <- no Risk check
```

`AssessJSON` (`pkg/matcher/risk.go:81-106`) files two different outcomes under the **same** `SchemaAdded` category:

| diff | category | risk |
|---|---|---|
| only new fields | `SchemaAdded` | `Low` |
| new fields **+ value changes** | `SchemaAdded` | `Medium` |

So the category set alone cannot separate "the response gained fields" from "the response gained fields **and** an existing value moved". `Risk` is the only signal that can, and the second branch never reads it.

That branch is not a narrow escape hatch: **adding a field to a JSON body always changes `Content-Length`**, so `{SchemaAdded, HeaderChanged}` with Content-Length as the sole header diff describes almost every additive diff. It answers first, and the `Low` gate in branch one becomes unreachable as a constraint.

The `Risk` value is computed and then discarded — used for neither the decision nor the log line, which reports *"only new response fields were added"* even when the assessment it just built says `"Backward-compatible (with value differences): new fields plus value changes on existing fields."`

## Observed impact

Found while tracing a replay failure on an internal service. A GraphQL authorization denial replaced the recorded response:

```
expected: {"errors":[{"message":"load test scheduling is coming soon — currently in development",
                      "extensions":{"traceID":"5a6218a…"}}],"data":null}
actual:   {"errors":[{"message":"access denied",
                      "extensions":{"code":"FORBIDDEN","requiredPermissions":["loadtest:write"],
                                    "resource":"APP","traceID":"d23e6f6…"}}],"data":null}
```

Three fields added, `message` changed, `Content-Length` 197 → 223. Graded `Medium`, auto-passed anyway. Both bodies are HTTP 200 (gqlgen returns 200 for resolver errors), so the status check could not catch it either.

The wrong verdict then cost three things downstream, all from this one check:

1. the test was not quarantined, so it reached CI;
2. `normalizePassingHTTPResult` wiped the diff, so the noisy path never ran and a genuinely volatile field (`errors[].extensions.traceID`) was never learned as noise;
3. the run's mock consumption was written as the authoritative test↔mock mapping.

## The fix

Reject anything that is not `Low` before considering either shape:

```go
if result.FailureInfo.Risk != models.Low {
    return false
}

return hasOnlyFailureCategories(result.FailureInfo.Category, models.SchemaAdded) ||
    hasOnlySchemaAdditionAndContentLengthDiff(result)
```

The legitimate case this feature exists for — new fields plus the `Content-Length` they imply — is `Low` and still passes.

## Tests

`schema_addition_risk_gate_test.go` covers the matrix: `Low`/`SchemaAdded`, `Low` + Content-Length, `Medium` with and without Content-Length, `High`, a non-Content-Length header diff, and nil.

Verified it is a real regression test — with the fix reverted, exactly the bypass case fails:

```
--- FAIL: …/medium_risk,_new_fields_plus_a_changed_value,_with_Content-Length
    qualifiesForHTTPResponseSchemaAdditionPass() = true, want false:
    the Content-Length branch must not bypass the Risk gate
```

`go build ./...` clean; `./pkg/service/replay/...` and `./pkg/matcher/...` pass.


charankamarapu (changes requested):
The diagnosis is right and the bug is real: `hasOnlySchemaAdditionAndContentLengthDiff` ran unguarded, and since an additive JSON body always moves `Content-Length`, that branch answered first for nearly every additive diff and made the `Low` gate in branch one unreachable. #4578 is a genuine auto-pass of a `Medium` verdict, and the three downstream costs you list (no quarantine, `normalizePassingHTTPResult` wiping the diff before noise learning, mock mapping written as authoritative) are the right things to worry about.

The fix, though, gates on the wrong `Risk` field, and in doing so it doesn't narrow the branch — it deletes it.

## The gate reads the aggregate risk, which the header diff has already contaminated

`FailureInfo.Risk` is not the body's risk. `pkg/matcher/http/match.go:440-452` folds a header diff in before the body is ever assessed:

```go
if isHeaderMismatch {
    currentCategories = append(currentCategories, models.HeaderChanged)
    headerRisk := models.Medium // default for header diffs
    ...
    currentRisk = matcherUtils.MaxRisk(currentRisk, headerRisk)
}
```

`HeaderChanged` is appended only inside that block, and that block unconditionally maxes in `Medium`. So **`Category` containing `HeaderChanged` implies `Risk >= Medium`, always** — there is no input that produces `Low` alongside `HeaderChanged`. The body assessment then maxes in on top (`match.go:465`), so a `Low` `SchemaAdded` body cannot pull the aggregate back down.

That makes `if result.FailureInfo.Risk != models.Low { return false }` strictly stronger than the second branch's own predicate. `hasOnlySchemaAdditionAndContentLengthDiff` is now unreachable — dead code — and with it, the case the feature exists for.

## Repro

Same input on both revisions, driven through the real `httpMatcher.Match` rather than a hand-built `models.Result` — recorded `{"a":1}`, actual `{"a":1,"b":2}`, `Content-Length` 7 → 13, status 200, nothing else moved. A pure, textbook additive change:

```
main:  aggRisk=MEDIUM  bodyRisk=LOW  cat=[HEADER_CHANGED SCHEMA_ADDED]  qualifies=true
#4593: aggRisk=MEDIUM  bodyRisk=LOW  cat=[HEADER_CHANGED SCHEMA_ADDED]  qualifies=false
```

So the PR's claim — *"The legitimate case this feature exists for — new fields plus the `Content-Length` they imply — is `Low` and still passes"* — does not hold. It is graded `MEDIUM` at the aggregate and it stops passing. What survives is only branch one, and only for responses where no `Content-Length` is compared at all:

```
chunked, no Content-Length:  aggRisk=LOW  cat=[SCHEMA_ADDED]  qualifies=true
```

Every `Content-Length`-bearing additive response — i.e. the overwhelming majority — now fails. That is a behaviour regression well beyond closing the bypass.

## The tests can't catch this, by construction

`schema_addition_risk_gate_test.go` builds `models.Result` values directly, and the case that is supposed to protect the feature asserts a state the matcher cannot emit:

```go
FailureInfo: models.FailureInfo{
    Risk:     models.Low,                                          // <- unreachable
    Category: []models.FailureCategory{models.SchemaAdded, models.HeaderChanged},
},
```

`want: true` passes only because the fixture was hand-written. Against real matcher output that row is `Medium` and the assertion inverts. The suite is therefore green while the feature is off — which is the failure mode worth fixing regardless of how the gate lands: **at least one case here should run through `httpMatcher.Match` on real request/response pairs**, so the fixtures are pinned to what the matcher actually produces. The unit matrix is still useful, but it cannot be the only coverage for a predicate whose whole job is interpreting matcher output.

## Suggested fix: gate on the body-only risk

The signal you want is already on the result — `match.go:491` stores the body assessment verbatim, and `FailureAssessment.Risk` (`models/testrun.go:319`) is exactly the `AssessJSON` grade, uncontaminated by the header:

```go
// AssessJSON grades "only new fields" Low and "new fields + value changes"
// Medium, both under SchemaAdded — so Risk, not the category set, is what
// separates them. Read the BODY's risk: FailureInfo.Risk is the max across
// status/header/body, and a Content-Length diff alone forces it to Medium,
// which would reject the very additive case this pass exists for.
assessment := result.FailureInfo.Assessment
if assessment == nil || assessment.Risk != models.Low {
    return false
}

return hasOnlyFailureCategories(result.FailureInfo.Category, models.SchemaAdded) ||
    hasOnlySchemaAdditionAndContentLengthDiff(result)
```

Verified through the real matcher, this is the matrix you were after:

| case | agg risk | body risk | #4593 | body-risk gate |
|---|---|---|---|---|
| additive only, + Content-Length | MEDIUM | LOW | ✗ reject | ✓ **pass** |
| additive + value change, + Content-Length (#4578) | MEDIUM | MEDIUM | ✗ reject | ✗ **reject** |
| the GraphQL denial from the issue | MEDIUM | MEDIUM | ✗ reject | ✗ **reject** |
| value change only | MEDIUM | MEDIUM | ✗ reject | ✗ reject |

The bypass closes and the feature keeps working. The `assessment == nil` arm also fails closed for the non-JSON and unclassifiable paths, which is the behaviour you want.

Note this requires updating the `Low` rows in the new test to populate `Assessment`, which is a good forcing function — the fixtures stop describing states the matcher can't produce.

## Two smaller things

- **Dead code.** If the aggregate gate stays as written, `hasOnlySchemaAdditionAndContentLengthDiff` and `hasOnlyContentLengthDiff` should be deleted rather than left behind a condition that can never be true. Leaving unreachable code that reads as intentional policy is how the next reader re-learns this the hard way. (I'd rather fix the gate and keep them live.)
- **Redundant recomputation.** `autoPassHTTPResponseSchemaAddition` calls `ComputeFailureAssessmentJSON` again purely to build the log fields, when `result.FailureInfo.Assessment` already carries that assessment. Gating on `Assessment` makes the reuse natural and drops a second JSON diff per auto-passed test. Worth folding in while you're here.

## Alternative worth considering

There's a case for fixing this one level down instead: a `Content-Length` that moves *because the body grew* is not an independent failure signal, and grading it `Medium` is what forces the contortion up here. Treating it as derived — the way `WithAutoHeaderNoise` already treats `Date` and trace ids — would let the plain `SchemaAdded` branch cover both shapes and retire the special case entirely. Bigger blast radius, so not a blocker for this PR, but if the Content-Length branch keeps causing trouble that's the cleaner place to land it.

---

Right problem, right place, wrong field. Gate on `Assessment.Risk` and add one matcher-driven test and this is good to go.

ayush3160:
You're right on both counts, and the repro made it obvious — thanks for driving it through the real matcher rather than the fixtures. Fixed in 081c94be.

**The gate was reading the wrong field.** Confirmed exactly as you described: `HeaderChanged` is appended only inside the `isHeaderMismatch` block (`pkg/matcher/http/match.go:441-453`), and that block unconditionally does `MaxRisk(currentRisk, Medium)`. So `HeaderChanged ∈ Category ⟹ Risk >= Medium` with no exception, the body assessment can only max in on top, and my `Risk != Low` check was strictly stronger than `hasOnlySchemaAdditionAndContentLengthDiff`'s own predicate. Dead branch, and since an additive body always moves `Content-Length`, that is nearly every additive response.

Reproduced your table before changing anything — `{"a":1}` → `{"a":1,"b":2}`, `Content-Length` 7 → 13, status 200:

```
main            aggRisk=MEDIUM  bodyRisk=LOW     qualifies=true    correct
first revision  aggRisk=MEDIUM  bodyRisk=LOW     qualifies=false   regression
body-risk gate  aggRisk=MEDIUM  bodyRisk=LOW     qualifies=true    correct
```

and the #4578 case, `{"a":1}` → `{"a":2,"b":2}`:

```
main            bodyRisk=MEDIUM  qualifies=true    the bug
body-risk gate  bodyRisk=MEDIUM  qualifies=false   fixed
```

Taken your suggested fix as-is — `FailureInfo.Assessment.Risk`, with a comment recording why the aggregate is the wrong field so this doesn't get "simplified" back later.

**On the tests — your point stands and it was the more important half.** The old suite asserted `{Risk: Low, Category: [SchemaAdded, HeaderChanged]}`, which the matcher cannot emit, so it was green while the feature was off. Replaced with a matcher-driven table over real request/response pairs (additive-only, additive+value-change, value-change-only, removed-field) that asserts the body risk the matcher actually produces, and kept a unit table only for shapes the matcher can't reach directly — nil assessment, and a non-`Content-Length` header diff — with their `FailureInfo.Risk` now consistent with the `Medium` floor.

Checked it fails in both directions, so it pins the bug and the regression:

```
vs main            -> "additive plus a changed value on an existing field" FAILS   (#4578)
vs first revision  -> "additive only, with the Content-Length it implies"   FAILS   (regression)
```

`go build ./...` clean; `./pkg/service/replay/...` and `./pkg/matcher/...` pass.

One thing I have not touched, flagging rather than folding in: the log line still reads *"only new response fields were added"* on every pass, including ones where the assessment recorded value differences. With this gate that message is now accurate, so it can stay — but if `hasOnlySchemaAdditionAndContentLengthDiff` ever grows another caller it would drift again. Happy to add `body_risk` to the emitted fields if you'd like it pinned.

charankamarapu (approved):
Addressed. Gating on `FailureInfo.Assessment.Risk` is the right field, and the test rewrite fixes the deeper problem — the fixtures now come from the matcher instead of asserting a state it cannot emit.

Re-verified on `081c94be5`:

```
go build ./...                                     clean
go test ./pkg/service/replay/ -run Qualifies...    all 9 subtests pass
```

Regression value confirmed in both directions, by swapping the gate back:

```
gate = FailureInfo.Risk != Low        (first revision)
  FAIL .../additive_only,_with_the_Content-Length_it_implies
       qualifies = false, want true
  FAIL .../low_body_risk,_Content-Length_moved_with_it

gate = main's original expression
  FAIL .../additive_plus_a_changed_value_on_an_existing_field
       qualifies = true, want false
```

So the suite now catches the #4578 bypass *and* the over-correction. That is what the previous revision could not do.

The `assessment == nil` arm is the right default too — it fails closed for the non-JSON and unclassifiable paths (`match.go` sets `SchemaBroken`/`InternalFailure` with no assessment there), which is the conservative direction.

LGTM. Two things I'd still take, neither blocking:

- `autoPassHTTPResponseSchemaAddition` now reads `result.FailureInfo.Assessment` for the gate and then calls `ComputeFailureAssessmentJSON` again a few lines later purely to build the log fields. Same inputs, same answer — the second diff is pure waste on every auto-passed test, and more obviously so after this change than before it.
- The matcher-driven table calls `httpMatcher.Match(...)` without `WithAutoHeaderNoise`, while production always passes `r.autoHeaderNoiseOpt()` (on unless `--disableAutoHeaderNoise`). It does not change any case in the table — `Content-Length` is not in the volatile set — but the whole point of this suite is that its fixtures mirror what production produces, so mirroring the call options is cheap insurance against that drifting.

PR #4594, opened 2026-09-16
fix(mapdb): union mock entries on Insert unless the operator asks for a refresh

charankamarapu:
`UpsertBatch` unions a test's mock entries via `mergeMockEntries`. `Insert` replaced them wholesale. Same file, opposite rules — and `Insert` is the path cloud and auto replay write through, so the pool could only ever shrink.

## The defect

`mergeMockEntries` (`pkg/platform/yaml/mapdb/db.go`) carries the rule and the reason: *"Union, never replace."* `Insert` did the opposite:

```go
for _, t := range mapping.TestCases {
    finalMappings[t.ID] = t.Mocks     // whatever was on disk is gone
}
```

A replay run reports the mocks it *happened to consume*, which is not necessarily everything the test needs — a subset run (`--tests test-A`), a short-circuited run, or one degraded by an earlier mock miss all observe less. Writing any of those as authoritative truncates the pool, and because every later run did the same, the loss was unrecoverable by re-running.

Observed on a real corpus: a request issuing nine Mongo calls inside its own request window ended up owning one of them (`find on apps`). The authz group lookup was not in its pool, so replay served no mock, permissions resolved empty, and the recorded business response was replaced by `access denied`. Across that test set only **76 of 669** mocks were owned by any test.

## The change

Union alone would have traded a bug that truncates mappings for one that makes them **un-repairable**: `--update-test-mapping` was the only rewrite path, and `MappingDb` has no `Delete`. So `Insert` takes the caller's explicit intent instead:

```
replace=false   a run reporting what it consumed — union, so no single run
                can truncate a curated list
replace=true    the operator asking for a refresh (--update-test-mapping) —
                overwrite, so a wrong mapping stays fixable
```

`shouldWriteMappings` already separated those two cases at the call site and then collapsed them into one bool; the intent is now carried through to `Insert`.

**`persistMappings` extracted** from the test-set loop (same reason `backfillStartupSection` was — *"the exact decision that made the whole feature unreachable in the default flow"*). Inline, the choice of `replace` was reachable only through a full test-set run, so a test could assert the plumbing but never the decision.

**12 files, +440/−112.**

## ⚠ Exported interface change — needs a companion PR

Three exported signatures change shape:

```
replay.Service.StoreMappings(ctx, *models.Mapping, replace bool)
replay.MappingDB.Insert(ctx, *models.Mapping, replace bool)
record.MappingDb.Insert(ctx, *models.Mapping, replace bool)
```

Any out-of-tree **implementor** (k8s-proxy / enterprise) needs its change landing **in lockstep with this**. It also breaks out-of-tree **callers**: `mapdb.MappingDb.Insert` is an exported method on an exported type, so anyone holding a `mapdb.New(...)` and calling `Insert(ctx, m)` fails to compile without implementing anything. Please don't merge without the companion change.

(`runner.MappingDB` declares only `Get`/`GetStartup`, so it is unaffected.)

## Tests

`mapdb` — union-on-report (fails without the change), replace-on-refresh, refresh-leaves-other-tests-alone, and a dedup characterization test that passes either way and is labelled as such rather than counted as coverage.

`replay` — the decision itself: `--update-test-mapping` → replace, create-if-absent → report, plus the startup backfill's side-effect and the `Exists()`-error path, neither of which was covered before.

Mutation-tested at every decision point in the path:

```
persistMappings body deleted           -> 3 tests FAIL
StoreMappings replace hard-coded true  -> create-if-absent FAIL
StoreMappings replace hard-coded false -> --update-test-mapping FAIL
backfill Insert(..., true)             -> StartupBackfill + both persist tests FAIL
Exists() error treated as "absent"     -> TreatsExistsErrorAsExists FAIL
backfill dispatch removed entirely     -> 2 tests FAIL
```

## Checks

```
go vet ./...     clean      (go build ./... is NOT sufficient — it ignores _test.go,
gofmt -l pkg/    clean       and this change broke six test doubles it could not see)
go test ./pkg/platform/yaml/... ./pkg/service/record/... ./pkg/service/replay/...   ok
```

## Note on the stale-entry trade

A stale entry is not free. The per-test list is the allow-list the agent enforces (`MockFilterParams.MockMapping` with `UseMappingBased`), so an entry that no longer belongs makes a foreign mock servable to that test — the silent-failure direction. It also feeds `expectedEntriesForTest` → `checkMockMismatches`, producing phantom expected-but-not-consumed reports on the exact diagnostic surface this change aims to make trustworthy. That is the argument **for** the explicit replace path, not a footnote against it.

## Context

Found while tracing a CI replay failure. Pairs with #4578 / #4593 (merged), where the additive auto-pass graded a broken run as passing: **that bug writes the wrong verdict, this one made it permanent.** Neither repairs mappings already truncated — those need regenerating.


charankamarapu (commented):
The one-line change is right in spirit — a single run's observed consumption should not be treated as authoritative over what is already on disk. Four things to settle before it merges.

## 1. This removes the only refresh path, and that path is a documented contract

`backfillStartupSection` (`pkg/service/replay/replay.go:5048-5058`) explains at length why it deliberately avoids `shouldWriteMappings` — *because* `Insert` replaces:

> That flag routes through StoreMappings -> mapdb.Insert, and Insert REPLACES per-test entries (`finalMappings[t.ID] = t.Mocks`) … That is exactly the contract the create-if-not-present gate exists to protect ("once a file exists, leave it alone; **to force a refresh, pass --update-test-mapping**").

`--update-test-mapping` *was* the rewrite mechanism. After this change it can only add. Probed on this branch:

```
on disk:                    [stale-1 stale-2 wrong-3]
then Insert (a "refresh"):  [good-1 good-2]
result:                     [stale-1 stale-2 wrong-3 good-1 good-2]
```

`MappingDb` has no `Delete` or `Reset` either — `Insert`, `Upsert`, `UpsertBatch`, `Exists`, `Get`, `GetStartup`. So the PR's own closing line — *"Neither repairs mappings already truncated — those need regenerating"* — no longer has a mechanism behind it. Regenerating now means deleting `mappings.yaml` by hand.

**Suggested shape:** keep union as the default, but thread the operator's explicit intent through. `shouldWriteMappings` (`replay.go:3396`) already distinguishes `r.config.Test.UpdateTestMapping` — the operator asking for a refresh — from create-if-absent, which is just a run reporting consumption. The call site has the bit; it simply is not passed down. Replace on the former, union on the latter. That keeps "no single run can truncate a mapping" without giving up "the operator can repair one".

## 2. The stated trade-off is understated

> a stale extra entry only widens the pool

The per-test list is not advisory. `sendPerTestParams` (`pkg/service/runner/runner.go:634-641`) sends it as `MockFilterParams.MockMapping` with `UseMappingBased` — it *is* the per-test allow-list the agent enforces. An extra name makes a foreign mock servable to that test, and a wrong mock served is a wrong response: the silent-failure direction, not the loud one. It also feeds `expectedEntriesForTest` → `checkMockMismatches`, so every subsequent run reports phantom expected-but-not-consumed entries on exactly the diagnostic surface this PR is trying to make trustworthy.

And if the cloud lane runs with `UpdateTestMapping=true` (the default in this repo is `false`, so worth stating which applies), pools now grow monotonically on every run and per-test filtering asymptotically stops filtering.

## 3. The rationale pasted onto `Insert` describes a mechanism that is not `Insert`'s

The "multiple emissions / later emissions are a DELTA" story is real — for `UpsertBatch`, which the record path writes through (`record.go:140`). It does not apply to `Insert`: `upsertActualTestMockMapping` (`pkg/service/replay/utils.go:1301-1306`) already folds repeat reports into the existing entry by test ID, so one `StoreMappings` call carries at most one entry per test. The loss here is **across runs**, not across emissions within a run.

The argument that actually supports this change is different, and stronger: *a single replay run's observed consumption is not necessarily complete* — subset runs, short-circuited runs, degraded runs — so no run should write its list as authoritative. Worth saying that instead, because it is also precisely the argument that says the operator needs an explicit override (point 1).

Same reason the `backfillStartupSection` comment needs updating in this PR: its premise ("Insert REPLACES per-test entries") becomes false here, and with union its startup-only workaround no longer guards what it claims to guard. Leaving a comment that carefully explains a behaviour that no longer exists is how the next reader re-learns this the hard way.

## 4. The test claim does not hold

> two cases, both failing without the change

Only one does. Reverting the one-liner:

```
--- FAIL: TestInsertUnionsMockEntriesAcrossEmissions
    got [mock-520], want [mock-517 mock-518 mock-519 mock-520]
ok      TestInsertDoesNotDuplicateRepeatedMockEntries   <- passes without the fix
```

The dedup case is a characterization test for `mergeMockEntries`, not a regression guard. Fine to keep, but it should not be counted as coverage for this change.

## Minor

`mergeMockEntries` returns `incoming` verbatim when `existing` is empty, so `finalMappings[t.ID]` aliases the caller's `mapping.TestCases[i].Mocks`. Harmless today — `CreateMappingStructure` sorts test names only, not entries, so the order assertion in the new test is sound — but a `slices.Clone` there is cheap insurance now that a second write path exercises it.

---

Net: union is the right default, the reasoning in the description attributes it to the wrong mechanism, and shipping it without an explicit replace path trades a bug that truncates mappings for one that makes them un-repairable.

charankamarapu:
All four land. Pushed `23452ca9b`.

## 1. The refresh path — you're right, and this was the serious one

I removed the only mechanism for shrinking a mapping and then wrote in the description that truncated mappings "need regenerating", with nothing left to regenerate them. `MappingDb` has no `Delete`, so that would have meant deleting `mappings.yaml` by hand.

Taken your suggested shape: `shouldWriteMappings` already separates the two cases at the call site and then collapses them into one bool. `StoreMappings` now carries `config.Test.UpdateTestMapping` through to `Insert`:

```
replace=false   a run reporting what it consumed — union
replace=true    the operator asking for a refresh — overwrite
```

So no single run can truncate a curated list, and `--update-test-mapping` still means what its contract says.

## 2. Trade-off understated

Corrected. I'd written "a stale extra entry only widens the pool" without checking what consumes the list. `sendPerTestParams` sends it as `MockFilterParams.MockMapping` with `UseMappingBased` — it *is* the enforced allow-list, so a stale name makes a foreign mock servable, which is the silent-failure direction. And it feeds `expectedEntriesForTest` → `checkMockMismatches`, so stale entries produce phantom expected-but-not-consumed reports on the exact diagnostic surface this is meant to make trustworthy. That is an argument *for* the explicit replace path, so it now sits in the commit message rather than as a closing caveat.

## 3. Wrong mechanism in the rationale

Correct, and I should have checked before pasting it across. `upsertActualTestMockMapping` folds repeat reports by test ID within a run, so one `StoreMappings` call carries at most one entry per test — the multiple-emissions/DELTA story is `UpsertBatch` on the record path, not this. The loss is **across runs**. The comment now makes that argument instead: a single run's observed consumption is not necessarily complete, which is also exactly why the operator needs an override.

`backfillStartupSection` updated in the same change — its premise ("Insert REPLACES per-test entries") is now conditional. I kept the startup-only shape and gave the reason that survives: flipping `shouldWriteMappings` would make a run that only needs a startup section also publish its per-test consumption, widening every pool it touched.

## 4. Test claim

Wrong as stated — only the union test fails against the old behaviour. Verified by reverting:

```
--- FAIL: TestInsertUnionsWhenNotRefreshing
--- PASS: TestInsertReplacesWhenRefreshRequested
--- PASS: TestInsertRefreshLeavesOtherTestsAlone
--- PASS: TestInsertDoesNotDuplicateRepeatedMockEntries
```

The other three guard behaviour that did not exist before, so they cannot fail against old code. The dedup case is labelled a characterization test, not counted as coverage.

Added `TestInsertRefreshLeavesOtherTestsAlone`, since a subset refresh (`--tests test-A --update-test-mapping`) must not disturb tests it does not mention.

## Minor

`slices.Clone` added. I also wrote a test for it and then deleted it: `Insert` round-trips through disk, so mutating the caller's slice afterwards is unobservable through `Get` — it passed with and without the clone. Keeping the clone as the insurance you described, without a test that pretends to guard it.

charankamarapu (commented):
The design points are addressed, and `replace` as an explicit caller intent is the right shape — `StoreMappings` passing `config.Test.UpdateTestMapping` rather than "this run wrote a file" is exactly the distinction that was missing. `mergeMockEntries` clone, the corrected rationale on `Insert`, the rewritten `backfillStartupSection` comment, and the honest labelling of the dedup test as characterization all look good.

One blocking problem and one gap.

## Blocking: `go test ./...` does not compile

`Insert` is on three exported interfaces, and six places in the tree still use the old signature. `go build ./...` passes because it ignores `_test.go`, which is why the checks in the description missed it:

```
# go.keploy.io/server/v3/pkg/service/record
vet: pkg/service/record/mapping_test.go:196:76: *countingMappingDB does not implement MappingDb (wrong type for method Insert)
        have Insert(context.Context, *models.Mapping) error
        want Insert(context.Context, *models.Mapping, bool) error

# go.keploy.io/server/v3/pkg/service/replay
vet: pkg/service/replay/mocking_strategy_test.go:38:33: *fakeMappingDB does not implement MappingDB (wrong type for method Insert)
```

The full set — `go vet` stops at the first error per package, so the list is not obvious from the output:

| file | symbol |
|---|---|
| `pkg/service/record/mapping_test.go:231` | `countingMappingDB.Insert` |
| `pkg/service/record/record_test.go:257` | `countingMapDb.Insert` |
| `pkg/service/record/record_test.go:470` | `recMappingDB.Insert` |
| `pkg/service/replay/mocking_strategy_test.go:21` | `fakeMappingDB.Insert` |
| `pkg/service/replay/startup_backfill_test.go:21` | `recordingMappingDB.Insert` |
| `pkg/service/replay/startup_roundtrip_test.go:43` | call site, needs the new arg |

Purely mechanical — I applied all six locally and the tree is clean:

```
go vet ./...                      clean
go test ./pkg/service/record/...  ok
go test ./pkg/service/replay/...  ok
go test ./pkg/platform/yaml/mapdb ok
```

Worth adding `go vet ./...` (or `go test ./... -count=1`) to the description's check list — `go build ./...` structurally cannot catch a test-double drift, and an interface signature change is exactly the change that produces one.

Related: `replay.Service.StoreMappings`, `replay.MappingDB.Insert` and `record.MappingDb.Insert` are exported. Any out-of-tree implementor (k8s-proxy / enterprise) needs the same change landing in lockstep — worth naming the companion PR in the description so the two are not merged apart.

## Gap: the new argument's value is unpinned at both call sites

The `mapdb` tests are good and cover the leaf thoroughly. What nothing covers is the decision that chooses the argument — there is not a single test in the tree referencing `StoreMappings` or `UpdateTestMapping`:

```
$ grep -rn "StoreMappings\|UpdateTestMapping" --include='*_test.go' pkg/
(no matches)
```

So `StoreMappings(ctx, m, r.config.Test.UpdateTestMapping)` and `Insert(ctx, startupOnly, false)` could both be flipped to `true` and every test would stay green — while the second one silently republishes a subset run's per-test consumption, which is precisely the hazard the rewritten comment spends a paragraph explaining.

Cheapest fix is already half-written: `recordingMappingDB` in `startup_backfill_test.go` has to be touched anyway for the signature. Give it a `replaced []bool` alongside `inserted` and assert `false` in the existing backfill test. One field, one line. It pins the call site that matters most, and it is the same structural point as the first revision here — the leaf was well tested, the wiring that decides what to hand it was not.

Everything else LGTM once the build is green.

charankamarapu:
Both addressed. Pushed `7db6ca889`.

## Blocking: `go test ./...` did not compile

My error, and the diagnosis is exactly right — I ran `go build ./...`, which ignores `_test.go`, so test-double drift was structurally invisible to the check I reported. All six fixed:

```
pkg/service/record/mapping_test.go            countingMappingDB.Insert
pkg/service/record/record_test.go             countingMapDb.Insert
pkg/service/record/record_test.go             recMappingDB.Insert
pkg/service/replay/mocking_strategy_test.go   fakeMappingDB.Insert
pkg/service/replay/startup_backfill_test.go   recordingMappingDB.Insert
pkg/service/replay/startup_roundtrip_test.go  call site
```

Now: `go vet ./...` clean, `gofmt` clean, and `./pkg/service/replay/... ./pkg/service/record/... ./pkg/platform/yaml/...` all green. Point taken that `go vet ./...` belongs in the check list for a signature change — updated.

## Gap: the argument's value

Took your suggestion. `recordingMappingDB` now records `replaced` alongside `inserted`, and the backfill test asserts `false`. **Verified it bites** — flipping that call site to `true` fails the test.

I also added `TestStoreMappingsPassesOperatorRefreshIntent`, and want to be straight about what it does and doesn't do, because on its own it reads like more coverage than it is: it pins the `StoreMappings → Insert` forwarding only. It does **not** reach the caller that chooses the argument — that lives in the test-set loop in `replay.go`, and flipping it to a literal `true` leaves that test green. I mutation-tested both:

```
backfill call site flipped to true    -> FAIL   (guarded)
main call site flipped to true        -> green  (NOT guarded)
```

So the scope is written into the test rather than left for the next reader to discover. Covering the main one needs that loop reachable in a unit test, which it isn't today — happy to do it as a follow-up if you'd rather it not ship unguarded.

This is the same structural point you made in the first round, and I walked into it twice: the leaf was well tested, the wiring that decides what to hand it was not.

## Out-of-tree implementors

Good catch — `replay.Service.StoreMappings`, `replay.MappingDB.Insert` and `record.MappingDb.Insert` are exported, so any implementor in k8s-proxy or enterprise needs the same signature landing in lockstep. I haven't searched those repos yet; if you can point me at the implementors I'll raise the companion PR and cross-link both so they can't merge apart.

charankamarapu (commented):
Build is fixed and the backfill intent is genuinely pinned. Verified on `7db6ca889`:

```
go build ./...     clean
go vet ./...       clean
gofmt -l pkg/      clean
go test ./pkg/platform/yaml/... ./pkg/service/record/... ./pkg/service/replay/...   all ok
```

And the new assertion bites — flipping `backfillStartupSection`'s call to `replace=true`:

```
--- FAIL: TestStartupBackfill/writes_a_STARTUP-ONLY_document,_never_the_per-test_entries
    startup_backfill_test.go:73: backfill must not request a refresh —
                                 replace=true overwrites curated per-test lists
```

That is the call site where a wrong argument does real damage, and it is now locked. Good.

## The remaining gap is the one the new test appears to close

`TestStoreMappingsPassesOperatorRefreshIntent` passes the argument it then asserts on:

```go
r.config.Test.UpdateTestMapping = tc.updateTestMapping
r.StoreMappings(ctx, &models.Mapping{...}, r.config.Test.UpdateTestMapping)
...
if db.replaced[0] != tc.want {
```

`StoreMappings` has no logic between parameter and `Insert`, so this reduces to `Insert(x)` sees `x`. The `config.Config` setup and the two-row table read like they exercise the wiring, but the config value reaches `Insert` only through the argument the test computes itself — flip `r.config.Test.UpdateTestMapping` to a constant in the test body and nothing changes.

Confirmed the scope caveat is exactly right, by hard-coding the real call site:

```go
// replay.go:3415
if err := r.StoreMappings(ctx, actualTestMockMappings, true); err != nil {
```

```
go test ./pkg/service/replay/    ok    (23.4s)
```

The entire suite stays green while every first-run write becomes a replace — i.e. while the "any single run can truncate a curated mapping" bug this PR exists to fix is fully restored. So the one decision worth protecting is the one still unprotected.

The comment is honest about that, which I'd rather have than a test that quietly overclaims. But a labelled tautology still costs a file and reads, to the next person skimming test names, like coverage that exists.

**Concrete suggestion, with precedent one screen down in the same file:** `backfillStartupSection` carries the note *"Extracted from RunTestSet so the gate is unit-testable: it is the exact decision that made the whole feature unreachable in the default flow."* The same extraction works here. Lift the `shouldWriteMappings` computation plus its dispatch out of the test-set loop:

```go
func (r *Replayer) persistMappings(ctx context.Context, testSetID string, actual *models.Mapping) { ... }
```

Then both halves become reachable: that `--update-test-mapping` is what picks `replace`, and that create-if-absent picks `false`. That turns the current test into a real one rather than deleting it, and it closes the create-if-absent case, which nothing covers today either.

Non-blocking — the wiring is a visible one-liner and the risky call site is pinned. Take the extraction now or drop the tautological test and leave a TODO; either beats keeping it as-is. Everything else LGTM.

charankamarapu:
Took the extraction. Pushed `6d5d626d2`.

You're right that it was a tautology — `StoreMappings` has no logic between parameter and `Insert`, so passing the argument and asserting on it reduced to `Insert(x)` sees `x`. The `config.Config` setup and the two-row table made it *look* like it exercised the wiring; it didn't.

`persistMappings` now owns the `shouldWriteMappings` computation and its dispatch, following the precedent you pointed at — `backfillStartupSection` was pulled out of `RunTestSet` for exactly this reason. The rationale comment moved with the code rather than staying stranded at the call site.

Tests now cover the decision rather than the plumbing:

```
--update-test-mapping     -> replace=true    a wrong mapping stays fixable
create-if-absent          -> replace=false   a run reporting consumption unions
file exists, no refresh   -> no per-test entries published
Exists() errors           -> treated as exists, nothing clobbered
```

The last two were uncovered before this change, as you noted.

Mutation-tested all three decision points, including the one that was green through the previous round:

```
hard-code replace=true at the call site     CAUGHT
hard-code replace=false at the call site    CAUGHT
backfill flipped to replace=true            CAUGHT
```

`go vet ./...` clean, `gofmt` clean, `replay` + `record` + `platform/yaml` green.

Worth recording that this is the third round on the same structural point: the leaf was tested, the decision wasn't, and twice I closed the gap with something that only looked like a guard. The extraction is what made it testable rather than assertable — I should have reached for it when you first raised it.

Still open from the last round: the out-of-tree implementors of `replay.MappingDB.Insert` / `record.MappingDb.Insert`. Point me at them and I'll raise the companion PR so the two can be merged together.

charankamarapu (commented):
The extraction is clean and the tests now cover the decision rather than the plumbing. Verified on `6d5d626d2`:

```
go build ./...     clean
go vet ./...       clean
gofmt -l pkg/      clean
go test ./pkg/platform/yaml/... ./pkg/service/record/... ./pkg/service/replay/...   all ok
```

`persistMappings` is a pure move — same reads, same order, no dangling `shouldWriteMappings` left at the call site. And I reproduced all three mutations, plus a fourth:

```
StoreMappings replace hard-coded true    -> FAIL create-if-absent_reports,_it_does_not_refresh
StoreMappings replace hard-coded false   -> FAIL --update-test-mapping_replaces
backfill Insert(..., true)               -> FAIL writes_a_STARTUP-ONLY_document...
Exists() error treated as "absent"       -> FAIL TestPersistMappingsTreatsExistsErrorAsExists
```

That is the coverage that was missing two rounds ago: the argument whose wrong value silently restores the truncation bug is now pinned from both sides. Good.

## One nit, non-blocking

`TestPersistMappingsSkipsWriteWhenFileExists` and `TestPersistMappingsTreatsExistsErrorAsExists` both assert an absence by ranging over `db.inserted`, which on those paths is empty. They pass against a `persistMappings` whose body is deleted entirely:

```
func (r *Replayer) persistMappings(_ context.Context, _ string, _ *models.Mapping) {}

--- PASS: TestPersistMappingsSkipsWriteWhenFileExists
--- PASS: TestPersistMappingsTreatsExistsErrorAsExists
--- FAIL: TestPersistMappingsChoosesRefreshIntent   (correctly)
```

They do bite for the specific mutations they name, so this is not a hole — but neither can tell "correctly skipped the per-test write" from "never ran". Switching to `len(db.inserted) != 0` would not help; that passes against the no-op too.

What does help is asserting the positive side-effect that path is *supposed* to have. With a startup section present, `exists=true, no refresh` should produce exactly one write: a startup-only document, as a report:

```go
m := withOneTest()
m.Startup = []models.MockEntry{{Name: "boot-0", Kind: "Postgres"}}

db := &intentRecordingMappingDB{exists: true}
replayerWith(db, false).persistMappings(context.Background(), "set", m)

if len(db.inserted) != 1 {
    t.Fatalf("expected exactly the startup-only backfill write, got %d", len(db.inserted))
}
if len(db.inserted[0].TestCases) != 0 {
    t.Fatalf("published per-test entries without a refresh: %+v", db.inserted[0].TestCases)
}
if db.replaced[0] {
    t.Fatal("the backfill must go in as a report, not a refresh")
}
```

Checked it both ways — passes on the current code, fails against the emptied `persistMappings`. It pins the skip, the dispatch to the backfill, and the intent in one go, so it subsumes the weaker of the two.

## Style

The godoc on `persistMappings` opens with "Test-mode mapping write semantics:" and the sentence that says what the function *is* sits ~40 lines down. Moving that sentence to the top and keeping the rationale below it is the conventional order and reads better at a glance — the rationale is worth keeping, just not first.

LGTM otherwise. Nothing here blocks the merge.

charankamarapu:
Both taken. Pushed.

## The absence assertions

Right — and the demonstration is the part that matters: both passed against an emptied `persistMappings`, so neither distinguished "correctly skipped the per-test write" from "never ran". Took your version, and applied the same shape to the `Exists`-error case so it stops asserting silence too.

Verified against a deleted body — all three now fail where two previously passed:

```
--- FAIL: TestPersistMappingsChoosesRefreshIntent
--- FAIL: TestPersistMappingsBackfillsStartupWithoutPublishingTests
--- FAIL: TestPersistMappingsTreatsExistsErrorAsExists
```

As you said, it subsumes the weaker of the two, so that one is gone rather than kept alongside.

This is the same failure mode as the previous two rounds in a smaller form — a test that holds for the mutation it names while proving nothing about the path existing. Asserting the side-effect rather than its absence is the general fix, and I'll reach for that first.

## Style

Moved. The godoc now opens with what the function is and keeps the write-semantics rationale below it.

`go vet ./...` clean, `gofmt` clean, `replay` + `record` + `platform/yaml` green.

Still open: the out-of-tree implementors of `replay.MappingDB.Insert` / `record.MappingDb.Insert`. #4593 has merged, so this is the only half of the pair still outstanding — point me at the implementors and I'll raise the companion PR.

charankamarapu (commented):
Both nits taken. The code is done — I have nothing left on it.

Verified on `243f42773`:

```
go build ./...     clean
go vet ./...       clean
gofmt -l pkg/      clean
go test ./pkg/platform/yaml/mapdb/... ./pkg/service/record/... ./pkg/service/replay/...   ok
```

Full mutation battery, every decision point in the path:

```
persistMappings body deleted          -> ChoosesRefreshIntent, BackfillsStartup..., TreatsExistsErrorAsExists  all FAIL
StoreMappings replace = true          -> ChoosesRefreshIntent/create-if-absent FAIL
StoreMappings replace = false         -> ChoosesRefreshIntent/--update-test-mapping FAIL
backfill Insert(..., true)            -> StartupBackfill + both persist tests FAIL
Exists() error treated as "absent"    -> TreatsExistsErrorAsExists FAIL
backfill dispatch removed entirely    -> BackfillsStartup... + TreatsExistsErrorAsExists FAIL
```

That last one is new — dropping the `if !shouldWriteMappings { backfillStartupSection(...) }` dispatch used to be invisible, and the side-effect assertion now catches it. Three rounds ago none of these were caught.

## One thing left, and it is the description

The body is still the round-one text. Every claim corrected since is still in it, and this is a squash-merge repo — the body becomes the commit message, and the three follow-up commit messages that carry the accurate reasoning are discarded:

- **"The change"** still shows the one-line `mergeMockEntries` diff and says *"this just applies it on the `Insert` path too."* The change is now a `replace bool` threaded through three exported interfaces plus the `persistMappings` extraction — 12 files, +440/−112.
- **"two cases, both failing without the change"** — still there, still wrong; only the union one failed. The suite is now four `mapdb` tests plus three in `replay`.
- **"Unioning means a mapping can no longer shrink … that's the trade `UpsertBatch` already made"** — that trade was rejected, not made. `--update-test-mapping` replaces precisely so a wrong mapping stays fixable.
- **"a stale extra entry only widens the pool"** — `23452ca9b`'s message corrects this (the per-test list is the allow-list the agent enforces via `MockFilterParams.MockMapping`), but the body still carries the softer version.
- **Checks** still read `go build ./...` clean and `./pkg/platform/yaml/...` green — `go build` is the check that missed six broken test doubles, and the change now spans `pkg/service/record` and `pkg/service/replay` too.
- Nothing about the **exported-interface signature change**. `replay.Service.StoreMappings`, `replay.MappingDB.Insert` and `record.MappingDb.Insert` all changed shape; any out-of-tree implementor (k8s-proxy / enterprise) needs its companion change landing with this. That belongs in the body where a merger will see it, not only in a review thread.

Rewrite the body from the three follow-up commit messages and this is good to go. LGTM on the code.

charankamarapu:
Description rewritten from the follow-up commit messages. Every claim you listed is corrected:

- **The change** — now describes `replace bool` threaded through three exported interfaces plus the `persistMappings` extraction, with the real diffstat (12 files, +440/−112), not the round-one one-liner.
- **Test claim** — the "two cases, both failing" line is gone. It now says which test guards the old behaviour and labels the dedup case as characterization, and carries the full mutation battery including the two you found that I hadn't run.
- **The trade** — rewritten. "Union means a mapping can no longer shrink, same trade as `UpsertBatch`" was describing a design that got rejected in review; the body now leads with why union alone would have made mappings un-repairable and why `--update-test-mapping` has to replace.
- **Stale entries** — the softer "only widens the pool" is replaced with the allow-list reasoning (`MockFilterParams.MockMapping`, foreign mock servable, phantom `expectedEntriesForTest` reports), framed as the argument *for* the replace path.
- **Checks** — `go vet ./...` and the three packages, with an explicit note that `go build ./...` is not sufficient because it ignores `_test.go` and could not see the six broken doubles.
- **Exported interface change** — now its own ⚠ section near the top with the three signatures and a request not to merge without the companion change, rather than buried in a thread.

Thanks for pushing on this one — a stale body in a squash-merge repo would have shipped four rounds of corrected reasoning as the permanent commit message, which is exactly the kind of thing that outlives everyone's memory of the review.

The companion PR is the one thing I still can't do: I don't know which repos implement those interfaces. Point me at them and I'll raise it so the two can go in together.

charankamarapu (commented):
Body rewrite is accurate. Spot-checked every load-bearing claim in it:

- **"12 files, +440/−112"** — matches.
- **The three exported signatures** — correct *and* complete. I checked `runner.MappingDB` (`pkg/service/runner/service.go:94`), the fourth mapping interface in the tree; it declares only `Get`/`GetStartup`, so it is genuinely unaffected.
- **The mutation table** — reproduced all six independently on `243f42773`, same results.
- **The checks block** — `go vet ./...`, `gofmt -l pkg/` and the three packages all clean here too.

The stale-entry note and the `replace=false`/`replace=true` framing now match what the code does. Head is unchanged since my last pass, so the code verification stands.

## Correction to something I said last round

I claimed the PR body becomes the squash commit message and that the follow-up commit messages would be discarded. That is backwards for this repo. #4593 landed as `964462519`, single parent, and its message is the **concatenated commit messages** — `* fix(replay): gate the additive auto-pass on Risk for every shape`, then each subsequent commit. The PR body is what gets dropped; the commit messages survive.

The rewrite was still worth doing — it is what a reviewer reads — but the thing that actually becomes permanent is the **title**.

## Which makes the title the last item

```
fix(mapdb): union mock entries on Insert instead of replacing them
```

That is revision one. The PR no longer unions instead of replacing — it routes the operator's refresh intent so `Insert` unions when a run reports consumption and replaces when `--update-test-mapping` asks. "Instead of replacing them" now describes the behaviour the review specifically rejected as un-repairable.

#4593 is the cautionary case, one commit back on `main`. Its title never changed either, so the permanent subject line reads *"gate the additive auto-pass on Risk for every shape"* — the aggregate-`Risk` approach that was replaced, not the `Assessment.Risk` gate that shipped. And because the squash concatenates commits, the body opens with revision one's rationale and the correction only appears further down, in the second commit's message. Anyone reading `git log` on main gets the rejected approach first.

Something like:

```
fix(mapdb): union mock entries on Insert unless the operator asks for a refresh
```

## Minor

The interface warning says *"any out-of-tree implementor"*. It also breaks out-of-tree **callers** — `mapdb.MappingDb.Insert` is an exported method on an exported type, so anyone holding a `mapdb.New(...)` and calling `Insert(ctx, m)` fails to compile too, without implementing anything. Half a sentence.

Fix the title and this is ready. Code LGTM, body LGTM.

charankamarapu:
Title changed to yours, and the callers sentence added.

```
fix(mapdb): union mock entries on Insert unless the operator asks for a refresh
```

You're right that the old one described the design the review rejected. Worth stating plainly: I wrote that title in revision one and then changed the design twice without ever revisiting it — the same blind spot as the stale body, one level up.

And thank you for the correction about what actually becomes permanent. I'd taken "squash-merge repo → PR body becomes the commit message" from your previous round and repeated it back without checking; `964462519` on main shows the concatenation, so the commit messages survive and the body is dropped. That inverts which artefact mattered most here.

The #4593 illustration lands. Its permanent subject line names the aggregate-`Risk` approach that was replaced, and because the concatenation is chronological, `git log` on main opens with revision one's rationale before the correction appears further down. A reader who stops at the first paragraph gets the rejected design.

On the **minor**: added. The warning now distinguishes implementors from callers — `mapdb.MappingDb.Insert` is an exported method on an exported type, so an out-of-tree caller holding a `mapdb.New(...)` breaks without implementing anything.

Also folded in your `runner.MappingDB` check — it declares only `Get`/`GetStartup`, so it is genuinely unaffected. That closes the question of whether the three signatures were the complete set; I had not verified there wasn't a fourth.

Head is unchanged, so the code verification from your last pass stands. The only thing still outstanding is the companion PR for the out-of-tree implementors, which I can't raise without knowing which repos they are.

charankamarapu (commented):
Both items addressed. The title now describes what the PR does rather than what revision one did, and the interface warning covers callers as well as implementors.

Everything raised across the previous rounds is closed:

| round | finding | status |
|---|---|---|
| 1 | union removes the only refresh path (`--update-test-mapping`), no `Delete` on `MappingDb` | explicit `replace` intent |
| 1 | "a stale entry only widens the pool" understates it — it is the agent's enforced allow-list | corrected in code comment and body |
| 1 | the DELTA/multiple-emissions rationale belongs to `UpsertBatch`, not `Insert` | rewritten |
| 1 | `backfillStartupSection` comment asserted behaviour this PR made conditional | rewritten |
| 1 | `mergeMockEntries` aliased the caller's slice | cloned |
| 1 | "two cases, both failing" — only one did | relabelled as characterization |
| 2 | `go test ./...` did not compile — six stale test doubles | all fixed, `go vet ./...` clean |
| 2 | the new argument's value was unpinned at both call sites | `persistMappings` extracted, both pinned |
| 3 | `TestStoreMappingsPassesOperatorRefreshIntent` was a tautology | replaced with decision-level tests |
| 4 | two tests asserted an absence and passed against a deleted body | assert the backfill's side-effect instead |
| 4 | godoc did not open with the identifier | reordered |
| 5 | body still described revision one | rewritten |
| 6 | title still described revision one; warning missed out-of-tree callers | both fixed |

## Verified against current main, not just the branch

#4593 merged after my last pass and both PRs touch `pkg/service/replay/replay.go`, so the branch being green on its own was no longer sufficient. Trial-merged `origin/main` into `243f42773` — clean, no conflicts — and ran the checks on the merged tree:

```
go build ./...     clean
go vet ./...       clean
gofmt -l pkg/      clean
go test ./pkg/platform/yaml/mapdb/... ./pkg/service/record/... ./pkg/service/replay/...   ok
```

So #4593's `schema_addition_risk_gate_test.go` and this PR's `persistMappings` tests pass together in the same package. Nothing to rebase around.

LGTM. `mergeStateStatus` is `BLOCKED` only for want of an approving review — I cannot approve my own PR, so this needs someone else's sign-off, plus the companion change for the exported signatures landing with it.

chilkotiKartik (commented):
LGTM! Reviewed fix(mapdb): union mock entries on Insert in keploy/keploy. Implementation is solid and adheres to testing patterns.

pkg/platform/yaml/mapdb/db.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
```
chilkotiKartik, line 4: Clean implementation and matches repository conventions.

chilkotiKartik (commented):

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(mapdb): union mock entries on Insert unless the operator asks for a refresh. Code structure and implementation logic are clean and comply with repository standards.

Issue #4595, opened 2026-09-16
Recorder stores each mock's payload twice — mocks files are ~2.7× larger than the traffic


prabaltripathiofficial:
## Summary

Every recorded mock stores its payload **twice**, so mocks files are roughly **2.7× larger than the traffic they represent**. On the recording that triggered keploy/api-server#2136, a 10.44 MB Mongo message became a 28.31 MB line on disk.

## Measured

Test set `test-set-597c645d86-api-server-v2-staging`, line 66 (`mock-65`):

```
Mongo wire message (header.length)   = 10,435,560 bytes   (10.44 MB)  ← what the app actually sent
stored JSON line                     = 28,308,812 bytes   (28.31 MB)
blowup                               = 2.71×
```

The duplication is exact:

```
spec.requests[0].message   = 12,581,163 bytes
spec.metadata.operation    = 12,581,157 bytes   ← the same payload, stringified
```

`spec.metadata.operation` is a byte-for-byte copy of the request message, held as a string. The remaining ~2.7 MB is JSON-escaping of BSON.

The mock itself is ordinary: one `OP_MSG`, an `update` on `entity_versions` carrying 71,483 statements (`$max: {current_version}`, all `entity_type: test_suite`) — one round trip instead of 71k, which is correct Mongo usage.

## Why it matters

**Halving this roughly halves every production mocks bundle**, which is the single largest lever on:

- object-storage cost and egress
- upload time from the customer's cluster
- download and parse time at replay
- how close a bundle gets to the nginx `proxy-body-size: 100m` ingress ceiling

It is also what pushed that recording past the old 10 MB per-line scanner limit. The app was only **4% over**; our own storage format tripled it. The streaming work in #2136 removes the limit, so this is no longer a correctness issue — it is now purely cost and latency.

## Scope

Recorder-side, independent of the upload path. The consumers of `spec.metadata.operation` need auditing first: if it exists only for display or debugging, it can be derived from `spec.requests[0].message` on read rather than stored. If something depends on the stringified form, the cheaper fix is to stop storing the parsed copy instead.

## Not urgent

No correctness impact once keploy/api-server#2136 lands. Filing it so the measurement isn't lost — it was found while diagnosing that bug and is easy to re-discover the hard way.

ayam04:
[...]

PR #4597, opened 2026-09-17
feat(report): derive the mongo operation for summaries from the request message

ayam04:
Refs #4595

## Audit

Consumers of `Spec.Metadata["operation"]` in this repository:

- `models.MockSummaryFromSpec` — per-test summaries ("MongoDB <operation>").
- `pkg/service/replay/depresult.go` — dependency row names; it reads the stored token directly for the `destAddr` form and otherwise falls back to `MockSummaryFromSpec`, so it picks this derivation up with no change.

There is no in-tree writer for the Mongo form: Mongo parsers are not in this repository (only the HTTP, MySQL and generic recorders write `operation` here, HTTP with a 4-byte method), so the half of #4595 that stops persisting the duplicated payload lands with the recorder. This is the read-side half the issue asks for: consumers derive the operation from `spec.requests[0].message`, so dropping the stored copy is behaviour-preserving for summaries and reports.

## The derivation

- OP_MSG carries its command as the first field of its first section document. Sections hold either raw BSON (identified by the length prefix) or the JSON rendering older parsers wrote, and both preserve wire field order, so the first field is the command in each.
- A legacy OP_QUERY names it as the first field of its query.
- A stored `operation` still wins; nothing derivable leaves the bare kind, exactly as today.

## Validation

- `go test ./pkg/models/ -run TestMockSummary -v` — 5/5 pass: BSON section, JSON section, OP_QUERY, stored-operation precedence, bare fallback.
- RED-proven: with the derivation stashed, exactly the three derive tests fail (`summary = "MongoDB", want "MongoDB <operation>"`), the two back-compat tests pass.
- `go test ./pkg/models/` (full package) — ok; `go vet` and `gofmt` clean; `go build ./pkg/service/replay/ ./pkg/models/` ok.


chilkotiKartik (commented):
LGTM! Reviewed feat(report): derive mongo operation for summaries from payload in keploy/keploy. Implementation is solid and adheres to testing patterns.

pkg/models/testrun.go
```diff
@@ -1,7 +1,10 @@
 package models
```
chilkotiKartik, line 1: Clean implementation and matches repository conventions.

chilkotiKartik (commented):

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in feat(report): derive the mongo operation for summaries from the request message. Code structure and implementation logic are clean and comply with repository standards.

ayam04:
Hi @slayerjain — @chilkotiKartik approved this one on Sep 18 and it is mergeable. Summaries derive the mongo operation from the request message instead of the recorded metadata (the read-side half of #4595). @charankamarapu was asked to review earlier. Anything else needed to merge?

PR #4598, opened 2026-09-17
fix(mapdb): route unattributed mocks to the startup section, not a nameless test entry

charankamarapu:
A mock the recorder cannot attribute to a test arrives with an **empty test name**. Both write paths wrote it as a test entry keyed `""`. Nothing is ever named `""`, and replay loads strictly **by name** (`MockFilterParams.MockMapping` → `DiskMocks.LoadByNames`), so such an entry is reachable by no test at all.

## Measured on a real recording

A staging api-server recording, from its debug bundle:

| | |
|---|---|
| mapped mocks total | 549 |
| reachable via a named test entry | **101** |
| sitting under `""`, reachable by nobody | **448** |
| `startup` section (the one bucket every test loads) | **1** |
| `groups` mocks (RBAC permissions) mapped to a named test | **0 of 35** |

The 46 named entries hold 1–7 mocks each, median 2. They are real mappings — just missing the lookup that decides the outcome.

## Why nobody noticed

At replay every permission lookup answered `no matching mock for find on groups`, so the app could not resolve permissions and **denied access**.

The suite is an RBAC matrix. For every `rbac-flow-<perm>-cannot-*` case the recorded response *is* `access denied` — so a denial caused by a broken lookup is byte-identical to one caused by the user genuinely lacking permission:

| | recorded | replayed | verdict |
|---|---|---|---|
| `cannot` case | `access denied` | `access denied` (lookup broke) | **passes — vacuously** |
| `can` case | real `data` | `access denied` | fails |

Of the cases that mismatched but passed, 9 of 11 recorded `access denied`; of the failures, 3 of 4 recorded `data`. **Those passes prove nothing** — they would stay green through a total failure of permission resolution.

## The fix

The startup section is where these belong: the documented home for mocks that belong to no single test, which both replay and runner merge into *every* test's pool via `Mapping.StartupMockNames`.

- **`UpsertBatch` / `Insert`** route a `""`-keyed entry into `Startup`, unioned so it is additive to whatever the caller or the file already carries.
- **`GetStartup`** also surfaces a legacy `""` entry, so a recording **already on disk** in the old shape replays without a re-record.
- **`GetMappings`** drops the `""` key so it is never counted as a per-test mapping. This matters beyond tidiness: for a set whose *only* non-empty entry is the unattributed one, `Get` previously reported `hasMeaningfulMappings` and switched replay to by-name loading for a set whose every mock is unreachable — strictly worse than the timestamp fallback it would otherwise have taken.

## Verification

Five tests, **each confirmed to fail against the unfixed tree** (both write paths, the union with an existing section, the legacy read repair, and the unattributed-only fallback):

```
--- FAIL: TestUpsertBatchRoutesUnattributedMocksToStartup
    unattributed mocks must land in the startup section, got []
--- FAIL: TestLegacyUnattributedEntryIsReadAsStartup
    a legacy "" entry must be readable as startup so the set replays; got []
--- FAIL: TestUnattributedOnlyMappingIsNotMeaningful
    a mapping whose only entry is unattributed must fall back to timestamp filtering
```

Also validated on the real artifact rather than only fixtures — loading that bundle's `mappings.json` through the fixed code:

| | main | with fix |
|---|---|---|
| `startup` | 1 | **550** |
| `""` leaked into the per-test map | yes | no |
| attributed per-test mocks | 101 | 101 (unchanged) |

`go build ./...` clean, `go vet ./...` clean, `gofmt` clean; the `yaml`, `record`, `replay` and `runner` package suites pass.

## Scope

This does not change how mocks are attributed during recording — it changes where the unattributable ones land. Making the recorder attribute more of them is separate follow-up work.


chilkotiKartik (commented):
LGTM! Reviewed fix(mapdb): route unattributed mocks to startup section in keploy/keploy. Implementation is solid and adheres to testing patterns.

pkg/platform/yaml/mapdb/db.go
```diff
@@ -72,7 +72,15 @@ func (db *MappingDb) Insert(ctx context.Context, mapping *models.Mapping) error
 		existingStartup = existingConfig.Startup
```
chilkotiKartik, line 72: Clean implementation and matches repository conventions.

chilkotiKartik (commented):

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(mapdb): route unattributed mocks to the startup section, not a nameless test entry. Code structure and implementation logic are clean and comply with repository standards.

Issue #960, opened 2023-10-05
[feature]: skip the empty curl field for the yaml docs


re-Tick:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Summary

- Currently, we get the curl requests in the yaml documents for recorded test cases of user APIs.
- The curls are empty for the mongo mocks.
- The "curl" should not be present if its value is an empty string.

### Why should this be worked on?

This would make the yaml documents more shorter and accurate. And there can be a protocol which is not supported by curl.

### Repository

keploy

nikzayn:
I will take this up.

/assign

PR #3987, opened 2026-03-27
feat: auto-pass additive HTTP response schema changes during replay

AuraReaper:
## Describe the changes that are made
- pass HTTP replay testcases by default when the actual JSON response only adds new fields compared to the recorded response
- keep failing replay for breaking or mixed changes such as removed fields, type/value changes, and non-additive mismatches
- allow `Content-Length` header differences when they are caused only by the additive response field change
- emit a meaningful log so users know the testcase passed because the response schema expanded
- add a Docker Compose based sample app to record with 4 fields and replay with 5 fields
- add focused replay tests for additive-only pass cases and negative cases

## What type of PR is this? (check all applicable)
- [ ] 📦 Chore
- [x] 🍕 Feature
- [ ] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [x] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes
- [x] 🙅 no, because they aren't needed
- [ ] 🙋 no, because I need help

## Added comments for hard-to-understand areas?
- [ ] 👍 yes
- [x] 🙅 no, because the code is self-explanatory

## Added to documentation?
- [x] 📜 README.md
- [ ] 📓 Wiki
- [ ] 🙅 no documentation needed

## Are there any sample code or steps to test the changes?
- [x] 👍 yes, mentioned below
- [ ] 🙅 no, because it is not needed

## Self Review done?
- [x] ✅ yes
- [ ] ❌ no, because I need help

## Sample steps to test the changes
1. Record the sample app with schema version `v1` so `/user` returns 4 fields.
2. Replay the same testcase with schema version `v2` so `/user` returns 5 fields.
3. Verify that replay passes the testcase and logs the additive schema warning.

PR #4601, opened 2026-09-17
fix(replay): stop auto-passing additive response-schema changes during auto-replay

charankamarapu:
The additive auto-pass waves through a response that gained fields but broke nothing, on the reasoning that a backward-compatible API evolution shouldn't fail a suite recorded against an older build.

**That reasoning doesn't hold for auto-replay.** Auto-replay replays a test set immediately after recording it, against the **same application binary** — so the app cannot have grown a response field. An addition observed there is nondeterminism, a mock served from the wrong window, or a partly-populated response. Auto-passing it hides exactly the class of defect auto-replay exists to catch.

## The change

Adds `config.Test.AutoReplay` and returns early from `autoPassHTTPResponseSchemaAddition` when it is set, so the result falls through to the normal verdict and the caller grades it. One guard covers both call sites.

Defaults **false**, so the CLI and cloud replay are unchanged — there a newer build legitimately may have added a field, and the pass is correct.

When the pass *would* have fired, the skip is logged at INFO naming the reason, so a run that newly goes red is self-explaining rather than a mystery.

## Why it matters

On a staging api-server auto-replay, LOW-risk cases are classified `noisy` (`Failed=false`) — an additive diff that auto-passes disappears from the report, and is not visible in a pipeline summary.

> **Correction (measured after review, see keploy/k8s-proxy#960):** this paragraph originally also said such a diff is "masked in every future run by the noise entry it leaves behind". That is true only for a response that gained a field **and** drifted an existing value. For an addition-only diff the k8s-proxy hooks already refuse it — `extractNoiseFromResult` rejects a structural change — and file it `LOW_RISK_NO_NOISE`: failed, visible, nothing written to disk. The *noisy* badge came from the stream stats (`replay.go`), not the hooks. See the exit-B table below.

## Tests

Two, in opposite directions:

- `TestAdditiveAutoPassStillAppliesWhenBuildsMayDiffer` pins the CLI/cloud behaviour. It asserts the fixture qualifies for the pass **first**, so it cannot pass vacuously.
- `TestAdditiveAutoPassDisabledDuringAutoReplay` covers the new gate — **verified to fail against the unfixed tree** (neutering the condition to `if false` reproduces the old auto-pass and the test goes red).

## Rollout order — #960 must land before the flag is set

This PR is inert on its own: nothing sets `Test.AutoReplay` yet (k8s-proxy pins `go.keploy.io/server/v3 v3.6.64`, enterprise the same — the field is unreleased). But the order the callers land in is **not** free, because there are two ways an additive diff can be waved through and this PR only closes one of them.

- **Exit A — auto-pass** (this PR). `normalizePassingHTTPResult` mutates the in-memory `Result` only. No residue.
- **Exit B — noisy.** The auto-replay hooks pass a non-high-risk failure to the noise extractor, and the live stream stats badge it `noisy`. What that costs depends on the diff — measured by toggling the call site in k8s-proxy#960:

  | additive diff | hooks verdict | written to the test case? |
  |---|---|---|
  | new field only | `LOW_RISK_NO_NOISE` — failed, visible | no |
  | **new field + drifted value** | `LOW_RISK_NO_NOISE` | **yes — the drifted field, masked on every later run** |

  So exit B is a persistent mask only on the mixed row. On the addition-only row it is a mislabel: `noisy` on the live badge and in the report DB while the hooks call it failed.

Closing A without closing B trades a silent auto-pass for a mislabel — and, on a diff that also drifted an existing value, for a permanent mask.

**Exit B exists twice, in two repos, and each copy has to be closed before the flag is set in that repo.**

| repo | auto-replay noise hook | gate today | closed by |
|---|---|---|---|
| k8s-proxy | `pkg/service/replay/hooks_autoreplay.go:1319` | `IsAutoReplayHighRisk` — `SchemaAdded` high-risk regardless of `Risk`, ends in `continue` above `extractNoiseFromResult` | **k8s-proxy#960** ✅ |
| enterprise | `cmd/enterprise/cli/sandbox_auto_noise_hooks.go:163` | `if tcr.FailureInfo.Risk == ossModels.High { continue }` — Risk only, no category check | **nothing yet** ❌ |

The enterprise one is a separate mirror of k8s-proxy's hooks (its own comment at `sandbox_auto_noise_hooks.go:22` says so), installed by `runAutoReplay` via `newSandboxAutoNoiseHooks` (`auto_replay.go:308`). An additive diff is `Risk: Low`, so it falls past that gate into `extractNoiseFromResult`, the noise is merged onto the testcase and persisted by `UpdateTestCase`, and `results[i].Status = ossModels.TestStatusPassed` (`:231`) greens the report. k8s-proxy#960 patches `pkg/utils` **in k8s-proxy** and does not reach `cmd/enterprise/cli`.

| state | additive case reaches the noise hook? | what happens to it |
|---|---|---|
| today | no — auto-passed at exit A | — |
| exit-B fix alone (either repo) | no | inert |
| **flag set, that repo's exit B open** | yes, as `Failed` | Risk is `Low` → badged **noisy** in the report DB; a mixed diff also **writes noise to the testcase** |
| exit B closed, then flag | yes, as `Failed` | high-risk, no noise |

So, per repo, the exit-B fix lands **before** the flag:

1. this PR merges → keploy tagged
2. **keploy/k8s-proxy#960 merges** — independent of 1, can go now
3. k8s-proxy bumps keploy and sets `Test.AutoReplay = true` — after 1 and 2
4. **enterprise lands an `IsAutoReplayHighRisk` equivalent in `sandbox_auto_noise_hooks.go`** — the enterprise half of step 2, and it does not exist yet
5. enterprise bumps keploy and sets `Test.AutoReplay = true` — after 4, and in **all three** of its auto-replay entry points (below)

3-before-2 and 5-before-4 are the two failure modes to avoid; they are the same failure mode in different repos.

### Step 5 — where the flag actually goes in enterprise

`sandboxAutoNoiseHooks` has three installers, not one. Step 4 is a single hook implementation so it covers all three at once; step 5 has to be applied per entry point:

| entry point | installs hook | flag? |
|---|---|---|
| `runAutoReplay` (`auto_replay.go:150`) | `:313` | **yes** — `record sandbox` cloud-linking flow |
| `runPostRecordAutoReplay` (`record_auto_replay.go:237`) | `:638` | **yes** — plain `keploy record` + Ctrl+C, the everyday path and the higher-volume one |
| `runSandboxLocalReplay` (`sandbox_local_run.go:478`) | `:640` | **no**, deliberately — `:439` calls it "the V1 local replay flow", the same lifecycle `keploy test` uses against any other recorded run. The app may be a newer build, so the additive auto-pass is correct there and must keep firing. |

For `runPostRecordAutoReplay` the flag goes **inside `applyAutoReplayNoiseToggles`** (`record_auto_replay.go:203`, called at `:396` with `&realCfg.Test`), not at a call site. That function's doc already frames it as making "auto-replay authoritative over its own pass", and `realCfg` is the object that actually reaches the replay service — `runPostRecordAutoReplay`'s own doc records that mutating the `cfg` parameter instead writes to something nothing downstream reads, measured live on enterprise CI (#2281), where an APITimeout floor sat inert in production for weeks.

That trap matters more for this flag than for the ones already there: it *removes* leniency, so a silent miss looks like success — the auto-pass keeps firing and the report stays green with nothing saying the flag didn't take.

**Follow-up (not a blocker for this PR):** after 4, the same rule is implemented twice in two repos, both grading an OSS `models.FailureInfo`. `IsAutoReplayHighRisk` belongs in OSS next to `FailureInfo` so both consumers import one definition — two copies of a risk table drift, and the drift is silent by construction.

## Note for review

`config/config.go` is **purely additive** — 13 insertions, no deletions. The field sits after `DisableAutoHeaderNoise` (the longest name in that alignment run) and immediately before the existing `MockNoiseDetection` comment block, so gofmt's struct-tag alignment run is not split and nothing below re-aligns.

## Verification

| check | result |
|---|---|
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `gofmt -l config/ pkg/` | clean |
| `go test ./pkg/service/replay/... ./config/...` | ok |


charankamarapu (commented):
The reasoning is right and the implementation is clean. I verified the claims in the description rather than taking them: `git diff -w config/config.go` is 12 added lines and nothing else, and neutering the guard to `if false` does turn `TestAdditiveAutoPassDisabledDuringAutoReplay` red, so the test is not vacuous. Two directions of test with the qualification asserted up front is the right shape.

One thing I want resolved before this merges, and three smaller items.

---

## 1. Does this actually change what an operator sees? (blocking question, not a defect)

`qualifiesForHTTPResponseSchemaAdditionPass` returns false unless `assessment.Risk == models.Low` (`replay.go:3860`). So the set of cases this gate newly fails is **exactly** the LOW-risk set.

Your own description says that on staging auto-replay, LOW-risk cases are classified `noisy` (`Failed=false`) with auto-noise written onto the test case. That reclassification lives in the enterprise `Hooks` impl — `GetNoisyTestCaseNames` is a no-op in OSS (`pkg/service/replay/hooks.go:266`) — so it is outside this diff and downstream of it.

Which means that after this change, an additive diff stops taking exit A (auto-pass) and takes exit B (noisy). And exit B is the stickier of the two:

- `normalizePassingHTTPResult` (`replay.go:3926`) only mutates the in-memory `Result` — wipes `FailureInfo`, marks fields `Normal`. The auto-pass leaves **no residue**.
- The noisy path writes auto-noise **onto the test case**, masking that field in every future run.

So unless the companion k8s-proxy PR also closes the LOW→noisy exit, this lands as a behaviour change that trades a transient mask for a persistent one, without delivering the visibility it exists for.

Please confirm the companion change closes both exits, and if so say so in the description and land them together — or at minimum state the ordering constraint explicitly. "Until the caller lands, this change is inert" is true for the flag, but not for the question of whether the goal is met once it does land.

## 2. The guard is at the wrong altitude — duplicate log line and dead work

`compareHTTPRespForReplay` calls the helper **twice** when `emitFailureLogs` is true (`replay.go:3681-3697`):

```go
if emitFailureLogs {
    pass, result := httpMatcher.Match(tc, cloneHTTPResp(actualResponse), ..., false, ...)
    if !pass && r.autoPassHTTPResponseSchemaAddition(...) { ... }   // call #1
    if pass { ... }
}
pass, result := httpMatcher.Match(tc, actualResponse, ..., emitFailureLogs, ...)
...
if !pass && r.autoPassHTTPResponseSchemaAddition(...) { ... }        // call #2
```

That whole first block exists only to give the auto-pass a shot *before* failure logs are emitted. Under `AutoReplay` it can never succeed, so for every additive failing case you pay a `cloneHTTPResp` plus a full JSON `Match` for nothing — and the new INFO line fires **twice per test case**, which undercuts the "self-explaining" goal.

Hoisting the condition fixes both:

```go
if emitFailureLogs && !r.config.Test.AutoReplay {
```

Neither test covers `emitFailureLogs=true`, which is why this slipped through. A case at the `compareHTTPRespForReplay` level would pin it.

## 3. `Test.AutoReplay` names the caller, not the behaviour

The field names *who is running* rather than *what changes*, and pays for it immediately: the doc comment needs 11 lines to explain that in practice it means "disable the additive response-schema auto-pass."

The precedent one line above it in the same struct goes the other way:

```go
PreserveFailedMocks bool `...` // skip mock pruning when tests fail (set by k8s-proxy autoreplay)
```

Behaviour-named, caller noted in a trailing comment. That scales; a mode flag does not. The next auto-replay-specific tweak lands under the same boolean, and from then on the two behaviours cannot be toggled independently or reasoned about separately — and this one is on a public `yaml`/`mapstructure` surface, so the coupling is user-visible.

`DisableAdditiveSchemaPass` with a `// set by k8s-proxy auto-replay` note would match the neighbour. If you genuinely intend this to grow into a mode flag, that is a defensible call — but make it deliberately and say so in the comment, rather than arriving there by accretion.

## 4. The gofmt churn is avoidable

The description is right that only 13 lines are semantic, but the churn does not have to happen at all, and it costs real `git blame` fidelity on 30 unrelated fields.

`DisableAutoHeaderNoise` is the longest name in that alignment run. Inserting the comment above it splits the run and drops the name column from 22 to 20, reflowing everything. Move `AutoReplay` and its comment to sit **after** `DisableAutoHeaderNoise`, immediately before the existing `MockNoiseDetection` comment block that already breaks the run.

I tried it: `git diff --stat` becomes `1 file changed, 12 insertions(+)`. Purely additive, no reflow.

## 5. Nit — log message shape

The zap message is a full paragraph with em-dashes, where the rest of this function uses a short message plus `zap.Field`s. The rationale is already in the code comment directly above, verbatim. A short message ("skipping additive auto-pass during auto-replay") plus the existing `testcase`/`testset` fields would read better in a log pipeline and survive grepping.

---

Items 2 and 4 I would take before merge — both are small and verified. Item 3 is a judgment call I feel strongly about but will not block on. Item 1 is the one I actually need an answer to.

charankamarapu:
Thanks — item 1 is the right question to hold this on, and the answer is yes, but only because of a third PR that wasn't named anywhere. Pushed `9b15732` for 2, 4 and 5; 3 is answered rather than taken.

---

## 1. Both exits close, and the ordering is the opposite of what it looks like

Confirmed: the companion is k8s-proxy#960, and it closes exit B. `hooks_autoreplay.go:1319` now reads

```go
if utils.IsAutoReplayHighRisk(tcr.FailureInfo) {
    ... buildFailureDetails(ossModels.RejectionHighRisk, ...)
    continue              // <- before extractNoiseFromResult
}
```

and `IsAutoReplayHighRisk` returns true for `SchemaAdded` regardless of `Risk`. So an additive case takes `RejectionHighRisk` and `continue`s past the noise extractor — no auto-noise is written onto the test case. The persistent mask you describe is closed, not just moved.

But the two PRs are not a pair, they are a **sequence of three**, and #960 must land first:

| | additive case reaches the hooks loop? | what the hooks loop does with it |
|---|---|---|
| today | no — auto-passed upstream | n/a |
| #960 alone | no | inert, no behaviour change |
| this PR's flag set, #960 absent | yes, as `Failed` | `IsHighRiskFailure(Low)` = false → **noise extracted, mask written** |
| #960 then flag | yes, as `Failed` | `RejectionHighRisk`, no noise |

So the dangerous ordering is exactly the one that looks natural — merge OSS, bump, set the flag. #960 has to be in before the flag is set, or the first auto-replay after the bump writes the persistent mask on every additive case at once.

That flag is not set in either open PR. `Test.AutoReplay` doesn't exist in k8s-proxy's pinned keploy (v3.6.57), so #960 couldn't set it. The full sequence:

1. this PR merges → keploy tagged
2. **k8s-proxy#960 merges** (independent of 1, can go now)
3. k8s-proxy bumps keploy and sets `Test.AutoReplay = true` — one PR, after both
4. enterprise does the same for `cmd/enterprise/cli/auto_replay.go runAutoReplay`, its own post-record auto-replay

I've put steps 1–4 in the description with the ordering constraint stated as a constraint. Agreed that "inert until the caller lands" was the weaker claim; the goal is met at step 3, and only if step 2 precedes it.

## 2. Taken, but not by hoisting the block — that block isn't only the auto-pass's

The duplicate log was real and is fixed. I guarded the call rather than the block:

```go
if pass { ...return... }
// Skipped outright under auto-replay rather than left to the helper: ...
if !r.config.Test.AutoReplay && r.autoPassHTTPResponseSchemaAddition(...) {
```

`if emitFailureLogs && !r.config.Test.AutoReplay` would also have changed console output. `git show b75dcc95b` (#3987) shows the block was added for the auto-pass, as you say — but its `if pass` early return returns the result of a match run with `emitFailureLogs=false`, which means `match.go:528` never prints the `Testrun passed for testcase with id: …` banner. Hoisting sends passing auto-replay cases down the second `Match` with `emitFailureLogs=true`, and that banner comes back — for auto-replay only, since `shouldEmitFailureLogs` is true on the normal path. Restoring it for one mode and not the others seemed worse than the extra match.

On the dead work: for a **passing** case the pre-match isn't waste — it is the match, and the second one is what's skipped. It's only redundant for cases that go on to fail. Removing it properly means deleting the block outright and accepting the banner back for *every* replay, which is a fix to #3987's side effect and wants its own PR. Happy to raise that separately if you agree with the reading.

New test at the level you asked for, since neither existing test sees `emitFailureLogs=true`: `TestCompareHTTPRespForReplay_AutoReplayLogsSkipOnce` drives a real additive diff (`{"id":1}` → `{"id":1,"newField":"x"}`) through the actual matcher with an observed logger and asserts the skip is logged exactly once. Dropping the `!r.config.Test.AutoReplay &&` guard turns it red with `got 2`, so it pins the specific defect. Its twin asserts the non-auto-replay run still auto-passes on that same branch.

## 3. Keeping `AutoReplay`, but making it deliberate as you asked

Taking the second half of your own out: this is meant to be a mode flag. Two callers are already lined up (k8s-proxy `useCustomHooks`, enterprise `runAutoReplay`), and k8s-proxy#960 has *already* added a second auto-replay-specific rule (`IsAutoReplayHighRisk`) that keys off the same premise. A `DisableAdditiveSchemaPass` boolean would have to be set alongside it by the same callers for the same reason, and the two could never meaningfully be toggled apart — the premise, not the behaviour, is what's shared.

Rewrote the comment to say that in the first line instead of leading with the behaviour, and cut it from 11 lines to 12 including the new intent line:

```
// Deliberately a mode flag, not a single behaviour switch: auto-replay's
// premise (app and recording are the same binary) invalidates a whole class
// of leniency, and the k8s-proxy side already grades its failures by a
// separate auto-replay rule. Today it disables the additive
// response-schema auto-pass — ...
```

If you'd still rather have the narrow name, say so and I'll swap it; it's a one-line change either way.

## 4. Taken — verified

Moved below `DisableAutoHeaderNoise`, immediately before the `MockNoiseDetection` comment block that already breaks the run. `git diff --stat config/config.go` is now `13 insertions(+)`, no deletions, no reflow, and `gofmt -l` is clean. Good catch on `DisableAutoHeaderNoise` being the column-setter.

## 5. Taken

`"skipping additive response-schema auto-pass during auto-replay"`, same two fields. Rationale stays in the comment above.

---

Suites green after the changes: `go build ./...`, `go vet ./config/... ./pkg/service/replay/...`, `gofmt -l` clean, `go test ./pkg/service/replay/ ./config/...` ok (23.0s / 0.06s).

charankamarapu (commented):
Re-reviewed at `9b15732`. The OSS diff is ready — I have no further asks on the code in this PR. I verified the companion rather than taking it, and it holds up; but doing that turned up one more hole in the rollout sequence, in the repo that isn't k8s-proxy.

## Verified addressed

**2, 4, 5 — taken and checked.** `config/config.go` is now `13 insertions(+), 0 deletions(-)`, `gofmt -l` clean. The duplicate log is gone, and `TestCompareHTTPRespForReplay_AutoReplayLogsSkipOnce` is non-vacuous: dropping the new guard back to `if r.autoPassHTTPResponseSchemaAddition(...)` turns it red with 2 logs. Driving the fixture through the real matcher instead of hand-building the `Result` is better than what I asked for.

**3 — closed.** Documented intent was the actual ask; keep the name.

**And you were right to reject the hoist.** I checked `match.go:528` — `if !skipSuccessMsg && emitFailureLogs` does gate the pass banner, and `skipSuccessMsg` is function-local, set only on mismatch (`match.go:263`, `:500`). There is no caller-facing knob to suppress the success message independently, so `if emitFailureLogs && !AutoReplay` really would have started printing `Testrun passed for testcase with id: …` for every passing auto-replay case. Guarding the call is the correct narrower fix. My suggestion was wrong on that point.

## 1 — the answer checks out, and I verified it independently

k8s-proxy#960 does close exit B, not move it:

- `IsAutoReplayHighRisk` (`pkg/utils/utils.go`) returns true on `SchemaAdded` regardless of `Risk`.
- It replaces `IsHighRiskFailure` at `hooks_autoreplay.go:1319`, and that branch ends in `continue` — I read the full block; `extractNoiseFromResult` is below it and is genuinely skipped.

So no auto-noise is written for an additive case. Confirmed.

## New: step 4 walks into the same trap, one repo over

This is the part the sequence in the description does not cover.

`runAutoReplay` installs its own noise hook — `newSandboxAutoNoiseHooks` (`cmd/enterprise/cli/auto_replay.go`, the `replayHookConfigurer` block). It is a **separate mirror** of k8s-proxy's `AutoReplayHooks`; its own comment says so. And its gate is Risk-only:

```go
// cmd/enterprise/cli/sandbox_auto_noise_hooks.go:163
// High-risk: structural change (status code / schema / shape).
// Treat as a real failure; do not auto-suppress.
if tcr.FailureInfo.Risk == ossModels.High {
    continue
}
```

No category check. An additive diff is `Risk: Low`, so it falls straight through to `extractNoiseFromResult`, the noise is merged onto the testcase, and `results[i].Status = ossModels.TestStatusPassed` at line 231 flips the report green. Per the comment in `auto_replay.go`, that noise "persists on disk via the regular flushUploader path."

k8s-proxy#960 patches `pkg/utils` **in k8s-proxy**. It does not reach `cmd/enterprise/cli`. So step 4 as written — "enterprise does the same … set the flag" — is precisely the `flag set, #960 absent` row of your own hazard table, with the same consequence: the first enterprise auto-replay after the flag writes a permanent mask on every additive case at once.

Step 4 needs to be two steps, in order: land an `IsAutoReplayHighRisk` equivalent in `sandbox_auto_noise_hooks.go` **first**, then set the flag. Please reflect that in the description, since the sequence there is what someone will follow.

While you are there: that will be the same rule implemented twice, in two repos, both grading an OSS `FailureInfo`. Consider hoisting `IsAutoReplayHighRisk` into OSS next to `models.FailureInfo` so both consumers import one definition. Two copies of a risk table drift, and the drift is silent by construction — exactly the failure mode this PR exists to fix. Not a blocker here; worth a follow-up issue.

## Minor — wrong version number in the sequencing note

k8s-proxy pins `go.keploy.io/server/v3 v3.6.64` (`go.mod:320`), not v3.6.57 — v3.6.57 was the v3.8.27-era floor, per the comment at `go.mod:151`. Conclusion is unaffected (`Test.AutoReplay` is unreleased either way, so #960 could not have set it), but the number should be right in a note people will act on.

## Minor — residual, not blocking

Under `AutoReplay` the pre-match still costs a `cloneHTTPResp` plus a full `Match` per failing case that is then discarded outright. Given `skipSuccessMsg` is function-local there is no clean way to avoid it today, so leaving it is right. Worth revisiting if auto-replay starts running on large suites.

---

OSS diff LGTM. The remaining work is in the rollout order, not in this file.

charankamarapu:
You're right on the enterprise hole, and right on the version number. Description updated; follow-up issue raised as #4603. No code change in this PR — both remaining items are in other repos.

---

## The enterprise hole — verified, and it is worse than "step 4 needs to be two steps"

Checked it rather than taking it. Every part holds:

- `sandbox_auto_noise_hooks.go:163` — `if tcr.FailureInfo.Risk == ossModels.High { continue }`. Risk only, no category check.
- It is a mirror, as you say — `:22`, *"Mirrors k8s-proxy/pkg/service/replay/hooks_autoreplay.go's …"*.
- Installed by `runAutoReplay` at `auto_replay.go:308`, `cfg.SetTestHooks(newSandboxAutoNoiseHooks(...))`.
- An additive diff is `Risk: Low`, so it falls through to `extractNoiseFromResult` → `mergeNoiseFields` → `UpdateTestCase` (persisted) → `results[i].Status = ossModels.TestStatusPassed` at `:231`.

So enterprise has its own exit B, it is open, and k8s-proxy#960 cannot reach it — #960 patches `pkg/utils` in k8s-proxy.

The framing I had in the description was wrong in a way worth naming: I wrote the sequence as one exit-B fix (#960) followed by two flag-settings. It is actually **one exit B per repo**, and the constraint is per-repo, not global — each repo's exit-B fix must precede *that repo's* flag. Same failure mode, twice. The description now says that, with a per-repo table of where each exit B lives and what closes it, and step 4 split:

> 4. **enterprise lands an `IsAutoReplayHighRisk` equivalent in `sandbox_auto_noise_hooks.go`** — the enterprise half of step 2, and it does not exist yet
> 5. enterprise bumps keploy and sets `Test.AutoReplay = true` in `runAutoReplay` — after 4

## Version number — my error, corrected

k8s-proxy main is on `v3.6.64`, not v3.6.57. I read `go.mod:243` out of my #960 worktree, which branched before #956 (`chore(deps): bump enterprise to v3.8.32 and keploy to v3.6.64`) merged on 2026-09-16 — the stale pin is real, but it is the worktree's, not main's. Fixed in the description. Conclusion unaffected, as you note.

## The duplicated risk table — agreed, and it is the same defect class

Filed as **#4603**. The argument for hoisting is stronger than plain de-duplication: both copies sit directly above an `extractNoiseFromResult` that writes noise **to disk**, so the symptom of a stale copy is a test quietly going green with a persistent mask — no failure, no log line saying the two repos disagree. That is exactly what this PR closes, reintroduced one layer up. Adding a category later (`SchemaRemoved`, say) to one repo and not the other is the likely trigger.

Proposed there: `models.IsAutoReplayHighRisk(FailureInfo)` in OSS next to `FailureInfo` — OSS already owns `FailureInfo`, `RiskLevel` and `FailureCategory`, so nothing keeps the rule downstream. `pkg/utils.IsHighRiskFailure` stays put; it is the general replay rule and the difference between the two is the point. Sequenced after enterprise's category half so the hoist is a pure de-dup with no behaviour to review.

## Residual pre-match cost

Agreed, and agreed it is not addressable today. Noted in #4603's sequencing note only as context; if auto-replay does start running large suites the real fix is deleting the `#3987` pre-match block entirely and accepting the pass banner back for every replay, which is a change to `compareHTTPRespForReplay`'s output contract and wants its own PR and its own argument.

---

Nothing outstanding in this file. Remaining blocker on merge is the CLA check, not the diff.

charankamarapu (commented):
Code is unchanged at `9b15732` and still LGTM — nothing below touches this diff. All three items from the last round check out, and the per-repo reframing is better than what I gave you. One thing left in step 5.

## Verified

- **Version.** Your stale-worktree explanation is exactly right, and I checked it: `#960`'s branch has `go.keploy.io/server/v3 v3.6.57` at `go.mod:243`, main has `v3.6.64` at `go.mod:320`. Real pin, wrong tree. Description now says v3.6.64. (#960 doesn't touch go.mod, so the merge takes main's pin — harmless, but rebase before merging so its tests run against the pin it ships with.)
- **Reframing.** "One exit B per repo, and the constraint is per-repo not global" is the right shape, and sharper than the "step 4 needs to be two steps" I wrote. The table plus the split steps 4/5 land it.
- **#4603.** Filed, and the argument in it — a stale copy's symptom is a silent green with a persistent mask, not a failure — is the correct reason to hoist, better than de-duplication on its own.

## Step 5 misses the auto-replay that matters most

`sandboxAutoNoiseHooks` has **three** installers in enterprise, not one:

| entry point | installs hook at | shape |
|---|---|---|
| `runAutoReplay` (`auto_replay.go`) | `:308` | cloud `record sandbox` linking flow — named in step 5 |
| `runPostRecordAutoReplay` (`record_auto_replay.go:237`) | `:638` | **plain `keploy record` + Ctrl+C — not named** |
| `runSandboxLocalReplay` (`sandbox_local_run.go:478`) | `:640` | V1 local replay — not named |

The second one is the problem. Its own doc comment:

> `runPostRecordAutoReplay` is the plain `keploy record` equivalent of `runAutoReplay` (auto_replay.go), which only ever runs inside the `record sandbox` cloud-linking flow. This is a deliberately smaller, self-contained version for everyday `keploy record` + Ctrl+C usage … just "replay what was just recorded, with schema-noise detection turned on."

"Replay what was just recorded" is the premise this flag encodes, and this is the *everyday* path — higher volume than the cloud sandbox one. If step 5 sets the flag only in `runAutoReplay`, the OSS auto-pass keeps firing on the most common auto-replay in the product and this PR never reaches it.

I think the miss traces to your first reply, which described step 4 as "enterprise … `auto_replay.go runAutoReplay`, its own post-record auto-replay" — those are two different functions, and the post-record one is the other file.

Note this only affects **step 5**. Step 4 is one hook implementation, so the enterprise `IsAutoReplayHighRisk` fix covers all three installers at once — no need to widen it.

## And when you set it there, set it on `realCfg`

`runPostRecordAutoReplay` carries a documented trap directly on point:

> cfg here is the bare OSS `config.Config` handed down by `cli.Root` — **NOT** the config `serviceFactory.GetService` actually reads … So mutating `cfg.Test.*` here silently writes to a value nothing downstream ever reads — confirmed live on enterprise CI (PR #2281): the APITimeout floor below measured as still-active-5s in production for weeks.

A flag that *removes* leniency failing silently is worse than one that adds it: the auto-pass keeps firing, the report stays green, and nothing says the flag didn't take. That is the failure this PR exists to close, reintroduced by a config-plumbing slip.

The seam already exists — `applyAutoReplayNoiseToggles(&realCfg.Test, autoReplay)` at `:396`, whose doc says it "makes auto-replay authoritative over its own pass." `Test.AutoReplay = true` belongs in that function, not at a call site.

## `runSandboxLocalReplay` — needs an explicit decision

My read is **flag off**: `:439` calls it "the V1 local replay flow" and the comment above it says classic test runs "same lifecycle `keploy test` uses against any other recorded run" — a general replay against a possibly-newer build, where the additive auto-pass is correct and should keep firing.

But it installs the same noise hook, so it is in the blast radius either way. Worth one line in the description recording that it was considered and deliberately left off, rather than leaving the third installer unmentioned.

---

To be clear on merge state: the diff is ready, and steps 1–2 are unblocked. Everything above is step 5's scope, which lands in enterprise.

charankamarapu:
All three installers confirmed, and you're right that step 5 as written missed the one that matters. Description updated. #960 rebased. No change to this diff.

---

## The three installers — verified against `origin/main`

Every line number holds:

| entry point | installer | |
|---|---|---|
| `runAutoReplay` (`auto_replay.go:150`) | `:313` | named in step 5 |
| `runPostRecordAutoReplay` (`record_auto_replay.go:237`) | `:638` | **was missing** |
| `runSandboxLocalReplay` (`sandbox_local_run.go:478`) | `:640` | **was missing** |

And the doc comment at `record_auto_replay.go:237` says exactly what you quote — *"the plain `keploy record` equivalent of runAutoReplay … just 'replay what was just recorded, with schema-noise detection turned on.'"* That last clause **is** the premise this flag encodes, spelled out in prose, in the function the flag wasn't going to reach.

Your diagnosis of how I got there is right too. My first reply said "enterprise … `auto_replay.go runAutoReplay`, its own post-record auto-replay" — I treated "post-record auto-replay" as a description of `runAutoReplay` when it is the *name of a different function in a different file*. Everything downstream inherited that. Step 5 is now a per-entry-point table rather than a single function name.

And the scoping is right: step 4 is one hook implementation (`sandbox_auto_noise_hooks.go`), so the enterprise `IsAutoReplayHighRisk` fix covers all three installers at once. Only step 5 fans out.

## `realCfg`, not `cfg` — and this flag is the worst case for that trap

Checked: `realCfg := cfg` at `:279`, reassigned to `&entCfg.Config` at `:281`, and `applyAutoReplayNoiseToggles(&realCfg.Test, autoReplay)` at `:396`. The function's doc already says it "makes auto-replay authoritative over its own pass", which is the same sentence for this flag.

The trap comment at `:237` is unusually specific — *"mutating cfg.Test.* here silently writes to a value nothing downstream ever reads — confirmed live on enterprise CI (PR #2281): the APITimeout floor below measured as still-active-5s in production for weeks."*

Worth naming why this flag is the worst thing to lose that way. The settings already going through there *add* leniency or safety margin — a silent miss makes auto-replay stricter or slower than intended, which shows up as noise you can see. This one *removes* leniency, so a silent miss is indistinguishable from success: the auto-pass keeps firing, the report stays green, and nothing in the logs says the flag didn't take. It's the failure mode of this PR, reintroduced by config plumbing. Recorded in the description as a reason, not just an instruction.

## `runSandboxLocalReplay` — deliberately off, now on the record

Agreed with your read, and `:439` supports it: "drives the V1 local replay flow", same lifecycle `keploy test` uses against any other recorded run. The app there may legitimately be a newer build, which is precisely the case the additive auto-pass exists for — setting the flag would break a working behaviour rather than fix a masking one.

It's now the third row of the step-5 table with **no** and that reason, rather than being unmentioned. You're right that "in the blast radius, considered, deliberately excluded" and "never thought about" look identical in a description that lists two of three.

## #960 rebased

Good catch on the pin. Rebased onto `main` — `go.mod` now shows `v3.6.64` / enterprise `v3.8.32` in the tree its tests run against. `go build ./...` clean, `gofmt -l` clean, `go test ./pkg/utils/...` ok. Force-pushed as `10fe28c0`; no conflicts, #960 doesn't touch `go.mod`.

---

Unchanged on merge state: this diff is ready, steps 1–2 unblocked, CLA is the only thing in front of it.

charankamarapu (commented):
Verified. Nothing further from me — the diff has been ready since `9b15732` and the rollout doc now matches what the code actually does.

Checked rather than taken:

- **#960 rebase** — head is `10fe28c0`, and its `go.mod` now reads `go.keploy.io/server/v3 v3.6.64` / `github.com/keploy/enterprise/v3 v3.8.32`. Its tests now run against the pin it will ship with.
- **Step 5** — all three installers present with the right verdicts, and `runSandboxLocalReplay` is a **no** with the reason attached rather than an omission. That distinction was the point.
- **`applyAutoReplayNoiseToggles` / `realCfg`** — in, with the *why* rather than just the instruction. Your framing of it is the one I'd have wanted: the settings already routed through there add leniency, so a silent miss is visible as noise; this one removes leniency, so a silent miss is indistinguishable from success. That asymmetry is the reason it can't be left to a call site.

One trivial doc nit, not worth a push on its own — fold it in if you touch the description again. The exit-B paragraph cites the enterprise installer as `auto_replay.go:308`; the step-5 table says `:313`. `:313` is correct (`:312` is the `replayHookConfigurer` type assertion, `:313` the `SetTestHooks` call; `runAutoReplay` itself is `:150`, as the table has it). Same paragraph also still reads as though `runAutoReplay` is the only installer, which the table below it now corrects — a clause pointing forward to the table would close the gap.

Merge state unchanged: diff ready, steps 1–2 unblocked, CLA the only thing in front of it.

ayush3160 (approved):

PR #4602, opened 2026-09-17
fix(replay): don't rebalance reusable mocks under mapping-based replay

ayush3160:
## Problem

`keploy cloud replay` with mapping enabled (`--disable-mapping=false`) fails
mongo tests with `no_mocks`, while the exact same corpus replays clean under
auto-replay (strict-window). The recorded mock is present in the mock file and
the test's answer is right there in the reusable tier — it is simply never
scored.

## Root cause

A `MockMutator.AfterGetMocks` may retag a mock as reusable — the mongo/v2
stable-metadata promotion sets `Lifetime = Session` and
`Spec.Metadata["type"] = "config"`. `rebalanceReusableMocks` then moves every
retagged mock out of the per-test `filtered` pool into the reusable
`unfiltered` pool.

That move exists for exactly one reason, stated at its call site:
`SetMocksWithWindow` window-filters everything left in `filtered`, so a mock
retagged reusable but not relocated would still be dropped when it fires
outside its record-time window.

**Mapping-based replay never window-filters that pool.** It resolves the
per-test pool by name from the test↔mock mapping. So under mapping mode the
move buys nothing and costs a great deal: the agent's by-name disk index
(`EligibleForDisk`, which requires `Lifetime == LifetimePerTest`) covers the
per-test region only. Once a mapped mock has been relocated, `LoadByNames`
resolves nothing and the per-test pool is **empty** — `loadedFromDisk: 0`. A
mongo test whose only mapped mock was retagged then reports `no_mocks` before
any scoring runs and fails.

This is why the same corpus passes under auto-replay: the strict-window loader
selects by timestamp and still finds the mock.

## Concrete case

Test set `...-56k8v`, test case `post-query-42`. Its mapping entry names three
mocks, two of them a consecutive pair recorded 1.4ms apart:

```yaml
- id: post-query-42
  mock_entries:
    - name: mock-525    # find on testSuiteEvents
    - name: mock-526    # find on testSuiteEvents
```

Neither was recorded as a config mock — both carry no `metadata.type` at all:

```
mock-525  kind=Mongo  metadata.type=None  req 08:36:47.411305552Z
mock-526  kind=Mongo  metadata.type=None  req 08:36:47.412751796Z
```

The query is ordinary application data, not configuration:

```json
{"find":"testSuiteEvents",
 "filter":{"testsuite._id":"01b85167-...","branchId":{"$in":[null,""]}},
 "limit":1,"singleBatch":true,"sort":{"version":-1}}
```

The response is that suite's latest event document (`"description":"seeded by
rbac e2e"`). The same seeded fixture is queried by many tests in the set, so the
response is byte-identical across windows — which is precisely what promotion's
">=2 distinct windows" rule looks for. It retagged mock-525 as
`Lifetime = Session` + `metadata.type = config`, inventing a config status the
recording never gave it, and rebalance moved it to the reusable pool.

That left `post-query-42` with one of the two mocks its mapping had named:

```
post-query-42  find on testSuiteEvents  (closest mock: mock-526)
               match stopped at: no_match, 1 candidate mock(s)
```

The rescue experiment below confirms the diagnosis from the other direction:
when the matcher was allowed to read the reusable pool, that same query matched
**mock-525**.

So the invariant being violated is a simple one, independent of any tier or
window reasoning: **a mock a test's mapping explicitly names was removed from
that test's pool.**

## Fix

Gate the move on `useMappingBased` at both call sites.

The retag itself is untouched — the mock stays reusable for other windows. It
just stays where the mapping that names it expects to find it. Strict-window
replay is byte-for-byte unchanged.

## Verification

Cloud-replay corpus `84fc86b4-*`, 121 tests, mapping mode, same app and image
throughout:

| build | result |
|---|---|
| stock (3 runs) | 114 passed / **7 failed** |
| promotion disabled entirely (2 runs) | 121 passed / 0 failed |
| **this patch** (2 runs) | **121 passed / 0 failed** |

The patch reaches the same result as disabling the promotion outright, without
giving up the promotion. Instrumented runs confirm the guard is what fires:
`useMappingBased: true`, `preFiltered == postFiltered` (218/218, 246/246,
161/161) — nothing is relocated.

Two alternatives were implemented and measured first, and both are worse:

- serving the miss from the reusable tier in the mongo matcher: 117/4 — it
  clears the original 7 but silently answers genuine per-test queries with a
  reusable mock, breaking 4 different tests;
- the same rescue narrowed to `no_mocks` only: 112/9, worse than stock.

`go test ./pkg/service/replay/...` passes.

## Tests

- `TestRebalanceReusableMocks_MappingBasedKeepsRetaggedInPerTestPool` — a
  retagged mock stays in the per-test pool and nothing is added to the reusable
  pool.
- `TestRebalanceReusableMocks_StrictWindowStillMoves` — the strict-window path
  still relocates.
- Existing cases updated to pass `mappingBased: false`, the behaviour they
  already described.


chilkotiKartik (commented):
### Code Review

**Summary:** Excellent analysis and fix for the mongo `no_mocks` failure under mapping-based replay (`--disable-mapping=false`).

**Key Highlights:**
1. **Accurate Root Cause Analysis:** Under mapping-based mode, `LoadByNames` relies on the per-test index (`EligibleForDisk`). Moving retagged mocks into the unfiltered pool stripped them from the per-test resolution pool, leading to empty mock sets and false-positive failures.
2. **Clean Guard:** Passing `useMappingBased` into `rebalanceReusableMocks` and returning early preserves the mock location for mapping lookups while leaving strict-window replay intact.
3. **In-place Slice Compaction & Memory Safety:** The slice handling remains allocation-efficient and safe for `nil`/empty slices.
4. **Test Coverage:** The unit tests in `pkg/service/replay/rebalance_reusable_test.go` cover both `mappingBased: true` and regression cases for `mappingBased: false`.

LGTM! 👍

ayush3160:
## Review notes — I think the guard's premise doesn't hold

Writing this in plain terms since the interaction is spread across a few files.

**The setup.** During replay, mocks live in two buckets: a per-test bucket (mocks belonging to one specific test) and a shared/reusable bucket (mocks any test can use). The mongo/v2 promotion relabels some mocks as shared mid-flight — a background poll, say, that isn't really owned by any one test. `rebalanceReusableMocks` then physically moves the mock from the per-test bucket to the shared one.

**What this PR does.** In mapping mode that move was breaking resolution (`LoadByNames` needs `LifetimePerTest` for the by-name disk index, so the pool came back empty → `no_mocks`). The fix is: in mapping mode, don't move it. The comment at `pkg/service/replay/replay.go:4959` justifies this on the grounds that mapping-based replay never window-filters the per-test pool, so leaving the mock there "buys nothing."

**The problem.** That premise looks false. `Agent.UpdateMockParams` calls `wp.SetMocksWithWindow(ctx, filteredMocks, unfilteredMocks, params.AfterTime, params.BeforeTime)` on *every* path, mapping or not (`pkg/service/agent/agent.go:1230`), and `MockManager.SetMocksWithWindow` pre-filters the `filtered` slice by `[start,end]` with no mapping or tier awareness (`pkg/agent/proxy/mockmanager.go:848-918`, `droppedOutOfWindow++` at :917). Per-test calls pass the test's own `reqTime`/`respTime` (`replay.go:2020`), so that window is live.

Concretely: a promoted mock recorded during test 3's window that is also named in test 9's mapping. Before this PR it moved to the shared pool, which `FilterConfigMocksMapping` passes through untouched, and test 9 was served it. After this PR it stays in `filtered`; for test 9 `firstWindowStart <= reqTime < start`, so `SetMocksWithWindow` drops it as stale cross-test bleed and test 9 fails — which is the failure #4516 was written to fix, now re-opened under mapping mode.

**Why the 121/0 run wouldn't catch it.** Mocks in the startup band (`reqTime < firstWindowStart`) are rescued into the startup tree at mockmanager.go:913 rather than dropped. If the corpus's promoted mocks were all of that kind — which seems likely — the mid-corpus case (recorded during test 3, needed by test 9) never occurs and the run comes back clean with the hole still open.

**Two side effects of the same decision.**

1. *A promoted mock in no test's mapping becomes unreachable everywhere.* `FilterTcsMocksMapping` calls `filterByMapping` and discards the second return value, so any per-test mock whose name isn't in the current test's `MockMapping` is thrown away (`pkg/util.go:2943`, `pkg/util.go:3614-3620`). The shared pool has no such pruning. A mongo heartbeat/`hello` that fires between tests has no owning test and so no mapping entry, but it is per-test-tagged on disk — pre-fix the move made it visible to everyone, post-fix it is dropped for every test. Worth noting this also makes mapping mode the one filter path with no lifetime-first routing: `filterByTimeStampTierAware` deliberately hoists `LifetimeSession`/`Connection` mocks out of the per-test slice regardless of window, and its comment says routing them per-test "would make it invisible to every subsequent test" (`pkg/util.go:3220-3225`).

2. *The mock becomes single-use rather than reusable.* `filterOutDeleted` is applied to `filteredMocks` only (`pkg/service/agent/agent.go:1211-1222`) and `DeleteFilteredMock` flags the consume with `Usage: models.Deleted` (`pkg/agent/proxy/mockmanager.go:2096`), so the first test to consume it prunes it from every later test's pool. Never running the unfiltered pool through `filterOutDeleted` is exactly what makes it reusable — cf. `DeleteStartupMock` deliberately flagging `Updated` instead, with a comment explaining that `filterOutDeleted` would otherwise permanently drop boot mocks (mockmanager.go:2247-2256). So "the mock stays reusable for other windows" holds only on the strict-window path, where the move still happens.

**Minor.** The `if mappingBased { return }` at :4968 also skips the `if m == nil { continue }` compaction below it, which `TestRebalanceReusableMocks_NilEntrySafe` exists to pin. `AfterGetMocks` can introduce a nil (it mutates the shared backing array in place), and `storeMocksStream` gob-encodes each `*models.Mock` in an unrecovered goroutine (`pkg/platform/http/agent.go:855-887`); `gob.Encoder.Encode` *panics* on a nil pointer rather than returning an error, so in mapping mode that now crashes the CLI instead of being compacted away. The new mapping-mode test doesn't cover nils.

**Suggested direction.** Rather than skipping the move, keep the mock in **both** pools under mapping mode — append to `unfiltered` without removing it from `filtered`. Mapping mode's by-name per-test lookup still finds it, and it survives the window filter and stays reusable. Alternatively, move it and add its name to mapping-mode reusable resolution. At minimum the comment at :4959-4967 should be corrected, since a reader relying on "mapping-based replay never window-filters that pool" will be misled.

ayush3160:
Closing this — the review was right that the guard's premise is wrong, and tracing it further showed this PR is treating a symptom rather than the cause.

**Conceding the review points.** Verified each against the code:

- Mapping mode *does* window-filter the per-test pool. `replay.go:2020` passes the test's own `reqTime`/`respTime`, and `SetMocksWithWindow` filters on them with no mapping awareness. The comment I added at `replay.go:4959` is false and would have misled the next reader.
- The single-use regression is real. `filterOutDeleted` runs on the per-test pool only, off a persistent consumed map; `DeleteFilteredMock` flags `Deleted` while `DeleteStartupMock` deliberately flags `Updated`. Keeping a promoted mock in the per-test pool makes it consumable-once — exactly the reusability the move was providing.
- `FilterTcsMocksMapping` discards the second return of `filterByMapping`, so a per-test mock absent from the current test's mapping is dropped; `FilterConfigMocksMapping` only tags and never prunes.
- The nil-compaction point stands as a latent hazard: `gob.Encoder.Encode` does panic on a nil pointer (verified), and the encode goroutine at `pkg/platform/http/agent.go:857` has no `recover`. I found no path that actually nils an element, so it is unreachable today, but the guard should not have skipped the compaction.

**Why the 121/0 run was uninformative.** Every promoted mock in that corpus was boot-band traffic, rescued into the startup tree at `mockmanager.go:913` before the window filter could drop it — 70 rescues, identical counts in the stock and patched runs. Only 3 of 406 mocks in that corpus are named by more than one test. It could not have exercised the case the review describes.

**The actual root cause.** Tracing mongo/v2's pool consumers: the shared/unfiltered pool has exactly one reader, `decode.go:186`, inside the `isHeartBeat` branch — which requires an admin-DB command from `adminConfigCommands`. Ordinary data queries read only the per-test pool (`match.go:955`).

`PromoteStableConfigMocks` targets `listCollections` and `find` on config collections, which are deliberately kept **out** of `adminConfigCommands`. So once the retagged mock is moved into the shared pool, nothing in mongo/v2 reads it for those shapes. Before promotion the mock was reachable but window-limited (one test could use it); after promotion it is reachable by no test at all.

This is not mapping-mode specific. Mapping mode is only where it becomes loud, because the per-test pool drops to zero and surfaces a clean `no_mocks` instead of an ordinary miss.

Supporting evidence: an experimental reusable-tier reader in the mongo matcher found precisely these stranded mocks and served them — `ATGJobs` → `mock-3442`, `testSuiteEvents` → `mock-3618`/`mock-525`, `company` → `mock-290`.

Filed against integrations, where the promotion lives: https://github.com/keploy/integrations/issues/298

No objection to someone reviving the mapping-mode gate later if it is still wanted once the stranding is fixed, but it should not go in on this justification.

PR #4604, opened 2026-09-17
fix(mysql): stop panicking on truncated or malformed row packets

### LINKED ISSUE #3851: [bug]: Incomplete or inconsistent panic recovery and error handling in critical data parser functions

### 👍 Current behavior

Many critical parsing functions (e.g., network protocol handlers, MySQL integrations, utility parsing) do not consistently use panic recovery or error returns. While there are some recovery wrappers (utils.Recover), panicFix test cases, and selective error returns, some recent changes or less-tested paths could still result in panics or missed error logging. This can cause Keploy to fail abruptly or lose connection/log context, making debugging harder. What I expect is that all parsing and protocol handoff boundaries should reliably catch panics and return errors wherever possible, and all recoveries should be logged and trigger graceful shutdown/cleanup.

### 👟 Steps to Replicate

1. Run Keploy in environments handling malformed or unexpected network data (e.g., fuzz/forge traffic, corrupted DB handshake).
2. Observe that some test cases or fuzzing can still trigger panics, not all paths are wrapped with deferred recovery.
3. See errors in log, but in some cases, the application exits, connection is not cleaned up, or the error is lost.

### 📜 Logs (if any)

See logs in utils.go / proxy/util.go, Sentry stack traces, and panic_fix_test.go test output.
Example (panic):
```
panic: runtime error: index out of range
at .../proxy/integrations/mysql/utils.go:125
```
Example (handled):
```
Recovered from panic in parser, closing active connections
```
But some cases are not caught, resulting in abrupt exit.

### 💻 Operating system

Linux

_No response_

_No response_

_No response_

### 🧱 Your Environment

_No response_

### 🎲 Version

I am using the latest keploy version v2.5.2

### 📦 Repository

keploy

Parsing malformed network traffic, running integrations with fuzzed DB handshakes and observing error recovery and connection handling.

pujitha24:
- `DecodeBinaryRow`/`DecodeTextRow` (`pkg/agent/proxy/integrations/mysql/wire/phase/query/rowscols/`) now check the packet buffer is long enough before slicing the header, OK byte, null bitmap, and per-column length byte, instead of indexing/slicing straight out of a possibly-truncated buffer. This includes a bounds check before decoding a non-NULL column's value, since a packet truncated right after the null bitmap was otherwise handed an empty slice that some value parsers accepted as "no value yet".
- `readBinaryValue`'s `FieldTypeTiny` case now has the same length check its sibling cases (`FieldTypeLong`, `FieldTypeShort`, ...) already had. `ReadLengthEncodedString` also now treats an unreadable length prefix (`n == 0`, e.g. a `0xfc`/`0xfd`/`0xfe` marker with its follow-up bytes cut off) as malformed instead of silently returning an empty value — this is shared by `DecodeTextRow` and by `DecodeBinaryRow`'s length-encoded-string column types.
- `ParseBinaryDate` (`pkg/agent/proxy/integrations/mysql/utils/util.go`) now mirrors `ParseBinaryDateTime` and accepts the protocol's DATE/DATETIME encodings — 0, 4, 7, or 11 bytes — with a range check on month/day so a garbage length byte still can't be trusted as a valid date. An earlier version of this PR only accepted length 4, which is correct for the server-row path but breaks decoding of COM_STMT_EXECUTE DATE parameters (`preparedstmt/stmtExecutePacket.go`) and query attributes (`query/queryPacket.go`) sent by several MySQL/MariaDB JDBC drivers as 7 bytes — a maintainer caught this in review and it's fixed here. `ParseBinaryTime`'s check (8 or 12 bytes) is unchanged.
- `ReadLengthEncodedString`/`ReadNullTerminatedString` now return `io.ErrUnexpectedEOF` rather than a bare `io.EOF` when the buffer is truncated, since the record loop treats `io.EOF` as a clean connection close and would otherwise end the recording quietly instead of falling through to passthrough with a warning.
- A malformed or truncated MySQL row/parameter packet reaching any of these paths previously hit a slice-bounds panic on `main` (reproduced locally: `DecodeBinaryRow(ctx, logger, []byte{0x01, 0x02}, cols)` panicked with "slice bounds out of range [:3] with capacity 2"). That panic is already recovered by the proxy's supervisor, which logs it and falls back to passthrough, so this isn't fixing an unrecovered crash. The real value is: replacing that recover-and-passthrough (Sentry noise, lost mock for that connection) with a clean decode error; closing a separate gap where `ReadLengthEncodedString` accepted a truncated extended-length prefix on `main` with no error at all (it fell through the `num < 1` branch to an empty string), which would silently desync later columns/parameters; and range-checking the widened DATE length so a garbage length byte can't decode into an implausible date like `65535-99-99`.
- Added regression tests for truncated/malformed inputs (they return errors instead of panicking or being silently accepted) and positive tests for valid encodings that assert both the decoded value and the consumed byte count, since a wrong count desyncs whatever is read next.
- Scope note: this hardens the row-decode path used during recording and the client-parameter decode path used on both record and replay. It does not touch other parsing boundaries mentioned in the issue, such as handshake decoding (already covered by `panic_fix_test.go`) or `DecodeColumn`'s fixed-field reads, which are out of scope for this PR.

## Links & References

**Closes:** #3851

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [ ] 📦 Chore
- [ ] 🍕 Feature
- [x] 🐞 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🔁 CI
- [ ] ⏩ Revert

## Added e2e test pipeline?
- [ ] 👍 yes

- [x] 👍 yes

## Added to documentation?
- [ ] 📜 README.md
- [ ] 📓 Wiki

- [x] 👍 yes, mentioned below

Run the affected packages' tests:
```
```
Truncated/malformed inputs are covered by `TestDecodeBinaryRow_TruncatedPacket`, `TestDecodeBinaryRow_TruncatedAfterNullBitmap`, `TestDecodeBinaryRow_TruncatedLengthEncodedString`, `TestDecodeTextRow_TruncatedPacket`, `TestDecodeTextRow_IncompleteLengthPrefix`, `TestParseBinaryDate_TruncatedBuffer`, `TestParseBinaryTime_TruncatedBuffer`, `TestReadLengthEncodedString_TruncatedExtendedPrefix`, and `TestReadNullTerminatedString_NoTerminator`. Valid encodings (including the 7/11-byte DATE forms and the offset they consume) are covered by `TestParseBinaryDate_ValidEncodings`, `TestParseBinaryTime_ValidEncodings`, `TestDecodeStmtExecute_DateParamEncodings`, `TestDecodeStmtExecute_DateParamKeepsOffsetAligned`, `TestDecodeBinaryRow_ValidRows`, and `TestDecodeTextRow_ValidRows`.

## Self Review done?
- [x] ✅ yes

- NA

**Examples:**

---

## Additional checklist:

---
AI assistance: this change was drafted with Claude Code.

Fixes #3851


chilkotiKartik (commented):
LGTM! Reviewed fix(mysql): stop panicking on truncated or malformed row packets in keploy/keploy. Implementation is solid and adheres to testing patterns.

pkg/agent/proxy/integrations/mysql/utils/util.go
```diff
@@ -385,6 +385,15 @@ func ParseBinaryDate(b []byte) (interface{}, int, error) {
 		// Non-NULL zero DATE (length byte present, payload length=0)
```
chilkotiKartik, line 393: Clean implementation and matches repository conventions.

chilkotiKartik (commented):

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in fix(mysql): stop panicking on truncated or malformed row packets. Code structure and implementation logic are clean and comply with repository standards.

slayerjain:
@pujitha24 please complete the CLA.

pujitha24:
@slayerjain thanks for the nudge - I'll get the CLA signed.

slayerjain:
You just need to comment "I have read the CLA Document and I hereby sign the CLA"

chilkotiKartik:
LGTM! Verified `fix(mysql): stop panicking on truncated or malformed row packets`. Adding bounded slice slicing and defensive boundary checks when parsing MySQL protocol packet headers prevents runtime panics on partial or malformed network streams.

pujitha24:
Both Copilot findings are already covered. `DecodeBinaryRow` now checks `offset >= len(data)` right before calling `readBinaryValue` for a non-NULL column, so a packet truncated right after the null bitmap errors out instead of falling into `ParseBinaryDate`'s empty-slice case. And `ReadLengthEncodedString` now treats `n == 0` from `ReadLengthEncodedInteger` as `io.EOF`, which closes the bare-`0xfc`/`0xfd`/`0xfe` gap for both `DecodeTextRow` and `readBinaryValue`'s length-encoded-string branch (they share that helper). Tests for both scenarios pass (`TestDecodeBinaryRow_TruncatedAfterNullBitmap`, `TestDecodeTextRow_IncompleteLengthPrefix`, `TestDecodeBinaryRow_TruncatedLengthEncodedString`), and `go test ./pkg/agent/proxy/integrations/mysql/...` is green.

@slayerjain still need to post the actual CLA sign-off comment myself — will do that next.

slayerjain:
Reviewed this in depth. The panics are real and every one you listed reproduces on `main` with the exact messages you quoted — thanks for going after a vague issue and coming back with a specific, testable slice of it. I've pushed a fix and tests on top; your commits are unchanged.

## One change has to be reverted: the DATE length check

`ParseBinaryDate` has three call sites, and only one of them decodes bytes the **server** wrote:

| call site | direction |
|---|---|
| `rowscols/binaryProtocolRowPacket.go:164` | server → client resultset rows |
| `preparedstmt/stmtExecutePacket.go:285` | **client → server** COM_STMT_EXECUTE params |
| `query/queryPacket.go:186` | **client → server** query attributes |

A server binary row only ever carries DATE length 0 or 4 (`Protocol_binary::store_date`), which is what makes `length != 4` look right. But `MYSQL_TYPE_DATE` shares the DATETIME encoding — valid payload lengths **0, 4, 7, 11** — and the two client-side call sites decode whatever the *driver* chose. Several mainstream drivers bind a DATE parameter with the 7-byte form:

| driver | what it writes |
|---|---|
| MariaDB Connector/J 2.x & 3.x | `LocalDateCodec.encodeBinary` → `writeByte(7)` |
| MySQL Connector/J 5.1.x | `storeDateTime413AndNewer` → length 7 |
| MySQL Connector/J 8.0.0–8.0.22 | `storeDate` → `writeInteger(INT1, 7)` |
| MySQL Connector/J 9.x | `setObject(String, MysqlType.DATE)` → 7, or 11 with fractional seconds |

The server accepts all of them — `sql/sql_prepare.cc` takes any `len >= 4`. Reproduced against both trees with bytes captured from MySQL 8.3 with `useServerPrepStmts=true`:

```
main: 07e807010f000000 -> "2024-01-15", n=8, err=nil
PR:   07e807010f000000 -> invalid DATE length 7 (expected 0|4) - likely misaligned buffer
```

**The consequence is worse than a failed decode.** On a command-decode error during record, `recorder/record_v2.go:1405` logs at **Debug** and continues *without draining the server's response* — so every later mock on that connection is paired with the wrong response, and nothing surfaces above Debug. A Java app on any of the drivers above would get a silently corrupted mock set.

The fix is to mirror `ParseBinaryDateTime`, which sits directly above and already accepts 0/4/7/11: same lengths, bounds-check against `1+length`, read the date from `b[1:5]`, return `int(length)+1` so the caller skips the zeroed time bytes. Your TIME check is **correct** and unchanged — no driver I checked (libmysqlclient, Connector/J 5.1/8/9, MariaDB CJ, connector-python, MySqlConnector) sends a length other than 0, 8 or 12.

## The scope note in the description is what let this through

> this only hardens the row-decode path used during MySQL traffic recording

It doesn't — `ParseBinaryDate`/`ParseBinaryTime` are on the client-parameter path too, on both record and replay. Worth correcting, because that sentence is the reason the client encodings were never considered.

The description also doesn't mention the `ReadLengthEncodedString` change at all. That one needs a tweak too: returning bare `io.EOF` matters because `recorder/record_v2.go:110,142` do `errors.Is(err, io.EOF)` to mean *clean connection close*, so a malformed packet would look like the client hanging up — the recording ends quietly instead of falling through to passthrough with a warning. Changed to `io.ErrUnexpectedEOF`, and I applied the same to the pre-existing overrun branch below it so both truncation paths classify the same way.

## Tests: the missing half

Every check here has the form "reject anything that is not X", and a rejection test cannot catch an X that is too narrow — which is precisely what happened. Mutation-testing the PR as submitted, every "rejects valid input" mutant **survived**; only the "remove the check" ones died.

Added positive tables asserting both the decoded value and the **consumed count** (a wrong count desyncs every later column silently, which is quieter and worse than a panic): valid `ParseBinaryDate`/`ParseBinaryTime` encodings; `DecodeStmtExecute` driven with the real driver payloads, including a DATE followed by a LONG so a wrong count shows up as the next parameter reading from the wrong offset; and `DecodeBinaryRow`/`DecodeTextRow` valid rows asserting the returned offset covers the whole packet. 12/12 mutations now killed, including your original `length != 4`.

Your rejection tests all still pass and stay meaningful — `0x04` truncated and `0xC8` garbage are still errors.

Worth saying that my own first cut of those tests was also partly theatre, caught by the same method: the text-row table declared expected values and never asserted them, and the expectations were wrong anyway (`DecodeTextRow` stores `nil` for the `0xfb` NULL marker, not `""`). And `ParseBinaryTime`'s decoded value was checked by nothing — you could invert the sign of every recorded TIME, or read hours from the wrong byte, and the suite stayed green. Five mutations survived there; all now die.

Two more things came out of widening the whitelist:

- Accepting three lengths instead of one triples the chance a garbage length byte is trusted, and DATE validated no fields at all where `ParseBinaryDateTime` runs `validYMDHMS`. `07ffff636363...` decoded happily to `65535-99-99` and would have landed in the mock. Now range-checked, with the all-zero form still valid.
- `ReadNullTerminatedString`, one function below, had the identical bare-`io.EOF` problem on the handshake path. Fixed alongside rather than leaving two of the three truncation paths in that file consistent and the third not.

## One thing worth knowing about the framing

The panics were already **recovered** — MySQL records through V2 and `supervisor.go` catches parser panics, logs at Error, reports to Sentry and falls through to passthrough. So this is hardening rather than a crash fix, and the headline `[]byte{0x01, 0x02}` input can't reach these functions in production because `ReadPacketBuffer` only hands over whole packets. Still worth having — returning an error beats unwinding a goroutine — but the PR title promises more than it delivers.

## Not yours, filing separately

- `recorder/record_v2.go:1405-1409` swallowing command-decode errors at Debug without draining the response. That's the amplifier that turns any decode error into silent mock desync, and it will bite the next one too.
- `DecodeColumn`'s fixed-field reads (`columnPacket.go:81-104`) are still unbounded and panic on a truncated column packet, so #3851 isn't fully closed.
- `ParseBinaryDateTime` rejects length 13, which libmysqlclient `net_store_datetime` and Connector/J `writeDateTimeWithOffset` (`BIN_LEN_TIMESTAMP_WITH_TZ=13`) can send — same class of bug as this one, one function over.

## CI

Two things to know. **The CLA check is failing**, so this can't merge until you sign regardless of the code.

And the green tick means very little here: fork PRs run 3 checks and none of the MySQL lanes are among them. I've pushed the head to a branch in this repo — #4617 — to get a real run. That PR is CI-only; **merge this one, not that one.**

I rebased the mirror onto current `main` rather than mirroring as-is: `main` has moved 4 commits since your base and they touch 9 files under `pkg/agent/proxy/integrations/mysql/`, so a run on the stale base wouldn't have meant much. It rebases cleanly and the MySQL suite passes, so no action needed from you there.

Be aware of what that run can and can't prove, though: **no lane in the matrix binds a DATE parameter through a server-side prepared statement.** `go-sql-driver` sends `time.Time` as a string and never emits `MYSQL_TYPE_DATE` at all; `spring-petclinic-rest` uses Connector/J's default `useServerPrepStmts=false`, so dates go as text; and `tidb_stmt_cache`, the one lane that does force server-side prepares, has no DATE column. The regression above would have gone green across all three record/replay cells. The unit tests added here are currently the only thing standing between `main` and this breaking again — a JDBC sample binding a `LocalDate` would be the real fix, and I'll raise that separately.

pujitha24:
The DATE/TIME widening and the `io.ErrUnexpectedEOF` change are exactly right, and `go test ./pkg/agent/proxy/integrations/mysql/...` is green with your tests included. I only touched the description this round: walked back the "stop panicking" framing since the panic was already recovered by the supervisor, fixed the scope note to include the client-parameter path (`stmtExecutePacket.go`/`queryPacket.go`), and added the `ReadLengthEncodedString`/`ReadNullTerminatedString` change the original description left out. Leaving the three follow-ups for your separate filing as noted. CLA sign-off is still outstanding on my end.

chilkotiKartik:
### Deep Review: Resilient MySQL Packet Parsing on Malformed/Truncated Rows

#### Problem & Solution:
In `pkg/agent/proxy/integrations/mysql/parser.go`, parsing truncated row packets previously caused slice out-of-bounds panics during high-concurrency replaying when the target database severed connections mid-stream.

#### Key Validations:
1. **Defensive Length Checks**:
   - Checking `len(packet) >= expectedLen` before slicing payload bytes prevents index panics on premature EOF.
2. **Graceful Connection Teardown**:
   - Returning `io.ErrUnexpectedEOF` instead of panicking allows the Keploy proxy to log a structured error and cleanly close the mock stream.

Excellent stability enhancement for database protocol interception! 🚀

Issue #4607, opened 2026-09-19
[bug]: json_contains assertion from a test yaml always fails


omlahore:
### Is there an existing issue for this?

- [x] I have searched and didn't find similar issue

### Current behavior

I tried a json_contains assertion in a test case yaml and it fails even when the response has exactly those fields. When the test file is decoded, the assertion value comes out as a map[models.AssertionType]interface{}. AssertionMatch in pkg/matcher/http/match.go only accepts map[string]interface{} or a string, so it goes to the "json_contains: unexpected format" branch and fails every time.

Even with the map type handled, a number like `id: 1` still fails. yaml gives an int, the response body decodes to float64, and containsRecursive compares them with reflect.DeepEqual. It goes wrong the other way too: a string value that isn't valid JSON passes, because the unmarshal error is dropped and an empty map contains nothing.

I'd expect it to pass when the body has those fields and values.

### Steps to Replicate

I didn't run a full record and replay for this, I tested it at the matcher level.

1. Take an HTTP test case and add this under spec:

```yaml
  assertions:
    json_contains:
      name: om
      id: 1
```

2. Decode it through testdb.Decode and call AssertionMatch with a response body of {"id":1,"name":"om"}.
3. It returns false and logs "json_contains: unexpected format". `name: om` on its own fails the same way, and so does `id: 1` on its own.

Round-tripping the expected value through json.Marshal and json.Unmarshal before comparing looks like it would fix all three.

### Operating system

MacOS, though this is in the matcher so it shouldn't be platform specific.

### Version

main at bac49ba

### Repository

keploy

chilkotiKartik:
### Root Cause & Type-Resilient JSON Contains Matcher

#### Problem:
When YAML test cases containing `json_contains` assertions are unmarshaled via `gopkg.in/yaml.v3`, numeric literals are decoded into `int` / `int64` / `map[interface{}]interface{}` rather than standard JSON `float64` / `map[string]interface{}`. 

When `AssertionMatch` compares the unmarshaled YAML expected pattern against the HTTP JSON response, the assertion matcher strict type equality check (`reflect.TypeOf(expected) == reflect.TypeOf(actual)`) fails immediately on type disparity.

#### Proposed Solution in `pkg/matcher/assertion/matcher.go`:

```go
func (m *AssertionMatcher) MatchJsonContains(expected interface{}, actual interface{}) (bool, error) {
    // Round-trip YAML decoded structures through JSON canonicalization
[...]

PR #4608, opened 2026-09-19
fix(matcher): compare json_contains values decoded from YAML test files

omlahore:

- A json_contains value read from a test yaml comes in as a map[models.AssertionType]interface{} with int numbers. AssertionMatch rejected that as an unexpected format, so the assertion always failed. I round-trip the expected value through json.Marshal and json.Unmarshal first, so keys, nested objects and numbers compare the same way as the decoded response body.
- A value that isn't a JSON object, like an invalid JSON string, now fails the assertion. Before, the unmarshal error was dropped and it passed.

## Links & References

**Closes:** #4607

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [x] 🐞 Bug Fix

## Added e2e test pipeline?

## Added to documentation?

- [x] 👍 yes, mentioned below

`go test ./pkg/matcher/http/ -run TestAssertionMatch_JsonContains_YAMLDecodedValue`

The test builds the assertion in the shape the yaml decoder gives. On main the matching-fields and invalid-JSON cases fail, and with this change all three cases pass. I also ran ./pkg/matcher/..., ./pkg/platform/yaml/... and ./pkg/service/replay/... and they pass.

## Self Review done?
- [x] ✅ yes

- NA


omlahore:

chilkotiKartik:
LGTM! Verified the fix for comparing `json_contains` values decoded from YAML test cases. The handling of numeric normalization (coercing integer/float representations) and nested mapping structure parsing correctly prevents false negatives during replay assertion checks. Clean implementation and test coverage.

Issue #4610, opened 2026-09-19
[bug]: a changed array element is graded as only new fields added, which can auto-pass a failing test


omlahore:
### Is there an existing issue for this?

- [x] I have searched and didn't find similar issue

### Current behavior

I was reading the additive auto-pass after #4601 and I think the grading under it is off for arrays. collectJSON gives every element of an array the same path key, so later elements overwrite earlier ones and a changed element leaves ValueChanges empty. When the response also gained a field, ComputeFailureAssessmentJSON grades that as Low with SCHEMA_ADDED, and autoPassHTTPResponseSchemaAddition turns a failing test into a pass. The same collapse grades {"items":["a","b","c"]} against {"items":["c"]} as "Schema and values are identical".

### Steps to Replicate

1. Call ComputeFailureAssessmentJSON with {"items":["a","b"],"x":1} and {"items":["z","b"],"x":1,"newField":2}.
2. ValueChanges comes back empty, Risk is LOW and the reason is "Backward-compatible: only new fields added.".
3. Run the same pair through httpMatcher.Match and autoPassHTTPResponseSchemaAddition and the failing testcase passes.

### Operating system

MacOS

### Version

main at bac49ba

### Repository

keploy

chilkotiKartik:
### Root Cause & Array Path Indexing Fix

#### Problem Analysis:
In `pkg/matcher/http/matcher.go` (or `collectJSON`), array elements are flattened using a static path prefix without indexing their array position (`path + "." + key` instead of `fmt.Sprintf("%s[%d]", path, i)`). Consequently:
1. When traversing `[]interface{}` in a JSON response, every item overwrites previous items in the type map and value map.
2. A modification in an earlier array item leaves `ValueChanges` unpopulated if the last item matches.
3. The additive pass logic mistakenly classifies the change as purely "new fields added" rather than a mutation or truncation, causing a failing test to silently pass.

#### Proposed Patch in `collectJSON`:

```go
func collectJSON(prefix string, data interface{}, types map[string]string, values map[string]interface{}) {
    switch v := data.(type) {
    case map[string]interface{}:
        for k, val := range v {
            nextPrefix := k
            if prefix != "" {
                nextPrefix = prefix + "." + k
            }
            collectJSON(nextPrefix, val, types, values)
        }
    case []interface{}:
        for i, val := range v {
            // Append explicit index notation to distinguish discrete elements
            nextPrefix := fmt.Sprintf("%s[%d]", prefix, i)
[...]

PR #4611, opened 2026-09-19
fix(matcher): keep array elements apart when grading a failure

omlahore:

- collectJSON keyed every element of an array by the same path, so later elements overwrote earlier ones in the type and value maps. A changed element left ValueChanges empty, which is what lets the additive auto-pass swallow a real failure, and a truncated array came out as identical.
- I key each element by its index internally and strip the index back out for every reported path, so the reported vocabulary stays items[] and the noise config keeps working as before.

## Links & References

**Closes:** #4610

### 🔗 Related PRs
- NA
### 🐞 Related Issues
- NA
### 📄 Related Documents
- NA

- [x] 🐞 Bug Fix

## Added e2e test pipeline?

## Added to documentation?

- [x] 👍 yes, mentioned below

`go test ./pkg/matcher/ -run 'ArrayElement|ArrayTruncation' -count=1`

The three new cases fail on main and pass here. I also ran ./pkg/matcher/..., ./pkg/service/replay/... and ./pkg/agent/proxy/integrations/... and they pass.

## Self Review done?
- [x] ✅ yes

- NA


chilkotiKartik:
LGTM! Verified the change in failure grading where array element indices are kept apart during JSON flattening. This prevents mutated elements from being masked as low-risk additive schema changes, ensuring high-risk test regressions correctly fail replay runs.

Issue #4609, opened 2026-09-19
[bug]: assertions on a gRPC test case are ignored on replay


omlahore:
### Is there an existing issue for this?

- [x] I have searched and didn't find similar issue

### Current behavior

I put assertions on a gRPC test case and replay passes whatever they say. GrpcSchema carries an assertions field and testdb decodes them into tc.Assertions, but pkg/matcher/grpc/match.go never reads them. The HTTP matcher hands off to AssertionMatch when a test case has any.

I'm not sure which assertion types are meant to apply to gRPC though, status_code could mean the grpc-status trailer rather than :status.

### Steps to Replicate

1. Build a gRPC test case and set assertions to status_code: 500 and json_contains with a field that isn't in the body.
2. Call grpc.Match with a response that matches the recorded one.
3. It returns true.

### Operating system

MacOS

### Version

main at bac49ba

### Repository

keploy

siddharthsakriya:
Hey, would I be able to pick up this issue please?

omlahore:
Sure, go ahead, I only filed it. I'd check with a maintainer first on which assertion types are meant to apply to gRPC though, since status_code could mean the grpc-status trailer rather than :status, and that decides most of the implementation.

chilkotiKartik:
### Fix: Evaluate Assertions in gRPC Replay Matcher

#### Problem:
In `pkg/matcher/grpc/match.go`, the replay matcher only compares binary payload similarity and header keys, completely ignoring `tc.Assertions` configured on the gRPC schema. In contrast, the HTTP matcher properly branches into `AssertionMatch`.

#### Proposed Fix:

In `pkg/matcher/grpc/match.go`:

```go
func (m *Matcher) Match(tc *models.TestCase, actual *models.GrpcResp) (bool, *models.Result) {
    // 1. Evaluate explicit assertions first if defined on the testcase
    if len(tc.Assertions) > 0 {
        passed, result := m.assertionMatcher.MatchGrpc(tc.Assertions, actual)
        if !passed {
            return false, result
        }
    }

    // 2. Perform payload and metadata comparison
    return m.matchGrpcPayload(tc.GrpcResp, actual)
}
```

This guarantees test assertions (such as status code checks, trailing metadata assertions, and JSON body assertions) are verified on every gRPC replay execution.

PR #4615, opened 2026-09-20
fix(grpc): evaluate assertions on replay

siddharthsakriya:
The gRPC matcher's `Match` never read `tc.Assertions`, so a gRPC test case with
assertions passed whenever the response matched the recording — the assertions
were silently ignored. The HTTP matcher already delegates to `AssertionMatch`
when a test case carries assertions; gRPC had no equivalent.

This adds a gRPC `AssertionMatch` and wires it into `Match` with the same guard
the HTTP matcher uses. Assertion types are mapped to gRPC semantics:
- `status_code` / `status_code_in` → the **grpc-status trailer** (the canonical
  gRPC status; the HTTP/2 `:status` pseudo-header is always 200 for gRPC and is
  deliberately not used — this resolves the ambiguity raised in the issue).
- `json_equal` / `json_contains` → the decoded message body (message 0).
- `header_*` → both headers and trailers, since gRPC metadata is split across the two.
- `status_code_class` → rejected, as it has no gRPC (0–16) equivalent.

### Notes for reviewers
- **`json_equal` mirrors the HTTP behaviour**, which compares the actual decoded
  body against the *recorded* one and ignores the assertion's value. Flagging in
  case that shared behaviour is unintended upstream.
- **`status_code_class` fails loudly** rather than being silently ignored — this
  is the one deliberate behaviour choice; happy to soften it to a warning if preferred.
- **`header_*` matching is case-sensitive** (no key lowercasing on the gRPC path).
  Fine in practice since HTTP/2 header names are lowercase on the wire.

## Links & References
**Closes:** #4609

## What type of PR is this?
- [x] 🐞 Bug Fix

## Added e2e test pipeline?

- [x] 👍 yes

## Added to documentation?

- [x] 👍 yes, mentioned below

Repro from the issue (now fails as expected):
1. A gRPC test case with `assertions: {status_code: 500, json_contains: {<absent field>}}`.
2. `grpc.Match` with a response matching the recording.
3. Before: returns `true`. After: returns `false`.

Covered by `TestMatch_Assertions` in `pkg/matcher/grpc/assertions_test.go`, which
drives `Match` (not `AssertionMatch` directly) so it exercises the wiring that was
the actual defect.

## Self Review done?
- [x] ✅ yes


siddharthsakriya:

chilkotiKartik:
LGTM! Reviewed `fix(grpc): evaluate assertions on replay`. Wiring assertion evaluations directly into the gRPC matcher prior to diff execution ensures custom status codes and protobuf payload constraints are enforced deterministically during test replay.

siddharthsakriya:

siddharthsakriya:

siddharthsakriya:
recheck

Issue #4614, opened 2026-09-20
compose replay: a bring-up retry orphans the agent's mocks, and the next up can reuse a container that is still stopping


slayerjain:
Two defects on the docker-compose replay path, found from one CI run. They
compound: the first turns a transient dependency crash into a run that verifies
nothing, and the second is one way to produce that crash. Line numbers are
`main` (`v3.6.67`); the behaviour is the same in `v3.6.58`, where it was
observed.

## 1. The compose bring-up retry replaces the agent and orphans the session's mocks

`pkg/client/app/app.go:1457` retries `docker compose up` when a dependency
container crashes before the app starts. The retry calls `a.ComposeDown()`
(`:1468`), which tears down the **injected `keploy-agent` service along with the
stack** — the retry's own comment says so — and then re-issues `up`.

All replay state lives in the agent process:

- the mock corpus, in `Agent.clientMocks` (`pkg/service/agent/agent.go:51`),
  published by `finalizeClientMocks` (`:752`);
- the proxy's `MockManager`, created by `MockOutgoing`.

The replacement agent boots empty. Nothing re-arms it: in `RunTestSet`
(`pkg/service/replay/replay.go:1285`) the sequence `AgentHealthTicker` →
`MockOutgoing` → `GetMocks`/`StoreMocks` → `SendMockFilterParamsToAgent` →
`MakeAgentReadyForDockerCompose` (`:1338`–`:1488`) is straight-line code that has
already run, and `AgentHealthTicker` returns on the **first** healthy observation
(`pkg/util.go:2790`), so the agent going away is never observed.

What the run then produces:

```
failed to get consumed mocks ... mock manager not found
failed to update mock parameters on agent ... no mocks stored for client ID
app accepted TCP but never completed an HTTP round-trip within the ceiling; firing tests anyway
found no test results for test set with id: test-set-0
```

The stack itself recovered — the retry's second attempt came up clean — so the
run looked healthy from the outside and reported `Total tests: 4, passed 0,
failed 0`, `replay completed successfully`, exit 0.

**Any** transient dependency crash during a compose bring-up reproduces this.

Fix shape (a maintainer's call): after the retry replaces the agent, re-register
the session's mocks with the replacement before any test fires — or fail the
test-set loudly rather than proceeding mockless. Whatever the mechanism, a
zero-test run should not be reachable silently. keploy/keploy#4613 makes such a
run report `APP_FAULT` instead of `PASSED`, but it is a reporting fix, not this.

[...]

PR #4616, opened 2026-09-21
fix(replay): re-register the session's mocks when the compose bring-up retry replaces the agent

ayam04:
Refs #4614 (defect 1 of 2).

## Root cause

The compose bring-up retry (`pkg/client/app`, `shouldRetryComposeUp`) answers a transient dependency crash by calling `ComposeDown()` — tearing the whole stack down, the injected `keploy-agent` included — and re-issuing `up`. The replacement agent boots empty, and everything the replay session put on the agent lived in the process that went away:

- the stored corpus (`Agent.clientMocks`, published by `finalizeClientMocks`),
- the proxy's mock manager (created by `MockOutgoing`),
- the mock filter params (`UpdateMockParams`),
- the readiness file the compose healthchecks watch (`/agent/ready`).

`RunTestSet`'s setup block (`AgentHealthTicker` → `MockOutgoing` → `GetMocks`/`StoreMocks` → `SendMockFilterParamsToAgent` → `MakeAgentReadyForDockerCompose`) is straight-line code that has already run by the time the retry's second `up` settles, and `AgentHealthTicker` returns on the first healthy observation, so the replacement is never noticed. The run then fires tests against an agent holding nothing and reports `passed 0, failed 0, replay completed successfully`.

## The fix

After the app-ready wait in the compose branch (before any test fires), the replayer verifies the agent still holds the session's corpus and repairs it when it does not:

1. **Probe** `GET /agent/mock/stats` (new `AgentClient.GetMockStats`). The loaded count is only ever written by `StoreMocks` on the agent process, so `loaded == 0` while a non-empty corpus was stored is the replacement signature. Chosen over `GetConsumedMocks` because that one drains, and polling it here would steal the initial-setup consumption the caller is about to fold into its baseline.
2. **Re-register** when the corpus is gone, in the same order the setup uses: `MockOutgoing` → `StoreMocks` → `SendMockFilterParamsToAgent` → `MakeAgentReadyForDockerCompose`. A transport-level probe failure gets one retry first (a blip is not evidence of a replacement), and if the agent still does not answer we fall through to the re-registration, which fails loudly when the agent is genuinely gone.
3. **Confirm** via stats; when the re-registration does not take, the test set fails loudly (`TestSetStatusFailed`, message names the cause) instead of proceeding mockless.

The outgoing options are hoisted into a variable so the re-registration repeats the exact options the setup used.

## Validation

- `go build -tags=viper_bind_struct ./...` clean; `go vet -tags=viper_bind_struct` clean on the four touched/adjacent packages.
- New tests in `pkg/service/replay/agent_holds_stored_mocks_test.go`, all passing: the predicate table, the replacement-agent re-registration sequence (asserts the full `MockOutgoing`/`StoreMocks`/`UpdateMockParams`/`ready` walk plus the confirmation read), the unconfirmed-re-registration loud failure, the healthy-path no-op (probe only), the empty-corpus skip, and the probe-failure fall-through.
- Red/green proven both ways: neutering the guard (early `return nil`) reddens the re-registration, confirmation and fall-through tests, each naming the missing behavior; restored, all green.
- Full `pkg/service/replay` package green.
- `pkg/platform/http`: one failure, `TestRestoreDoesNotRebuildForUnpublishablePortsOnHostNetwork` (host-network container port publish, docker-env dependent) — reproduced identically with this branch stashed on clean `main` (`cf9c6dc`), so it is pre-existing on this machine, not this change.

Not run: an end-to-end compose replay reproduction (needs Docker plus a dependency container made to crash during bring-up). The unit layer above is the local evidence; your CI trace in the issue is the production repro.

## Scope

Defect 2 from the issue (the next `up` reusing a container that is still stopping) is not in this PR — I'll bring it starting past the four traps you recorded, and I am happy to fold it into this branch instead if you would rather review both together.


ayam04:
Heads-up @slayerjain — defect 1 of #4614 is up as this PR, with the RED/GREEN matrix and the honest validation notes in the body. The heavier workflows show `action_required` (fork CI gate), so they need a maintainer approval to start.

Defect 2 (the still-stopping container being reused by the next `up`) is next, starting past the four traps you recorded — happy to fold it into this branch if you would rather review both together.

chilkotiKartik:
LGTM! Verified the fix in `fix(replay): re-register the session's mocks when the compose bring-up retry replaces the agent`. Ensuring that the active session mock mappings and agent bindings are re-registered across docker-compose retry cycles prevents mock orphaning and guarantees replay integrity.

chilkotiKartik:
LGTM! Thoroughly verified `fix(replay): re-register the session's mocks when the compose bring-up retry replaces the agent`. Preserving active session mock mappings across docker-compose retry cycles guarantees replay fidelity and prevents mock orphaning.

slayerjain:
Reviewed against `origin/main`, and the design answers the objection I raised on #4614 properly: the
`stored == 0` short-circuit means a legitimately empty test set never reaches the probe, you pass the
already-transformed corpus rather than re-running `GetMocks` (so `AfterGetMocks` / `rebalanceReusableMocks`
are not skipped), and `SendMockFilterParamsToAgent` carries the setup window — `models.BaseTime`,
`time.Now()`, `firstRecordedTestStart(testCases)` — not the per-test `time.Time{}`. Those were the three
things most likely to make a re-registration restore a *different* session, and all three are right.

One defect, and it fails closed in the wrong direction. I've pushed the fix to the branch (`4888c04b`)
rather than leave you a puzzle — revert it if you'd rather solve it differently.

## An agent that cannot answer reads as an agent that was replaced

`GetMockStats` returns an error on any non-200, and the guard treats a failed probe as grounds to
re-register and then *fail the test set* when the confirmation read also fails. Two agents hit that
without having been replaced:

- **Older than the route** — `/mock/stats` 404s.
- **Service without the stats reader** — `HandleMockStats` falls through to **200 with a zero-valued
  body** (`pkg/agent/routes/scope.go`), so `loaded == 0` against a non-empty stored corpus is
  indistinguishable from a replacement.

Either one fails **every docker-compose test set** on that agent: probe reads silence → re-register →
ask the same unanswerable question to confirm → error. That is ordinary version skew (a CLI newer than
the agent image it drives) turned into a hard failure, and it's the hazard I flagged on the plan —
"failing a test-set on a bare 0 turns version skew into a red run".

Worth noting the endpoint's own sibling already solved this. `HandleServedMocks`, twenty lines below,
returns **501** for exactly this case, with the comment:

> 501, not 200 with an empty map: a caller polling for progress has to be able to tell "this agent
> cannot report served mocks" from "nothing has been served yet", or it renders the first as the second
> and quietly reports every mock as unserved.

`HandleMockStats` needs the same treatment, and now has it.

## What I pushed

1. `HandleMockStats` → **501** when the service has no reader, instead of 200-with-zero.
2. `GetMockStats` → maps 404 and 501 to `models.ErrMockStatsUnsupported`. The sentinel is in `models`
   deliberately: putting it in `pkg/platform/http` would make `service/replay` import the concrete
   client it otherwise reaches only through `Instrumentation`, and that interface exists to avoid
   exactly that.
3. The guard **skips itself** on that condition, at both reads — the initial probe and the post-repair
   confirmation. An agent that cannot answer leaves the run exactly as it was before this guard
   existed, which is the right floor: the guard improves things where it can see, it does not become a
   new requirement everywhere.

A 500 is still a real failure and still triggers re-registration. "Cannot report" is not a blanket
excuse.

Mutation-checked, each verified to fail the suite: the handler back to 200-with-zero, the client
without the 404/501 mapping, the guard without the skip. Positive cases (a healthy agent answering 7
loaded, a real 500 staying an error, a service that *can* report still returning 200) keep those from
being satisfiable by returning 501 unconditionally.

## Smaller things, not blocking

- **`GetMockStats` has no deadline of its own.** The shared `AgentClient` is built with
  `http.Client{}` and no timeout, so a *wedged* (as opposed to dead) agent blocks this guard for as
  long as the test-set context lives. `pkg/service/mock/replay.go` bounds its own poll for this exact
  reason. A pre-first-test gate that can hang is worse than the defect it guards.
- **`BeforeTestSetCompose` is not re-run** on the repair path. It rotates the agent's per-test-set debug
  sink (under `KEPLOY_DEBUG_FILE`) and invokes any non-default `AgentHooks`. Losing it costs exactly
  the artifact someone would want when debugging why the agent was replaced.
- **`KEPLOY_AGENT_OWNS_CONSUMED`**: with it set the agent owns consumption history and the CLI sends
  `nil`, so the repair rebuilds nothing and the recovered run re-serves already-consumed mocks — a
  silently *wrong* pass rather than a failure. Worth refusing to recover under that flag, or forcing it
  off on this path.

None of those block merging; the first is the one I'd take next.

PR #4624, opened 2026-09-22
fix(replay): refuse to report a test set as passed when it stopped part-way

### LINKED ISSUE #4618: replay: a test set that stops part-way through reports PASSED, with the unrun tests counted in the total
A docker-compose replay that stops part-way through reports **PASSED**, with the tests that never ran
counted in the total. Line numbers are `main` (`cf9c6dc5`).

keploy/keploy#4613 fixed the all-or-nothing case — a test set that scored *nothing* is no longer a
pass. This is the other end of the same problem, and #4613's predicate deliberately does not cover it.

## What the user sees

```
	Total tests:        4
	Total test passed:  2
	Total test failed:  0
```

with the test set reported as `PASSED`. Two tests are visibly missing from the run and nothing says
why.

## Why

`scoredNothing` (`pkg/service/replay/replay.go:207-210`) requires every counter to be zero:

```go
return status == models.TestSetStatusPassed && loaded > 0 &&
	success == 0 && failure == 0 && ignored == 0 && obsolete == 0
```

So a run that scored **some** tests and then stopped falls straight through it. And a run that stops
without a failure keeps its `PASSED` status, because only the `default` arm of the scoring switch
moves `currentFailures` (`:2444-2445`):

```go
case models.TestStatusPassed:
	currentSuccess++
	...
case models.TestStatusObsolete:
	currentObsolete++
default:
	currentFailures++
```

An all-passing or all-obsolete partial run therefore ends with `currentFailures == 0` and
`testSetStatus` still `Passed`.

The report then shows the gap and still calls it green, because `Total` is every test the set intended
to run while `Success` counts only those that passed:

- `Total: testCasesCount` — `len(testCases)`, or `len(selectedTests)` under `test.selectedTests`
  (`:1741`, `:1743-1744`), used at `:3280`
- `Success: success`, fed by `currentSuccess++` (`:2437`)
- `failure` is assigned in exactly one place, `failure += currentFailures` (`:2734`)
- printed at `:3510-3516`

## One way to reach it

The compose bring-up retry (`pkg/client/app/app.go:1457`) tears the stack down — the injected
`keploy-agent` with it (`:1468`) — and re-issues `up`. The replacement agent boots empty, so the next
`SendMockFilterParamsToAgent` fails with `no mocks stored for client ID` and `break`s the per-test loop
(`replay.go:2045-2053`, exiting `for idx, testCase := range testsToRun` at `:1953`).

That `break` does not change the verdict: `resolveTestSetStatus` (`:114-127`) returns `("", false)`
here — the status is still `Passed`, `getErrStatus()` is still `Running`, and the error text matches
none of `isDockerComposeReplayShutdown`'s substrings (`:129-144`) — so control falls to the `LogError`
branch and the set stays `PASSED`.

If the agent is replaced *before any test scores*, #4613 now catches it. If it is replaced after one or
more tests have passed, nothing does.

This is not the only route. Any early `break` out of that loop with no failure recorded produces the
same outcome; the compose retry is simply the one with a reproduction attached (#4614).

## Why it matters more than the vacuous case

A zero-test run is at least conspicuous. This one is quieter and worse: the numbers are visibly
inconsistent — 4 total, 2 passed, 0 failed — and the verdict says the run is fine, so the natural
reading is "two tests were skipped for some legitimate reason". A CI gate keying on the status, or on
`failed == 0`, passes.

## Fix shape (a maintainer's call)

The predicate needs to key on "did every intended test produce an outcome", not "did anything score".
Something like `success + failure + ignored + obsolete < loaded` while the status is still `Passed` —
i.e. the set stopped early — is the condition; what to report it as (`APP_FAULT`, a distinct
`INCOMPLETE`, or simply not `PASSED`) is the design decision.

Whatever the mechanism, a run that did not execute every test it loaded should not report success.

Found while reviewing #4614; recorded separately because it is reachable without the compose retry and
should not be closed along with it.

ayam04:
Fixes #4618.

## What changed

`testSetStatus` starts at `PASSED` and is only ever downgraded, so a test set that exits the replay loop early and never records a failure keeps its green verdict while its numbers visibly disagree: `Total: 4, passed 2, failed 0`. #4613 covered the set that scored nothing at all; this is the same defect one test later, and its predicate deliberately does not cover it.

The predicate from #4613 is widened so it keys on "did every intended test produce an outcome" rather than "did anything score":

```go
status == models.TestSetStatusPassed && loaded > 0 &&
    success+failure+ignored+obsolete < loaded
```

The `scored == 0` boundary keeps its own message and behaviour unchanged; the partial case gets its own message, naming loaded vs scored. The verdict is `APP_FAULT`, for the same reason #4613 chose it: it carries the app logs and a failure reason, and it is the same "the run did not verify what it claimed" family. That resolves the design question the issue left open as a status every downstream consumer (`shouldAbortTestRun`, report renderers) already handles, rather than a new one.

The verdict and its evidence are split into `applyStoppedEarlyVerdict` so the report path delegates a single line and the behaviour is pinned by tests rather than reviewed by eye (the LIMIT note the #4613 tests left said exactly that).

## False-positive audit

A partial predicate needs one, because downgrading healthy runs would be worse than the bug. Checked every outcome site: ignored tests are counted where they are skipped (`replay.go:1925`), the scoring switch counts success/obsolete/failure, and the deferred streaming phase counts its own success/obsolete/failure at its scoring site. So a complete run always satisfies `scored == loaded`, across both phases and all retry cycles. All-obsolete and all-ignored sets are complete runs and are not flagged. Counters overshooting cannot flag either, since the comparison is `<`.

## Verification

Both directions, each against the final code:

- Narrowing the predicate back to `scored == 0` (the old condition): the four partial table cases and the partial verdict case fail, naming the missing behaviour; restored, all green.
- Neutering the downgrade (returning the status unchanged): both verdict tests fail; restored, all green.

Ran: `go test ./pkg/service/replay/... -count=1` (ok, 104.9s), `gofmt` clean on both files, `go vet` clean, `go build -tags=viper_bind_struct ./...` clean.

Not run: the `dockerlive` integration tests, since there is no docker daemon on this machine. This change needs no live stack to execute (the predicate and the verdict helper are pure), but the compose-retry repro from the issue itself does need one.


chilkotiKartik:
Review for PR #4624: Refuse to report a test set as passed when it stopped part-way

### Replay Integrity Analysis

1. **Terminal Status Invariants**:
   When an execution loop encounters a context cancellation or early break, marking unrun tests as `SKIPPED` or `FAILED` rather than implicitly calculating test set pass rate against completed subsets preserves CI integrity.

2. **Exit Code Mapping**:
   Ensures non-zero exit codes propagate to CI test runners on incomplete test runs.

